PR #15, opened 2018-07-21
Barebones WAR support

NamelessFaceless:
Just PRing the barebones WAR support I ported from the old events system, etc. Some of the code is super messy right now, and there's a lot to be added, still. But I thought we should have the barebones, at the very least, for when the website goes live. I'll be developing it a lot over the next week, and it should be functional for what's already there. The output is sane.

Accounted for currently:
- GCD Uptime
- Wasted rage (I'm not calling it beast gauge or whatever the fuck they call it)
- Rage generated by each GCD (goes together with the previously mentioned wasted rage)


src/data/ACTIONS/SHARED.js
```diff
@@ -2,18 +2,18 @@ export default {
 	SPRINT: {
 		id: 3,
 		name: 'Sprint',
-		icon: process.env.PUBLIC_URL + '/icon/action/sprint.png',
+		icon: process.env.PUBLIC_URL + '/icon/action/sprint.png'
```
ackwell: Dunno what happened here. Was gonna fix it up myself, but dropping a comment instead to go with some others.

src/parser/jobs/war/About.js
```diff
@@ -0,0 +1,21 @@
+import React, {Fragment} from 'react'
+import {Icon, Message} from 'semantic-ui-react'
+
+import CONTRIBUTORS from 'data/CONTRIBUTORS'
+import CoreAbout from 'parser/core/modules/About'
+
+export default class About extends CoreAbout {
+	description = <Fragment>
```
ackwell: Can I just get an update on the `description` to reflect the current state of the module group? Something something Please Look Forward To It something.

src/parser/jobs/war/Gauge.js
```diff
@@ -0,0 +1,108 @@
+import React, {Fragment} from 'react'
+
+import {ActionLink} from 'components/ui/DbLink'
+import ACTIONS from 'data/ACTIONS'
+//import STATUSES from 'data/STATUSES'
+import Module from 'parser/core/Module'
+import {Suggestion, SEVERITY} from 'parser/core/modules/Suggestions'
+
+// General actions that give Rage (this is how I'm referring to the Warrior gauge) -- Except Storm's Path, since it's a fringe case that gives +20 instead of 10.
```
ackwell: Hm. If you consolidate this into `[id]: <rage delta>`, it'll cut down on some boiler plate and make maintaining the numbers easier. Something like

```js
const RAGE_ACTIONS = {
	[ACTIONS.MAIM.id]: 10,
	[ACTIONS.STORMS_PATH]: 20,
	[ACTIONS.INFURIATE]: 50,
	// ...
}
```

and then you can consolidate the `_onCast` hook to check if there's a key, and handle adding to gauge and tracking wastage using it if there is one.

src/parser/jobs/war/Gauge.js
```diff
@@ -0,0 +1,108 @@
...
+const RAGE_ACTIONS = [
+	ACTIONS.MAIM.id,
+	ACTIONS.STORMS_EYE.id,
+	ACTIONS.SKULL_SUNDER.id,
+	ACTIONS.BUTCHERS_BLOCK.id,
+]
+
+// Actions that reduce Infuriate's cooldown.
+const INFURIATECD_ACTIONS = [
```
ackwell: `INFURIATE_CD_ACTIONS` :blobknife:

src/parser/jobs/war/Gauge.js
```diff
@@ -0,0 +1,108 @@
...
+
+	_onComplete() {
+		if (this._wastedRage >= 50) {
+			this.suggestions.add(new Suggestion({
+				icon: ACTIONS.INFURIATE.icon,
+				content: <Fragment>
+					You used <ActionLink {...ACTIONS.STORMS_PATH}/>, <ActionLink {...ACTIONS.STORMS_EYE}/>, <ActionLink {...ACTIONS.INFURIATE}/>, or any gauge generators in a way that overcapped you.
+					And you lost at least one Fell Cleave due to it.
+				</Fragment>,
+				severity: SEVERITY.MEDIUM,
```
ackwell: Is this worth upgrading to a `MAJOR` after a breakpoint or nah? Iuno much about WAR :blobsweat:

NamelessFaceless: It might be, yeah. I think if you're over like, idk, 100 it should be upgraded. I don't know any fights that you'd end up like that.

PR #19, opened 2018-07-22
Update installation instructions

charles-shao:
Minor update to the instructions while I was setting this up on my local dev environment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,8 +46,8 @@ yarn
 
 While `yarn` is running, copy the `.env.local.example` file in the project root, and call it `.env.local`. Make a few changes in it:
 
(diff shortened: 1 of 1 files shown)


ackwell:
The `TODO_FINAL_DEPLOY_URL` and `INSERT_API_KEY_HERE` are intentionally referencing the value, not the key. Don't want someone unfamiliar with `.env` files accidentally editing the key and leaving the placeholder values in place :notlikeblob:.

Oh and yeah sorry I just pushed a fix for the profile url, hah.

charles-shao:
Oh ok, fair enough. Closing PR.

PR #20, opened 2018-07-22
Add missing 'Cleric Stance' to shared role actions

charles-shao:
This was missing when I was analysing SCH.

### DIFF
--- a/src/data/STATUSES/SCH.js
+++ b/src/data/STATUSES/SCH.js
@@ -1,10 +1,4 @@
 export default {
-	CLERIC_STANCE: {
-		id: 145,
-		name: 'Cleric Stance',
-		icon: 'https://secure.xivdb.com/img/game/010000/010402.png'
-	},
-
 	FEY_WIND: {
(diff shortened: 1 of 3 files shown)


src/data/STATUSES/SCH.js
```diff
@@ -1,4 +1,10 @@
 export default {
+	CLERIC_STANCE: {
```
ackwell: This should probably go in `ROLE.js`, with the other role actions.

I've not actually got any role actions in for SCH - it's not supported yet per se.

charles-shao: Sure, I'll break out shared statuses. For clarification purposes, was it intended to mix statuses that afflict both players and enemies in the same file?

ackwell: Yeah. I'm only classifying actions/statuses by the job that actually uses them (or role, etc). Despite the different icon styles, XIV treats both buffs and debuffs as a homogeneous 'Status'.

PR #30, opened 2018-07-23
Add Context to Raven Error Handling

KhloeLeclair:
This pull request adds environment and release tagging for all Raven exceptions. It also adds a good deal of contextual data to exceptions captured by the parser within a module's event handlers.

Currently, this code checks for an `onError` function on the module itself. I chose to implement it this way rather than through another event because there's no good way to return data from an event handler. This method probably needs a better name.

If that method exists, it's called and the result is used. If the method does not exist, or it returns `undefined`, a function is run that iterates all the keys of the module and grabs out all the primitive data types. For a lot of modules, this will still grab some useful information.

All exceptions captured in the event handling code include tags for the current job and module. In addition to the data gathered from the module, the event that triggered the error is included as well as the report, fight, and player being viewed.


NamelessFaceless:
lgtm :blobmorning:

PR #40, opened 2018-07-24
Two new modules for NIN and minor cleanup

Toastdeib:
This NIN update includes a module for tracking Dream Within A Dream usage outside of Trick Attack windows and Armor Crush usage in them, as well as module for tracking Duality usage on any non-Aeolian Edge GCDs. I also tidied up a couple of things here and there to keep the code as clean and consistent as possible.

### DIFF
--- a/src/parser/jobs/nin/Duality.js
+++ b/src/parser/jobs/nin/Duality.js
@@ -3,7 +3,7 @@ import React, {Fragment} from 'react'
 
 import {ActionLink} from 'components/ui/DbLink'
 import ACTIONS from 'data/ACTIONS'
-//import STATUSES from 'data/STATUSES'
+import STATUSES from 'data/STATUSES'
 import Module from 'parser/core/Module'
 import {Suggestion, SEVERITY} from 'parser/core/modules/Suggestions'
 
@@ -13,16 +13,17 @@ const DUALITY_GCDS = {
 	[ACTIONS.AEOLIAN_EDGE.id]: 0,
 	[ACTIONS.SHADOW_FANG.id]: 1,
 	[ACTIONS.ARMOR_CRUSH.id]: 1,
+	[ACTIONS.THROWING_DAGGER.id]: 1,
 }
 
 export default class Duality extends Module {
 	static handle = 'duality'
 	static dependencies = [
+		'combatants',
 		'suggestions',
 	]
 
 	_badDualityUses = 0
-	_dualityActive = false
 
 	constructor(...args) {
 		super(...args)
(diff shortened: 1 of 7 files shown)


src/parser/jobs/nin/Duality.js
```diff
@@ -0,0 +1,58 @@
+import React, {Fragment} from 'react'
+//import {Icon, Message} from 'semantic-ui-react'
+
+import {ActionLink} from 'components/ui/DbLink'
+import ACTIONS from 'data/ACTIONS'
+//import STATUSES from 'data/STATUSES'
+import Module from 'parser/core/Module'
+import {Suggestion, SEVERITY} from 'parser/core/modules/Suggestions'
+
+const DUALITY_GCDS = {
```
ackwell: Maybe consider using a whitelist? Something like

```js
const CORRECT_DUALITY = [
	ACTIONS.AEOLIAN_EDGE.id
]

// then you can do...

if (!CORRECT_DUALITY.includes(abilityId)) {
	this._badDualityUses++
}
```

src/parser/jobs/nin/Duality.js
```diff
@@ -0,0 +1,58 @@
...
+	constructor(...args) {
+		super(...args)
+		this.addHook('cast', {by: 'player'}, this._onCast)
+		this.addHook('complete', this._onComplete)
+	}
+
+	_onCast(event) {
+		const abilityId = event.ability.guid
+		
+		if (abilityId === ACTIONS.DUALITY.id) {
```
ackwell: Event order is `applybuff` -> `cast` -> `removebuff`, so you can probably use the `combatants` module and just check `combatants.selected.hasStatus(STATUSES.DUALITY.id)`

src/parser/jobs/nin/Duality.js
Toastdeib: The problem with that approach is that it would increment `_badDualityUses` on abilities that don't actually consume the buff (e.g. other OGCDs or Death Blossom). I set it up the way I did specifically because it's only a subset of abilities that actually work with Duality. That said, it occurs to me now that I may have missed one - gotta test whether Duality works on Throwing Dagger. I'll add another comment when I verify that.

src/parser/jobs/nin/Duality.js
Toastdeib: I'll twiddle that and see if I get the same results. I'll definitely be cleaner than my current approach. And while I'm at it - would that also work for checking for the TA status (with the enemies module) in my TA window module? If yes, I should probably switch that as well - my current approach for that doesn't account properly for adds anyway, now that I think about it.

src/parser/jobs/nin/Duality.js
ackwell: Ahhhh, of course. Good point.

src/parser/jobs/nin/Duality.js
ackwell: It _should_, yeah. I'm a bit rusty on `combatants` and `enemies`, but have a look around. I think `this.enemies.getEntity(event.targetID).hasStatus(...)` might work? (assuming targetID is a mob)

src/parser/jobs/nin/Duality.js
Toastdeib: Confirmed Throwing Dagger consumes Duality, so that'll be on the list when I update the PR.

PR #41, opened 2018-07-24
Fixed rage gauge calculation and added 5-GCD IR Usage

NamelessFaceless:
- Fixed the longstanding bug in the rage calculation
- Added the 5GCD IR Viewer (OGCDs not excluded)


ackwell:
'at'll do fer now

PR #43, opened 2018-07-24
AST action and status dump (for real this time);

suShirow:
AST arcana undraw module - checks if they're not using macros for undraw;
Added ast/ArcanaGroups.js for future sanity


src/data/ACTIONS/AST.js
```diff
@@ -14,22 +14,366 @@ export default {
...
+
+	GRAVITY: {
+		id: 3615,
+		name: 'Gravity',
+		icon: 'https://secure.xivdb.com/img/game/003000/003123.png',
+		onGcd: true,
+		castTime: 3,
+	},
+
+	DIGNITY: {
```
ackwell: Full names please and thank

src/data/ACTIONS/AST.js
```diff
@@ -14,22 +14,366 @@ export default {
...
+		name: 'The Spire',
+		icon: 'https://secure.xivdb.com/img/game/003000/003115.png',
+		cooldown: 30
+	},
+
+	// -----
+	// Expanded arcana cast from draw slot
+	// -----
+
+	THE_BALANCE_EXPANDED: {
```
ackwell: Today I learned that square has different action IDs for all the royals. What the flying fuck.

src/parser/jobs/ast/ArcanaUndrawUsage.js
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	_onComplete() {
+		
+		const badUndraws = this._badUndraws
+			console.warn(this._badUndraws, 'undraws')
+		if (badUndraws.length) {
+			this.suggestions.add(new Suggestion({
+				icon: 'https://secure.xivdb.com/img/game/003000/003108.png', // Undraw action
+				content: <Fragment>
```
ackwell: This content block is too big, if we're gonna keep this suggestion, will need to move the macros into some other format.

src/parser/jobs/ast/ArcanaUndrawUsage.js
```diff
@@ -0,0 +1,97 @@
...
+import {Suggestion, SEVERITY} from 'parser/core/modules/Suggestions'
+import {ActionLink} from 'components/ui/DbLink'
+
+const OGCD_ARCANA_REMOVAL = [
+	ACTIONS.UNDRAW_SPREAD.id,
+	ACTIONS.EMPTY_ROAD.id,
+	ACTIONS.UNDRAW.id,
+]
+
+export default class ArcanaUndrawUsage extends Module {
```
ackwell: This module seems a bit suss to me.

I'm no healer main, but while I'd certainly _recommend_ using a shadowflare macro on SMN, it's not something I'd necessarily raise an issue about.

Are we 100% sure that not using a macro is always a bad idea?

src/parser/jobs/ast/AstWeaving.js
```diff
@@ -0,0 +1,22 @@
+import Weaving from 'parser/core/modules/Weaving'
+import ACTIONS from 'data/ACTIONS'
+
+export default class AstWeaving extends Weaving {
```
ackwell: Malefic III is a 1.5s cast, which I realise I was missing from the core weaving handling. That's been added now, allowing for 1 weave after a 1.5s cast - is this override still required?

suShirow: No, we can remove this override now. I had brought this up the other day, glad we could keep it neat. There's no other reason to check for weaving to my knowledge

src/parser/jobs/ast/ArcanaUndrawUsage.js
suShirow: So the funny thing is that if you use the given actions, you can't remove cards in your spread while casting another spell. There's no animation lock but it still delays the next GCD by a fraction of a second, even if you mash it. I would value getting more opinions on this, but that's my observation.

src/data/ACTIONS/AST.js
suShirow: That's what I thought at first, but it's not as bad! They actually they only differentiate between Expanded/Single-target, and then another set for Cast from draw or Cast from spread. I wonder if this will affect your Arcanum.js in core

src/data/ACTIONS/AST.js
suShirow: Will do

PR #53, opened 2018-07-25
More BLM things

rexatlanti:
mostly bugfixes with stuff that I found and new suggestions. Changes include:
-Fire1 counter as minor suggestion
-downtime module thingy. Seems to be actually useful for other jobs as well
-Skipping B4 in the last fire cycle suggestion
Guess that's about it


src/data/ACTIONS/BLM.js
```diff
@@ -1,192 +1,192 @@
...
-		onGcd: true,
-	},
-	THUNDER_IV_PROC: {
-		id: '742PROC',
-		name: 'Thunder IV',
-		icon: process.env.PUBLIC_URL + '/icon/action/t4p.png',
-		onGcd: true,
-	},
-}
+export default {
```
ackwell: Looks like something's gone wrong w/ line endings.

src/parser/core/modules/Cooldowns.js
```diff
@@ -1,184 +1,184 @@
...
-			(time, status) => time + Math.min(status.length, currentTimestamp - status.timestamp),
-			cd.current? Math.min(cd.current.length, currentTimestamp - cd.current.timestamp) : 0
-		)
-	}
-
-	get used() {
-		return Object.keys(this._cooldowns)
-	}
-}
+import React from 'react'
```
ackwell: Samesies, line endings are jank

src/parser/jobs/blm/BlmCooldowns.js
```diff
@@ -0,0 +1,148 @@
...
+
+		//new Rule and adds the array of Requirements that just got generated
+		this.checklist.add(new Rule({
+			name: 'Use your OGCDs',
+			description: 'Always make sure to use your OGCDs when they are up but don\'t clip them. Utilize your procs or fast Blizzard III or Fire IIIs.',
+			requirements: OGCDRequirements,
+		}))
+	}
+
+	//cool function that I invented that just sets ok usage as 98% and falls very quickly
```
ackwell: While I understand what you're trying to do with this, the current setup doesn't lend itself to being very clear.

Someone looks at a suggestion and they see a really low score when they know they used it a _reasonable_ amount, and will think it's bugging out - possibly even ignore it.

rexatlanti: well a reasonable ammount is 98% even slightly below a reasonable amount is 95%. I didn't want to use the percent at all but no idea how to make something different with Rule.js

PR #56, opened 2018-07-25
Basic MNK analysis

ayyaruq:
Super simple MNK analysis. Can be merged as-is or I can keep adding stuff from the NOTES. Very open to feedback on what I have in mind or things I might have forgotten about.

### Uptime
- [ ] Don't clip or drop Demolish by more than 2 seconds, unless reapplying would be on GL0
- [ ] Avoid GL0 Demolish applications, if opener or downtime ignore this
- [ ] GL3 uptime
- [ ] General auto-attack uptime

### Forms
- [x] Reset forms by using Opo-Opo form skills in a different form
- [x] Form expiry (taking over 10s to transition)
- [x] Form Shifting under PB (WHY IS THIS A THING?)
- [x] Using formless DK or AotD
- [x] Form Shifting mid-combo

### AoE
- [ ] Make sure there's enough enemies for RB (2+) or AotD (5+) effectiveness
- [ ] Allow AotD for Silence if we can tell if the enemy is actually interrupted
- [ ] Demolish on multiple targets?

### Tornado Kick
- [ ] Minor on 8th RoF GCD instead of 9th
- [ ] Major if just after RoF drops
- [ ] Check recovery doesn't take more than 4 GCDs (PB is 3, ending before Snap and using RoW is 4)
- [ ] Use before forced downtime longer than RoE can be used, this might be fight specific and is hard to track since FFLogs is weird about GL refreshes

### Positionals
- [ ] Can kinda guess this from the damage mainly because MNK potency is high but it's a bit giri-giri
- [ ] Crit the Boot

### Buff Windows
- [ ] RoF basically on CD, take downtime into account or fights that derp a window
- [ ] Brotherhood, same as RoF besides slight variance for duration

### Chakras
How do we track this? There's a game packet for meditative brotherhood procs but FFLogs doesn't use it
Maybe check for FBC being used at low GL when the next GCD gives a stack

### DIFF
--- a/src/data/ACTIONS/MNK.js
+++ b/src/data/ACTIONS/MNK.js
@@ -1,0 +1,135 @@
+export default {
+	// -----
+	// Player GCDs
+	// -----
+
+	ROCKBREAKER: {
+		id: 70,
+		name: 'Rockbreaker',
+		icon: 'https://secure.xivdb.com/img/game/002000/002529.png',
+		onGcd: true,
+	},
+
+	ONE_ILM_PUNCH: {
+		id: 72,
+		name: 'One-Ilm Punch',
+		icon: 'https://secure.xivdb.com/img/game/000000/000214.png',
+		onGcd: true,
+	},
+
+	DRAGON_KICK: {
+		id: 74,
+		name: 'Dragon Kick',
+		icon: 'https://secure.xivdb.com/img/game/002000/002528.png',
+		onGcd: true,
+	},
+
+	FORM_SHIFT: {
+		id: 4262,
+		name: 'Form Shift',
+		icon: 'https://secure.xivdb.com/img/game/002000/002536.png',
+		onGcd: true,
+	},
+
+	MEDITATION: {
+		id: 3546,
+		name: 'Meditation',
+		icon: 'https://secure.xivdb.com/img/game/002000/002534.png',
+		onGcd: true,
+	},
+
+	// -----
+	// Player OGCDs
+	// -----
+
+	SHOULDER_TACKLE: {
+		id: 71,
+		name: 'Shoulder Tackle',
+		icon: 'https://secure.xivdb.com/img/game/002000/002526.png',
+		cooldown: 30,
+	},
+
+	FISTS_OF_FIRE: {
+		id: 63,
+		name: 'Fists of Fire',
+		icon: 'https://secure.xivdb.com/img/game/000000/000205.png',
+		cooldown: 3,
+	},
+
+	THE_FORBIDDEN_CHAKRA: {
+		id: 3547,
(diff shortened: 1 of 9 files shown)


src/data/ACTIONS/MNK.js
```diff
@@ -0,0 +1,135 @@
...
+	},
+
+	THE_FORBIDDEN_CHAKRA: {
+		id: 3547,
+		name: 'The Forbidden Chakra',
+		icon: 'https://secure.xivdb.com/img/game/002000/002535.png',
+		cooldown: 5,
+	},
+
+	MEDITATION: {
```
Toastdeib: Shouldn't this technically count as `onGcd: true`? It's a short CD, but it still hits all your regular GCD abilities.

ayyaruq: You're right, after a chat in Discord it's moving to GCD and going to implement a module to deal with the different timer on it, similar to SMN/RDM.

src/parser/jobs/mnk/Forms.js
```diff
@@ -0,0 +1,191 @@
...
+				if (action === ACTIONS.FORM_SHIFT.id) {
+					this._poorForm++
+				}
+				return
+			}
+
+			// If we're in downtime or had poor HP uptime, ignore Form Shift
+			if (action === ACTIONS.FORM_SHIFT.id) {
+				// Recovery
+				if (RAISE_STATUSES.find(dead => this.combatants.select.hasStatus(dead))) {
```
ackwell: I'd imagine the recovery from death wouldn't take the full duration of a weakness buff.

Would it be worth hooking the `death` event, and then only allowing form shifts until such a point that they're back in GL3?

src/parser/jobs/mnk/Forms.js
```diff
@@ -0,0 +1,191 @@
...
+
+			// Handle relevant actions per form
+			switch (current_form) {
+			case STATUSES.OPO_OPO_FORM.id:
+				break
+			case STATUSES.RAPTOR_FORM.id:
+				// Using Opo-Opo skills resets form
+				if (OPO_OPO_SKILLS.includes(action)) { this._resetForms++ }
+				break
+			case STATUSES.COEURL_FORM.id:
```
ackwell: Can probably afford to let the switch fall through for these two, saves duping the handling.

src/parser/jobs/mnk/Forms.js
ayyaruq: Maybe, I think actually the raise event is more important. The main thing is to get stacks back ASAP, and what I want to discourage is people wasting 3 GCD's on Form Shift to get into Coeurl when they can just start from Bootshine unless they're out of melee range. Might need to rethink this a bit.

PR #58, opened 2018-07-25
BLM things and added fight duration into the header

rexatlanti:
People wanted to see the current fight duration in the header so I implemented it


src/components/Header.js
```diff
@@ -23,6 +23,11 @@ class Header extends Component {
 		collapseMenu: PropTypes.bool.isRequired,
 	}
 
+	formatDuration(duration) {
```
ackwell: Probably worth moving to `utilities` to keep implementation consolidated.

src/parser/jobs/blm/BlmCooldowns.js
```diff
@@ -0,0 +1,148 @@
...
+			description: 'Always make sure to use your OGCDs when they are up but don\'t clip them. Utilize your procs or fast Blizzard III or Fire IIIs.',
+			requirements: OGCDRequirements,
+		}))
+	}
+
+	//cool function that I invented that just sets ok usage as 98% and falls very quickly
+	_percentFunction(actionId, downtime, fightlength) {
+		const cooldown = getAction(actionId).cooldown
+		const numberOfUses = Math.floor(fightlength/(cooldown*1000))
+		return Math.exp(-downtime*0.03/(DOWNTIME_OK_TIME*numberOfUses))*100
```
ackwell: As before, I'm concerned about how opaque this is. The end-user will probably think it's bugging out or something, and possibly ignore it.

src/parser/jobs/blm/FireCounter.js
```diff
@@ -65,7 +69,36 @@ export default class FireCounter extends Module {
...
+		}
+
+		//suggestion for unneccessary extra F1s.
+		if (this._extraF1s) {
+			this.suggestions.add(new Suggestion({
+				icon: ACTIONS.FIRE_I.icon,
+				content: <Fragment>
+					Casting more than one <ActionLink {...ACTIONS.FIRE_I}/> per Astral Fire cycle is a crutch that should be avoided by better pre-planning of the encounter.
+				</Fragment>,
+				severity: SEVERITY.MINOR,
```
ackwell: Is this _always_ going to be minor? What if I'm a total idiot and chain cast F1 in the fire cycle?

src/components/Header.js
rexatlanti: I c/p it from some other module since I couldn't access it in header

src/parser/jobs/blm/FireCounter.js
rexatlanti: good Idea. I'll add something

src/components/Header.js
ackwell: Exactly

src/parser/jobs/blm/BlmCooldowns.js
rexatlanti: The function itself isn't opaque at all. But all it does is:
```
Downtime: 0s => 100%
Downtime: ALLOWEDOGCDDOWNTIME * MAXPOSSIBLENUMBEROFUSES => 98%
Downtime: way more than that => low%
```
above 95% is green. I've tested various logs from different percentages.
Good people stay easily at ~98%. People around blue/purple get their 95%+.
I would give it a try and if I hear from people who use it "it's confusing" "I feel like it's worthless" etc. then I would get rid of it or replace it with something else.

PR #69, opened 2018-07-26
AST: Buff Extension module. 

suShirow:
It's looking halfway decent now, going to park this here for implementation review (sketchy) and for other devs to input while I work on other modules.
Major concern is how to wait for the delayed aoe effect of CO to resolve on every target.
Minor concern is having to write another set of checks for buff refreshes on pets.

Added getStatuses() to Entity.js


src/parser/core/Entity.js
```diff
@@ -38,4 +38,27 @@ export default class Entity {
...
+
+
+		const currentTimestamp = forTimestamp < this.parser.currentTimestamp ? forTimestamp
+			: this.parser.currentTimestamp
+
+		if (forTimestamp < this.parser.currentTimestamp) {
+			console.warn('Entity:getStatuses timestamp given is later than current timestamp')
+		}
+
+		return this.buffs.filter(buff =>
```
ackwell: Might be worth consolidating this last bit of logic w/ the `getStatus` method in some way?

src/parser/jobs/ast/About.js
```diff
@@ -1,18 +1,33 @@
...
+import ACTIONS from 'data/ACTIONS'
+
+import CONTRIBUTORS from 'data/CONTRIBUTORS'
+
 import CoreAbout from 'parser/core/modules/About'
 
 export default class About extends CoreAbout {
 	description = <Fragment>
-		<p>This analyser aims to identify some of the low-hanging fruit that could be used to improve your gameplay, as well as (eventually) (hopefully) give deeper insight into optimizing your healing and card usage.</p>
+		<p>The biggest <ActionLink {...ACTIONS.DRAW} /> to an AST is their ability to buff their party DPS with Arcanum.
```
ackwell: I should have you arrested for this.

src/parser/jobs/ast/BuffExtensions.module.css
```diff
@@ -0,0 +1,24 @@
+.buffextensions__headerItem {
```
ackwell: BEM :blobknife:

No need to namespace css classes, the names get mangled to include module info and a hash during build.

src/parser/jobs/ast/BuffExtensions.js
```diff
@@ -0,0 +1,270 @@
...
+	 * @return {void} null
+	 */
+	_onBuffRefresh(event) {
+		const statusID = event.ability.guid
+
+		// Ignore if timestamp is after aoe effect grace period
+		if (this._oppositionEvent && event.timestamp < (this._oppositionEvent.event.timestamp + this._envEffectGracePeriod)) {
+
+			// Ignore refreshes on protects, royal road statuses,
+			if (statusID !== STATUSES.PROTECT.id
```
ackwell: You can consolidate this up top of the file with something like

```js
const IGNORE_STATUSES = [
	STATUSES.PROTECT.id,
	STATUSES.COLLECTIVE_UNCONSCIOUS.id,
	...ROYAL_ROAD_STATES,
	...HELD_ARCANA,
	...DRAWN_ARCANA,
]
```

That way you only need to do the one `IGNORE_STATUSES.includes(statusId)` down here.

src/parser/jobs/ast/BuffExtensions.js
```diff
@@ -0,0 +1,270 @@
...
+	 * Populates the _oppositionEvent created in _onOpposition with refresh events on friendlies
+	 *
+	 * @param {obj} log event of a Celestial Opposition cast
+	 * @return {void} null
+	 */
+	_onBuffRefresh(event) {
+		const statusID = event.ability.guid
+
+		// Ignore if timestamp is after aoe effect grace period
+		if (this._oppositionEvent && event.timestamp < (this._oppositionEvent.event.timestamp + this._envEffectGracePeriod)) {
```
ackwell: You could flatten this function a lot by reversing the `if`s and shortcutting out, i.e.:

```js
if (cond) {
	// ...
}

// vs

if (!cond) { return }
// ...
```

src/parser/jobs/ast/BuffExtensions.js
```diff
@@ -0,0 +1,270 @@
...
+				})
+			}
+		}
+
+
+
+	}
+
+	output() {
+		const panels = this._dilationUses.map(dilation => {
```
ackwell: I'd try to flatten this a little. A lot of logic looping around here.

src/parser/jobs/ast/BuffExtensions.js
suShirow: I really needed this. Always wondered which was the cleaner way to write these

suShirow:
I've updated `AoE.js` to make aoe events for `heal, applybuff, refreshbuff` events. I learned a lot about how AoEs are presented in the log events.

Problem: Then I tried to use `aoerefreshbuff` for my Celestial opposition. I found out that buff events have a longer pulse gap than damage/heal events. I also realized that it returned separate events for each status, which is actually working as intended. It got pretty sticky when I tried to merge them into a single `OnCelestialOpposition` event, because each event now has the pulse thresholds and now we have to worry about correlating them all over again.

Conclusion: The logic is simpler if I just identified the players affected by CO through the `refreshbuff` events, and then checked their statuses for what they had from the Astro in the moments around the CO cast. 
This is accurate to a reasonable degree, the module works, and we can come up with some suggestions derived from the information pulled. 
I just hope the `AoE.js` is working well.

I've spent more than a couple of days on this and would like to move onto Minor Arcanas usage. 
We can come back to this when the implementation causes seeing organs to bleed spontaneously.

PR #71, opened 2018-07-26
Flexible Checklist

Vulcwen:
Job modules can now freely modify:

- success conditions
- how far the progress bar is filled
- The info next to the progress bar (doesn't need to show a percentage anymore)
- Info for requirements (note, you'll have to round percentages yourself if you change it, use `percentage.toFixed(decimals)` for this)
- How the checklist aggregates percentages (default mean)

Question: Do we want score as an additional thing for tracking non-percentage values, or should we just use percentage to track scores, even if it's not really a percentage?

PS: changed the default success check to be equal or larger than the target, rather than larger than the target.

EDIT, look at comment below for updated changes


Vulcwen:
As discussed in discord, simplified the system a lot, and added 2 more modes.
Changes from master (not from the earlier version of the PR):
- An inbetween value for decent results, shown with a yellow progress bar. Specify this by setting decentTarget on the rule. This value defaults to 95, which equals the default of target.
- percent can be set directly on the rule, this controls the progress bar, and is used for success check
- success check is changed to >= (also works this way for decentTarget)
- text may be added to the rule and the requirements, these can take percentage as argument
- displaying percentage can be opted out on (showPercent = false)

Vulcwen:
Last commit dealt with the issue that the `text` property only took strings (or functions that return a string), while the localization effort requires JSX Trans entries to function. You can still use strings if nothing needs to be translated.

Vulcwen:
A certain weird issue makes this version unable to be merged, a fix, made from scratch, is to be provided in a new PR.

PR #76, opened 2018-07-27
More BLM stuff

rexatlanti:


src/parser/jobs/blm/RotationWatchdog.js
```diff
@@ -8,16 +8,15 @@ import {ActionLink} from 'components/ui/DbLink'
...
 import {Suggestion, SEVERITY} from 'parser/core/modules/Suggestions'
 
 const EXPECTED_FIRE4 = 6
 const FIRE4_FROM_CONVERT = 2
 
 const DEBUG_LOG_ALL_FIRE_COUNTS = false
 
-export default class FireCounter extends Module {
-	static handle = 'firecounter'
+export default class rotation extends Module {
```
MikeMatrix: Please name your classes the same as your File, as well as use Pascal case for classes. => `RotationWatchdog`

src/parser/jobs/blm/RotationWatchdog.js
```diff
@@ -133,23 +139,25 @@ export default class FireCounter extends Module {
...
-				this._history.push(this._fireCounter)
-				if (this._lastStop && this._UH > 0 && this._fireCounter.missingCount === 2) {
-					const missedF4s = this._fireCounter.missingCount --
+			if (this._rotation.missingCount.missing > 0 || DEBUG_LOG_ALL_FIRE_COUNTS) {
+				this._rotation.fire4Count = fire4Count
+				if (this._rotation.casts.length > 3) {
+					this._history.push(this._rotation)
+				}
+				if (this._lastStop && this._UH > 0 && this._rotation.missingCount === 2) {
+					const missedF4s = this._rotation.missingCount --
```
MikeMatrix: You can inline this expression

src/parser/jobs/blm/RotationWatchdog.js
```diff
@@ -158,14 +166,15 @@ export default class FireCounter extends Module {
...
-		this._fireCounter = {}
+		this._rotation = {}
 		this._lockedBuffs = false
 	}
 
 	_getMissingFire4Count(count, hasConvert) {
 		const NotEnoughUH = this._UH  < 2
-		const missingFire4 = EXPECTED_FIRE4 + (hasConvert ? FIRE4_FROM_CONVERT : 0) - (NotEnoughUH ? 1 : 0) - count
-		return missingFire4
+		const ExpectedFire4 = EXPECTED_FIRE4 + (hasConvert ? FIRE4_FROM_CONVERT : 0) - (NotEnoughUH ? 1 : 0)
```
MikeMatrix: Please use camel case for variable names => `expectedFire4`

src/parser/jobs/blm/RotationWatchdog.js
```diff
@@ -158,14 +166,15 @@ export default class FireCounter extends Module {
...
+		const missingFire4 = ExpectedFire4 - count
+		return {missing: missingFire4, expected: ExpectedFire4}
```
MikeMatrix: might be cleaner to name the variables the same as the object keys, which would allow you do do something like

```javascript
const expected = [ ... ]
const missing = [ ... ]
return {missing, expected}
```

src/parser/jobs/blm/RotationWatchdog.js
```diff
@@ -193,7 +202,7 @@ export default class FireCounter extends Module {
 					key: 'title-' + rotation.start,
 					content: <Fragment>
 						{this.parser.formatTimestamp(rotation.start)}
-						&nbsp;-&nbsp;{this._renderCount(rotation.fire4Count, rotation.missingCount)} Fire IVs
+						&nbsp;-&nbsp;{this._renderCount(rotation.fire4Count, rotation.missingCount.missing)} / {rotation.missingCount.expected} Fire IVs
```
MikeMatrix: Try to avoid using `&nbsp;` since they tend to make code a bit unreadable.
You could use `<span> - </span>` to the same effect.
It also helps grouping your lines in that into different span groups, which might help later down the line with styling them more easily.

MikeMatrix:
Oh also I think your import in `index.js` is incorrect at this point

src/parser/jobs/blm/RotationWatchdog.js
ackwell: I'm guilty as charged on this :blobsweat:.

I've seen some other projects use `{' - '}` to achieve what I assume this does - thoughts on that approach?

MikeMatrix: Also works fine, but has the downside that it requires both sides of that expression to have a string that's not empty to work

rexatlanti: everything addded now

PR #80, opened 2018-07-28
Switched the Shadow Fang module over to use tiered suggestions

Toastdeib:
Pretty much what it says on the box. I also added a duration property to the Shadow Fang status so there are fewer magic numbers flying around.


src/parser/jobs/nin/ShadowFang.js
```diff
@@ -74,14 +75,19 @@ export default class ShadowFang extends Module {
 		}))
 
 		// Suggestion for DoT clipping
-		const severity = this.getDotClippingSeverity()
-		if (severity) {
-			this.suggestions.add(new Suggestion({
+		const clipping = this.getDotClippingAmount()
+		if (clipping >= CLIP_SECS_PER_MIN_THRESHOLD) {
```
ackwell: You can remove this `if` - `TieredSuggestion` will automatically `SEVERITY.IGNORE` if it doesn't meet the minimum tier.

Toastdeib: Ooh, nice. I'll update that.

PR #81, opened 2018-07-28
{Please use the Auto-Translate Function}

KhloeLeclair:
{Achievements} {Auto-Translation Dictionary}
{Please check it.}

{Recommendations}: {Combine}

This integrates jsLingui into the code base. It adds a language select menu that appears when there is more than one available language. Changing a language happens as soon as the new language file is loaded. Tooltips work. All core strings should be localized.

It's a bit ugly in places as I had to work around some jsLingui behavior that ends up unhelpful when building dynamic content like the project does.

There are new yarn commands for extracting translation keys and compiling translations. Extract will be useful to module developers wanting to localize their work at the same time. All developers should extract before they commit changes if they add new localization keys.

Compile isn't really useful at this point because we're using the lingui webpack loader, but it's there if anyone wants it for some reason.

{I had fun today!} {Good-bye.}


NamelessFaceless:
I think this is good to merge, yeah. Don't see any issue with it.

PR #84, opened 2018-07-29
Moved valid target check into Enemies

ackwell:
This means other modules can utilise the `enemies` module for validity checks, instead of it being solely exposed via the `aoedamage` event.


src/parser/core/modules/Enemies.js
```diff
@@ -43,18 +45,31 @@ export default class Enemies extends Entities {
 
 	getEntities() {
 		// Don't need to init this, getEntity will always be called prior with enough deets
-		return this._enemies
+		return _.filter(this._enemies)
```
MikeMatrix: Do you use this here to remove falsy?

ackwell: Yep

PR #85, opened 2018-07-29
Cleric stance module

xiankai:
So, this is a modified copy-pasted version of @MikeMatrix's handy Fight or Flight module. I may have ended up over-engineering it a wee bit while trying to make it more generic though whoops. Custom handling code beyond x usages per cleric stance window have been shafted into `_getExpectedCastsPerRotation` and `_getActualCastsPerRotation`.

And since cleric stance is so useless as I've discovered in the past couple of days :blobshocked:, some of the text may also require tweaks. Below are some samples:

image

**Suggestions**
image


src/parser/jobs/sch/ClericStance.js
```diff
@@ -0,0 +1,193 @@
...
+					this._getExpectedCastsPerRotation(timestamp, id, count)
+				, 0)
+
+			const actual = Object.values(this._buffRotations).reduce(
+				(total, rotation) => total + Math.min(
+					this._getActualCastsPerRotation(rotation, id)
+				), 0)
+
+			const diff = expected - actual
+			if (diff < 1) {
```
MikeMatrix: No need to drop out early tbh, since `TieredSuggestion` will receive `undefined` value in those cases, which will not display anything.

src/parser/jobs/sch/ClericStance.js
```diff
@@ -0,0 +1,193 @@
...
+
+	_canShadowFlareBeUsed() {
+		const {id, cooldown} = getAction(ACTIONS.SHADOW_FLARE.id)
+
+		const cooldownRemaining = this.cooldowns.getCooldownRemaining(id)
+		if (cooldownRemaining - (STATUSES.CLERIC_STANCE.duration * 1000) < this.gcd.getEstimate()) {
+			// cooldown will come off before CS duration expires
+			// this probably won't be accurate if the CS duration gets extended, idk.
+			return true
+		} if ((cooldown * 1000 - cooldownRemaining) < this.gcd.getEstimate()) {
```
MikeMatrix: Pretty sure this is intended to be a `else if`, but if not, this needs to be in a new line.

src/parser/jobs/sch/ClericStance.js
```diff
@@ -0,0 +1,193 @@
...
+import STATUSES from 'data/STATUSES'
+
+// all these suggestions for CS are ultimately trivial :notlikeblob:
+const defaultSeverityTiers = {
+	1: SEVERITY.MINOR,
+}
+
+// list of things to track for CS usage.
+// can override the suggestions from the default template here
+const EXPECTED_CASTS = [
```
MikeMatrix: Try formatting these a little cleaner and consistent, so they're easier to read.
Along the lines of

```javascript
{
    ...getAction(ACTIONS.CELRIC_STANCE.id),
    name: 'GCD',
    count: 6,
}
```

is probably better, since this will, especially when there is big properties like `content` make keys easier to find.

src/parser/jobs/sch/ClericStance.js
```diff
@@ -0,0 +1,193 @@
...
+		return false
+	}
+
+	_onEnd() {
+		this._buffStart = null
+	}
+
+	_onComplete() {
+		EXPECTED_CASTS.forEach(({id, name, icon, count, why, content, tiers}) => {
+			const expected = Object.keys(this._buffRotations).reduce(
```
MikeMatrix: Minor thing, that could be considered here, since we have now access to it (this is purely optional and for code cleanliness) would be using the lodash reduce, since that one supports objects right out of the gate.

```javascript
_.reduce(this._buffRotations, (total, rotation, timestamp) => [ ... ], 0)
```

src/parser/jobs/sch/ClericStance.js
xiankai: I did have an `else if` there but looks like auto-lint on save removed it. New line it is!

PR #93, opened 2018-07-29
Create a pre-commit linting hook

KhloeLeclair:
This uses the modules `husky` and `lint-staged` to run `eslint --fix` on all staged files match `src/**/*.js`. It should stop people from commiting code with syntax errors and also apply style tweaks for people that don't have their editors doing it on save.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,12 +46,12 @@
     "hooks": {
       "pre-commit": "lint-staged"
     }
-	},
-	"lint-staged": {
-		"linters": {
-			"src/**/*.js": ["eslint --fix", "git add"]
-		}
-	},
+  },
+  "lint-staged": {
+    "linters": {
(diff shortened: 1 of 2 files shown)


KhloeLeclair:
It was pointed out in #development in Discord that `husky` and `lint-staged` were used before and taken out. That's because it was wrongly configured before and ran eslint on the entire `src/` tree. This will **only** lint staged files with `eslint --fix` and, if they're changed via the fix, add the modifications to the commit.

MikeMatrix:
Probably something @ackwell needs to look at.

@ackwell As per discussion in discord, seemingly the last time we had this, it was set up incorrectly and forced a check on all files in src, while this would only check over the files that are staged in src. This should accelerate the check significantly and should also not block for test code that's not getting commited.

PR #99, opened 2018-07-30
added an description override for CooldownDowntime.js

rexatlanti:


src/parser/core/modules/CooldownDowntime.js
```diff
@@ -22,6 +22,7 @@ export default class CooldownDowntime extends Module {
 		this.allowedDowntime = 0
 		//Default Target to hit
 		this.target = 95
+		this.description = <Trans id="core.cooldowndowntime.ogcd-cd-metric">Always make sure to use your OGCD CDs when they are up but don't clip them.  {this.allowedDowntime === 0 ? '' : <Trans id="core.cooldowndowntime.ogcd-cd-buffer">To account for random factors you are given a buffer of {this.allowedDowntime/1000} seconds per instance to use your CD</Trans>}</Trans>
```
Toastdeib: Minor and non-blocking, but I have a couple of phrasing suggestions while you're moving this text around anyway:
1) "OGCD CDs" is redundant (off-global cooldown cooldowns), I would shorten that to just "OGCDs".
2) You should probably use `this.parser.formatDuration()` for the "you are given a buffer of..." part for consistency with other modules.
3) The second sentence should be punctuated (currently missing a period at the end) for consistency. Also, I might change that ending to "...to hold your cooldowns." instead of "...to use your CD".

rexatlanti: I am not the author and wasn't huge fan either of the text which is why I made this PR :blobsweat: I'll address the issue real quick

src/parser/core/modules/CooldownDowntime.js
```diff
@@ -22,6 +22,7 @@ export default class CooldownDowntime extends Module {
 		this.allowedDowntime = 0
 		//Default Target to hit
 		this.target = 95
+		this.description = <Trans id="core.cooldowndowntime.ogcd-cd-metric">Always make sure to use your OGCDs when they are up but don't clip them.  {this.allowedDowntime === 0 ? '' : <Trans id="core.cooldowndowntime.ogcd-cd-buffer">To account for random factors you are given a buffer of {this.parser.formatDuration(this.allowedDowntime/1000)} seconds per instance to hold your cooldowns.</Trans>}</Trans>
```
Toastdeib: Sorry, I should've clarified. `this.parser.formatDuration()` takes milliseconds and includes "s" to denote seconds (e.g. `this.parser.formatDuration(10000)` produces `'10s'`), then uses `m:ss` notation if it exceeds a minute. You'd want to turn this:
```js
{this.parser.formatDuration(this.allowedDowntime/1000)} seconds
```
into this:
```js
{this.parser.formatDuration(this.allowedDowntime)}
```
to utilize it properly.

rexatlanti: ah ok. I thought it took the value in seconds. Didn't doublecheck :blobsweat:

PR #108, opened 2018-07-31
SMN module i18n integrations

ackwell:
* Adds the `TransMarkdown` component, used to translate and then parse markdown strings. Useful for content such as module group `About` descriptions.
* Integrates i18n components throughout most of the SMN module group.


src/parser/jobs/smn/DoTs.js
```diff
@@ -83,20 +84,26 @@ export default class DoTs extends Module {
 		// Checklist rule for dot uptime
 		this.checklist.add(new Rule({
 			name: 'Keep your DoTs up',
```
KhloeLeclair: The name here still needs i18n support.

PR #111, opened 2018-08-01
Jinpu/Shifu Uptime + bug fixes

Rjb99:
-Added Jinpu/Shifu uptime 
-Target uptime is set at 95% based on feedback from Boxer, will be fine tuning this number in the coming days
(invuln does make them quite abit borked)

-Updated About.js to be more about sam then copypasta. 

(has a please note about sen/kenki gen, but that's disabled so it's actually meaningless. :blobsweat:)

-Fixed a bug that made Higanbana be a minor issue at 0.00 seconds clipping.

I've got Gauge.js in this PR because I'm stupid. but it's currently disabled so it won't work anyway.

(because I'm apparently very bad at kenki management)


src/parser/jobs/sam/Buffs.js
```diff
@@ -0,0 +1,136 @@
...
+			}))
+		}
+
+	}
+
+	//
+	getJinpuUptimePercent() {
+		const statusUptime = this.combatants.getStatusUptime(STATUSES.JINPU.id, this.parser.player.id)
+		const fightUptime = this.parser.fightDuration - this.invuln.getInvulnerableUptime()
+
```
NamelessFaceless: You could honestly reduce this from two to one method, by changing it to

`getUptimePercent(statusId)` and passing the `STATUSES.JUNPU.id` as a parameter instead. Same for the `STATUSES.SHIFU.id`.

src/parser/jobs/sam/Gauge.js
```diff
@@ -0,0 +1,208 @@
...
+
+	_onComplete() {
+		//kenki suggestions
+		if (this._wastedKenki >= 20) {
+			this.suggestions.add(new Suggestion({
+				icon: ACTIONS.HAKAZE.icon,
+				content: <Fragment>
+					You used kenki builders in a way that overcapped you.
+				</Fragment>,
+				severity: this._wastedKenki === 20? SEVERITY.MINOR : this._wastedKenki >= 50? SEVERITY.MEDIUM : SEVERITY.MAJOR,
```
NamelessFaceless: Could use `TieredSuggestions` for this instead.

src/parser/jobs/sam/Gauge.js
```diff
@@ -0,0 +1,208 @@
...
+		}
+
+		//sen suggestions
+		if (this._wastedsen >= 1) {
+			this.suggestions.add(new Suggestion({
+				icon: ACTIONS.MEIKYO_SHISUI.icon,
+				content: <Fragment>
+					You used <ActionLink {...ACTIONS.GEKKO}/>, <ActionLink {...ACTIONS.KASHA}/>, <ActionLink {...ACTIONS.YUKIKAZE}/>, at a time when you already had that sen, thus wasting a combo because it did not give you sen.
+				</Fragment>,
+				severity: this._wastedsen === 1? SEVERITY.MINOR : this._wastedsen >= 3? SEVERITY.MEDIUM : SEVERITY.MAJOR,
```
NamelessFaceless: Same as above, `TieredSuggestions`.

src/parser/jobs/sam/Gauge.js
```diff
@@ -0,0 +1,208 @@
...
+	_Sen2Kenki() {
+		this._kenki += ((this._gekkosen + this._kashasen + this._yukikazesen) * 20)
+		if (this.kenki > MAX_KENKI) {
+			const waste = this._kenki - MAX_KENKI
+			this._wastedKenki += waste
+			this._kenki = MAX_KENKI
+			return waste
+		}
+
+		this._gekkosen = 0
```
MikeMatrix: `_Sen2Kenki()` converts current sen to a kenki value and adds that, then removes all current sen.
Meaning this really should call `_removeSen()` as the last call instead of handling that logic itself.

src/parser/jobs/sam/Buffs.js
```diff
@@ -0,0 +1,136 @@
...
+	constructor(...args) {
+		super(...args)
+
+		const filter = {
+			by: 'player',
+			abilityId: [STATUSES.JINPU.id, STATUSES.SHIFU.id],
+		}
+		this.addHook('applybuff', filter, this._onShifuApplication, this._onJinpuApplication)
+		this.addHook('refreshbuff', filter, this._onShifuApplication, this._onJinpuApplication)
```
MikeMatrix: I'm like 90% sure this doesn't actually call `_onJinpuApplication` ever...

NamelessFaceless: I'm pretty sure you need a hook per buff at that point.

MikeMatrix: Yea I think signature of `addHook` is 
```javascript
addHook(events, filter, cb) {
```

Which means the last callback there would be ignored...

Actually maybe a future improvement?
Thoughts @ackwell?

ackwell: Yeah the sig is

```js
addHook(events[, filter], cb)
```

I'm hesitant to allow multiple callbacks per event. The snipping we're commenting on could theoretically be written as

```js
this.addHook(['applybuff', 'refreshbuff'], filter, event => {
	this._onShifuApplication(event)
	this._onJinpuApplication(event)
})
```

You could take it a step further and combine the two callbacks, too - they're doing the same thing.

MikeMatrix: Fair enough. I was just thinking it might be an interesting pattern to consider, splitting the logic in that manner.
But just using a combine function is suitable enough for that...
Maybe provide that boilerplate as a helper function?

src/parser/jobs/sam/About.js
```diff
@@ -1,12 +1,17 @@
...
+
 import CONTRIBUTORS, {ROLES} from 'data/CONTRIBUTORS'
+
 import CoreAbout from 'parser/core/modules/About'
+//import {ActionLink} from 'components/ui/DbLink'
 
 export default class About extends CoreAbout {
 	description = <Fragment>
-		<p>This analyzer aims to identify some of the low-hanging fruit that could be used to improve your SAM gameplay, as well as give a deeper insight into what happened during an encounter.</p>
+		<p> So you study the blade do you? Well consider this analysis the exam to see exactly how much you've learned about the basics of Samurai. This tool will track your sen and kenki gains/uses to see if you are missing possible resources to gain or you have failed to make the most out of what you gained over the course of the fight.</p>
```
ackwell: Don't forget to i18nify this when you get a chance :blobmorning:

Doesn't need to be done for this one, though.

src/parser/jobs/sam/Buffs.js
```diff
@@ -0,0 +1,132 @@
...
+
+	constructor(...args) {
+		super(...args)
+
+		const filter = {
+			by: 'player',
+			abilityId: [STATUSES.JINPU.id, STATUSES.SHIFU.id],
+		}
+		this.addHook(['applybuff', 'refreshbuff'], filter, event => {
+			this._onShifuApplication(event)
```
ackwell: This is still listening for jinpu _or_ shifu, and then running clip checks for _both_ - probably artificially inflating clip amount. Using existing code, this'll need to be more along the lines of

```js
this.addHook(['applybuff', 'refreshbuff'], {by: 'player', abilityId: STATUSES.JINPU.id}, this._onJinpuApplication)
this.addHook(['applybuff', 'refreshbuff'], {by: 'player', abilityId: STATUSES.SHIFU.id}, this._onShifuApplication)
```

src/parser/jobs/sam/Gauge.js
```diff
@@ -0,0 +1,208 @@
...
+	_wastedsen = 0
+
+	constructor(...args) {
+		super(...args)
+		this.addHook('cast', {by: 'player'}, this._onCast)
+		this.addHook('death', {to: 'player'}, this._onDeath)
+		this.addHook('complete', this._onComplete)
+	}
+	//check for kenki value changes, then sen changes
+	_onCast(event) {
```
ackwell: This would almost be cleaner as multiple hooks, but I'm not sure.

src/parser/jobs/sam/Gauge.js
Toastdeib: I'd say should, not just could. Or at the very least fix the bug that flags the 21-49 range as `MAJOR` and the 50+ range as `MEDIUM`. :blobmorning:

src/parser/jobs/sam/Gauge.js
```diff
@@ -0,0 +1,208 @@
...
+		}
+		if (abilityId === ACTIONS.YUKIKAZE.id) {
+			this._addYukikazeSen()
+		}
+
+	}
+
+	//kenki quick maths
+
+	_addKenki(abilityId) {
```
Toastdeib: Certainly not a blocker, but you may want to remove the return values from the Kenki helper functions since they all appear to be unused. They don't hurt anything, it just reduces code clarity.

src/parser/jobs/sam/Gauge.js
```diff
@@ -0,0 +1,208 @@
...
+
+		return 0
+	}
+
+	//sen calcs
+
+	_addGekkoSen() {
+		this._gekkosen += 1
+		if (this._gekkosen > MAX_GEKKO_SEN) {
+			const waste = this._gekkosen - MAX_GEKKO_SEN
```
Toastdeib: I think you'll want to cap `this._gekkosen` down to `MAX_GEKKO_SEN` after calculating waste here (and same goes for Kasha/Yukikaze). Otherwise, it'll just compound. Like, if someone used 3 Gekko finishers in a row, the first would be fine, the second would be +1 waste, and the third would be +2 waste. Also it'll throw off the Sen -> Kenki calculation, since you'd be getting 60 out of Gekko alone in this scenario.

src/parser/jobs/sam/Buffs.js
```diff
@@ -0,0 +1,126 @@
...
+		super(...args)
+
+		this.addHook(['applybuff', 'refreshbuff'], {by: 'player', abilityId: STATUSES.JINPU.id}, this._onJinpuApplication)
+		this.addHook(['applybuff', 'refreshbuff'], {by: 'player', abilityId: STATUSES.SHIFU.id}, this._onShifuApplication)
+
+		this.addHook('complete', this._onComplete)
+	}
+
+	_onJinpuApplication(event) {
+		this._JinpuUses.unshift(event)
```
ackwell: Not going to block, but `unshift` is a peculiar choice.

src/parser/jobs/sam/Buffs.js
```diff
@@ -0,0 +1,126 @@
...
+
+				new Requirement({
+					name: <Fragment><ActionLink {...ACTIONS.SHIFU} /> uptime</Fragment>,
+					percent: () => this.getUptimePercent(STATUSES.SHIFU.id),
+				}),
+
+			],
+		}))
+
+		if (this._earlyJinpuApplications) {
```
ackwell: These can likely be combined into one suggestion in future revisions

src/parser/jobs/sam/Gauge.js
```diff
@@ -0,0 +1,214 @@
...
+		}
+
+		this._removeSen()
+
+		return 0
+	}
+
+	//sen calcs
+
+	_addGekkoSen() {
```
ackwell: These func can probably be combined relatively easily in future revisions.

PR #112, opened 2018-08-01
Speedmod

LaqiThish:
Speedmod:
- Uses lookup for combined buffs. Formula being used previously is proven to be wrong
- Extended by WHM, BLM, MNK to handle job-specific speed buffs
- Filled in MNK version to handle Greased Lightning II/III and Riddle of Fire

SpeedmodConsts:
- Big lookup table of datamined magic numbers for all buff combinations. Would really prefer this all stays in one place

GlobalCooldown:
- Estimates more-correctly for BLM/casters. Still needs BLM-specific work


ackwell:
If you set the `cooldown` of meditation to `1.2`, then the `MIN_GCD` in the calcs could be changed to `Math.min(MIN_GCD, action.cooldown)`. That way anything sub-1.5 will be essentially locked to its cooldown, which I'm pretty sure reflects the way it works in-game.

ayyaruq:
That works, I thought `cooldown` was only valid for non-GCD skills.

src/parser/core/modules/GlobalCooldown.js
```diff
@@ -133,7 +141,14 @@ export default class GlobalCooldown extends Module {
 		}
 
 		// Calculate the lengths of the GCD
-		const lengths = this.gcds.map(gcd => gcd.length)
+		// TODO: Ideally don't explicitly check only instants and 2.5s casts. Being able to use 2.8s casts would give tons more samples to consider for more accurate values
```
ackwell: Copying from discord for posterity:

>just thinking
>the `getUptime`/checklist entries are multiplying the recorded GCD length by a cooldown ratio
>would you be able to reverse that when saving >2.5s GCDs?
>that way stuff like F4 would get normalised to the 2.5s GCD length

ackwell: Nah, just thoughts for future improvements.

src/parser/core/modules/SpeedmodConsts.js
```diff
@@ -0,0 +1,155 @@
+import STATUSES from 'data/STATUSES'
```
ackwell: This file is :blobknife:, but I realise lookup tables are basically the only accurate info we've got right now.

Looking forward to a proper formula to replace all this ;-;

PR #115, opened 2018-08-01
Intl polyfill

### LINKED ISSUE #113: Add Intl polyfill
Fair few errors sliding through from mobile safari.

https://sentry.io/xivanalysis/client/issues/624517900/

```
ReferenceError: Can't find variable: Intl
  at a (./locale/en/messages.json?b888:1:4299)
  at call (components/Analyse/Analyse.js:195:8)
  at i (components/Analyse/Analyse.js:195:28)
  at handler (components/Analyse/Analyse.js:195:28)
...
(22 additional frame(s) were not displayed)
```

ackwell:
Also a fix for smn text because rolling two disparate things together is totally my style.

Should resolve #113

### DIFF
--- a/src/components/I18nLoader.js
+++ b/src/components/I18nLoader.js
@@ -38,7 +38,8 @@ export class I18nLoader extends Component {
 		)]
 
 		// Polyfill
-		if (!window.Intl) {
+		const needsPolyfill = !window.Intl
+		if (needsPolyfill) {
 			promises.push(
 				import(
 					/* webpackMode: 'lazy' */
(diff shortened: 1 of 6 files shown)


src/components/I18nLoader.js
```diff
@@ -36,6 +36,20 @@ export class I18nLoader extends Component {
 			`../../locale/${language}/messages.json`
 		)
 
+		if (!window.Intl) {
+			await import(
+				/* webpackMode: 'lazy', webpackChunkName: 'intl-polyfill' */
+				'intl'
+			)
+			// TODO: This is also including `kde` and I've got no idea how to get rid of it
+			await import(
```
KhloeLeclair: You're waiting for three chunks to be pulled in, in sequence. You should probably be doing an `await Promise.all(...)` with the three separate import statements to let the client load all chunks in parallel. Faster client load times. You can then grab the catalog we want from the array returned from Promise.all.

ackwell: Ah shit yeah.

ackwell:
Pushed up a fix, it should now load our `messages.json` and the `intl` polyfill (if required) at the same time.

Unfortunately, I can't include the polyfill locale data in with that grouping, as it makes calls to the main polyfill at import time.

PR #119, opened 2018-08-02
Rewrite <DbLink /> to ditch the XIVDB Tooltip JS Library

KhloeLeclair:
This PR implements `<DbLink />` in React. It performs better and updates more reasonably when something like the current language changes. It:

* Creates a XIVDBTooltipProvider class that handles debouncing and caching requests to be nice to XIVDB's API.
* Rewrites the `<DbLink />` component to use that provider and to format its own output.
* Removes the script tag loading the existing XIVDB tooltip JS library.

Behavior is slightly different. Rather than following your mouse cursor, the tool-tip that appears (which is implemented via Semantic UI's `<Popup />`) has a static position somewhere around the link that it will be fully visible.

### DIFF
--- a/src/components/ui/DbLink.js
+++ b/src/components/ui/DbLink.js
@@ -1,114 +1,235 @@
 import PropTypes from 'prop-types'
-import React, {Component} from 'react'
+import React, {PureComponent} from 'react'
 import {connect} from 'react-redux'
-import Observer from 'react-intersection-observer'
+import {Popup, Icon} from 'semantic-ui-react'
+import {Trans} from '@lingui/react'
+
+import _ from 'lodash'
+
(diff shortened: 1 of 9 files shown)


src/components/ui/DbLink.js
```diff
@@ -118,11 +260,9 @@ const Wrapped = connect(state => ({
 
 export default Wrapped
 
-// -----
-// Helpers 'cus i'm lazy
-// -----
-export const ActionLink = (props) => <Wrapped type="action" {...props} />
-export const StatusLink = (props) => <Wrapped
+// Helpers, because ack is lazy.
```
ackwell: :blobknife:

PR #120, opened 2018-08-02
Tweaks to event filter used in invul period guesses

ackwell:
This PR tweaks the event filter used when guessing invuln periods.

- Fixed stupid mistake that was preventing heal filter from working
- Expanded heal to _exclude_ self-heals
- Added `begincast` event filter that uses same check as heal

These changes do make it a little _too_ lax about downtime, but it's the difference between being hyper trigger happy and a _bit_ too lax, which... well, it's better, i guess.

**Before (Guardian):**

**After (Guardian):**

### DIFF
--- a/src/parser/core/modules/Invulnerability.js
+++ b/src/parser/core/modules/Invulnerability.js
@@ -4,7 +4,7 @@ import {Item} from 'parser/core/modules/Timeline'
 // Ms of inactivity before an enemy is considered invuln
 const MIN_INVULN_INACTIVITY = 5000
 
-const ACTIVITY_EVENT_TYPES = ['damage', 'heal']
+const ACTIVITY_EVENT_TYPES = ['begincast', 'damage', 'heal']
 
 export default class Invulnerability extends Module {
 	static handle = 'invuln'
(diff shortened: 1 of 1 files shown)


src/parser/core/modules/Invulnerability.js
```diff
@@ -61,13 +61,22 @@ export default class Invulnerability extends Module {
 
 		for (let i = 0; i < events.length; i++) {
 			const event = events[i]
+			if (event.type === 'begincast') {
+				console.log(event)
```
MikeMatrix: Bad Boy :thonkop:

src/parser/core/modules/Invulnerability.js
```diff
@@ -62,12 +62,18 @@ export default class Invulnerability extends Module {
...
+				// Make sure it's got all the info we care about, and we're tracking it
 				!ACTIVITY_EVENT_TYPES.includes(event.type) ||
-				(event.type === 'damage' && event.targetIsFriendly) ||
-				(event.type === 'heal' && event.sourceIsFriendly)
+				!event.targetID ||
+				!event.sourceID ||
+				// we only care about casts and heals by players onto something other than themselves
+				(
+					['begincast', 'heal'].includes(event.type) &&
+					(!event.sourceIsFriendly || event.sourceID === event.targetID)
```
ackwell: Hm, on thinking about it a bit, this will possibly allow heals from healers onto current target to prevent downtime. Hold off merging, I'll run some tests in a bit.

PR #121, opened 2018-08-03
Combos

Toastdeib:
For use with any jobs that have combos. Except MNK because forms are fucky. The core module provides hooks for `comboHit()` (invoked every time a successful combo action is executed) and `addJobSpecificSuggestions()` (invoked during `_onComplete()`, allows subclasses to provide job-specific suggestion content in addition to the generic one).

### DIFF
--- a/src/data/ACTIONS/ROG.js
+++ b/src/data/ACTIONS/ROG.js
@@ -8,6 +8,10 @@ export default {
 		name: 'Spinning Edge',
 		icon: 'https://secure.xivdb.com/img/game/000000/000601.png',
 		onGcd: true,
+		potency: 150,
+		combo: {
+			start: true,
+		},
 	},
 
 	GUST_SLASH: {
@@ -15,6 +19,11 @@ export default {
 		name: 'Gust Slash',
 		icon: 'https://secure.xivdb.com/img/game/000000/000602.png',
 		onGcd: true,
+		potency: 100,
+		combo: {
+			from: 2240,
+			potency: 200,
+		},
 	},
 
 	AEOLIAN_EDGE: {
@@ -22,6 +31,12 @@ export default {
 		name: 'Aeolian Edge',
 		icon: 'https://secure.xivdb.com/img/game/000000/000605.png',
 		onGcd: true,
+		potency: 100,
+		combo: {
+			from: 2242,
+			potency: 340, // TODO - *Cries in positionals*
+			end: true,
+		},
 	},
 
 	SHADOW_FANG: {
@@ -29,6 +44,12 @@ export default {
 		name: 'Shadow Fang',
 		icon: 'https://secure.xivdb.com/img/game/000000/000606.png',
 		onGcd: true,
+		potency: 100,
+		combo: {
+			from: 2242,
+			potency: 200,
+			end: true,
+		},
 	},
 
 	DEATH_BLOSSOM: {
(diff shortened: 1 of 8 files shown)


src/parser/jobs/nin/NinCombos.js
```diff
@@ -0,0 +1,13 @@
+import ACTIONS from 'data/ACTIONS'
+import Combos from 'parser/core/modules/Combos'
+
+export default class NinCombos extends Combos {
```
ackwell: Preference renaming the parent at import time. Instead of `nin/NinCombos`, you can have `nin/Combos` by doing:

```js
import CoreCombos from 'parser/core/modules/Combos'
export default class Combos Extends CoreCombos {
}
```

Oh, and the parent class already has the handle, no need to re-declare.

src/parser/core/modules/Combos.js
```diff
@@ -0,0 +1,118 @@
...
+	_lastGcdTime = this.parser.fight.start_time
+	_brokenComboCount = 0
+	_uncomboedGcdCount = 0
+
+	_comboBreakers = []
+	_uncomboedGcds = []
+
+	constructor(...args) {
+		super(...args)
+		this.addHook('cast', {by: 'player'}, this._onCast) // TODO - Can we filter to onGcd=true? That would be neat.
```
ackwell: Not really, it's not part of the event data. Maybe something to look into normalising in but don't hold your breath.

src/parser/core/modules/Combos.js
```diff
@@ -0,0 +1,118 @@
...
+			// Combo breaking action, that's a paddlin'
+			this._lastAction = NO_COMBO
+			this._brokenComboCount++
+			this._comboBreakers.push(event)
+		}
+	}
+
+	_onComplete() {
+		if (this._brokenComboCount > 0 || this._uncomboedGcdCount > 0) {
+			const why = [
```
ackwell: Soz, gonna block on this - going to try and keep `core` as close to fully i18nified as possible.
This `why` will be particularly annoying to implement, I'd bet - might need to simplify the text generation logic a bit to keep the i18n impl sane.

src/parser/core/modules/Combos.js
Toastdeib: Eh, not really worth normalising. I'll delete the TODO.

src/parser/jobs/nin/NinCombos.js
Toastdeib: Noted, will update.

src/data/ACTIONS/ROG.js
```diff
@@ -8,41 +8,66 @@ export default {
...
 
 	AEOLIAN_EDGE: {
 		id: 2255,
 		name: 'Aeolian Edge',
 		icon: 'https://secure.xivdb.com/img/game/000000/000605.png',
 		onGcd: true,
+		potency: 100,
+		combo: {
+			from: 2242,
+			potency: 340, // TODO - *Cries in positionals*
```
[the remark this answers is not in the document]
Toastdeib: /ritualprayer

src/parser/core/modules/Combos.js
Toastdeib: S'all good, I'll slap this junk into a <Trans> tag. Not going to bother with NIN quite yet, though - I'll tackle all of that at once in a separate PR.

PR #127, opened 2018-08-03
Initial RDM Procs - Verstone, Verfire, Impactful

Cephid:
Implementation of Proc tracking for RDMs.

Tracks Missed Verstone, Impactful, Verfire, Enhanced Scatter Procs
Tracks Verstone, Verfire, Impactful, Enhanced Scatter Procs used against a boss while a boss was invuln
Tracks Overwritten Verstone, Verfire, Impactful Procs
Handles for not counting a Verstone or Verfire as overwritten if Impact was cast to trigger the overwrite and Impactful was <= 8 seconds in duration at the time of cast.

This uses tiered Suggestions, but I did not put the Trans tags in yet, I'll do another push later this week with Trans tags added to all existing RDM modules.

I am specifically looking for feedback on the messages, as I'm not 100% happy with them, but couldn't come up with anything better.

Also updated the Notes file, to represent what's done so far.


src/parser/jobs/rdm/Procs.js
```diff
@@ -0,0 +1,322 @@
...
+		this.addHook('complete', this._onComplete)
+	}
+
+	_onCast(event) {
+		const abilityID = event.ability.guid
+		const invuln = this.downtime.getDowntime(this._previousCast||0, event.timestamp)
+		this._previousCast = event.timestamp
+		this._bossWasInvuln = invuln > 0
+
+		switch (abilityID) {
```
MikeMatrix: I think this could be hugely simplified, by just going to use the ACTION id instead of this custom enum.
I don't see any reason for the enum to be there quite honestly.

To simplify this, you could make a bunch of fall through switch cases, then assign the cast state being the casts id, defaulting to setting it to null and making special treatment to impact.

src/parser/jobs/rdm/Procs.js
```diff
@@ -0,0 +1,322 @@
...
+			this._castState = CAST_STATE.NA
+			break
+		}
+	}
+
+	_onGain(event) {
+		const statusID = event.ability.guid
+
+		switch (statusID) {
+		case STATUSES.VERSTONE_READY.id:
```
MikeMatrix: This could likely be simplified by instead having a map from STATUS -> some counter object combining a lot of this, which would generalize this entire switch statement.

src/parser/jobs/rdm/Procs.js
```diff
@@ -0,0 +1,322 @@
...
+		//Nothing to see here
+			break
+		}
+	}
+
+	_onRemove(event) {
+		const statusID = event.ability.guid
+
+		switch (statusID) {
+		case STATUSES.VERSTONE_READY.id:
```
MikeMatrix: Similar to here, what I wrote on the `_onGain` switch statement.

src/parser/jobs/rdm/Procs.js
Cephid: Note - no language I've worked on prior has allowed fall through switch cases, so I'm not sure what it's functional application is.

In this particular instance, I'm not certain what you would fall through to though, each case beyond impactful is only setting the current cast state so I can accurately determine what was cast immediately prior to a buff being removed, the buff removing happens in a different event so I wouldn't be able to query the onRemove event for the spell that was cast to trigger the removal.  I could remove the enum, I prefer working with subsets, but not opposed to working with the full set.  Or are you stating I should just ditch the State capture and capture the previously cast ability ID and then query it from ACTIONS in my other functions?

src/parser/jobs/rdm/Procs.js
Cephid: If I understand you correctly soemthing like below?  Or are you thinking something more akin to property iteration of this._currentProcs with some mapping to know which overWritten property to update?  Or as a tertiary option instead of just a list of counters are you wanting more of an array of occurrences with the type of occurrence (Overwrite, Missed, etc) and the StatusID so I know which spell it belonged to?

STATUSES.map(ID => {
//Just copy the if statements current in cases and string them here, removing the switch/cast lines?
})

MikeMatrix: Something along the lines of

```javascript
const _missedProcs = {};
[...]

const missed = _currentProcs[statusId] > 0 && !this._impactfulProcOverride
_missedProcs[statusId] = (_missedProcs[statusId] || 0) + (missed ? 1 : 0)
```

src/parser/jobs/rdm/Procs.js
MikeMatrix: Something like this, although honestly I'd just not do a switch statement and make some logic to ignore other `_castState` values and make the impact stuff just an if block.

```javascript
switch(actionId) {
case ACTIONS.IMPACT.id:
    this._castState = actionId
    // [...] Impact specific stuff that you already have
    break
case ACTIONS.VERSTONE.id:
case ACTIONS.VERFIRE.id:
case ACTIONS.VERTHUNDER.id:
case ACTIONS.VERAREO.id: // is this a typo in the ACTIONS map?
case ACTIONS.SCATTER.id:
    this._castState = actionId
    break
default:
    this._castState = null
    break
}
```

src/parser/jobs/rdm/Procs.js
```diff
@@ -0,0 +1,322 @@
...
+	}
+
+	_onComplete() {
+		this.verifyVerStone()
+		this.verifyVerFire()
+		this.verifyImpactful()
+		this.verifyScatter()
+	}
+
+	verifyVerStone() {
```
Toastdeib: verVerStone for short? :blobmorning:

src/parser/jobs/rdm/Procs.js
```diff
@@ -0,0 +1,169 @@
...
+			// console.log(util.inspect(this._history[statusID], {showHidden: true, depth: null}))
+			this.generateSuggestions(statusID)
+		}
+	}
+
+	generateSuggestions(statusID) {
+		if (this._history[statusID].missed) {
+			this.suggestions.add(new TieredSuggestion({
+				icon: this._castStateMap[statusID].icon,
+				why: `${this._history[statusID].missed} ${this._castStateMap[statusID].name}${this._history[statusID].missed !== 1 ? 's' : ''} casts missed due to buff falling off`,
```
ackwell: For future PRs - might be worth  pulling a lot of these lookups out into a few consts at the start of the function. Will shorten up the code a lot, and make the i18n strings look nicer when that gets implemented.

PR #132, opened 2018-08-07
Feature/earthlystar

suShirow:
Tiny tiny module to count the number of uncooked Earthly Star they made, then spit out a suggestion. Hoping for more detail on Earthly in future sprints.
Test log: `KZX4CkrBwvntMT3A/6/35/`


src/parser/jobs/ast/DoTs.js
```diff
@@ -87,9 +87,9 @@ export default class DoT extends Module {
 			this.suggestions.add(new Suggestion({
 				icon: ACTIONS.COMBUST_II.icon,
 				content: <Fragment>
-					Avoid refreshing <ActionLink {...ACTIONS.COMBUST_II} /> significantly before it expires. Aim to refresh it between 2 to 0 seconds remaining on the duration.
+					Avoid refreshing <ActionLink {...ACTIONS.COMBUST_II} /> significantly before it expires.
 				</Fragment>,
-				severity: maxClip < 10000? SEVERITY.MINOR : maxClip < 30000? SEVERITY.MEDIUM : SEVERITY.MAJOR,
+				severity: maxClip < 10000? SEVERITY.MINOR : maxClip < STATUS_DURATION[ACTIONS.COMBUST_II.id] ? SEVERITY.MEDIUM : SEVERITY.MAJOR,
```
NamelessFaceless: Non-blocking, and I'll probably merge this in already -- But I'd probably try to use a `const` instead of `10000`.

NamelessFaceless: Non-blocking, and I'll probably merge this in already -- But I'd probably try to use a `const` instead of `10000`.

PR #134, opened 2018-08-08
RDM gauge graph

ackwell:
Currently no support for zoom/pan, the chartjs plugin that handles it is trash and my attempts at fixing it were dodgy at best.

Core UI module added to centralise the config for this sort of thing, should make it a bit easier to deal with / adjust them down the line.

PR also fixes RDM gauge to track deaths because hey graphs make this noticeable.

### DIFF
--- a/src/components/ui/TimeLineChart.js
+++ b/src/components/ui/TimeLineChart.js
@@ -2,27 +2,27 @@ import PropTypes from 'prop-types'
 import React, {PureComponent} from 'react'
 import {Line} from 'react-chartjs-2'
 
+const DEFAULT_OPTIONS = {
+	scales: {
+		xAxes: [{
+			type: 'time',
+			time: {
+				displayFormats: {
+					minute: 'm:ss',
+					second: 'm:ss',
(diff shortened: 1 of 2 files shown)


src/components/ui/TimeLineChart.js
```diff
@@ -0,0 +1,28 @@
...
+import React, {PureComponent} from 'react'
+import {Line} from 'react-chartjs-2'
+
+export default class TimeLineChart extends PureComponent {
+	static propTypes = {
+		data: PropTypes.object.isRequired,
+	}
+
+	render() {
+		const options = {
```
MikeMatrix: Should probably be either a const outside the class or a member of the class instead of being re-eval'd every render step (potentially causing a rerender down the line because it's a new instance).

ackwell: mm, point

ackwell:
I feex.

Was a left-over from computed options for min/max on the zoom/pan, but _damn_ that plugin was jank.

PR #139, opened 2018-08-09
Added a module that keeps track of your Ley Line usage

rexatlanti:
Ley Line usage in percent is tracked (aka "how much time am I really in that circle?") and translation is added too. (Still can't `yarn extract`sadly)


src/parser/jobs/blm/Leylines.js
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	_onCastLeyLines(event) {
+		this.ll = this._LeyLineHistory
+		this.ll.push(event.timestamp)
+
+	}
+
+	_onApplyCircleOfPower(event) {
+		this.cop = this._CircleOfPowers
```
Toastdeib: What's the reasoning for the assignments here and at the end of the function? I'm pretty sure this is no different from manipulating `this._CircleOfPowers` directly - the assignment should be by reference, so modifying `this.cop` should also modify `this._CircleOfPowers` at the same time.

src/parser/jobs/blm/Leylines.js
```diff
@@ -0,0 +1,100 @@
...
+
+		this.cop.current = {
+			start: event.timestamp,
+			stop: null,
+		}
+		this._CircleOfPowers = this.cop
+	}
+
+	_onRemoveCircleOfPower(event) {
+		this.cop = this._CircleOfPowers
```
Toastdeib: Same as above.

src/parser/jobs/blm/Leylines.js
```diff
@@ -0,0 +1,100 @@
...
+
+export default class Leylines extends Module {
+	static handle = 'leylines'
+	static i18n_id = i18nMark('blm.leylines.title')
+	static title = 'Ley Lines'
+
+	static dependencies = [
+		'checklist',
+	]
+	_CircleOfPowers = {
```
Toastdeib: Non-blocker, but I'd recommend standardizing your variable style to keep things more clean and consistent. Half of them seem to start with upper case letters (e.g. this one) and the other half with lower (e.g. `const circleOfPowerDurations` down in `_onComplete()`). Lower is my personal preference, but all I really care about is that you adhere to a standard, whichever it may be. I put the comment here since it's the first non-static variable in the file, but it applies to all of them going down from here (including some function params).

src/parser/jobs/blm/Leylines.js
```diff
@@ -0,0 +1,100 @@
...
+		}
+		this._CircleOfPowers = this.cop
+	}
+
+	_percentFunction(numberOfLeyLines, SumOfCoPUpime) {
+		return (SumOfCoPUpime/(numberOfLeyLines*LEYLINE_DURATION))*100
+	}
+
+	_onComplete(event) {
+		this.cop = this._CircleOfPowers
```
Toastdeib: And once more here.

src/parser/jobs/blm/Leylines.js
```diff
@@ -0,0 +1,100 @@
...
+		if (!this.cop.history.length) { return }
```
Toastdeib: This check should probably be after the one below it - if you hit some inane edge case where the only time they were in Ley Lines was at the very end of the fight, you'd have a value in `current` to add into history, but you'd short circuit before you got the chance because there were no _other_ history events.

Alternatively, it might be worth a warning in and of itself - no Circle of Power history implies that the player didn't use Ley Lines once in the entire encounter, which is a pretty big mistake to be making.

src/parser/jobs/blm/Leylines.js
```diff
@@ -0,0 +1,100 @@
...
+		if (this.cop.current) {
+			if (!this.cop.current.stop) {
```
Toastdeib: This one's a non-blocker, but - if I'm understanding it correctly, I think you could invert this snippet of logic here so that it only pushes to history if the current event _does_ have a stop time, which should remove the need for the `history.filter()` line below, since you wouldn't be pushing CoP events that hit the end of the fight in the first place.

src/parser/jobs/blm/Leylines.js
```diff
@@ -0,0 +1,100 @@
...
+				this.cop.current.stop = event.timestamp
+			}
+			this.cop.history.push(this.cop.current)
+		}
+		//filter out the last possible LL usage because it would make things weird.
+		this.cop.history.filter(cops => cops.start < (event.timestamp - LEYLINE_DURATION))
```
Toastdeib: These two `.filter()` calls are actually doing nothing - they return a filtered array rather than doing it in place, so you'd have to self-assign the result (e.g. `this.cop.history = this.cop.history.filter(...)`).

src/parser/jobs/blm/Leylines.js
```diff
@@ -0,0 +1,100 @@
...
+		this._LeyLineHistory.filter(timestamps => timestamps < (event.timestamp - LEYLINE_DURATION))
```
Toastdeib: I'm not sure what leaving them in does to your results, but filtering out LL events that run into the end of the fight have the potential to give you >100% uptime results. If, say, you had 100% uptime on all your mid-fight LL casts and cast your last one with ~20s left in the fight, stood in it for 15s, then moved out, that CoP event wouldn't get cut but the LL would, leaving you with more time in CoP than time with LL active. Not a super major issue since it's kind of an edge case, but worth considering.

src/parser/jobs/blm/Leylines.js
rexatlanti: that one is for laziness. cop is less to write : ^)

src/parser/jobs/blm/Leylines.js
rexatlanti: You are right with the first part. We do have more of the issue that CoP simmply isn't tracked in fflogs sometimes

src/parser/jobs/blm/Leylines.js
rexatlanti: I also filter out all CoP events (given that the filter works) for exactly that reason

src/parser/jobs/blm/Leylines.js
Toastdeib: In that case, I'd recommend either doing `const cop` rather than `this.cop` to shorthand it or just rename the actual `_CircleOfPowers` variable in the first place so you don't feel a need to shorthand it at all.

src/parser/jobs/blm/Leylines.js
Toastdeib: Ah, but that's the thing - you're only filtering CoP events that hit the end of the fight. If one of them ends 5s before the fight ends, it won't get cut (which is the scenario I outlined above).

src/parser/jobs/blm/Leylines.js
```diff
@@ -0,0 +1,93 @@
...
+
+export default class Leylines extends Module {
+	static handle = 'leylines'
+	static i18n_id = i18nMark('blm.leylines.title')
+	static title = 'Ley Lines'
+
+	static dependencies = [
+		'checklist',
+	]
+	_circleOfPowers = {
```
ackwell: Circle**s** of Power.

English :tm:

PR #142, opened 2018-08-11
Kassatsu module for NIN

Toastdeib:
On today's episode of "how to not be a bad NIN", we have a module for tracking Kassatsu use. It'll flag any non-Fuma/Raiton Ninjutsus used under Kassatsu with varying degrees of badness, as well as watching for Kassatsu falling off naturally without getting spent on a Ninjutsu cast at all.


src/parser/jobs/nin/Kassatsu.js
```diff
@@ -0,0 +1,174 @@
...
+		if (!this._kassatsuSpent) {
+			this._kassatsuWastes++
+		}
+
+		// Reset the flag for the next time it's cast
+		this._kassatsuSpent = false
+	}
+
+	_onComplete() {
+		if (this._kassatsuWastes > 0) {
```
ackwell: Not gonna block on it, but this seems like a _lot_ of potential suggestion spam if someone's dicking up their kassatsus.

Is it feasible to consolidate them a bit, or would it make the suggestions themselves meaningless? Iuno enough about NIN to say much.

Toastdeib: I was hemming and hawing over keeping them grouped as a single suggestion, but I broke them out for two reasons - varying severity and ease of translation. Like, using Kassatsu on Suiton is way less bad than using it on Huton (especially if setting up a TA so they don't fall off schedule). I could maybe consolidate some of them, but in practice I wouldn't expect more than one, maybe two of these suggestions to show per log, and if more than two _do_ show, the player really needs to get their shit together.

PR #143, opened 2018-08-11
Add MNK DoT/Gauge uptime and basic AoE checks

ayyaruq:
The change for the TimeChart should probably merge the default but idk Javascript. There's a bunch of TODOs in the MNK job modules, I mean to fix those in later PRs so this one should be good to go.

### DIFF
--- a/src/parser/jobs/mnk/Demolish.js
+++ b/src/parser/jobs/mnk/Demolish.js
@@ -1,0 +1,100 @@
+import React, {Fragment} from 'react'
+
+import {ActionLink, StatusLink} from 'components/ui/DbLink'
+import ACTIONS from 'data/ACTIONS'
+import STATUSES from 'data/STATUSES'
+
+import Module from 'parser/core/Module'
+import {Rule, Requirement} from 'parser/core/modules/Checklist'
+import {TieredSuggestion, SEVERITY} from 'parser/core/modules/Suggestions'
+
+const DEMO_DURATION_MILLIS = STATUSES.DEMOLISH.duration * 1000
+
+export default class Demolish extends Module {
+	static handle = 'demolish'
+	static dependencies = [
+		'checklist',
+		'enemies',
+		'invuln',
+		'suggestions',
+	]
+
+	_clipDemo = 0
+	_lastDemo = {}
+
+	constructor(...args) {
+		super(...args)
+
+		const filter = {
+			by: 'player',
+			abilityId: STATUSES.DEMOLISH.id,
+		}
+
+		this.addHook(['applydebuff', 'refreshdebuff'], filter, this._onApply)
+		this.addHook('complete', this._onComplete)
+	}
+
+	// Aggregate Demo uptime
+	_onApply(event) {
+		const lastDemo = this._lastDemo[event.targetID] = this._lastDemo[event.targetID] || 0
+
+		if (!lastDemo) {
+			this._lastDemo[event.targetID] = event.timestamp
+			return
+		}
+
+		let clip = DEMO_DURATION_MILLIS - (event.timestamp - lastDemo)
+
+		clip -= this.invuln.getUntargetableUptime('all', event.timestamp - DEMO_DURATION_MILLIS, event.timestamp)
+		clip -= this.invuln.getInvulnerableUptime('all', event.timestamp, event.timestamp + DEMO_DURATION_MILLIS + clip)
+
+		this._clipDemo += Math.max(0, clip)
+		this._lastDemo[event.targetID] = event.timestamp
+	}
+
+	_onComplete() {
+		this.checklist.add(new Rule({
+			name: 'Keep Demolish up',
+			description: <Fragment>
+				<ActionLink {...ACTIONS.DEMOLISH}/> is your strongest finisher (assuming at least 4 DoT ticks hit).
+			</Fragment>,
+			requirements: [
+				new Requirement({
+					name: <Fragment><ActionLink {...ACTIONS.DEMOLISH}/> uptime</Fragment>,
+					percent: () => this.getDotUptimePercent(),
+				}),
+			],
+			// TODO: calculate the number of good Demolishes a fight should have
+			//       and set target to allow dropping without losing a tick
+			target: 85,
+		}))
+
+		this.suggestions.add(new TieredSuggestion({
+			icon: ACTIONS.DEMOLISH.icon,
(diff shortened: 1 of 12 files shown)


src/parser/jobs/mnk/Forms.js
```diff
@@ -130,7 +130,7 @@ export default class Forms extends Module {
 				icon: ACTIONS.FORM_SHIFT.icon,
 				severity: SEVERITY.MAJOR,
 				content: <Fragment>
-					Avoid using <ActionLink {...ACTIONS.DRAGON_KICK}/> and <ActionLink {...ACTIONS.ARM_OF_THE_DESTROYER}/> outside of <StatusLink {...STATUSES.OPO_OPO_FORM}/>. Their special effects only activate when in the correct form.
+					Avoid using <ActionLink {...ACTIONS.DRAGON_KICK}/> and <ActionLink {...ACTIONS.ARM_OF_THE_DESTROYER}/> outside of <StatusLink {...STATUSES.OPO_OPO_FORM}/>. Their special effects only activate when in the correct form and <ActionLink {...ACTIONS.BOOTSHINE} /> has higher potency.
```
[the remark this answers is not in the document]
ayyaruq: If you're actually hitting the positional, you crit the boot so the effective potency is higher. I do actually have a check in positionals module for critting Bootshine but some people think it's a silly thing to verify.

src/parser/jobs/mnk/NOTES.md
```diff
@@ -1,28 +1,43 @@
...
 - [ ] Minor on 8th RoF GCD instead of 9th
 - [ ] Major if just after RoF drops
 - [ ] Check recovery doesn't take more than 4 GCDs (PB is 3, ending before Snap and using RoW is 4)
 - [ ] Use before forced downtime longer than RoE can be used, this might be fight specific and is hard to track since FFLogs is weird about GL refreshes
+- [ ] Implement setup/recovery scenarios? Usually this is just "fix your opener"
 
 ### Positionals
 - [ ] Can kinda guess this from the damage mainly because MNK potency is high but it's a bit giri-giri
-- [ ] Crit the Boot
+- [x] Crit the Boot
```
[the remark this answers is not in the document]
ayyaruq: Derp, I committed my global notes. The positionals module currently tracks only crit Bootshines, and makes a checklist for crit rate. I'll probably remove the checklist because it's not super useful but I'm still not too sure on how effective that module is besides for obviously missed positionals, it depends a lot on fights too.

I had an idea of tracking the damage amount for diff skills and normalising (not the way parser modules normalise) the damage across buffs, and using that for positional info. MNK has significant potency differences for positional that other jobs don't, plus special effects, so it's easier to tell when we hit positionals. There's also the position of character and enemy data that could be used but it becomes fight specific to calculate position plus we'd need to know where the aggro holder is to determine direction, and that only works assuming the boss is facing that player in the first place.

src/parser/jobs/mnk/MnkAoE.js
```diff
@@ -0,0 +1,78 @@
...
+		}
+	}
+
+	_onComplete() {
+		if (this._badAotDs.length >= 1) {
+			this.suggestions.add(new Suggestion({
+				icon: ACTIONS.ARM_OF_THE_DESTROYER.icon,
+				severity: SEVERITY.MEDIUM,
+				content: <Fragment>
+					<ActionLink {...ACTIONS.ARM_OF_THE_DESTROYER}/> is only TP efficient when there are more than {AOTD_MIN_TARGETS} targets.
```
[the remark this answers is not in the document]
ayyaruq: I might just prune the TP part entirely. AotD should be TP-efficient at 3 mobs, the 5 is to deal with potency crits on boot. Rockbreaker is potency efficient at 2 mobs and TP efficient at 3.

src/parser/jobs/mnk/Demolish.js
```diff
@@ -0,0 +1,100 @@
...
+
+		if (!lastDemo) {
+			this._lastDemo[event.targetID] = event.timestamp
+			return
+		}
+
+		let clip = DEMO_DURATION_MILLIS - (event.timestamp - lastDemo)
+
+		clip -= this.invuln.getUntargetableUptime('all', event.timestamp - DEMO_DURATION_MILLIS, event.timestamp)
+		clip -= this.invuln.getInvulnerableUptime('all', event.timestamp, event.timestamp + DEMO_DURATION_MILLIS + clip)
```
ackwell: lookaheads :notlikeblob:

remind me to research how the hell boss specific invulns are gonna work because i have noooooo idea

src/parser/jobs/mnk/Fists.js
```diff
@@ -0,0 +1,69 @@
...
+		this._stanceApplied.push(event)
+		return
+	}
+
+	_onRemove(event) {
+		this._stanceDropped.push(event)
+	}
+
+	_onComplete() {
+		if (this._stanceless > 0) {
```
ackwell: For ref, you can remove the guard  if you're using `TieredSuggestion`. Values under the smallest severity in your tier list will automatically fall back to `SEVEIRY.IGNORE`, which just hides it.

src/parser/jobs/mnk/GreasedLightning.js
```diff
@@ -0,0 +1,305 @@
...
+				borderColor: Color(JOBS.MONK.colour).fade(0.2),
+				steppedLine: true,
+			}],
+		}
+
+		const options = {
+			legend: {display: false},
+			tooltips: {enabled: false},
+			scales: {
+				xAxes: [{
```
ackwell: With the tweaks to the core chart component, you can probably completely remove most of these options. Anything defined in the component's default options will be set unless explicitly overwritten. :blobmorning:

src/parser/jobs/mnk/MnkAoE.js
```diff
@@ -0,0 +1,80 @@
...
+import ACTIONS from 'data/ACTIONS'
+
+import Module from 'parser/core/Module'
+import {Suggestion, SEVERITY} from 'parser/core/modules/Suggestions'
+
+const ROCKBREAKER_MIN_TARGETS = 2
+const AOTD_MIN_TARGETS = 5
+
+export default class MnkAoE extends Module {
+	static handle = 'mnkaoe'
```
ackwell: Remind me to rename the AoE event generation module so actual aoe modules don't have to have silly prefixes

PR #146, opened 2018-08-14
Added suggestions for Doton use

Toastdeib:
3 new suggestions for NIN regarding Doton usage. The logic on the `badStds` one is a little goofy at a glance, but the idea on it was to filter out uses where it started multi-target and ended single-target, since those are _usually_ justified (e.g. O7S adds).

### DIFF
--- a/src/parser/jobs/nin/Ninjutsu.js
+++ b/src/parser/jobs/nin/Ninjutsu.js
@@ -3,22 +3,31 @@ import React, {Fragment} from 'react'
 
 import {ActionLink} from 'components/ui/DbLink'
 import ACTIONS from 'data/ACTIONS'
+import STATUSES from 'data/STATUSES'
 import Module from 'parser/core/Module'
 import {Suggestion, SEVERITY} from 'parser/core/modules/Suggestions'
+
+const DOTON_TICK_TARGET = 6
+const TCJ_DOTON_TICK_TARGET = 7
+const JUSTIFIABLE_DOTON_TICKS = 10
 
 export default class Ninjutsu extends Module {
 	static handle = 'ninjutsu'
 	static dependencies = [
+		'combatants',
 		'suggestions',
 	]
 
 	_hyotonCount = 0
 	_rabbitCount = 0
+	_dotonCasts = []
 
 	constructor(...args) {
 		super(...args)
 		this.addHook('cast', {by: 'player', abilityId: ACTIONS.HYOTON.id}, this._onHyotonCast)
 		this.addHook('cast', {by: 'player', abilityId: ACTIONS.RABBIT_MEDIUM.id}, this._onRabbitCast)
+		this.addHook('cast', {by: 'player', abilityId: ACTIONS.DOTON.id}, this._onDotonCast)
+		this.addHook('aoedamage', {by: 'player', abilityId: STATUSES.DOTON.id}, this._onDotonDamage)
 		this.addHook('complete', this._onComplete)
 	}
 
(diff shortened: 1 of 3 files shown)


src/parser/jobs/nin/Ninjutsu.js
```diff
@@ -30,8 +39,54 @@ export default class Ninjutsu extends Module {
...
+	}
+
+	_appraiseDotonCasts() {
+		const result = {
+			badTcjs: 0, // TCJ Dotons that fell short of 7 ticks (TCJ Suiton will be more damage)
+			badAoes: 0, // AoE Dotons that fell short of 6 ticks (Katon will be more damage)
+			badStds: 0, // Single-target regular Dotons (do not do this)
+		}
+
+		for (let i = 0; i < this._dotonCasts.length; i++) {
```
ackwell: `Array.prototype.forEach` is a friendo. This will work fine, though.

src/parser/jobs/nin/Ninjutsu.js
```diff
@@ -30,8 +39,54 @@ export default class Ninjutsu extends Module {
...
+				// If it's a fully single-target TCJ that doesn't land every Doton tick, flag it
+				if (cast.ticks.length < TCJ_DOTON_TICK_TARGET) {
+					result.badTcjs++
+				}
+			} else if (cast.ticks.every(tick => tick > 1)) {
+				// If it's a fully multi-target Doton that misses at least 2 ticks, flag it
+				if (cast.ticks.length < DOTON_TICK_TARGET) {
+					result.badAoes++
+				}
+			} else if (cast.ticks.reduce((accum, value) => accum + value) < JUSTIFIABLE_DOTON_TICKS) {
```
ackwell: I like to be super anal and always provide an initial value to reducers. But that's just me.

src/parser/jobs/nin/Ninjutsu.js
```diff
@@ -64,5 +119,57 @@ export default class Ninjutsu extends Module {
 				</Fragment>,
 			}))
 		}
+
+		const {badTcjs, badAoes, badStds} = this._appraiseDotonCasts()
+		if (badTcjs > 0) {
+			this.suggestions.add(new Suggestion({
+				icon: ACTIONS.DOTON.icon,
+				content: <Fragment>
+					<Trans id="nin.ninjutsu.suggestions.tcj-doton.content">Avoid using <ActionLink {...ACTIONS.DOTON}/> under <ActionLink {...ACTIONS.TEN_CHI_JIN}/> unless at least {TCJ_DOTON_TICK_TARGET} ticks will hit or you&apos;re up against multiple targets. On a single target that&apos;s about to jump or move, using the <ActionLink {...ACTIONS.SUITON}/> combo will do more damage even if <ActionLink {...ACTIONS.TRICK_ATTACK}/> is on cooldown.</Trans>
```
ackwell: You can get away with apostrophes w/o the xml escape now.

src/parser/jobs/nin/Ninjutsu.js
Toastdeib: I'll start using this instead, gotta get over my trust issues one of these days.

src/parser/jobs/nin/Ninjutsu.js
Toastdeib: Oh neat. I'll stop using it then.

Toastdeib:
I'll make the TCJ suggestion tiered (and Hyoton while I'm at it), but I think the others are fine. Fights don't generally have enough adds for more than 1-2 AoE Dotons so it's not really worth tiering that one, and regular STDs should always be major in my opinion.

PR #147, opened 2018-08-14
Fixed two minor bugs in the NIN modules

Toastdeib:
Hellfrog was me being a dumb. For weaving, though, let me know if disagree with the change. I think it's the best way to handle it, since the only time you should _ever_ be double-weaving on NIN is if Huton is down and in that case you have bigger problems to worry about anyway.


src/parser/jobs/nin/Weaving.js
```diff
@@ -84,6 +84,6 @@ export default class Weaving extends CoreWeaving {
 			return weaveCount > 1
 		}
 
-		return super.isBadWeave(weave, maxWeaves)
+		return super.isBadWeave(weave, 1)
```
ackwell: This doesn't concern me at all, that's what it's there for. I'll rely on you to know what is(n't) appropriate for NIN itself.

PR #148, opened 2018-08-14
Updated RDM Gauge, added Trans tags

Cephid:
This is a small PR, I updated the RDM gauge to expose White/Black mana to other modules (not yet consumed), updated to remove all magic numbers from gauge and also added full Trans tags to all RDMgauge text.   Only exception is the labels on the Graph, it would not take JSX text, Trans, or Fragment.  So I left it as is for now.


src/parser/jobs/rdm/Gauge.js
```diff
@@ -223,94 +243,69 @@ export default class Gauge extends Module {
 					content: <Fragment>
 						<Message warning icon>
 							<Icon name="warning sign"/>
-								Due to a missing cast at the start of the log, mana calculations might be off.
-								Additionally 1 or more finishers might have been incorrectly flagged as wrongly used.
+							<Trans id="rdm.gauge.missing-cast.warning.content">Due to a missing cast at the start of the log, mana calculations might be off.
```
Toastdeib: Potentially minor (ackwell can have the final say here), but I think the standard we've been using for string IDs is `[job].[module].[section if applicable].[the rest of it]`. So for all of these ones, you'd want to be using `rdm.gauge.suggestions...` as the start of the key instead of just `rdm.gauge...`.

ackwell: That's mostly me being anal. Helps split between "gauge checklist" and "gauge suggestions" and "gauge output" (say).

I won't block on it, but going to leave this overnight.

src/parser/jobs/rdm/Gauge.js
```diff
@@ -223,94 +243,69 @@ export default class Gauge extends Module {
...
-			// 			You were missing an <ActionLink {...ACTIONS.VERTHUNDER}/> at the start of the Log
-			// 		</Fragment>,
-			// 		severity: SEVERITY.MAJOR,
-			// 		why: <Fragment>
-			// 			You were the first damage event, so it doesn&apos;t log your first cast as cast by you
-			// 		</Fragment>,
-			// 	}))
-			// }
-
 			if (this._whiteManaWasted && this._whiteManaWasted > 0) {
```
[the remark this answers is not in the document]
Cephid: Good point, removing.

src/parser/jobs/rdm/Gauge.js
Cephid: rdm.gauge.suggestions.missing-cast-warning-content
rdm.gauge.suggetsions.missing-cast-warning-why
etc.

Gonna push this change up, I honestly don't care what they are named one way or another.

PR #150, opened 2018-08-16
Initial supported patches handling

ackwell:
As per discussion, this:

* Modifies the supported patch meta info to display a range of patches
* Provides a fallback for module groups using the old format
* Compares the timestamp of the fight to the patches and shows a prominent warning if it occurs during an unsupported patch
* Updates the SMN module group to use the new format
* Extracts a bunch of other strings because someone forgot to :blobknife:

**Examples:**

image
image


ackwell:
Given that (almost) all updates square release are small changes like that, I can forsee it being a valid option. Means we can "maintain" wider patch support w/o ridiculous amounts of extra work.

I'll add it to this PR, should be easy.

ackwell:
Added suggestion from toast, and used it to give a concrete patch version in the error. Example:
image

PR #152, opened 2018-08-17
Fix version of fsevents used by react-scripts

Kovensky:
Fix installing on node 10; react-scripts requests a version of fsevents that's too old for node 10.


Kovensky:
Why does specifying a resolution turn an optional package into required >:(

PR #157, opened 2018-08-19
Ammo and Heat modules for MCH

Toastdeib:
New MCH stuff! This time I added two simulation modules, one for tracking ammo usage and one for tracking heat (since both are gauges and not hard data and fuuuuuuuuck). The heat module has a means of correcting itself if it detects that the null assumption (Gauss is on at the start of the fight) was incorrect, but I didn't bother with anything similar for ammo since it should sort itself out within 3 GDCs tops if it's inaccurate.

Also I remembered the extract this time. :blobmorning:


src/parser/jobs/mch/Heat.js
```diff
@@ -0,0 +1,271 @@
...
+			icon: ACTIONS.COOLDOWN.icon,
+			content: <Fragment>
+				<Trans id="mch.heat.suggestions.cooldown.content">Avoid using <ActionLink {...ACTIONS.COOLDOWN}/> if it would bring you below {HEATED_POINT} heat. Dipping too low will cost you heated shots and may delay overheat windows, throwing off your rotation and OGCD alignment.</Trans>
+			</Fragment>,
+			tiers: {
+				1: SEVERITY.MEDIUM,
+				2: SEVERITY.MAJOR,
+			},
+			value: this._badCooldowns,
+			why: <Fragment>
```
ackwell: You can scrap the `<Fragment>` in this case.

Toastdeib: Honestly I should probably do that on most of my modules. I think I have hella unnecessary `<Fragment>` tags all over the place right now.

PR #161, opened 2018-08-22
GCD estimate/uptime fixes and BLM things

LaqiThish:
Big fixes to GCD estimate and uptime calc. BLM Gauge updates to facilitate GCD fixes.

ACTIONS
- Add default values to castTime and cooldown to onGcd actions when not specified

GlobalCooldown
- Fix pairing GCDs with duration, speedmod, instant flag, etc
- Fix GCD estimate for non-2.5s casts
- Fix uptime calc for BLM's 50% castTime-only speed modifier (very hacky, see note, will clean up in later PR)
- Add caster tax to help with GCD and uptime estimates

Weaving
- Don't call isBadWeave for 0-length weaves

BLM Stuff
- Log fast-cast modifier in Speedmod
- Add custom Gauge events
- Update modules to use Gauge events
- Fix some module bugs


src/data/ACTIONS/index.js
```diff
@@ -57,13 +60,25 @@ const ACTIONS = {
...
+}
+
 export const COOLDOWN_GROUPS = _.groupBy(ACTIONS, 'cooldownGroup')
 
 export const HIT_TYPES = {
 	CRIT: 2,
 }
 
 addExtraIndex(ACTIONS, 'id')
+addExtraIndex(addDefaultValues(ACTIONS), 'id')
```
ackwell: Feels hacky to me, but I can see _why_ you're doing it this way. Should be fine for now.

WoWA has an intermediary between the spell (action) data and the consumers via their `Spellbook` module, IIRC - I'll have a think about some alternative implementations for Later:tm:.

src/parser/core/modules/GlobalCooldown.js
```diff
@@ -7,15 +7,25 @@ import {getAction} from 'data/ACTIONS'
...
-const MIN_GCD = 1500
-const MAX_GCD = 2500
+const MIN_GCD_IN_MILLIS = 1500
+const MAX_GCD_IN_MILLIS = 2500
+const BASE_GCD_IN_SECONDS = 2.5
+const CASTER_TAX_IN_MILLIS = 100
+
+const DEBUG_LOG_SAVED_GCDS = false && process.env.NODE_ENV !== 'production'
+
+// NOTE: Caster tax refers to spells taking 0.1s longer than their tooltip claims if their cast time is at least as long as their recast time.
```
ackwell: Squareeeeeeeeeeeeeeeeee

src/parser/core/modules/GlobalCooldown.js
```diff
@@ -59,20 +70,40 @@ export default class GlobalCooldown extends Module {
...
 
 			case 'cast':
-				if (this._castingEvent && this._castingEvent.ability.guid === action.id) {
-					this.saveGcd(this._castingEvent, false)
-				} else {
-					this.saveGcd(event, true)
-				}
+				const hasBeginCast = this._castingEvent !== null && this._castingEvent.ability.guid === action.id
+				const relevantEvent = hasBeginCast ? this._castingEvent : event
+				this.saveGcd(this._lastGcd, relevantEvent.timestamp) // Save last gcd with current timestamp
```
ackwell: Objects pass by reference. While the implementation of `saveGcd` _currently_ doesn't get effected by that, I can see future changes might hit that, and it'd be a pretty nasty debug to run through.

Gonna ask for the `this._lastGcd` to be passed as a copy to prevent those potential down-road issues - simplest impl. would be `{...this._lastGcd}`.

src/parser/core/modules/GlobalCooldown.js
```diff
@@ -59,20 +70,40 @@ export default class GlobalCooldown extends Module {
...
+
+				this._lastGcd.isInstant = !hasBeginCast
+				this._lastGcd.event = relevantEvent
 
 				this._castingEvent = null
 				break
 			}
 		}
 
+		this.saveGcd(this._lastGcd, events[events.length - 1].timestamp)
```
ackwell: See above

src/parser/core/modules/GlobalCooldown.js
```diff
@@ -59,20 +70,40 @@ export default class GlobalCooldown extends Module {
...
+
 		return events
 	}
 
+	_debugLogSavedGcds() {
+		if (!DEBUG_LOG_SAVED_GCDS) {
+			return
+		}
+
+		// NOTE: Please sanity-check results when changing normalise or saveGcd. Good test cases include:
```
ackwell: :notlikeblob:

src/parser/core/modules/GlobalCooldown.js
```diff
@@ -95,66 +126,77 @@ export default class GlobalCooldown extends Module {
...
+		const action = getAction(gcdInfo.event.ability.guid)
+		let speedMod = this.speedmod.get(gcdInfo.event.timestamp)
+		let castTime = action.castTime
 
-		if (this._lastGcd >= 0) {
-			// GCD is only to two decimal places, so round it there. Storing in Ms.
-			// eslint-disable-next-line no-magic-numbers
-			gcdLength = Math.round((event.timestamp - this._lastGcd)/10)*10
+		// HACK NOTE TODO: Need to properly account for abilities that alter only the cast or recast of attacks.
+		// Thinking of moving this into a module like speedmod, that can be called with a timestamp to grab modified base castTime/cooldown values
```
ackwell: Astral/umbral handling is technically supposed to be handled by `castTime`, though I recognise the current framework in there doesn't actually support what you need to do.

I'll let this pass for now despite the 🤢 , but ping me on deeskerd or w/e to chat about potential impl resolutions if you'd like.

src/parser/core/modules/GlobalCooldown.js
```diff
@@ -7,15 +7,25 @@ import {getAction} from 'data/ACTIONS'
 import Module from 'parser/core/Module'
 import {Group, Item} from './Timeline'
 
-const MIN_GCD = 1500
-const MAX_GCD = 2500
+const MIN_GCD_IN_MILLIS = 1500
+const MAX_GCD_IN_MILLIS = 2500
+const BASE_GCD_IN_SECONDS = 2.5
```
ackwell: I'd almost prefer all this be stored in ms, without the var suffixes - `1000` is a whitelisted magic if nessecary.

Not gonna  kick a stink about it, the names just feel dirty.

src/parser/jobs/blm/Gauge.js
```diff
@@ -7,210 +7,291 @@ import ACTIONS from 'data/ACTIONS'
...
-			const numberOfFouls = Math.floor(enoRunTime/ENOCHIAN_DURATION_REQUIRED)
-			if (numberOfFouls > 0) {
-				const offSet = enoRunTime % ENOCHIAN_DURATION_REQUIRED
-				this._enoTimer = event.timestamp - offSet
-				this._poly ++
-				if (this._poly > 1) {
-					this._poly = 1
-					this._overwrittenFoul ++
-				}
+	normalise(events) {
```
ackwell: I'm heavily questioning the use of normalise for this. Any reason the events can't be sent out with `this.parser.fabricateEvent`?

src/parser/jobs/blm/Gauge.js
```diff
@@ -7,210 +7,291 @@ import ACTIONS from 'data/ACTIONS'
...
+			this._overwrittenFoul--
 		}
+		this._hasPolyglot = false
+		this.addEvent()
+	}
 
-		//keep track of UH
-		if (abilityId === ACTIONS.BLIZZARD_IV.id) {
-			this._UH = 3
+	onGainUmbralIceStack(event) {
```
ackwell: Lots of repetition in these handlers. Can probably DRY them up a bit by passing stack counts around as args and so on.

src/parser/jobs/blm/Gauge.js
```diff
@@ -7,210 +7,291 @@ import ACTIONS from 'data/ACTIONS'
...
-			}
-			if (this._UI > 0) {
-				this._UI = 0
-				this._AF = 1
-				this._AFUITimer = event.timestamp
+	_onCast(event) {
+		const abilityId = event.ability.guid
+
+		switch (abilityId) {
+		case ACTIONS.ENOCHIAN.id:
```
ackwell: Holy switch batman.

w/o the normalise, this could probably be done with a bunch of hooks, though I don't particularly mind either way.

src/parser/jobs/blm/Speedmod.js
```diff
@@ -1,19 +1,63 @@
...
+	_isFastCast = false
 
 	constructor(...args) {
 		super(...args)
 		this.SPEED_BUFF_STATUS_IDS.push(
 			STATUSES.CIRCLE_OF_POWER.id, // Ley Lines buff
 		)
 	}
+
+	jobSpecificNormaliseLogic(event) {
```
ackwell: Ah, this must be why you need that logic in normalise. Still a bit uncomfortable with that... the more we put in normalise, the more needs to be put in normalise, the more needs... you get the point. Slippery slope and all but I'm concerned we'll end up with no hooks at all soon :blobsweat:.

Not got the brain space to suggest alternative implementations for the whole speedmod stack, but worth discussing modifications to unlock massive chunks of stuff from normalisers.

src/data/ACTIONS/index.js
LaqiThish: We had a talk about this a few weeks back, just fyi.

```
αckwell08/05/2018
i can see the sanity behind it honestly
saves filling in data and saves keeping on toes with defaulting
take a look at data/STATUSES/index```

Though if the implementation's not how you imagined, I can fix that.

src/parser/core/modules/GlobalCooldown.js
LaqiThish: Will make these pass as copy changes

src/parser/core/modules/GlobalCooldown.js
LaqiThish: That's fair. Even if they're all in ms, I'd prefer the suffix, but it wasn't my file to begin with. I'll make these changes.

src/parser/core/modules/GlobalCooldown.js
LaqiThish: Yes

src/data/ACTIONS/index.js
ackwell: Hahahaha, that's my memory for you.

This implementation is fine for now - that comment is more re: future development.

src/parser/jobs/blm/Gauge.js
ackwell: As per discussion on discord, going to let this go for now. Need to re-consider how some GCD based calculations are run so we can pull some of the normalise logic out of normalisation in the future.

PR #162, opened 2018-08-23
QoL tweaks

ackwell:
Nothing particularly visible, just a few tweaks for end user QoL

* Made the "Kills Only" and "Show Minor" switches persist across navigation and reloads
* Capped checklist %s at 100 in production builds - dev server will still show >100 for debugging purposes
* Added helper function `module.removeHook`, not that people really need it

In looking around for a test case for the second - most of the ones that have been brought up recently have been fixed already, so great work to all involved w/ speedmod changes!

### DIFF
--- a/src/components/Find/FightList.js
+++ b/src/components/Find/FightList.js
@@ -1,12 +1,12 @@
 import PropTypes from 'prop-types'
 import React, {Component, Fragment} from 'react'
+import {connect} from 'react-redux'
 import {Checkbox, Header, Icon, Menu} from 'semantic-ui-react'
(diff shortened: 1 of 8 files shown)


Toastdeib:
:+1: from me as well, and I already have a use queued up for `removeHook` so I'm excited for that.

MikeMatrix:
Use could be something like event toggled processing of events.
Some event happened -> Add Hook and process something that would be considered at that point -> Something else happens -> We no longer need to consider, so remove hook and skip execution of function entirely for other events.

PR #165, opened 2018-08-25
Trick Attack usage module

Toastdeib:
Back to NIN! This time I've got a module that yells at you about TA usage. It contains three suggestions: One for holding TA and losing potential casts, one for using it at a bad point in your opener, and one for failing to use it entirely. As tempted as I was, I kept the harshness for that last one in the comments and not in the actual suggestion text.


src/parser/jobs/nin/TrickAttackUsage.js
```diff
@@ -0,0 +1,106 @@
...
+	_onCast(event) {
+		const action = getAction(event.ability.guid)
+		if (action.onGcd) {
+			this._gcdCount++
+		}
+	}
+
+	_onTrickAttack(event) {
+		if (this._castHook !== null) {
+			this.removeHook(this._castHook)
```
ackwell: :bloblove:

PR #169, opened 2018-08-27
MNK buff uptime tracking

ayyaruq:
Mostly building on @nfriendly's work in #110, I also added FoF uptime checking and fixed normalisation for speedmod a little. The checklist percentages here are very hit and miss and I'm on the fence about the severity but the logic is good so derp. If this is good to go, it also closes the older PR.

### DIFF
--- a/src/parser/jobs/mnk/BuffUptime.js
+++ b/src/parser/jobs/mnk/BuffUptime.js
@@ -57,11 +57,11 @@ export default class BuffUptime extends Module {
 		// Only include GCDs, but don't double increment either
 		if (action.onGcd && !BUFF_CHECK_SKILLS.includes(action)) {
 			if (this._lastDragonKickUse[event.targetID]) {
-        this._gcdsSinceDK[event.targetID] += 1
+				this._gcdsSinceDK[event.targetID] += 1
 			}
 
 			if (this._lastTwinSnakesUse !== null) {
-        this._gcdsSinceTS++
+				this._gcdsSinceTS++
 			}
 		}
 	}
--- a/src/parser/jobs/mnk/BuffUptime.js
+++ b/src/parser/jobs/mnk/BuffUptime.js
@@ -5,7 +5,7 @@ import ACTIONS, {getAction} from 'data/ACTIONS'
 import STATUSES from 'data/STATUSES'
 import Module from 'parser/core/Module'
 import {Rule, Requirement} from 'parser/core/modules/Checklist'
-import {Suggestion, SEVERITY} from 'parser/core/modules/Suggestions'
+import {Suggestion, TieredSuggestion, SEVERITY} from 'parser/core/modules/Suggestions'
 
 const GCD_CYCLE_LENGTH = 6
 
@@ -57,7 +57,7 @@ export default class BuffUptime extends Module {
 		// Only include GCDs, but don't double increment either
 		if (action.onGcd && !BUFF_CHECK_SKILLS.includes(action)) {
 			if (this._lastDragonKickUse[event.targetID]) {
-				this._gcdsSinceDK[event.targetID] += 1
+				this._gcdsSinceDK[event.targetID]++
 			}
 
 			if (this._lastTwinSnakesUse !== null) {
@@ -143,12 +143,16 @@ export default class BuffUptime extends Module {
 		if (this._earlyTwinSnakes) {
 			const lostTruePotency = this._earlyTwinSnakes * (ACTIONS.TRUE_STRIKE.potency - ACTIONS.TWIN_SNAKES.potency)
 
-			this.suggestions.add(new Suggestion({
+			this.suggestions.add(new TieredSuggestion({
 				icon: ACTIONS.TWIN_SNAKES.icon,
 				content: <Fragment>
 					Avoid refreshing <ActionLink {...ACTIONS.TWIN_SNAKES} /> signficantly before its expiration as you're losing uses of the higher potency <ActionLink {...ACTIONS.TRUE_STRIKE} />.
 				</Fragment>,
-				severity: SEVERITY.MAJOR,
+				tiers: {
+					1: SEVERITY.MEDIUM,
+					4: SEVERITY.MAJOR,
+				},
+				value: this._earlyTwinSnakes,
 				why: <Fragment>
 					{lostTruePotency} potency lost to {this._earlyTwinSnakes} early refreshes.
 				</Fragment>,
--- a/src/parser/jobs/mnk/BuffUptime.js
+++ b/src/parser/jobs/mnk/BuffUptime.js
@@ -7,7 +7,12 @@ import Module from 'parser/core/Module'
 import {Rule, Requirement} from 'parser/core/modules/Checklist'
 import {Suggestion, SEVERITY} from 'parser/core/modules/Suggestions'
 
-const GCD_CYCLE_LENGTH = 5
+const GCD_CYCLE_LENGTH = 6
+
+const BUFF_CHECK_SKILLS = [
+	ACTIONS.DRAGON_KICK.id,
+	ACTIONS.TWIN_SNAKES.id,
+]
 
 export default class BuffUptime extends Module {
 	static handle = 'BuffUptime'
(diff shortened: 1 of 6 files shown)


src/parser/jobs/mnk/Fists.js
```diff
@@ -50,6 +54,20 @@ export default class Fists extends Module {
...
+			description: <Fragment>
+				Fists of Fire is a low effort 5% buff to all your damage.
+			</Fragment>,
+			requirements: [
+				new Requirement({
+					name: <Fragment><ActionLink {...ACTIONS.FISTS_OF_FIRE} /> uptime</Fragment>,
+					percent: () => this.getBuffUptimePercent(STATUSES.FISTS_OF_FIRE.id),
+				}),
+			],
+			target: 90,
```
Toastdeib: I am by no means an expert at MNK, but I feel like this could maybe be higher? I know RoE is a thing but in my understanding, you usually don't want to linger in FoE after you get the stack refresh. Not a blocker, just something to think about.

src/parser/jobs/mnk/Fists.js
```diff
@@ -60,4 +78,11 @@ export default class Fists extends Module {
 			value: this._stanceless,
 		}))
 	}
+
+	getBuffUptimePercent(statusId) {
+		const statusUptime = this.combatants.getStatusUptime(statusId, this.parser.player.id)
+		const fightUptime = this.parser.fightDuration - this.invuln.getInvulnerableUptime()
```
Toastdeib: Unless `this.combatants.getStatusUptime()` filters out invuln windows, I don't think you need (or actually want) to exclude them from the overall fight duration, since MNK Fist buffs don't rely on hitting things (unlike just about every other buff of their ilk that I can think of). They're just... active.

src/parser/jobs/mnk/BuffUptime.js
```diff
@@ -0,0 +1,161 @@
...
+
+	_onCast(event) {
+		const action = getAction(event.ability.guid)
+
+		if (!action) {
+			return
+		}
+
+		if (action.onGcd) {
+			this._gcdsSinceDK[event.targetID] += 1
```
Toastdeib: Won't this result in nulls during the opener/when adds show up? `_gcdsSinceDK[event.targetID]` is only initialized in `_onDragonKick()`, but the standard opener has 3 GCDs before your first DK. Also for accuracy, shouldn't it increment all the targets you're tracking on each GCD cast, regardless of the target you actually hit?

src/parser/jobs/mnk/BuffUptime.js
```diff
@@ -0,0 +1,161 @@
+import React, {Fragment} from 'react'
+
+import {ActionLink} from 'components/ui/DbLink'
+import ACTIONS, {getAction} from 'data/ACTIONS'
+import STATUSES from 'data/STATUSES'
+import Module from 'parser/core/Module'
+import {Rule, Requirement} from 'parser/core/modules/Checklist'
+import {Suggestion, SEVERITY} from 'parser/core/modules/Suggestions'
+
+const GCD_CYCLE_LENGTH = 5
```
Toastdeib: Shouldn't this actually be 6?

src/parser/jobs/mnk/BuffUptime.js
```diff
@@ -0,0 +1,161 @@
...
+				</Fragment>,
+				severity: SEVERITY.MEDIUM,
+				why: <Fragment>
+					{this._earlyDragonKicks} rotation cycles were interrupted by early refreshes.
+				</Fragment>,
+			}))
+		}
+
+		if (this._earlyTwinSnakes) {
+			const _lostTruePotency = this._earlyTwinSnakes * (ACTIONS.TRUE_STRIKE.potency - ACTIONS.TWIN_SNAKES.potency)
```
Toastdeib: Minor, but this doesn't need to be marked as private since it's local.

src/parser/jobs/mnk/BuffUptime.js
```diff
@@ -0,0 +1,161 @@
...
+		// Make sure we're tracking for this target
+		const lastApplication = this._lastDragonKickUse[event.targetID]
+
+		// If it's not been applied yet set it and skip out
+		if (!lastApplication) {
+			this._lastDragonKickUse[event.targetID] = event.timestamp
+			return
+		}
+
+		// This logic is busted if the player's GCD is under 1.67s
```
Toastdeib: _Enhanced Arrow PTSD flashbacks._ :internalscreaming:

src/parser/jobs/mnk/Fists.js
ayyaruq: This number kinda came out of just averaging diff FFLogs but it's actually really variable. For a player doing a basic rotation, it should be easily 95+%, but for players that do frequent Tornado Kicks and use RoW for recovery, the number ends up as low as 60%.

src/parser/jobs/mnk/Fists.js
ayyaruq: I actually didn't think of that and just copypasted from the other main module in this PR. Downtime is often enough FoE anyway though hmm. I'll adjust this one.

src/parser/jobs/mnk/BuffUptime.js
ayyaruq: It's zero indexed. If we initialise it at 1, then it should be 6. Maybe for clarity that's a good idea though, but then it looks weird when we have "1 GCD since DK/TS" when they reset it.

src/parser/jobs/mnk/BuffUptime.js
ayyaruq: Is that a thing? Most other modules seem to just yolo consts so I have no idea.

src/parser/jobs/mnk/BuffUptime.js
ayyaruq: Except for Shinryu's Tail, I can't think of any time DK applies to more than 1 target on refresh. It's a debuff on the target so they get refreshed independently. The uptime calculation is also kinda wonky depending on how many targets. I'm open to ideas on how to work this, how do you handle Slashing on multiple targets for NIN?

I think to deal with initialisation/opener, we can check if it's null, or maybe if DK is on the target and it dropped. Thoughts? Thanks for the feedback!

src/parser/jobs/mnk/BuffUptime.js
ayyaruq: Oops, yes.

src/parser/jobs/mnk/BuffUptime.js
```diff
@@ -0,0 +1,172 @@
...
+		const action = getAction(event.ability.guid)
+
+		if (!action) {
+			return
+		}
+
+		// Only include GCDs, but don't double increment either
+		if (action.onGcd && !BUFF_CHECK_SKILLS.includes(action)) {
+			if (this._lastDragonKickUse[event.targetID]) {
+				this._gcdsSinceDK[event.targetID] += 1
```
ackwell: could be `++` too

src/parser/jobs/mnk/BuffUptime.js
ackwell: Could always check the estimate etc, but that's a lot of logic for little gain.

src/parser/jobs/mnk/BuffUptime.js
```diff
@@ -0,0 +1,172 @@
...
+
+		if (this._earlyTwinSnakes) {
+			const lostTruePotency = this._earlyTwinSnakes * (ACTIONS.TRUE_STRIKE.potency - ACTIONS.TWIN_SNAKES.potency)
+
+			this.suggestions.add(new Suggestion({
+				icon: ACTIONS.TWIN_SNAKES.icon,
+				content: <Fragment>
+					Avoid refreshing <ActionLink {...ACTIONS.TWIN_SNAKES} /> signficantly before its expiration as you're losing uses of the higher potency <ActionLink {...ACTIONS.TRUE_STRIKE} />.
+				</Fragment>,
+				severity: SEVERITY.MAJOR,
```
ackwell: Do you really want to shout a MAJOR at people who make the mistake once over an entire fight?

ayyaruq:
Took the feedback into account and made a call on tiers, we can always change it later. Thanks @ackwell :bloblove:

src/parser/jobs/mnk/BuffUptime.js
```diff
@@ -0,0 +1,176 @@
...
+					Avoid refreshing <ActionLink {...ACTIONS.DRAGON_KICK} /> signficantly before its expiration as you're trading it for the higher damage <ActionLink {...ACTIONS.BOOTSHINE} />.
+				</Fragment>,
+				severity: SEVERITY.MEDIUM,
+				why: <Fragment>
+					{this._earlyDragonKicks} rotation cycles were interrupted by early refreshes.
+				</Fragment>,
+			}))
+		}
+
+		if (this._earlyTwinSnakes) {
```
ackwell: Tiered will fall back to being invisible fwiw.

PR #170, opened 2018-08-28
Basic DRG support (v2)

Toastdeib:
No one told me not to, so here we are. I've added basic support for DRG and a BotD/LotD simulation module for starters. I also twiddled the core Death module to fabricate `'raise'` events and track how much time you actually spent dead over the course of the fight.


src/parser/jobs/drg/About.js
```diff
@@ -0,0 +1,25 @@
+import {Trans} from '@lingui/react'
+import React, {Fragment} from 'react'
+import {Icon, Message} from 'semantic-ui-react'
+
+import CONTRIBUTORS, {ROLES} from 'data/CONTRIBUTORS'
+import CoreAbout from 'parser/core/modules/About'
+
+export default class About extends CoreAbout {
+	description = <Fragment>
+		<Trans id="drg.about.description"><p>This analyzer aims to help you beat the sterotypes, stay off the floor, and dish out some big juicy numbers. As the DRG rotation is pretty strict, the suggestions you see will focus mostly on keeping that rotation flowing smoothly, as well as picking out issues related to your Life of the Dragon windows.</p></Trans>
```
ackwell: I'd flip the `<Trans>` and `<p>` tags around so you don't end up with the `<0>` in the i18n json.

Also, have a look at `TransMarkdown` - not necessary in this case at all, but it'll be handy for stuff like the warning below and so on.

src/parser/jobs/drg/About.js
```diff
@@ -0,0 +1,25 @@
...
+				<Trans id="drg.about.description.warning"><strong>The module is a work in progress.</strong> It currently covers very few aspects of DRG gameplay, and while the analysis below should be reasonably accurate, this system <em>is</em> still in development, and may get a little mixed up sometimes. If you notice any issues, or have any concerns, please drop by our Discord channel or report a bug on our github repository!</Trans>
+			</Message.Content>
+		</Message>
+	</Fragment>
+	supportedPatches = {
+		from: '4.05',
+		to: '4.36',
+	}
+	contributors = [
+		{user: CONTRIBUTORS.TOASTDEIB, role: ROLES.MAINTAINER},
```
ackwell: You're a madman.

src/parser/jobs/drg/Combos.js
```diff
@@ -0,0 +1,209 @@
...
+import {Rule, Requirement} from 'parser/core/modules/Checklist'
+import {TieredSuggestion, SEVERITY} from 'parser/core/modules/Suggestions'
+
+const DRAGON_MAX_DURATION_MILLIS = 30000
+const DRAGON_DEFAULT_DURATION_MILLIS = 20000
+const BLOOD_EXTENSION_MILLIS = 10000
+
+const MAX_EYES = 3
+
+export default class Combos extends CoreCombos {
```
ackwell: Not sure how comfortable I am with combo handling and gauge tracking being rolled into one module, though I can see the benefit from the perspective of the combo requirements.

src/parser/jobs/drg/About.js
Toastdeib: :blobmorning:

src/parser/jobs/drg/About.js
ackwell: Weeeeeiiirrdddddd

PR #171, opened 2018-08-29
Ugly hack to update everything to babel 7

Kovensky:
There are like 3 different versions of `@babel/*` being requested everywhere so use overrides to request the latest one.

Use a path override to rewrite `babel-preset-react-app` to a local fork that doesn't feed `@babel/runtime` an option that was removed in babel 7. The entire thing had to be forked because react-scripts actually uses all those different entry points.


preset-react-app-custom/create.js
```diff
@@ -0,0 +1,135 @@
...
+          useBuiltIns: true,
+        },
+      ],
+      // Polyfills the runtime needed for async/await and generators
+      [
+        require('@babel/plugin-transform-runtime').default,
+        {
+          helpers: false,
+          regenerator: true,
+        },
```
Kovensky: The only change from the original code in the `react-scripts` repo is that the `polyfill: false,` line was deleted.

PR #177, opened 2018-09-06
AST: Arcana Tracking, i18n

suShirow:


src/parser/jobs/ast/ArcanaGroups.js
```diff
@@ -82,6 +82,37 @@ export const SPIRE_USED = [
...
+	ACTIONS.THE_SPIRE_FROM_SPREAD.id,
+	ACTIONS.THE_BALANCE_EXPANDED_FROM_SPREAD.id,
+	ACTIONS.THE_BOLE_EXPANDED_FROM_SPREAD.id,
+	ACTIONS.THE_ARROW_EXPANDED_FROM_SPREAD.id,
+	ACTIONS.THE_SPEAR_EXPANDED_FROM_SPREAD.id,
+	ACTIONS.THE_EWER_EXPANDED_FROM_SPREAD.id,
+	ACTIONS.THE_SPIRE_EXPANDED_FROM_SPREAD.id,
+]
+
+export const DRAWN_ARCANA_USE = [
```
NononaNona: DRAWN_ARCANA_USE is defined twice.

suShirow: You might be getting cross-eyed, there's DRAWN_ARCANA_USE which encompasses the actions, and DRAWN_ARCANA which are the Drawn statuses. If you can think of a better name for these variables I'll change them

NononaNona: I'm sorry I should have been clearer. What I meant is that there is DRAWN_CARD_USE, which I added, and DRAWN_ARCANA_USE, which you added, that both encompass the same cards. We only want one name. I think your name likely better fits the rest of the group, but that means the name in my module would need to be changed.

Regardless there should only be one.

suShirow: Oh I see. Totally missed that! I'll delete the extra then. Would you like me to change your module var names in this commit too?

NononaNona: Yes that would be great thank you!

src/parser/jobs/ast/ArcanaTracking.js
```diff
@@ -0,0 +1,661 @@
...
+					} else {
+						if (index >= lastIndex && index !== this._cardStateLog.length - 1) { stateItem[slot] = getStatus(this.arcanaActionToStatus(actionId)) }
+						if (index === this._cardStateLog.length - 1 && slot === 'rrAbility') { stateItem.lastEvent[slot] = getStatus(this.arcanaActionToStatus(actionId)) }
+					}
+				})
+		}
+	}
+
+	/**
+	 * Loops back to see if the specified card id was in possession without the possiblity of it being obtained via legal abilities.
```
NononaNona: Accidentally put in the wrong description?

src/parser/jobs/ast/ArcanaTracking.js
```diff
@@ -0,0 +1,661 @@
...
+				stateItem.lastEvent &&
+				abilityLookups.includes(stateItem.lastEvent.ability.guid)
+		)
+
+		if (lastIndex === -1) {
+			// There were none finds of specified abilities. They had it prepull, so assume this is 0
+			lastIndex = 0
+
+			// Modify log, they were holding onto this card since index
+			_.forEachRight(this._cardStateLog,
```
NononaNona: Was this meant to be outside of the if clause?

src/parser/jobs/ast/ArcanaTracking.js
```diff
@@ -0,0 +1,661 @@
...
+				(stateItem.lastEvent.ability.guid === ACTIONS.SLEEVE_DRAW.id
+				|| stateItem.lastEvent.ability.guid === ACTIONS.ROYAL_ROAD.id)
+			)
+
+			if (lastRoyalRoadIndex === -1) {
+				// There were no RRs or Sleeve Draws. They had it prepull, so assume this is 0
+				lastRoyalRoadIndex = 0
+
+				// Modify log, they were holding onto this rrAbility since index
+				_.forEachRight(this._cardStateLog,
```
NononaNona: Was this meant to be outside of the if clause?

src/parser/jobs/ast/ArcanaTracking.js
```diff
@@ -0,0 +1,661 @@
...
+				const cardStateItem = {..._.last(this._cardStateLog)}
+				const lastEvent = {
+					ability: {...ACTIONS.EMPTY_ROAD, guid: ACTIONS.EMPTY_ROAD.id},
+					timestamp: event.timestamp,
+				}
+
+				cardStateItem.lastEvent = lastEvent
+				cardStateItem.rrAbility = null
+
+				// TODO: insert a new action into the log
```
NononaNona: Would you mind explaining what the intention behind this TODO is?

src/parser/jobs/ast/ArcanaTracking.js
```diff
@@ -0,0 +1,661 @@
...
+
+				// Modify log, they were holding onto this rrAbility since index
+				_.forEachRight(this._cardStateLog,
+					(stateItem, index) => {
+						if (index >= lastRoyalRoadIndex && index !== this._cardStateLog.length - 1) { stateItem.rrAbility = getStatus(event.rrAbility.guid) }
+						if (index === this._cardStateLog.length - 1) { stateItem.lastEvent.rrAbility = getStatus(event.rrAbility.guid) }
+					})
+			}
+
+			// this.retconSearch([ACTIONS.SLEEVE_DRAW.id, ACTIONS.ROYAL_ROAD.id], 'rrAbility', event.ability.guid)
```
NononaNona: Why did you decide to not use retconSearch here?

PR #178, opened 2018-09-07
Added a Procs module for DRG

Toastdeib:
This module tracks misuse of Sharper Fang and Claw, Enhanced Wheeling Thrust, and Dive Ready procs with suggestions for dropping them and also for overwriting Dive Ready. I also apparently scooped up some AST strings in my extract.


src/parser/jobs/drg/Procs.js
```diff
@@ -0,0 +1,153 @@
...
+		this.addHook('refreshbuff', {by: 'player', abilityId: STATUSES.DIVE_READY.id}, this._onReadyOverwritten) // The other two can't be overwritten due to how they drop
+		this.addHook('removebuff', {by: 'player', abilityId: STATUSES.SHARPER_FANG_AND_CLAW.id}, this._onSharperRemoved)
+		this.addHook('removebuff', {by: 'player', abilityId: STATUSES.ENHANCED_WHEELING_THRUST.id}, this._onEnhancedRemoved)
+		this.addHook('removebuff', {by: 'player', abilityId: STATUSES.DIVE_READY.id}, this._onReadyRemoved)
+		this.addHook('complete', this._onComplete)
+	}
+
+	// For all of our cast/removal tracking, we only want to know if it happened outside of downtime to avoid errant penalization.
+	// The only one that doesn't need the check is overwriting Mirage Dive, since you can't get a proc from an invulnerable target.
+	_onFangAndClaw(event) {
```
ackwell: Could probably combine a few of these with a higher order function or just more general event hooking, unless you'd expect that some would need special handling in some manner.

src/parser/jobs/drg/Procs.js
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	_onComplete() {
+		const droppedFang = this._fangAndClaw.removed - this._fangAndClaw.casts
+		const droppedWheeling = this._wheelingThrust.removed - this._wheelingThrust.casts
+		const droppedMirage = this._mirageDive.removed - this._mirageDive.casts
+
+		this.suggestions.add(new TieredSuggestion({
+			icon: ACTIONS.FANG_AND_CLAW.icon,
+			content: <Trans id="drg.procs.suggestions.fang.content">
```
ackwell: This is duping up a bit with the suggestion below it - is there a particularly strong case to keep them apart? Might be able to tier it akin to `smn/DoTs` - it uses the worse of the cases as the tier `value`, and mentions both values in the `why`.

src/parser/jobs/drg/Procs.js
Toastdeib: Nah, none of them need anything special. I can twiddle it so the cast/remove events are generalized and make one object that keys on ability ID to wrap the three separate ones I have now.

src/parser/jobs/drg/Procs.js
Toastdeib: I could see a case for merging them. My only concern is that the `why` might be goofy if I include them both, since in most of the parses I've looked at, the numbers are 1 and 0 for dropped proc counts. It's less of an issue for SMN, since you very rarely have 0s of DoT clipping, but here it would wind up as "You dropped 1 Enhanced... proc and 0 Sharper... procs." If you think that sounds fine I'll go ahead and do it (although I'll probably use the sum for the tiers rather than the max).

ackwell: Mmm.

You might be able to use plural rules to hide one or the other, but I'd be fearful of how that'd hold up in other languages which _kinda_ defeats the purpose of i18n.

I'm not bothered by 0s as long as it's obvious why it _is_ there, but comes down to taste. Your call.

PR #183, opened 2018-09-08
Added custom weaving logic for DRG

Toastdeib:
Just a small one for jump weaving. Also cleaned up notes.md while I was at it.

### DIFF
--- a/src/parser/jobs/drg/Weaving.js
+++ b/src/parser/jobs/drg/Weaving.js
@@ -1,0 +1,21 @@
+import CoreWeaving from 'parser/core/modules/Weaving'
+import ACTIONS from 'data/ACTIONS'
+
+const JUMPS = [
+	ACTIONS.JUMP.id,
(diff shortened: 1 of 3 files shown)


src/parser/jobs/drg/Weaving.js
```diff
@@ -0,0 +1,21 @@
...
+
+const JUMPS = [
+	ACTIONS.JUMP.id,
+	ACTIONS.SPINESHATTER_DIVE.id,
+	ACTIONS.DRAGONFIRE_DIVE.id,
+]
+
+export default class Weaving extends CoreWeaving {
+	isBadWeave(weave/*, maxWeaves*/) {
+		for (let i = 0; i < weave.weaves.length; i++) {
```
ackwell: Could probably be simplified w/ a `weaves.some(...)`, but it'll do the trick 👍

Toastdeib: True, I may clean that up later in a maintenance PR.

PR #185, opened 2018-09-12
Maintenance for some NIN modules

Toastdeib:
Did some cleanup for six of my NIN modules. The majority of it should be invisible to end-users, but I did twiddle a couple of thresholds for my suggestions. Also I ran a yarn extract and noticed that it caught a couple of SMN strings that _someone_ missed. :blobmorning:


locale/de/messages.json
```diff
@@ -253,6 +253,8 @@
   "smn.aetherflow.total-drift": "",
   "smn.bahamut.ghost-disclaimer": "",
   "smn.bahamut.title": "",
+  "smn.contagion.suggestions.use-contagion.content": "",
```
ackwell: :blobmorning:

PR #186, opened 2018-09-12
 Show time between GCDs in the list of weaves 

Mikewando:
oGCDs are weaved between two GCDs (or the start of the fight and a
trailing GCD). The weaving module has been updated to track both the
leading and trailing GCDs so that the time between them can be
displayed to the user. This helps users determine if the weaving was
severe or not.

The GCD information can be read by the individual job modules, though
only the BLM module actually seemed to care which GCDs were used. It has
been updated.


Mikewando:
This looked good on my NIN parses, but looking more closely at some healer parses I noticed that code as-written doesn't take into account when casts _start_ so times are somewhat misleading for casters. I'll look further and see what can be done to help there.

PR #189, opened 2018-09-13
WHM Lilies

suShirow:
lilies branch from Vulcwen, revived and tweaked.
Fabricates an event `consumelilies` everytime lilies are consumed.
https://github.com/Vulcwen/xivanalysis/tree/lilies


src/parser/jobs/whm/Gauge.js
```diff
@@ -0,0 +1,53 @@
+/* eslint-disable */
```
ackwell: Nuh-uh.
Probably left over from dev or w/e, code looks fine at a glance.
Doubt it'll flag much but I am gonna req that be removed.

suShirow: I gotta be careful with those, thanks for spotting it

PR #190, opened 2018-09-13
AST: Lightspeed module

suShirow:
Lightspeed module to handle weaving. Also records MP cost savings and casts during lightspeed.
CastTime.js now interprets negative numbers as cast time reductions.

Minor additions:
Added display order for AST modules.
Added 0.5s to weaving cast times.
Added mpCost and mpCostFactor to AST GCDs.


src/parser/jobs/ast/ArcanaSuggestions.js
```diff
@@ -122,6 +124,51 @@ export default class ArcanaSuggestions extends Module {
 		//					<Trans id="ast.arcana-suggestions.messages.header4">Remarks</Trans></Table.HeaderCell>
 		// {this.RenderRemark(artifact)}
 
+		const cardDisplayPanel = [{
```
ackwell: Not a blocking issue - just thinking. Currently, this is being used to show/collapse the entire table.

There's existing precedence to use the accordion with one expando per "entry", though I'm not sure how applicable it is to this case, given the table gives context to itself. Bit 50/50 really, I wonder if there's a better widget altogether?

src/parser/jobs/ast/Lightspeed.js
```diff
@@ -0,0 +1,177 @@
...
+			abilityId: STATUSES.LIGHTSPEED.id,
+		}
+		this.addHook('applybuff', lsBuffFilter, this._onApplyLightspeed)
+		this.addHook('refreshbuff', lsBuffFilter, this._onRefreshLightspeed)
+		this.addHook('removebuff', lsBuffFilter, this._onRemoveLightspeed)
+
+		this.addHook('complete', this._onComplete)
+	}
+
+	_onCast(event) {
```
ackwell: I recognise that this is precisely what `smn/DWT` does, but I'm beginning to question starting the buff at the cast. Crap like swiftcast taking too long to apply and the next gcd still being a hardcast are an unfortunately common occurrence.

Not blocking, but worth considering.

src/parser/jobs/ast/Lightspeed.js
```diff
@@ -0,0 +1,177 @@
...
+
+		return <Fragment>
+			<p>
+				<Trans id="ast.lightspeed.messages.explanation">
+				Some of the applications of <ActionLink {...ACTIONS.LIGHTSPEED} /> include MP savings on heavy healing segments, keeping casts up while on the move and for weaving OGCDs.
+				To further complicate usage, <ActionLink {...ACTIONS.ESSENTIAL_DIGNITY} /> can reduce the cooldown, and the buff can be extended by <ActionLink {...ACTIONS.CELESTIAL_OPPOSITION} />.<br/><br/>
+				At this point of time it's difficult to identify what is optimal, since each fight calls for a different strategy.
+				</Trans>
+			</p>
+			{panels.length === 0 && noCastsMessage}
```
ackwell: I'd rather the output just hide itself completely if there were no casts. Not using lightspeed is possibly a suggestion instead?

src/parser/jobs/ast/Lightspeed.js
suShirow: Maybe start from the timestamp of the buff? I could even get a suggestion out to tell them not to mash too much if we see  `cast` `hardcaststart` `applybuff` in the order. But I have PTSD from logs out of order in `ArcanaTracking.js` so i didn't want to put too much thought into it.

src/parser/jobs/ast/ArcanaSuggestions.js
suShirow: I considered making a UI that would show the Aracana Spread, and let them click "next" and "back". And the info would slide in and out. It would compress it nicely. But I still feel that there are people who would find it easier to just scroll through it. 
Know any widget I could use that would fit the bill? Like a carousel widget or something

How about that PLUS an additional option to "View logs in a list" like what we have here?

PR #191, opened 2018-09-13
Tooltips

ackwell:
This replaces tooltips site-wide with `xivanalysis/tooltips`, ready for when xivdb inevitably isn't updated for 4.4. I got there. End me.


Toastdeib:
A lot of this just looks like black magic fuckery to me, but I didn't see anything obviously wrong. I'll pull down the branch tonight and test it out a bit with my WIP modules to see how she handles.

P.S. Until proven otherwise, I'm going to assume that `HOC` stands for Heart of (the) Cards.

ackwell:
Aite, I've fixed up some crap left over from last night, think we should be good to go.

I'll merge it tonight so I've got the whole weekend to Fix Shit™ - @KhloeLeclair / @MikeMatrix, any opinions would be 👌 if you've got the time.

ackwell:
This is pulling in the `optionalDependencies` which is annoying. There's only one, and it's pretty small, so I'm going to let it slide for now. Need to research how best to handle it for future updates.

PR #194, opened 2018-09-17
RDM Melee Combo and Finishers

Cephid:
I've updated the About to reflect that this is further along than our release.  This PR adds basic melee combo handling, and most of the Finisher logic based on Jump's requirements.  I'm not overly happy with how I had to calculate the finishers, but I couldn't think of a better way for this pass and I wanted to get finishers our before or the week of 4.4.

Added handling (I think?) for Enhanced Scatter 4.4 changes, I'm not 100% sure how es6 will handle that comparison but I don't have any logs to test it with until 4.4 drops.


src/parser/jobs/rdm/Gauge.js
```diff
@@ -301,6 +394,51 @@ export default class Gauge extends Module {
 					<Trans id="rdm.gauge.suggestions.black-mana-lost-why">You lost {this._blackManaLostToImbalance} Black Mana due to overage of White Mana</Trans>
 				</Fragment>,
 			}))
+
+			this.suggestions.add(new TieredSuggestion({
+				icon: ACTIONS.VERHOLY.icon,
+				content: <Trans id="rdm.gauge.suggestions.wasted-verholy-content">
+					Do not use <ActionLink {...ACTIONS.VERFLARE}/> when you should use <ActionLink {...ACTIONS.VERHOLY}/>
```
Toastdeib: Minor (and not a blocker) but you might want to elaborate a bit in the content sections for all 3 of the new suggestions. Maybe something more like
```
Avoid using <ActionLink {...ACTIONS.VERFLARE}/> over <ActionLink {...ACTIONS.VERHOLY}/> when [reasons].
```
where `reasons` is all the valid reasons for flagging a cast as bad (procs, mana imbalance, what have you).

src/parser/jobs/rdm/Combos.js
```diff
@@ -0,0 +1,132 @@
...
+				}
+			})
+		}
+
+		//Process Derped Combos
+		if (derpComboCount > 0) {
+			this.suggestions.add(new TieredSuggestion({
+				icon: ACTIONS.ENCHANTED_REDOUBLEMENT.icon,
+				why: <Plural
+					id="rdm.derpcombos.why"
```
Toastdeib: See the comment on `Gauge.js` about `Plural` tags. Also this is me nitpicking but translation ID conventions (should be something like `rdm.combos.suggestions.derpcombos.why`, and that applies to the other IDs in here too).

Cephid: I'll change my new Gauge suggestion ids with this change as well, however since I'm unsure how it would impact any existing translations I'm going to forgo changing the existing ones at this time.

src/parser/jobs/rdm/Combos.js
```diff
@@ -0,0 +1,112 @@
...
+import {TieredSuggestion, SEVERITY} from 'parser/core/modules/Suggestions'
+import React from 'react'
+import {Plural, Trans} from '@lingui/react'
+
+export default class Combos extends CoreCombos {
+	// Overrides
+	static suggestionIcon = ACTIONS.ENCHANTED_REDOUBLEMENT.icon
+
+	//These actions are considered a combo DERP
+	_derpComboActions = [
```
ackwell: I'd generally advise to keep this sort of meta outside the class itself - you can put variables in root scope, they get scoped to the current file.

src/parser/jobs/rdm/Gauge.js
```diff
@@ -231,6 +259,71 @@ export default class Gauge extends Module {
 			}
 		}
 
+		_handleFinisher(abilityId) {
+			if (abilityId !== ACTIONS.VERFLARE.id && abilityId !== ACTIONS.VERHOLY.id) {
```
ackwell: Personal taste, I'd put this conditional in the caller, not the callee.

src/parser/jobs/rdm/Gauge.js
```diff
@@ -231,6 +259,71 @@ export default class Gauge extends Module {
...
+			const isStoneReady = this.combatants.selected.hasStatus(STATUSES.VERSTONE_READY.id)
+			const white = this._whiteMana
+			const black = this._blackMana
+			const isAccelerationUp = this.cooldowns.getCooldownRemaining(ACTIONS.ACCELERATION.id) === 0
+			let useVerHoly = false
+			let useVerFlare = false
+			let doesntMatter = false
+			let useOnBadProc = false
+
+			if (isStoneReady &&
```
ackwell: Lint's not flagging it so not going to block, but this suite of `if`s could do with some TLC

src/parser/jobs/rdm/Gauge.js
```diff
@@ -301,6 +394,36 @@ export default class Gauge extends Module {
 					<Trans id="rdm.gauge.suggestions.black-mana-lost-why">You lost {this._blackManaLostToImbalance} Black Mana due to overage of White Mana</Trans>
 				</Fragment>,
 			}))
+
+			this.suggestions.add(new TieredSuggestion({
```
ackwell: Duping up a bit on suggestions - these are basically saying the same thing from the opposite side of the mana gauge. Consider consolidating them into a single metric.

src/parser/jobs/rdm/Gauge.js
```diff
@@ -180,8 +203,8 @@ export default class Gauge extends Module {
 
 		_onCast(event) {
 			const abilityId = event.ability.guid
-			//This just lets us determine if we've modified the current Mana numbers at all
-			//const isCalculated = false
+			//Handle Finisher BEFORE adjusting mana
+			this._handleFinisher(abilityId)
```
ackwell: This seems like a lot of stuff to be put in one module - finisher suggestions feel new-module-esque.

Not sure on the semantics of ensuring the finisher handler gets processed _before_ calculating gauge changes, ping me about it.

src/parser/jobs/rdm/Gauge.js
```diff
@@ -80,6 +92,17 @@ export default class Gauge extends Module {
...
 
+		//Finisher Handling
+		_incorrectFinishers = {
+			verholy: 0,
+			verflare: 0,
+			bothprocsup: 0,
+		}
+
+		// The report timestamp is relative to the report timestamp, and in ms. Convert.
+		_parseDate = getPatch(this.parser.parseDate)
```
ackwell: Apparently this doesn't work in es6, blame Khloe :blobmorning:

Suggested putting it in the constructor or a getter.

src/parser/jobs/rdm/Combos.js
Cephid: are you referencing the static suggestIcon or the _derpComboActions?  The first is done to override what's in the base class should I allow the base suggestion to appear.  Is there a functional benefit to keeping it outside the class?  I generally speaking will only put values outside the class that are truly static/const, since I'm modifying this array at runtime I didn't put it outside the class.

src/parser/jobs/rdm/Gauge.js
Cephid: Did someone else do this in a working fashion already?  If so I'll just copy what they did.

PR #201, opened 2018-09-21
Hotfix: E4E status entry

suShirow:
People actually use this now! And they even extend the buff! :(


src/data/STATUSES/ROLE.js
```diff
@@ -36,4 +36,9 @@ export default {
 		name: 'Surecast',
 		icon: 'https://xivapi.com/i/010000/010452.png',
 	},
+	EYE_FOR_AN_EYE: {
+		id: 185,
+		name: 'Eye For An Eye',
```
Toastdeib: Not sure if it's actually necessary to fix it, but I'm going to be a pedant and point out that the capitalization should be "Eye for an Eye" to match the in-game text. :blobmorning:

PR #206, opened 2018-09-25
Added a check for fizzled Wildfires

Toastdeib:
`Wildfire.js` will now include a suggestion if it detects that your Wildfires exploded while the boss was invulnerable (or otherwise did 0 damage because you did a stupid).


src/parser/jobs/mch/Wildfire.js
```diff
@@ -9,6 +9,8 @@ import STATUSES from 'data/STATUSES'
 import Module from 'parser/core/Module'
 import {TieredSuggestion, SEVERITY} from 'parser/core/modules/Suggestions'
 
+const DEBUFF_APPLICATION_BUFFER = 1000 // Buffer for the empty window check, since debuff applications always happen 9 times in the logs for some ungodly reason
```
ackwell: They _what_?

Toastdeib: You haven't seen that? There'll be like 9 consecutive `applydebuff` events (don't remember offhand if the same happens for `applybuff`) every time. Usually they're on the same millisecond but I ran into a few while testing this where some would be 1ms later than the first ones, so I added in wiggle room to be safe.

PR #216, opened 2018-10-03
Add MNK Riddle of Fire and Internal Release tracking

ayyaruq:
The RoF module telling users to fit a TK into RoF may be a bit controversial but even without fancy openers, it's pretty accepted as best practice at this point going by my own feedback and talking to other MNKs.

Pluralisation may be a bit derpy (eg: `<skill> was used 1 times in <buff>`; I intend to fix this soon when I add i18n support.

### DIFF
--- a/src/parser/jobs/mnk/InternalRelease.js
+++ b/src/parser/jobs/mnk/InternalRelease.js
@@ -1,0 +1,143 @@
+import React, {Fragment} from 'react'
+import {Accordion} from 'semantic-ui-react'
+
+import {ActionLink, StatusLink} from 'components/ui/DbLink'
+import Rotation from 'components/ui/Rotation'
+import ACTIONS, {getAction} from 'data/ACTIONS'
+import STATUSES from 'data/STATUSES'
+import Module from 'parser/core/Module'
+import {Suggestion, SEVERITY} from 'parser/core/modules/Suggestions'
+
+const IR_DURATION = STATUSES.INTERNAL_RELEASE.duration * 1000
+
+export default class InternalRelease extends Module {
+	static handle = 'internalrelease'
+	static dependencies = [
+		'suggestions',
+	]
+
+	static title = 'Internal Release'
+
+	_active = false
+	_history = []
+	_release = {}
+	_rushing = false
+
+	_badElixirs = 0
+	_badHowlings = 0
+
+	constructor(...args) {
+		super(...args)
+		this.addHook('cast', {by: 'player'}, this._onCast)
+		this.addHook('removebuff', {by: 'player', abilityId: STATUSES.INTERNAL_RELEASE.id}, this._onDrop)
+		this.addHook('complete', this._onComplete)
+	}
+
+	_onCast(event) {
+		const actionId = event.ability.guid
+
+		if (actionId === ACTIONS.INTERNAL_RELEASE.id) {
+			this._active = true
+			this._release = {
+				start: event.timestamp,
+				end: null,
+				casts: [],
+			}
+
+			const fightTimeRemaining = this.parser.fight.end_time - event.timestamp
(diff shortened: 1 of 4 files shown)


src/parser/jobs/mnk/InternalRelease.js
```diff
@@ -0,0 +1,143 @@
...
+		this.addHook('cast', {by: 'player'}, this._onCast)
+		this.addHook('removebuff', {by: 'player', abilityId: STATUSES.INTERNAL_RELEASE.id}, this._onDrop)
+		this.addHook('complete', this._onComplete)
+	}
+
+	_onCast(event) {
+		const actionId = event.ability.guid
+
+		if (actionId === ACTIONS.INTERNAL_RELEASE.id) {
+			this._active = true
```
Toastdeib: One recommendation here - you should probably use
```js
this.combatants.selected.hasStatus(STATUSES.INTERNAL_RELEASE.id)
```
when checking whether to add skills to a window rather than tracking an `_active` flag manually. It'll be more accurate (since there's a brief delay between hitting the OGCD and gaining the status) and it saves you from having to maintain the flag yourself.

src/parser/jobs/mnk/RiddleOfFire.js
```diff
@@ -0,0 +1,160 @@
...
+		this.addHook('cast', {by: 'player'}, this._onCast)
+		this.addHook('removebuff', {by: 'player', abilityId: STATUSES.RIDDLE_OF_FIRE.id}, this._onDrop)
+		this.addHook('complete', this._onComplete)
+	}
+
+	_onCast(event) {
+		const actionId = event.ability.guid
+
+		if (actionId === ACTIONS.RIDDLE_OF_FIRE.id) {
+			this._active = true
```
Toastdeib: Same comment here as above - `hasStatus()` is a safer bet than a manual `_active` flag.

src/parser/jobs/mnk/RiddleOfFire.js
```diff
@@ -0,0 +1,160 @@
...
+		// Aggregate GCDs under each RoF
+		const rofs = []
+		this._history.forEach(riddle => {
+			rofs.push(riddle.casts.filter(cast => getAction(cast.ability.guid).onGcd).length)
+		})
+
+		if (this._missedGcds > 0) {
+			this.suggestions.add(new TieredSuggestion({
+				icon: ACTIONS.RIDDLE_OF_FIRE.icon,
+				severity: SEVERITY.MAJOR,
```
Toastdeib: You can omit this if it's a `TieredSuggestion`.

src/parser/jobs/mnk/InternalRelease.js
ayyaruq: I think I did this so that double weaves show up. Could also filter on the status instead of action but I guess that's functionally the same as using the combatants `hasStatus` method anyway? I'll take a look when I get home if this was for a specific reason but this module was largely copied from WAR's Berserk tracking.

src/parser/jobs/mnk/RiddleOfFire.js
ayyaruq: Typo, part way through changing it to Tiered and left some old crap in, oops.

Toastdeib: I figured as much, just wanted to make sure it got cleaned up. :blobmorning:

src/parser/jobs/mnk/InternalRelease.js
Toastdeib: I'm pretty sure double weaves will still show up in your windows using `hasStatus()` - I'm using it in DRG for B4B/DS and you sometimes see OGCDs as the first cast of the window when they were double-woven. It actually caused a problem with one of my suggestions, which I have since sorted out.

PR #221, opened 2018-10-05
Add typescript support

Kovensky:
This is just enough to compile and test; but no lint support is added yet. This also doesn't use the TsChecker plugin for build speed.

But it builds and runs :ablobleylines:


package.json
```diff
@@ -94,6 +102,30 @@
...
+      "<rootDir>/src/**/__tests__/**/*.{ts,tsx}",
+      "<rootDir>/src/**/*.(spec|test).{ts,tsx}"
+    ],
+    "transform": {
+      "^.+\\.[tj]sx?$": "ts-jest"
+    },
+    "moduleFileExtensions": [
+      "ts",
+      "tsx"
+    ],
```
Kovensky: These all only specify ts/tsx because they get merged with the already existing jest config that supports js

package.json
```diff
@@ -94,6 +102,30 @@
...
+  "jest": {
+    "collectCoverageFrom": [
+      "src/**/*.{ts,tsx}"
+    ],
+    "testMatch": [
+      "<rootDir>/src/**/__tests__/**/*.{ts,tsx}",
+      "<rootDir>/src/**/*.(spec|test).{ts,tsx}"
+    ],
+    "transform": {
+      "^.+\\.[tj]sx?$": "ts-jest"
```
Kovensky: Jest doesn't run multiple transforms for the same file, and ts-jest will run babel for us because of the `jest.globals['ts-jest'].babelConfig`, so this is safe

package.json
```diff
@@ -5,6 +5,7 @@
   "dependencies": {
     "@babel/polyfill": "^7.0.0",
     "@lingui/react": "^2.5.0",
+    "@types/jest": "^23.3.3",
```
ackwell: Is this supposed to be in dev deps?

package.json
```diff
@@ -94,6 +102,30 @@
     "react-app-rewired": "^2.0.0",
     "react-scripts": "^2.0.3",
     "rimraf": "^2.6.2",
-    "source-map-explorer": "^1.6.0"
+    "source-map-explorer": "^1.6.0",
+    "ts-jest": "^23.10.3",
+    "ts-loader": "^5.2.1",
+    "typescript": "^3.1.1"
+  },
+  "jest": {
```
ackwell: If jest supports loading from a separate config file in this particular scenario, I'd prefer that. Not a huge fan of putting all the config in one file.

ackwell: It doesn't. I tried.

PR #223, opened 2018-10-07
Timeline order

ackwell:
Base code shamelessly stolen from toastyboi

Allows job modules to explicitly control cooldowns in the timeline view. Actions can be sorted, grouped, and merged in a recursive structure (though vis itself can't handle two levels of nesting, I'm leaving the functionality in-place in case vis ever gets replaced).

Example:

image

### DIFF
--- a/src/parser/jobs/nin/Cooldowns.js
+++ b/src/parser/jobs/nin/Cooldowns.js
@@ -7,7 +7,7 @@ export default class Cooldowns extends CoreCooldowns {
 			name: 'Ninjutsu',
 			actions: [
 				{
-					name: 'Mudra',
+					name: 'Mudras',
 					merge: true,
 					actions: [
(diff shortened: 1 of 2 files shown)


src/parser/jobs/nin/Cooldowns.js
```diff
@@ -0,0 +1,39 @@
+import CoreCooldowns from 'parser/core/modules/Cooldowns'
+import ACTIONS from 'data/ACTIONS'
+
+export default class Cooldowns extends CoreCooldowns {
+	static cooldownOrder = [
+		{
+			name: 'Ninjutsu',
+			actions: [
+				{
+					name: 'Mudra',
```
Toastdeib: Kinda nit-picky here but we should pluralize this and call it `Mudras`, I think. Also, we should probably add i18n into this.

PR #225, opened 2018-10-10
Refactor Fists module and add a pie chart

ayyaruq:


src/parser/jobs/mnk/Fists.js
```diff
@@ -28,28 +54,51 @@ export default class Fists extends Module {
...
-	_stanceless = 0
+	_activeFist = STANCES.find(fist => this.combatants.selected.hasStatus(fist)) || STANCELESS
+	_fistUptime = {[STANCELESS]: 0} // Initialise stanceless to prevent weird UI shit
+	_fistGCDs = {[STANCELESS]: 0} // Initialise empty
+
+	_lastFistChange = this.parser.fight.start_time
 
 	constructor(...args) {
 		super(...args)
+		this.addHook('cats', {by: 'player'}, this._onCast)
```
NamelessFaceless: You should probably fix this.

WE DON'T HAVE CATS IN THIS PROJECT.

src/parser/jobs/mnk/Fists.js
```diff
@@ -58,15 +107,105 @@ export default class Fists extends Module {
 			content: <Fragment>
 				Fist buffs are one of your biggest DPS contributors, either directly with <ActionLink {...ACTIONS.FISTS_OF_FIRE} /> or <StatusLink {...STATUSES.GREASED_LIGHTNING_I} /> manipulation with <ActionLink {...ACTIONS.FISTS_OF_EARTH} /> and <ActionLink {...ACTIONS.FISTS_OF_WIND} />.
 			</Fragment>,
-			why: `No Fist buff was active for ${this.parser.formatDuration(this._stanceless)}.`,
+			why: `No Fist buff was active for ${this._fistGCDs[STANCELESS]} GCDs.`,
```
NamelessFaceless: This is a bit nitpicky but, why not something like `${this._fistGCDs[STANCELESS]} GCDs had no fist buff active.` Or could use the plural i18n -- When you implement it.

src/parser/jobs/mnk/Fists.js
Kovensky: :blobcatmorning:

src/parser/jobs/mnk/Fists.js
ayyaruq: Sure, this is a pretty old module and better wording is always good. I intend to implement i18n on the weekend but worth fixing now.

src/parser/jobs/mnk/Fists.js
ayyaruq: Can we get cats though?

ayyaruq:
Changes have been addressed, this should be good to go.

src/parser/jobs/mnk/Fists.js
```diff
@@ -28,28 +54,51 @@ export default class Fists extends Module {
...
+
+		this._fistGCDs[this._activeFist]++
+	}
+
+	_onGain(event) {
+		this.handleFistChange(event.ability.guid)
+	}
+
+	_onRemove() {
+		// This is kinda flaky, there's a chance for remove to run after gain and effectively
```
ackwell: I can almost guarantee it's the diff between

```
...
wind cast
fire remove
wind apply
...
```

and 

```
...
wind cast
wind apply
fire remove
...
```

To work around/with that, I'd say it's probably safe to ignore cases where the stance being removed != current active stance.

src/parser/jobs/mnk/Fists.js
```diff
@@ -58,15 +107,105 @@ export default class Fists extends Module {
...
 			</Fragment>,
-			why: `No Fist buff was active for ${this.parser.formatDuration(this._stanceless)}.`,
+			why: `${this._fistGCDs[STANCELESS]} GCDs had no Fists buff active.`,
 			tiers: STANCELESS_SEVERITY,
-			value: this._stanceless,
+			value: this._fistGCDs[STANCELESS],
+		}))
+
+		// Semi lenient trigger, this assumes RoE is only used during downtime
+		this.suggestions.add(new TieredSuggestion({
```
ackwell: Is it worth having these as seperate suggestions, esp. given the pie is a thing?

src/parser/jobs/mnk/Fists.js
```diff
@@ -58,15 +107,105 @@ export default class Fists extends Module {
...
+			icon: ACTIONS.FISTS_OF_EARTH.icon,
+			content: <Fragment>
+				When using <ActionLink {...ACTIONS.RIDDLE_OF_EARTH} />, remember to change back
+				to <StatusLink {...STATUSES.FISTS_OF_FIRE} /> as soon as possible.
+			</Fragment>,
+			tiers: EARTH_SEVERITY,
+			why: <Fragment>
+				<StatusLink {...STATUSES.FISTS_OF_EARTH} /> was active for
+				{this._fistGCDs[STATUSES.FISTS_OF_EARTH.id]} GCDs.
+				Did you forget to swap back to <StatusLink {...STATUSES.FISTS_OF_FIRE} />?
```
ackwell: Unnecessary line, it'll be too long.

src/parser/jobs/mnk/Fists.js
```diff
@@ -1,22 +1,48 @@
...
-import STATUSES from 'data/STATUSES'
+import ACTIONS, {getAction} from 'data/ACTIONS'
+import JOBS from 'data/JOBS'
+import STATUSES, {getStatus} from 'data/STATUSES'
 
 import Module from 'parser/core/Module'
 import {TieredSuggestion, SEVERITY} from 'parser/core/modules/Suggestions'
 
+// Shamelessly import Muscle Mage, you must cast Fist to continue
+import styles from '../smn/Pets.module.css'
```
ackwell: Copy the styles across or abstract them into a UI component for me. This _will_ break in less than a day (supposing I get off my ass).

src/parser/jobs/mnk/Fists.js
ayyaruq: The situations are different enough I think it's worth highlighting them separately. Staying in Wind is usually a mistake, but Earth can be intentional (even if more damaging to DPS), and no stance is always bad but takes a full GCD to fix usually. By highlighting them individually, the user also can pay attention to the specific instance they do that instead of freaking out about ongoing FoF uptime I guess.

PR #227, opened 2018-10-10
Fix logs that ended on GL3 for uptime calculation

ayyaruq:
Still not really happy with this logic, it basically calculates total GL3 spanning the whole fight even where downtime occurred. Maybe just using `getStatusUptime` with a filter on stacks could work, not sure how that goes for `applybuffstack` buffs. This should at least fix the log for Jojo Zep of Tonberry. Open to any ideas for subtracting downtime segments.

### DIFF
--- a/src/parser/jobs/mnk/GreasedLightning.js
+++ b/src/parser/jobs/mnk/GreasedLightning.js
@@ -313,7 +313,7 @@ export default class GreasedLightning extends Module {
 
 		const statusUptime = this._stacks.reduce((duration, value, index) => {
 			const last = this._stacks[index-1] || {}
-			if (value.stack === 0 && last.stack === GL_MAX_STACKS) {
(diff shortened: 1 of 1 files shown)


ackwell:
>Still not really happy with this logic, it basically calculates total GL3 spanning the whole fight even where downtime occurred.

There's existing handling elsewhere in the project for this sort of case - downtime is not included in the initial uptime cals, and then full fight downtime is removed from the fight time when making the comparison.

Not super clean but it means it's not being ~judgy~ about downtime in downtime.

Going to merge this just to get it working in some capacity, but would be good to clear up the downtime handling when possible.

ayyaruq:
The trouble with the existing uptime stuff is the buff timing is all messed up due to fflogs data, so we need to rely on the module accounting for it. Unless the normaliser fixes the problem and I'm being dumb? Can talk more on discord.

PR #228, opened 2018-10-10
Rewrite of the Gauge module

NamelessFaceless:
This is rewrite of the gauge module. Instead of assuming the user has done every combo correctly, it now uses the `Combo` modules to only attribute `Rage` when the combo is done correctly. Also adds missed combos suggestion.

### DIFF
--- a/src/data/ACTIONS/MRD.js
+++ b/src/data/ACTIONS/MRD.js
@@ -7,7 +7,11 @@ export default {
 		id: 31,
 		name: 'Heavy Swing',
 		icon: 'https://xivapi.com/i/000000/000260.png',
+		potency: 160,
 		onGcd: true,
+		combo: {
+			start: true,
+		},
 	},
 
 	SKULL_SUNDER: {
@@ -15,6 +19,11 @@ export default {
 		name: 'Skull Sunder',
 		icon: 'https://xivapi.com/i/000000/000257.png',
 		onGcd: true,
+		potency: 100,
+		combo: {
+			from: 31,
+			potency: 210,
+		},
 	},
 
 	BUTCHERS_BLOCK: {
@@ -22,6 +31,12 @@ export default {
 		name: 'Butcher\'s Block',
 		icon: 'https://xivapi.com/i/000000/000262.png',
 		onGcd: true,
+		potency: 100,
+		combo: {
+			from: 35,
+			potency: 300,
+			end: true,
+		},
 	},
 
 	OVERPOWER: {
@@ -29,6 +44,8 @@ export default {
 		name: 'Overpower',
 		icon: 'https://xivapi.com/i/000000/000254.png',
 		onGcd: true,
+		potency: 130,
+		breaksCombo: true,
 	},
 
 	TOMAHAWK: {
@@ -36,6 +53,8 @@ export default {
 		name: 'Tomahawk',
 		icon: 'https://xivapi.com/i/000000/000261.png',
 		onGcd: true,
+		potency: 140,
+		breaksCombo: true,
 	},
 
 	MAIM: {
@@ -43,6 +62,12 @@ export default {
 		name: 'Maim',
 		icon: 'https://xivapi.com/i/000000/000255.png',
 		onGcd: true,
+		potency: 100,
+		combo: {
+			from: 31,
+			potency: 200,
+			end: true,
+		},
 	},
 
 	STORMS_PATH: {
@@ -50,6 +75,12 @@ export default {
 		name: 'Storm\'s Path',
 		icon: 'https://xivapi.com/i/000000/000258.png',
 		onGcd: true,
+		potency: 100,
+		combo: {
+			from: 37,
+			potency: 280,
+			end: true,
+		},
 	},
 
 	STORMS_EYE: {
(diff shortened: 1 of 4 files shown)


src/data/ACTIONS/MRD.js
```diff
@@ -7,56 +7,92 @@ export default {
...
 	},
 
 	SKULL_SUNDER: {
 		id: 35,
 		name: 'Skull Sunder',
 		icon: 'https://xivapi.com/i/000000/000257.png',
 		onGcd: true,
+		potency: 100,
+		combo: {
+			from: 31,
```
Kovensky: This is probably a bad idea, but here's a way to avoid having the `id` magic number repeated in more than one place:

```ts
export { actions as default }
const actions = {
  HEAVY_SWING: {
    ...
  },

  SKULL_SUNDER: {
    ...
    combo: {
      get from () { return actions.HEAVY_SWING }
      ...
    }
  }
}
```

Kovensky: Alternative:

```ts
const HEAVY_SWING = {
  ...
}

const SKULL_SUNDER = {
  ...
  combo: {
    from: HEAVY_SWING.id
    ...
  }
}

export default { HEAVY_SWING, SKULL_SUNDER, ... }
```

Toastdeib: Including combo data in the ACTIONS files is the current precedent, as seen here:
https://github.com/xivanalysis/xivanalysis/blob/master/src/data/ACTIONS/ROG.js#L24
https://github.com/xivanalysis/xivanalysis/blob/master/src/data/ACTIONS/RDM.js#L86
https://github.com/xivanalysis/xivanalysis/blob/master/src/data/ACTIONS/LNC.js#L23

While I generally agree with avoiding duplicate magic numbers, I feel like it's acceptable in this specific use case, especially since we're eventually hoping to automate building these files and combo data should come along with that.

src/data/ACTIONS/MRD.js
```diff
@@ -7,56 +7,92 @@ export default {
 		id: 31,
 		name: 'Heavy Swing',
 		icon: 'https://xivapi.com/i/000000/000260.png',
```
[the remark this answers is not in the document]
Toastdeib: It wouldn't be as simple as a single number - the old jobs all seem to have `000000` for the first part of the URL, but newer jobs don't:
https://github.com/xivanalysis/xivanalysis/blob/master/src/data/ACTIONS/RDM.js#L8
Also, same comment as above with generating the data files.

src/data/ACTIONS/WAR.js
```diff
@@ -8,27 +8,31 @@ export default {
 		name: 'Steel Cyclone',
 		icon: 'https://xivapi.com/i/002000/002552.png',
 		onGcd: true,
+		breaksCombo: false,
```
Toastdeib: By no means a blocker as it doesn't hurt anything, but specifying this as `false` isn't strictly necessary.

src/parser/jobs/war/Gauge.js
```diff
@@ -58,47 +58,43 @@ export default class Gauge extends Module {
 
 	constructor(...args) {
 		super(...args)
-		this.addHook('cast', {by: 'player'}, this._onCast)
 		this.addHook('death', {to: 'player'}, this._onDeath)
+		this.addHook('cast', {by: 'player', abilityId: Object.keys(RAGE_SPENDERS).map(Number)}, this._onSpenderCast)
+		this.addHook('cast', {by: 'player', abilityId: Object.keys(RAGE_GENERATORS).map(Number)}, this._onBuilderCast)
```
Toastdeib: You should be using the `'combo'` hook for the builders instead of `'cast'`. It should just work with no other changes, the combo events are fabbed as duplicates of the cast events.

src/data/ACTIONS/WAR.js
NamelessFaceless: I did it for completeness sake. If/when something that checks that ends up happening, it's there. I know it's mostly unnecessary, but figured I'd do it anyway.

src/parser/jobs/war/Gauge.js
NamelessFaceless: Was an oversight from me. Was testing something, and forgot to change it back to combos.

src/data/ACTIONS/WAR.js
Toastdeib: I meant it more in a sense of "missing property is falsey", but that's fair. If we do explicit checks it'll be good to have.

NamelessFaceless: Oh no, I understood. I just don't think it's a big deal to have (or to not have it), toastyboy.

src/data/ACTIONS/MRD.js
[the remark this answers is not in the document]
ayyaruq: @Toastdeib doesn't matter, if you fetch the actual action from xivapi it has an `Icon` key with the full URL path.

Embolden: https://xivapi.com/Action/7520?pretty=1&columns=ID,Name,Icon
Maim: https://xivapi.com/Action/37?pretty=1&columns=ID,Name,Icon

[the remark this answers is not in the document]
ackwell: The number is just rounded to partition the data into manageable folders, I believe. But as per @ayyaruq, this is a non-issue in the short term and automatable in the long.

PR #231, opened 2018-10-11
Convert utilities.js to ts

Kovensky:
Having these functions that are used in many places correctly typed helps type inference get more types right.

### DIFF
--- a/src/data/ACTIONS/index.js
+++ b/src/data/ACTIONS/index.js
@@ -30,7 +30,7 @@ import DRG from './DRG'
 const DEFAULT_GCD_CASTTIME = 0
 const DEFAULT_GCD_COOLDOWN = 2.5
 
-const ACTIONS = {
+const ACTIONS = addExtraIndex(addDefaultValues({
 	...SHARED,
 	...ROLE,
 	...DUTY,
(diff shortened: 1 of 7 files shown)


src/utilities.ts
```diff
@@ -80,39 +84,43 @@ export function extractErrorContext(object) {
...
 				[]
 
-		let closestIndex
-		let closest
+		let closestIndex: number|undefined
+		let closest: number|undefined
 
-		workingValues
-			.map(v => difference(v, value))
+		;(workingValues as ReadonlyArray<string|number>)
```
ackwell: I'm guessing this is required due to (a) the `as` necessitating the brackets, and (b) the brackets necessitating the `;` to prevent being read as a function call?

Is there any other way to handle this? Because my eyes are burning.

src/data/ACTIONS/index.js
```diff
@@ -30,7 +30,7 @@ import DRG from './DRG'
 const DEFAULT_GCD_CASTTIME = 0
 const DEFAULT_GCD_COOLDOWN = 2.5
 
-const ACTIONS = {
+const ACTIONS = addExtraIndex(addDefaultValues({
```
ackwell: I recall leylia having issues with intellisense not picking up the key names when `ACTIONS` was wrapped like this - does this still expose all the keys as before?

src/data/ACTIONS/index.js
```diff
@@ -81,11 +90,15 @@ export const HIT_TYPES = {
...
-addExtraIndex(ACTIONS, 'id')
-addExtraIndex(addDefaultValues(ACTIONS), 'id')
-
 export default ACTIONS
 
 // TODO: warn when falling back?
 // TODO: Return object (w/ caching?) with utility functions a-la wowa's Ability?
+// this actually should have 2 overloads: one for when `id` is `T extends keyof STATUSES`, and one for when `id` is numeric
+// this, and getStatuses, should just be allowed to return undefined for additional safety.
+// TODO: remove the '|| {}' once call sites are type-checked, or at least audited
```
ackwell: yes mom

src/data/ACTIONS/index.js
Kovensky: image

PR #240, opened 2018-10-14
Reworked WAR into the new modules format.

NamelessFaceless:


src/parser/jobs/war/modules/InnerRelease.js
```diff
@@ -28,7 +28,7 @@ export default class InnerRelease extends Module {
 	_ir = {}
 	_history = []
 	_isRushing = false
-	_irTime = 10000
+	_irTime = 10000 /* eslint-disable-line no-magic-numbers */
```
Toastdeib: Don't suppress, fix. You should add a `duration` property to the IR status and use it [thusly](https://github.com/xivanalysis/xivanalysis/blob/1a68c5ba83ea87e4d9b610556a5242c6beb74446/src/parser/jobs/mnk/Demolish.js#L11), or at least declare the raw 10000 in a constant like we're doing in [most other cases](https://github.com/xivanalysis/xivanalysis/blob/f627f475389d9eb90ebf3a81ddfc224d05faafea/src/parser/jobs/smn/modules/Rouse.js#L10) (and should probably also change but that's for another day).

src/parser/jobs/war/modules/InnerRelease.js
```diff
@@ -71,6 +71,7 @@ export default class InnerRelease extends Module {
 
 	_onRemoveIR() {
 		// TODO: You may need to make adjustments so the guard isn't necessary. The applybuff event is fab'd for things at the start of the fight, it may be a go.
+		/* eslint-disable-next-line no-magic-numbers */
```
Toastdeib: More things to fix instead of suppress. You should declare that 5 as a constant, and then you can also use it later if you so desire to display the target Fell Cleave count when generating the module output.

PR #243, opened 2018-10-15
Abstracted pie chart

ackwell:
I can't believe I just spent a chunk of my afternoon getting fucking typescript to read CSS classes. jfc.

Abstracts the pie chart originally used in `smn/Pets` and later ~~stolen~~ borrowed in `mch` and `mnk` into a shared UI component so everyone can use it _without_ breaking shit by referencing SMN's CSS.

This also comes with a lovely bonus of like 19 new files because ~ typescript css module definitions ~. Woo.

CC @Kovensky, @Toastdeib, and @ayyaruq 

Bonus points for anyone who can be bothered trying to sort out whatever causes the build to fail if the definitions aren't there. Would be good to be able to ignore 'em.


package.json
```diff
@@ -105,7 +106,8 @@
     "source-map-explorer": "^1.6.0",
     "ts-jest": "^23.10.3",
     "ts-loader": "^5.2.1",
-    "typescript": "^3.1.1"
+    "typescript": "^3.1.1",
+    "typings-for-css-modules-loader": "^1.7.0"
```
Kovensky: This is actually pretty cool, I just wish it didn't have to write to disk to work :blobcatverysad:

Would be nice if it'd work as a typescript plugin.

src/components/ui/PieChartWithLegend.tsx
```diff
@@ -0,0 +1,86 @@
...
+	additional?: React.ReactNode[],
+}
+
+type Props = ChartComponentProps & {
+	headers?: {
+		label?: React.ReactNode,
+		additional?: React.ReactNode[],
+	},
+	data: DataPoint[],
+}
```
Kovensky: `interface`s are nicer if you don't need to have mapped types or unions. They show up nicer on tooltips and can be extended without having to modify the original type definition.

```ts
interface DataPoint {
  value: number;
  ...
}

interface Props extends ChartComponentProps {
  headers?: ...
}
```

src/components/ui/PieChartWithLegend.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+		const options = merge({
+			responsive: false,
+			legend: {display: false},
+			tooltips: {enabled: false},
+		}, this.props.options)
+
+		return <>
+			<div className={styles.chartWrapper}>
+				<Pie
```
Kovensky: Pie charts are the worst :blobcatnotlikethis:

src/parser/jobs/mnk/Fists.js
```diff
@@ -11,7 +11,7 @@ import Module from 'parser/core/Module'
 import {TieredSuggestion, SEVERITY} from 'parser/core/modules/Suggestions'
 
 // Shamelessly import Muscle Mage, you must cast Fist to continue
-import styles from '../smn/modules/Pets.module.css'
+import styles from 'components/ui/PieChartWithLegend.module.css'
```
Kovensky: This is still a bit scary :woah:

src/parser/jobs/smn/modules/Pets.js
```diff
@@ -288,50 +287,28 @@ export default class Pets extends Module {
...
-						/></td>
-						<td>{this.getPetName(petId)}</td>
-						<td>{this.parser.formatDuration(this._petUptime.get(petId))}</td>
-						<td>{this.getPetUptimePercent(petId)}%</td>
-					</tr>)}
-				</tbody>
-			</table>
-		</>
+		return <PieChartWithLegend
+			headers={{
```
Kovensky: Note that this prevents `PureComponent` from working; `headers` will never be `===` the previous update's `headers`.

Toastdeib:
Having taken a look over this, I have nothing meaningful to contribute to the review. Everything looks fine as far as I can tell.

For real tho, the fuck is up with those file extensions. Typescript pls.

PR #251, opened 2018-10-19
Thank native browser smooth scrolling for less deps and good framerates

Kovensky:
Replaces the pretty scary `react-scroll` with using browser-native smooth scrolling which avoids all of the bad framedrops that spamming `window.scrollTo()` causes.

This uses `IntersectionObserver` to detect which header is currently active, which saves you from a `'scroll'` event listener, and `scrollBy` to scroll without having to worry about the current `scrollY`.


src/components/Analyse/components/ResultSegment.tsx
```diff
@@ -0,0 +1,88 @@
...
+	componentWillUnmount () {
+		this.observer.disconnect()
+		this.positionContext.unregister(this.props.index)
+	}
+
+	render () {
+		const { result } = this.props
+		return <Consumer>{value => {
+			this.positionContext = value
+			return <Segment vertical id={result.name}>
```
Kovensky: I changed this to an `id` for possible future support for anchors; e.g. to link directly to `#Timeline`.

src/components/Analyse/components/ResultSegment.tsx
```diff
@@ -0,0 +1,88 @@
...
+			this.positionContext.unregister(prevProps.index)
+		}
+
+		const ref = ReactDOM.findDOMNode(this) as HTMLElement
+
+		if (ref !== this.ref) {
+			this.observer.unobserve(this.ref!)
+			this.ref = ref
+			this.observer.observe(ref)
+		}
```
Kovensky: This is _probably_ unreachable in at runtime but it's here as a memory leak CYA.

src/components/Analyse/components/ResultSegment.tsx
```diff
@@ -0,0 +1,88 @@
...
+		// just to be ref'd, or we need the ReactDOM hacks. We _need_ the element to have a size so we can't
+		// just jam it in as a 0-size child that wouldn't cause any trouble.
+		this.ref = ReactDOM.findDOMNode(this) as HTMLElement
+		this.observer.observe(this.ref)
+	}
+
+	componentDidUpdate (prevProps: Readonly<Props>) {
+		if (this.props.index !== prevProps.index) {
+			this.positionContext.unregister(prevProps.index)
+		}
```
Kovensky: Could throw as an alternative; I doubt if the index changes that this is going to behave correctly.

src/components/Analyse/components/SegmentPositionContext.tsx
```diff
@@ -0,0 +1,72 @@
...
+			)
+
+			if (active !== this.state.active) {
+				this.setState({ active })
+			}
+		}
+	}
+
+	render () {
+		return <context.Provider value={this.state}>{this.props.children}</context.Provider>
```
Kovensky: Using `this.state` is kind of a dirty approach but avoids creating a new object on every render / having to have a separate memoized context object.

src/parser/core/Parser.js
```diff
@@ -434,10 +441,13 @@ class Parser {
 	 */
 	scrollTo(handle) {
 		const module = this.modules[handle]
-		scroller.scrollTo(module.constructor.title, {
-			offset: -50,
-			smooth: true,
-		})
+		const element = document.getElementById(module.constructor.title)
```
Kovensky: Changing to an `id` also had this nice side effect, instead of having to use a `querySelector` or somehow put the Provider at an even higher level and having to keep track of names in the Provider.

src/components/Analyse/components/ResultSegment.tsx
```diff
@@ -0,0 +1,88 @@
...
+			const active = entry.boundingClientRect.bottom >= OFFSET_FROM_VIEWPORT_TOP
+			this.positionContext.register(this, this.props.index, active)
+		}
+	}
+
+	scrollIntoView () {
+		// there actually is a this.ref!.scrollIntoView method, but it doesn't support offsets
+		scrollBy({
+			// the "+ 1" is needed to actually nudge it enough that the intersection observer detects it
+			// the alternative is to change the ">=" in handleIntersection with a ">"
```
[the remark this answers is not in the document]
Kovensky: Hmm...... I could do it as a static, but I'd need to keep track of each mounted instance in a `Set` 😓

Kovensky:
I merged https://github.com/xivanalysis/xivanalysis/pull/267 into this to get rid of the conflict :blobmorning:

PR #253, opened 2018-10-20
Module.ts + scope creep

ackwell:
This PR:

- Provides types for the FFLogs requests and responses
- Ports core's `Module` to typescript
- Adds the new `@dependency` decorator for use in both JS and TS modules
- And due to the previous, expands the `static dependencies` prop to allow for dependencies to be assigned to a different named prop on the module.

This does not introduce any breaking changes, but should hopefully vastly improve the QoL when working with modules in TS.


src/fflogs.ts
```diff
@@ -0,0 +1,228 @@
...
+	difficulty?: number
+	bossPercentage?: number
+	fightPercentage?: number
+	kill?: boolean
+	partial?: number
+	size?: number
+	standardComposition?: boolean
+
+	start_time: number
+	end_time: number
```
Kovensky: I like to annotate timestamps with a `/** */` doc comment to say in which format it is. I assume this is unix timestamp in seconds, so you'd write `/** Unix timestamp (s) */`.

It helps when you're also dealing with dates that might be ISO-8601 or some other format.

src/fflogs.ts
```diff
@@ -0,0 +1,228 @@
...
+
+export enum AbilityType {
+	PHYSICAL_DOT = 1,
+	HEAL = 8,
+	SPECIAL = 32,
+	MAGICAL_DOT = 64,
+	PHYSICAL_DIRECT = 128,
+	LIMIT_BREAK = 256,
+	MAGICAL_DIRECT = 1024,
+}
```
Kovensky: I discussing with the snek if these shouldn't be hex flags instead (0x1, 0x4, 0x20, 0x40, 0x80, 0x100, 0x400) but it seems fflogs' UI doesn't support hex 🤦‍♀️

src/fflogs.ts
```diff
@@ -0,0 +1,228 @@
...
+	type: string
+
+	sourceID: number
+	sourceIsFriendly: boolean
+	targetID: number
+	targetInstance: number
+	targetIsFriendly: boolean
+
+	// Hell if I know _every_ event key - cop out for anything not explicitly set
+	[key: string]: any
```
Kovensky: If I get motivated enough I'll just make a PR later to fix this... 😖 

FWIW, in my other codebases no `any` are allowed at all unless it really is an incomplete definition, which are normally marked by a `type TODO = any` instead

src/parser/core/Module.ts
```diff
@@ -0,0 +1,273 @@
...
+import 'reflect-metadata'
+
+import {Event, Ability} from 'fflogs'
+import Parser from './Parser'
+
+export const DISPLAY_ORDER = {
+	TOP: 0,
+	DEFAULT: 50,
+	BOTTOM: 100,
+}
```
Kovensky:
```ts
export enum DISPLAY_ORDER {
  TOP = 0,
  DEFAULT = 50,
  BOTTOM = 100,
}
```

src/parser/core/Module.ts
```diff
@@ -0,0 +1,273 @@
...
+	const dependency = Reflect.getMetadata('design:type', target, prop)
+	const constructor = target.constructor as typeof Module
+
+	// Make sure we're not modifying every single module
+	if (constructor.dependencies === Module.dependencies) {
+		constructor.dependencies = []
+	}
+
+	// If the dep is Object, it's _probably_ from a JS file. Fall back to simple handling
+	if (dependency === Object) {
```
Kovensky: Could also write warning in development with a suggested fix.

```js
import {About} from 'parser/modules/About'

/** @type {About} */
@dependency foo
```

Note that the decorator metadata needs to be able to access the concrete value of the constructor.

src/parser/core/Module.ts
```diff
@@ -0,0 +1,273 @@
...
+type HookCallback<T extends Event> = (event: T) => void
+
+interface Hook<T extends Event> {
+	events: T['type'][]
+	filter: Filter<T>
+	callback: HookCallback<T>
+}
+
+export default class Module {
+	static dependencies: Array<string | MappedDependency> = []
```
Kovensky: I prefer `(string | MappedDependency)[]`, but whatever goes here.

I also prefer using `ReadonlyArray` / `ReadonlyMap` / `ReadonlySet` wherever possible, but the decorator has to be able to `.push` to it so this one can't be readonly.

src/parser/core/Module.ts
```diff
@@ -0,0 +1,273 @@
...
+		if (!this._handle) {
+			const handle = this.name.charAt(0).toLowerCase() + this.name.slice(1)
+			console.error(`\`${this.name}\` does not have an explicitly set handle. Using \`${handle}\`. This WILL break in minified builds.`)
+			this._handle = handle
+		}
+		return this._handle
+	}
+	static set handle(value) {
+		this._handle = value
+	}
```
Kovensky: Just a compatibility note: in the current ES class properties spec, you can't just write `static handle = 'foo'` in a class body; that won't invoke setters and will just define a new own property that bypasses the setter. You'd need to do a `Foo.handle = 'foo'` after the class is defined.

This currently only works because `babel-preset-react-app` is forcing babel to transform class fields in the old spec mode: https://github.com/facebook/create-react-app/blob/master/packages/babel-preset-react-app/create.js#L115-L122

Fortunately, these setters don't rely on causing side effects so it'd probably be safe even if transforming in spec mode.

src/parser/core/Module.ts
```diff
@@ -0,0 +1,273 @@
...
+			this._title = this.handle.charAt(0).toUpperCase() + this.handle.slice(1)
+		}
+		return this._title
+	}
+	static set title(value) {
+		this._title = value
+	}
+
+	// DI FunTimes™
+	[key: string]: any;
```
Kovensky: http://www.nooooooooooooooo.com/

src/parser/core/Module.ts
```diff
@@ -0,0 +1,273 @@
...
+		return events
+	}
+
+	/**
+	 * This method is called when an error occurs, either when running
+	 * event hooks or calling {@link Module#output} in this module or
+	 * a module that depends on this module.
+	 * @param {String} source Either 'event' or 'output'
+	 * @param {Error} error The error that occurred
+	 * @param {Object} event The event that was being processed when the error occurred, if source is 'event'
```
Kovensky: You can remove the `{type annotation}` part of the JSDoc and it'll still work right. Right now they're just wrong 😅

src/parser/core/Module.ts
```diff
@@ -0,0 +1,273 @@
...
+	): Hook<T>
+	protected addHook<T extends Event>(
+		events: T['type'] | T['type'][],
+		filter: Filter<T> | HookCallback<T>,
+		cb: HookCallback<T>,
+	): Hook<T>
+	protected addHook<T extends Event>(
+		events: T['type'] | T['type'][],
+		_filter: Filter<T> | HookCallback<T>,
+		_cb?: HookCallback<T>,
```
Kovensky: I generally prefer to name them with an "Arg" suffix (`filterArg`, `cbArg`) because the `_` prefix is otherwise used to mark _unused_ arguments.

src/parser/core/Module.ts
```diff
@@ -0,0 +1,273 @@
...
+		hook.events.forEach(event => {
+			const hooks = this._hooks.get(event)
+			if (!hooks) { return }
+			hooks.delete(hook)
+		})
+	}
+
+	triggerEvent(event: Event) {
+		// Run through registered hooks. Avoid calling 'all' on symbols, they're internal stuff.
+		if (typeof event.type !== 'symbol') {
```
Kovensky: If you just believe the type definitions this looks like unreachable code; better change the definition of `type` to be `string | symbol`.

By the way, if the symbol is created with `Symbol()`, you can take its `typeof` to make a `unique symbol`: `const foo = Symbol(); type Foo = typeof foo`. The `typeof` of a symbol is pretty much the only type that's arguably nominal in typescript (no `unique symbol` is equal to another `unique symbol`).

src/parser/core/Module.ts
```diff
@@ -0,0 +1,273 @@
...
+			}
+
+			// Basic value check
+			return filterVal === eventVal
+		})
+
+		return match
+	}
+
+	output() {
```
Kovensky: `output(): React.ReactChild|false`

otherwise this will require all subclasses to only ever return `boolean`.

src/parser/core/Parser.js
```diff
@@ -122,6 +122,8 @@ class Parser {
 		})
 	}
 
+	_getDepHandle = (dep) => typeof dep === 'string'? dep : dep.handle
```
Kovensky: If you `export` the `interface MappedDependency`, you can refer to it with:

```ts
/** @param {import('./Module').MappedDependency} dep */
```

src/fflogs.ts
ackwell: Despite the comment, the event types are actually near-complete. Just needs some tweaks to filter handling before I can remove the index type.

src/parser/core/Module.ts
ackwell: That's unfortunate, a lot of stuff around the codebase uses that pattern.

I doubt CRA'll be changing that in the short term, so should be OK for now, but worth keeping in mind.

src/parser/core/Parser.js
ackwell: JS is JS. Will be resolved when porting `Parser`.

src/parser/core/Module.ts
ackwell: I _hate_ comments having a concrete effect on code. They shouldn't. Ever. Ever ever ever.

So nah. I'm not going to be pushing for `@dependency` to be used in JS - it's only really there so TS gets proper typing for module deps. That check's just there so people _can_ use it in JS without breaking everything.

PR #255, opened 2018-10-20
Add display order for MNK

ayyaruq:


src/parser/jobs/mnk/modules/DISPLAY_ORDER.js
```diff
@@ -0,0 +1,6 @@
+export default {
```
ackwell: I'd _almost_ prefer non-modules to be not-in-modules-folder now that there actually is a modules folder to not put them in. But that's an issue for another time.

PR #259, opened 2018-10-22
small bug fixes for OGCD module and LL module

rexatlanti:
Reopeners.js is still WIP but I forgot to switch branches. Pls no bulli


src/parser/core/modules/AlwaysBeCasting.js
```diff
@@ -35,7 +35,7 @@ export default class AlwaysBeCasting extends Module {
 			requirements: [
 				new Requirement({
 					name: <Trans id="core.always-cast.gcd-uptime">GCD Uptime</Trans>,
-					percent: this.gcd.getUptime() / fightDuration * 100,
+					percent: Math.min(this.gcd.getUptime() / fightDuration * 100, 100),
```
ackwell: I'd rather this be handled in a display-only manner on the requirement itself.

While it won't have much effect here, in a scenario with multiple requirements, capping one will cause the overall % to skew.

src/parser/core/modules/CooldownDowntime.js
```diff
@@ -59,7 +59,7 @@ export default class CooldownDowntime extends Module {
 		if (downtime < 0) {
 			downtime = 0
 		}
-		const possibleNumberOfUses = Math.floor(fightlength/(cooldown*1000))
+		const possibleNumberOfUses = Math.ceil(fightlength/(cooldown*1000))
```
ackwell: Nit; Looking at this in hindsight, you can probably get rid of some of the calculations if you factor the downtime out before calculating the usage count.

ayyaruq: Why ceiling? This makes sense if you can pre-pull the particular CD but otherwise you'd need to round down.

src/parser/jobs/blm/Reopeners.js
```diff
@@ -0,0 +1,108 @@
+import React, {Fragment} from 'react' // eslint-disable-line
```
ackwell: This entire file is a nope as-is. Never disable `eslint` without _good_ reason, and if you have one, explain it in a comment.

If this file is WIP, it shouldn't be in a PR without a WIP tag.

src/parser/core/modules/CooldownDowntime.js
rexatlanti: Imagine a ogcd with a 3 min cooldown. The fight is 2 min long. How often can you use said cooldown? Now imagine the fight being 4 min long. That's why I changed it

PR #262, opened 2018-10-23
Fix CoP-out-of-Bounds

ayyaruq:
I accidentally the whole Ley Lines.

The original bug was that percentages for CoP time in Ley Lines were overcapping (shouldn't happen), and CoP sometimes lasts longer than Ley Lines in FFLogs data because shit be bananas. While fixing the way CoP windows work relative to Ley Lines, I ended up the whole module.

This looks a lot crazier in the diff than it really is, probably the easiest way to review is to just look at the whole file. The main thing here is that methods are more generic, commented, less prone to errors with weird ACT/FFLogs behavior, and much more DRY.

This follows a pattern that other jobs (DRG/MNK and probably anything else Toast did) use which I think also helps make maintenance easier if someone else needs to look at a job.


ayyaruq:
Big thanks to @rexatlanti and @AkairyuGestalter on Discord for their help in review and suggestions.

src/parser/jobs/blm/modules/Leylines.js
```diff
@@ -58,130 +53,104 @@ export default class Leylines extends Module {
...
-	_onApplyCircleOfPower(event) {
-		if (this._circleOfPowers.current) { this._circleOfPowers.history.push(this._circleOfPowers.current) }
-
-		this._circleOfPowers.current = {
-			start: event.timestamp,
-			stop: null,
-		}
-		this._onApplyBuff(event) // Also track application time for UI
+	// A reminder of man's ability to generate electricity
+	_dontMovePercent(power, lines) {
```
ackwell: \>`dontMovePercent`

ackwell: I was subjected to The Castle so many times during various years of schooling that it's permanently blacklisted from my brain.

PR #264, opened 2018-10-24
Retroactively added changelog to Warrior

NamelessFaceless:


NamelessFaceless:
Personally I'd say it shouldn't matter much, but, I'd argue new entries to the bottom makes more sense (to me), for organization sake.

For the end user it wouldn't make as much sense.

PR #266, opened 2018-10-24
Proc handling cleanup

AkairyuGestalter:
Refactored/streamlined the BLM Proc handling code


src/parser/jobs/blm/modules/Procs.js
```diff
@@ -17,9 +17,16 @@ const THUNDER_ACTIONS = [
...
 
+const PROC_BUFFS = [
+	STATUSES.THUNDERCLOUD.id,
+	STATUSES.FIRESTARTER.id,
+]
+
 // Yeah they're the same duration now, but it could change...
-const THUNDERCLOUD_DURATION = 18000
-const FIRESTARTER_DURATION = 18000
+const BUFF_DURATIONS = {
```
ayyaruq: May want to put these in seconds in `src/data/STATUSES/BLM.js` and then just multiply by `1000` here. Seems to be precedent for other modules and avoids magic number BS.

src/parser/jobs/blm/modules/Procs.js
```diff
@@ -81,35 +75,23 @@ export default class Procs extends Module {
...
-					this.castTime.set(THUNDER_ACTIONS, 0, event.timestamp, event.timestamp) // Note that this cast was 0 time
-					if (event.ability.guid === ACTIONS.THUNDER_III.id) {
-						event.ability.overrideAction = ACTIONS.THUNDER_III_PROC // Mark this T3 as a proc for use elsewhere
-					} else if (event.ability.guid === ACTIONS.THUNDER_IV.id) {
-						event.ability.overrideAction = ACTIONS.THUNDER_IV_PROC // Might as well mark out T4 procs as well...
-					}
-				}
-			}
+	_onGainProc(event) {
+		const status = getStatus(event.ability.guid)
```
ayyaruq: If you literally only use the ID here, might be easier to just assign `statusId` directly without `getStatus()`.

src/parser/jobs/blm/modules/Procs.js
```diff
@@ -118,103 +100,106 @@ export default class Procs extends Module {
...
 	}
 
-	_onRemoveThundercloud(event) {
-		this._thundercloud = false
-		if (event.timestamp >= this._thundercloudWears) { // If this wore off because of time, note it
-			this._droppedT3Ps++
+	// Consolidate old onCast functions into one central function
+	_onCast(event) {
+		// Skip proc checking if we had a corresponding begincast event or the begincast we recorded isn't the same as this spell (ie. cancelled a cast, used a proc)
+		if (!this._castingSpell || this._castingSpell !== event.ability) {
```
ayyaruq: Tiny nitpick, you may want to assign specifically the name or guid just in case data gets weird. Passing around a full object isn't really going to be a performance hit but it is kinda weird.

ackwell: I mean, unless magic happens, this will always fail.
image

src/parser/jobs/blm/modules/Procs.js
```diff
@@ -118,103 +100,106 @@ export default class Procs extends Module {
...
+			if (event.ability.guid === ACTIONS.FIRE_III.id && this._buffWindows[STATUSES.FIRESTARTER.id].current) {
+				this.castTime.set([ACTIONS.FIRE_III.id], 0, event.timestamp, event.timestamp)
+				event.ability.overrideAction = ACTIONS.FIRE_III_PROC
+				this._stopAndSave(STATUSES.FIRESTARTER.id, event.timestamp, false) // Stop the buff and don't count is as a drop in case of latency/timing weirdness
```
ackwell: nit; avoid comments after code like this - it's pretty unreadable

src/parser/jobs/blm/modules/Procs.js
```diff
@@ -80,19 +74,18 @@ export default class Procs extends Module {
 	}
 
 	_onGainProc(event) {
-		const status = getStatus(event.ability.guid)
+		const statusId = getStatus(event.ability.guid).id
```
ayyaruq: If you literally only use the ID, then `event.ability.guid` is already that ID. You don't need to check if it exists in the next block either that way. If you do fetch the object, it's good to validate it before calling on members of it just in case the result of `getStatus()` is bad, so more like this:

```
const status = getStatus(event.ability.guid)

if (!status) {
  return
}

const tracker = this._buffWindows[status.id]
```

Or alternatively:
```
const statusId = event.ability.guid
const tracker = this._buffWindows[statusId]
```

src/parser/jobs/blm/modules/Procs.js
```diff
@@ -146,7 +142,7 @@ export default class Procs extends Module {
 		}
 
 		tracker.current.stop = endTime
-		if (tracker.current.stop >= tracker.current.wear && countDrops) {
+		if (tracker.current.stop - tracker.current.start >= getStatus(statusId).duration * 1000 && countDrops) {
```
ayyaruq: I would probably keep the durations array at the top of the file and do the math there.

src/parser/jobs/blm/modules/Procs.js
```diff
@@ -81,140 +80,149 @@ export default class Procs extends Module {
...
 	}
 
-	_onComplete(event) {
+	_onComplete() {
 		// Finalise buffs for timeline display
-		if (this._buffs[STATUSES.FIRESTARTER.id]) {
-			if (!this._buffs[STATUSES.FIRESTARTER.id].end) {
-				this.loseBuff(event.timestamp, STATUSES.FIRESTARTER)
-			}
+		if (this._buffWindows[STATUSES.THUNDERCLOUD.id].current) {
```
ackwell: Can probably simplify this with a loop over `PROC_BUFFS` down the road.

ackwell:
You'll need to reset to `upstream/master` - this PR was made off your own `master`, so that merge will have caused a divergence.

PR #268, opened 2018-10-25
Fix for the OGCD function in core and the Triple module

rexatlanti:


src/parser/jobs/blm/modules/DISPLAY_ORDER.js
```diff
@@ -3,4 +3,5 @@ export default {
 	LEY_LINES: 2,
 	TRIPLE: 3,
 	WEAVING: 4,
+	REOPENERS: 5,
```
ackwell: Doesn't look like this is in use atm?

src/parser/jobs/blm/modules/Triple.js
```diff
@@ -53,8 +54,25 @@ export default class Triple extends Module {
 			return
 		}
 
-		// Save the event to the DWT casts
-		this._triple.casts.push(event)
+		//stop tracking
```
ackwell: drop me a line on discord, not sure I follow why triple is being closed in the cast hook

src/parser/jobs/blm/modules/Triple.js
```diff
@@ -35,48 +40,55 @@ export default class Triple extends Module {
...
-				end: null,
-				casts: [],
-			}
+		const action = getAction(event.ability.guid)
 
-			this._ctIndex = this.castTime.set('all', 0)
-		}
-
-		if (!this._active || getAction(actionId).autoAttack) {
+		if (!this._active || getAction(action.id).autoAttack) {
```
ayyaruq: You can just `action.autoAttack` here.

src/parser/jobs/blm/modules/Triple.js
```diff
@@ -35,48 +40,55 @@ export default class Triple extends Module {
...
+		this._tripleFlag = false
+		this._triple.end = this.parser.currentTimestamp
+		this._history.push(this._triple)
+		this.castTime.reset(this._ctIndex)
 	}
 
 	output() {
 		const panels = this._history.map(triple => {
-			const numGcds = triple.casts.filter(cast => !getAction(cast.ability.guid).onGcd).length - 1 // 1 is Triplecast itself
+			const numGcds = triple.casts.filter(cast => !getAction(cast.ability.guid).onGcd).length
```
ayyaruq: I just realised I do this too but it's technically fine since the action is pre-validated before the event is pushed to the array.

src/parser/jobs/blm/modules/Triple.js
```diff
@@ -35,48 +40,55 @@ export default class Triple extends Module {
...
+	}
 
-			this.castTime.reset(this._ctIndex)
-		}
+	_stopRecording() {
+		this._active = false
+		this._tripleFlag = false
+		this._triple.end = this.parser.currentTimestamp
+		this._history.push(this._triple)
+		this.castTime.reset(this._ctIndex)
```
ackwell: By resetting `castTime` here in `_stopRecording` - which is called during the following `cast` event - any module that executes _before_ this module will be told that the current cast was instant.

You'll need to reset the cast time when the buff drops.

src/parser/jobs/blm/modules/Triple.js
```diff
@@ -35,48 +40,55 @@ export default class Triple extends Module {
...
-		}
-
-		if (!this._active || getAction(actionId).autoAttack) {
+		if (!this._active || action.autoAttack) {
 			return
 		}
-
-		// Save the event to the DWT casts
-		this._triple.casts.push(event)
+		//check if Triple window is active
```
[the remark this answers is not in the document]
rexatlanti: corrected

ackwell:
This was merged off your `master`, so you'll need to update your fork to match. Check #development for the recent convo, or ping moi or w/e if you need a hand.

I'm real questioning squashing at this point :blobsweat:

PR #280, opened 2018-10-29
Added a rotation watchdog module for DRG

Toastdeib:
New module for DRG that watches for rotational fuckery. It considers 3 rotations valid: CT => FT, FT => CT, and CT => CT (useful in the event of two targets). It also ignores short rotations (sub-11 GCDs) if they were cut short by downtime.


src/parser/jobs/drg/modules/RotationWatchdog.js
```diff
@@ -0,0 +1,176 @@
...
+		ACTIONS.FULL_THRUST.id,
+		ACTIONS.FANG_AND_CLAW.id,
+		ACTIONS.WHEELING_THRUST.id,
+		ACTIONS.IMPULSE_DRIVE.id,
+		ACTIONS.DISEMBOWEL.id,
+		ACTIONS.CHAOS_THRUST.id,
+		ACTIONS.WHEELING_THRUST.id,
+		ACTIONS.FANG_AND_CLAW.id,
+	],
+	[ // The two-target rotation: HT -> DoT -> more DoT (okay stop DoT)
```
ackwell: MANY WHELPS HANDLE IT

src/parser/jobs/drg/modules/RotationWatchdog.js
```diff
@@ -0,0 +1,176 @@
...
+	],
+]
+
+const AOE_ACTIONS = [
+	ACTIONS.DOOM_SPIKE.id,
+	ACTIONS.SONIC_THRUST.id,
+]
+
+const ROTATION_GCD_COUNT = 11
+const GCD_LENGTH = 2500
```
ackwell: I mean I know it's not hyper accurate, but is it worth using `gcd.getEstimate`?

Toastdeib: I considered it and I can make the change if you feel strongly, but I figured a generic 2.5s was fine here since it's just a "did we enter downtime" buffer.

src/parser/jobs/drg/modules/RotationWatchdog.js
Toastdeib: :blobmorning:

src/parser/jobs/drg/modules/RotationWatchdog.js
ackwell: Can resolve if it becomes troublesome down the track.

PR #289, opened 2018-11-05
Add normaliser for pre-pull fists

ayyaruq:
Also, fix a bug where stances can be negative or show at zero, mainly stanceless.

### DIFF
--- a/src/parser/jobs/mnk/modules/Fists.js
+++ b/src/parser/jobs/mnk/modules/Fists.js
@@ -83,32 +83,38 @@ export default class Fists extends Module {
 	normalise(events) {
 		for (let i = 0; i < events.length; i++) {
 			const event = events[i]
-			const abilityId = event.ability.guid
 
 			// Ignore any non-ability events
-			if (!abilityId) {
+			if (!event.ability) {
 				continue
 			}
 
 			// We got a remove (PrecastStatus will handle this), or an initial apply
-			if (['removebuff', 'applybuff'].includes(event.type) && STANCES.includes(abilityId)) {
+			if (['removebuff', 'applybuff'].includes(event.type) && STANCES.includes(event.ability.guid)) {
 				break
 			}
 
 			// Check for any specific casts that imply stance
 			if (event.type === 'cast') {
 				// They dun goofed
-				if (abilityId === ACTIONS.SHOULDER_TACKLE.id) {
+				if (event.ability.guid === ACTIONS.SHOULDER_TACKLE.id) {
 					break
 				}
 
 				// It was a legit Tackle, we know what's up
-				if (Object.keys(STANCE_MAP).map(Number).includes(abilityId)) {
+				if (Object.keys(STANCE_MAP).map(Number).includes(event.ability.guid)) {
+					const status = getStatus(STANCE_MAP[event.ability.guid])
+
 					events.splice(0, 0, {
 						...event,
 						ability: {
-							guid: STANCE_MAP[abilityId],
+							abilityIcon: status.abilityIcon,
+							guid: status.id,
+							name: status.name,
+							type: 1,
 						},
+						targetID: event.sourceID,
+						targetIsFriendly: true,
 						timestamp: this.parser.fight.start_time - 1,
 						type: 'applybuff',
 					})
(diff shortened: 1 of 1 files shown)


src/parser/jobs/mnk/modules/Fists.js
```diff
@@ -20,6 +20,16 @@ const STANCES = [
 	STATUSES.FISTS_OF_WIND.id,
 ]
 
+// Stance mapping for event splicing
+// We don't need Riddle of Wind because a tackle triggers it
+const STANCE_MAP = {
+	[ACTIONS.EARTH_TACKLE.id]: STATUSES.FISTS_OF_EARTH.id,
+	[ACTIONS.FIRE_TACKLE.id]: STATUSES.FISTS_OF_FIRE.id,
+	[ACTIONS.WIND_TACKLE.id]: STATUSES.FISTS_OF_WIND,
```
Toastdeib: `.id`

src/parser/jobs/mnk/modules/Fists.js
```diff
@@ -68,6 +78,48 @@ export default class Fists extends Module {
 		this.addHook('complete', this._onComplete)
 	}
 
+	normalise(events) {
+		for (let i = 0; i < events.length; i++) {
+			const event = events[i]
+			const abilityId = event.ability.guid
+			const startTime = this.parser.fight.start_time
```
Toastdeib: Minor, but you could move this outside of the loop. It's not going to change, so there's no reason to redeclare it every iteration. That, or just don't even bother to declare it as a local const since you only use it in one place anyway.

src/parser/jobs/mnk/modules/Fists.js
```diff
@@ -168,7 +220,7 @@ export default class Fists extends Module {
 	}
 
 	output() {
-		const uptimeKeys = Object.keys(this._fistUptime).map(Number)
+		const uptimeKeys = Object.keys(this._fistUptime).map(Number).filter(fist => fist > 0)
```
Toastdeib: I'm not sure this is doing what you want it to - won't it unconditionally drop `STANCELESS` from the chart? You're filtering the keys, not the values. Assuming the goal is to remove 0s entries, I think the simpler solution is to just drop the initial `[STANCELESS]: 0` when you create the `_fistUptime` object and instantiate it as an empty blob instead.

src/parser/jobs/mnk/modules/Fists.js
ayyaruq: Originally I used it more than once, forgot about it oops.

src/parser/jobs/mnk/modules/Fists.js
ayyaruq: I think my intent here was to do `this._fistUptime[fist] > 0` but I honestly can't remember, this was sitting on my HDD for a few days before I committed it. There was a reason we initialised stanceless as zero but can't quite recall it, I think it was because pre-normaliser, the initial value would get confused if the user was actually stanceless. Do you think just making it an empty blob would work?

src/parser/jobs/mnk/modules/Fists.js
```diff
@@ -68,12 +80,63 @@ export default class Fists extends Module {
 		this.addHook('complete', this._onComplete)
 	}
 
+	normalise(events) {
+		for (let i = 0; i < events.length; i++) {
+			const event = events[i]
+			const abilityId = event.ability.guid
```
ackwell: This could crash if `event.ability` isn't set on an event. Normaliser crashes will crash the entire application stack (intentionally) - let's avoid that.

Can't use  GH suggestions 'cus it's over multiple lines, but something along the lines of other normalisers would be

```js
if (!event.ability) { continue; }
```

src/parser/jobs/mnk/modules/Fists.js
```diff
@@ -68,12 +80,63 @@ export default class Fists extends Module {
...
+				if (abilityId === ACTIONS.SHOULDER_TACKLE.id) {
+					break
+				}
+
+				// It was a legit Tackle, we know what's up
+				if (Object.keys(STANCE_MAP).map(Number).includes(abilityId)) {
+					events.splice(0, 0, {
+						...event,
+						ability: {
+							guid: STANCE_MAP[abilityId],
```
ackwell: not blocking; this will remove other props such as the name and icon - it's not an issue in most cases, as the `guid` is used as the primary key for pretty much every use-case, but it may cause display issues down the road - just keep it in mind if you start seeing cure icons pop up or something hah

PR #296, opened 2018-11-21
SMN Tweaks

ackwell:
What it says on the tin.


src/parser/jobs/smn/modules/DWT.js
```diff
@@ -244,7 +248,8 @@ export default class DWT extends Module {
 						{this.parser.formatTimestamp(dwt.start)}
 						&nbsp;-&nbsp;{numGcds} GCDs
 						{dwt.rushing && <span className="text-info">&nbsp;(rushing)</span>}
-						{noDeathflare && <span className="text-error">&nbsp;(no Deathflare)</span>}
+						{noDeathflare && !dwt.died && <span className="text-error">&nbsp;(no Deathflare)</span>}
```
Toastdeib: Is there any way to i18n this stuff? Feels goofy to have this part be untranslated while the rest of the file is.

NamelessFaceless: I'd assume we could just use a trans tag instead of span? I dunno.

PR #310, opened 2018-12-09
[PLD] Sword Oath tracking

MikeMatrix:
Ought to be in here right?


src/parser/jobs/pld/modules/Oath.js
```diff
@@ -0,0 +1,103 @@
...
+		this.addHook('complete', this._onComplete)
+	}
+
+	_onSwordOathDamage() {
+		if (!this.firstSwordOathAttackFound) {
+			if (this.lastSwordOathApplication == null) {
+				// First hit, but no buff applied yet, so it was up before the fight
+				this.applySwordOath(this.parser.fight.start_time)
+			}
+			this.firstSwordOathAttackFound = true
```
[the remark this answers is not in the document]
MikeMatrix: Forgot about that option to be honest, but yea, makes the most sense to only run it once.

src/parser/jobs/pld/modules/Oath.js
```diff
@@ -0,0 +1,103 @@
...
+
+	_onApplySwordOath(event) {
+		this.applySwordOath(event.timestamp)
+	}
+
+	_onRemoveSwordOath(event) {
+		this.removeSwordOath(event.timestamp)
+	}
+
+	applySwordOath(timestamp) {
```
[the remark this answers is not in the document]
MikeMatrix: ez fix

MikeMatrix:
@Toastdeib I would love to do so, but the problem is the ambiguities that arrive around tracking all stances, or only Shield Oath. That's why I chose to resolve that with only tracking the uptime of Sword Oath.

I can't distinguish between Shield Oath and No Oath if no Oath was changed during the pull and the only way to allow for me to discern if Shield Oath is on, is by the player using Sheltron or Intervention, which is not guaranteed.

I think this approach tackles the same problem from a different angle, without having to tackle side issues like that at all.

Optimally I'd love to track all, so I can measure Gauge building and Gauge usage. But that's out of scope until I find a clear cut way to distinguish No Oath vs Shield Oath.

src/parser/jobs/pld/modules/Oath.js
```diff
@@ -0,0 +1,95 @@
...
+
+	_applySwordOath(timestamp) {
+		if (this._lastSwordOathApplication == null) {
+			this._lastSwordOathApplication = timestamp
+		}
+	}
+
+	_removeSwordOath(timestamp) {
+		if (this._lastSwordOathApplication != null) {
+			this._swordOathWindows = [
```
ackwell: _Should_ be safe to `.push` here I think?

src/parser/jobs/pld/modules/Oath.js
```diff
@@ -0,0 +1,95 @@
...
+		return Math.min(this._swordOathUptime() / fightDuration * 100, 100)
+	}
+
+	_onComplete() {
+		// Make sure we push the last uptime into our list
+		this._removeSwordOath(this.parser.currentTimestamp)
+
+		this.checklist.add(new Rule({
+			name: 'Keep in Sword Oath when possible',
+			description: <Fragment>
```
ackwell: Can use `<>Fragment Syntax</>` :blobmorning:

Also decorators, typescript, etc if ya want

MikeMatrix: Copy Pasta of something I'll change in a later PR anyways, so I didn't bother to change all or just this instance.

i18n soon™

src/parser/jobs/pld/modules/Oath.js
MikeMatrix: Should, but I played on the safe side and opted for an immutable approach

PR #319, opened 2018-12-23
[UI] Rotation Table

MikeMatrix:
A Table view to generally display partial rotations that are part of the parse.

- [x] Data driven UI element
- [x] Allow definitions and display of target values vs actual values
- [x] Jump to timeline
- [x] Document Features
- [x] Implement Rotation Tables into `PLD/Requiescat` and `PLD/FightOrFlight`


MikeMatrix:
Awww alright...
Yea no problem, but Luxon is more comfy T_T

ackwell:
If it's for durations, does `utilities/formatDuration` work? It's not great (nor as nice as the one in `Parser`), but probably best to keep impl consolidated so we can update in clean sweep later.

MikeMatrix:
I actually wanted to use Luxon to reimplement those neater... Because idk why, but the way it's currently done makes me cringe... especially on tables like the one in this PR it just looks weird to have different formatting done to the time, depending on how long the duration is.

MikeMatrix:
Removing the WIP tag, since I don't think I can make it hugely more compact and it seems to be an improvement over previous ways of doing this.

Migrations in case of changes shouldn't be to major so I think this is good to go.

MikeMatrix:
Wasn't integrity removed a few versions ago?

PR #324, opened 2019-01-09
update assize cooldown to be 45s instead of 60

Vulcwen:
Yes, I know the branch name is wrong


NamelessFaceless:
@ackwell how do we feel about adjusting shit for patches like this?

PR #326, opened 2019-01-12
Add DOT application times and tables to indicate clipping breakdown

AzariahAeroborn:
Provide more detail regarding DOT clipping by showing each DOT cast, amount of clipping, and cumulative clipping time similar to the Aetherflow drift table.


ackwell:
The flow drift table already isn't a particularly good resolution of displaying drift which I'd like to re-think. Displaying DoT drift is a good idea, though potentially a good chance to do said re-thinking and apply to both DoT and flow  modules.

In the case of dots, I'd be concerned with how large the table would end up - with "perfect" play, it would be ~2x longer than the flow table (a bit icky), but the worse they play, the more the table would bloat. Flow at least is bounded by the 60s CD.

Drop a line in the caster channel, Nem and/or I would welcome a chance to get more ideas on how to display drift~

PR #337, opened 2019-01-29
ABC module that shows you times where you did literally nothing

rexatlanti:


src/parser/jobs/blm/modules/ABC.js
```diff
@@ -0,0 +1,125 @@
...
+import DISPLAY_ORDER from './DISPLAY_ORDER'
+
+//value to be added to the gcd to avoid false positives
+const GCD_ERROR_OFFSET = 100
+/*since we look at end of a cast => beginning of a new one, we need to account for the time that's still on the gcd since we
+can literally not cast in that period. 1s is too much, but better than substracting not enough*/
+const DURATION_ERROR_OFFSET = 1000
+
+export default class AlwaysBeCasting extends Module {
+	static handle = 'blackmage_abc'
```
ackwell: Avoid prefixing handles. If it'd cause a conflict with another module, consider if it's been appropriately named. Is this ABC, or is it a report on when you _weren't_ ABC-ing?

Don't ask me for name suggestions tho I'm hopeless at that shit

src/parser/jobs/blm/modules/ABC.js
```diff
@@ -0,0 +1,125 @@
...
+		//filter out invuln periods
+		this._noCastWindows.history = this._noCastWindows.history.filter(windows => {
+			return this.invuln.getInvulns('all', windows.start, windows.stop).length === 0
+		})
+		//filter out negative durations
+		this._noCastWindows.history = this._noCastWindows.history.filter(windows => windows.stop - windows.start > DURATION_ERROR_OFFSET)
+	}
+
+	output() {
+		if (this._noCastWindows.history.length > 0) {
```
ackwell: can drop the extra indentation by reversing the conditional

src/parser/jobs/blm/modules/ABC.js
```diff
@@ -0,0 +1,125 @@
...
+			return <Table collapsing unstackable compact="very">
+				<Table.Header>
+					<Table.Row>
+						<Table.HeaderCell><Trans id="blm.abc.timestamp-header">Timestamp</Trans></Table.HeaderCell>
+						<Table.HeaderCell><Trans id="blm.abc.duration-header">Duration</Trans></Table.HeaderCell>
+						<Table.HeaderCell></Table.HeaderCell>
+					</Table.Row>
+				</Table.Header>
+				<Table.Body>
+					{this._noCastWindows.history.map(notCasting => {
```
ackwell: As with the other PR that's open at the moment - I'm really wary of how much this output could inflate. A good player may have sub-5 pops of this... but I've seen more than my fair share of BLMs who stop to pick their nose between each GCD. It may well easily inflate to more than a pageheight.

While I won't hard no this, It's worth discussing resolutions to this issue - as it _would_ be nice to resolve it in some manner.

cc @AzariahAeroborn

To me, this _feels_ like a timeline sorta shtick. But we all know how much the timeline sucks ass. There's already been some discussion about it in #caster_development, worth continuing it.

rexatlanti: Keep in mind the module is very lenient. I don't underestimate our users being bad but the module looks at the time between 2 actions (any actions) and sees if there has been gcd+0.1s inbetween. Any cast, even a canceled one or ogcd or sprint counts here as a try and resets the timer. My aim was that people preferably never see the output and if you do you really should wake the fuck up

rexatlanti: Maybe I explained that wrong. Since the module doesn't know if your last cast was instant or not it assumes a gcd ticking down after your last action and then expects another action a gcd later. If that doesn't happen the module triggers. So you need to basically do nothing for ~2.5-5s for it to trigger

src/parser/jobs/blm/modules/ABC.js
rexatlanti: called it "NotCasting". I'll probably name my children "kid1", "kid2", ... someday

src/parser/jobs/blm/modules/ABC.js
rexatlanti: fixed

rexatlanti:
also want a worst case scenario? Here. Have a max uptime BLM

https://www.fflogs.com/reports/qG6742dtQzWgjrbR#fight=2&source=4

src/parser/jobs/blm/modules/NotCasting.js
```diff
@@ -0,0 +1,126 @@
...
+		//filter out invuln periods
+		this._noCastWindows.history = this._noCastWindows.history.filter(windows => {
+			return this.invuln.getInvulns('all', windows.start, windows.stop).length === 0
+		})
+		//filter out negative durations
+		this._noCastWindows.history = this._noCastWindows.history.filter(windows => windows.stop - windows.start > DURATION_ERROR_OFFSET)
+	}
+
+	output() {
+		//if (!this._noCastWindows.history.length) { return } *dab*
```
ackwell: 👀

PR #343, opened 2019-02-18
Linter cleanup

ayyaruq:
Except that one bit of DRK because half that module needs a logical/maths rethinking I guess. Commits separated by actual change and module so it's easier to inevitably fix something. Also I tested literally none of this but it's prob good.


ayyaruq:
Hold a smidge for #345 so I can remove the DRK stuff here once that's merged.

ayyaruq:
@ackwell removed DRK stuff, idk what Sushi is doing with #344 so I guess merge this and healers adjust.

PR #345, opened 2019-02-19
DRK linter cleanup, improved oGCD tracking and bug fixes

AzariahAeroborn:
Remove normalized uptime calculations for Blood Weapon/Delirium, track as oGCDs instead
Add TieredSuggestion for using Delirium without extending BW/BP
Convert dropped GCD suggestion to TieredSuggestion, fix bug of not actually calling onComplete to add suggestion
Clean up magic constant values in resource simulator

### DIFF
--- a/src/parser/jobs/drk/index.js
+++ b/src/parser/jobs/drk/index.js
@@ -10,12 +10,12 @@ export default {
 	modules: () => import('./modules' /* webpackChunkName: "jobs-drk" */),
 
 	description: <Fragment>
-		<p>This analyzer focuses on blood and mana usage, and then further explores common resource generation problems such as Blood Weapon up-time and dropping GCDs.</p>
+		<p>This analyzer focuses on blood and mana usage, and then further explores common resource generation problems such as Blood Weapon usage and dropping GCDs.</p>
 		<p>Most of the data in this evaluation tool is simulated, and edge cases from FFLogs can cause weird results.  If something doesn't look right, please report it in the discord.</p>
 		<Message warning icon>
 			<Icon name="warning sign"/>
 			<Message.Content>
-				Full capabilities of this tool require logs to be from the current patch.  Logs from 4.2 and before will not be accurately evaluated for resources due to balancing.
+				Full capabilities of this tool require logs to be from the current patch.  Logs from 4.2 and before will not be accurately evaluated for resources due to job changes.
 			</Message.Content>
 		</Message>
 	</Fragment>,
@@ -25,12 +25,17 @@ export default {
 	},
 	contributors: [
 		{user: CONTRIBUTORS.ACRI, role: ROLES.THEORYCRAFT},
-		{user: CONTRIBUTORS.AZARIAH, role: ROLES.DEVELOPER},
+		{user: CONTRIBUTORS.AZARIAH, role: ROLES.MAINTAINER},
 	],
 	changelog: [
 		{
 			date: new Date('2019-01-25'),
 			changes: 'Update analyzer for 4.5 patch, remove inaccurate recommendation for DA usage during enmity combo.',
+			contributors: [CONTRIBUTORS.AZARIAH],
+		},
+		{
+			date: new Date('2019-02-18'),
+			changes: 'Change to tracking Blood Weapon & Delirium as oGCDs instead of normalized uptime.  Add suggestions for dropped GCDs and wasted Deliriums.',
 			contributors: [CONTRIBUTORS.AZARIAH],
 		},
 	],
--- a/src/parser/jobs/drk/modules/Buffs.js
+++ b/src/parser/jobs/drk/modules/Buffs.js
@@ -5,6 +5,8 @@ import {Rule, Requirement} from 'parser/core/modules/Checklist'
 import Module from 'parser/core/Module'
 import ACTIONS from 'data/ACTIONS'
 import STATUSES from 'data/STATUSES'
+import {TieredSuggestion, SEVERITY} from 'parser/core/modules/Suggestions'
+import {Plural} from '@lingui/react'
 
 /* CURRENTLY UNUSED. FUTURE IMPROVEMENT
 // Things that should eventually get flagged if they show up under grit and darkside
@@ -36,10 +38,11 @@ export default class Buffs extends Module {
 	static handle = 'buffs'
 	static title = 'Buffs and Stances'
 	static dependencies = [
-		'downtime',
 		'combatants',
 		'downtime',
 		'checklist',
+		'brokenLog',
+		'suggestions',
 	]
 	// -----
 	// Accessors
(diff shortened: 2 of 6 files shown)


src/parser/jobs/drk/modules/GCDs.js
```diff
@@ -56,13 +56,21 @@ export default class GCDs extends Module {
 	_lastComboGCDTimeStamp = undefined
 	_GCDChainDrops = []
 
+	// Recommendation severities
+	_severityDroppedGCDCombo = {
```
ayyaruq: Normally peeps seem to use a const for this since it's up top and easy to find, plus don't need to do `this._reeeee` to reference it.

src/parser/jobs/drk/modules/GCDs.js
```diff
@@ -95,17 +103,18 @@ export default class GCDs extends Module {
 		}
 	}
 
-	onComplete() {
+	_onComplete() {
 		//dropped combo chain
 		if (this._GCDChainDrops.length > 0) {
```
ayyaruq: I'd pull this block out of the conditional since the TieredSuggestion will handle display for you anyway.

src/parser/jobs/drk/modules/GCDs.js
```diff
@@ -95,17 +103,18 @@ export default class GCDs extends Module {
...
-			this.suggestions.add(new Suggestion({
+			this.suggestions.add(new TieredSuggestion({
 				icon: ACTIONS.SPINNING_SLASH.icon,
 				content: <Fragment>
 					You dropped your GCD combo, loosing out on potency and/or mana.
 				</Fragment>,
-				severity: this._GCDChainDrops.length <= (1) ? SEVERITY.MINOR : this._GCDChainDrops.length <= (6) ? SEVERITY.MEDIUM : SEVERITY.MAJOR,
+				tiers: this._severityDroppedGCDCombo,
```
ayyaruq: As above, could use global.

src/parser/jobs/drk/modules/GCDs.js
```diff
@@ -95,17 +103,18 @@ export default class GCDs extends Module {
...
+				value: this._GCDChainDrops.length,
 				why: <Fragment>
-					You wasted {this._GCDChainDrops} GCD chain actions.
+					You wasted {this._GCDChainDrops.length} GCD chain actions.
```
ayyaruq: I missed this on my side, nice catch :blobmorning:.

src/parser/jobs/drk/modules/ResourceSimulator.js
```diff
@@ -24,23 +24,30 @@ const BLOODSPILLER_BLOOD_COST = 50
...
 
-const BLOOD_PRICE_BLOOD_DAMAGE_TRIGGERED_BLOOD_AMOUNT = 1
-const BLOOD_PRICE_BLOOD_DAMAGE_TRIGGERED_MANA_AMOUNT = 120
+const BLOOD_PRICE_DAMAGE_TAKEN_BLOOD_AMOUNT = 1
+const BLOOD_PRICE_DAMAGE_TAKEN_MANA_AMOUNT = 120
 const BLOOD_PRICE_BLOOD_PASSIVE_RATE = 3000
 const BLOOD_PRICE_BLOOD_PASSIVE_AMOUNT = 4
 const BLOOD_PRICE_MAX_DURATION = 15000
 
+const CARVE_AND_SPIT_NO_DA_MANA_GAIN = 1200
```
ayyaruq: Why not use the `MANA_MODIFIERS` array? Can add `[ACTIONS.CARVE_AND_SPIT.id]: {value: 1200},` for later referencing, since other generators are there alongside spenders.

src/parser/jobs/drk/modules/ResourceSimulator.js
```diff
@@ -24,23 +24,30 @@ const BLOODSPILLER_BLOOD_COST = 50
...
+
 //using this as a lookup table instead of a collector
 const RESOURCE_STATUS_EFFECTS = {
 	[STATUSES.ANOTHER_VICTIM.id]: {
 		duration: 15000,
```
ayyaruq: This will still trip the linter I think. I disabled magic numbers for this whole const declaration.

src/parser/jobs/drk/modules/ResourceSimulator.js
```diff
@@ -24,23 +24,30 @@ const BLOODSPILLER_BLOOD_COST = 50
 // Meters
 // ------
 const MAX_MANA = 9480
+const MANA_REGEN_PERCENT_PER_TICK = 0.02
```
ayyaruq: Linter is dumb with constants and allows magic on single lines but not multi (like you have below). May want to use `2/100` here to be consistent but it's kinda yolo in this case.

src/parser/jobs/drk/modules/ResourceSimulator.js
```diff
@@ -251,7 +258,7 @@ export default class Resources extends Module {
 		if (ACTIONS.CARVE_AND_SPIT.id === abilityId && !this.buffs.darkArtsActive()) {
 			//great why are we here
 			this._noDACarve += 1
-			this.modifyMana(event.ability, 1200)
+			this.modifyMana(event.ability, CARVE_AND_SPIT_NO_DA_MANA_GAIN)
```
ayyaruq: If using the `MANA_MODIFIERS` array as mentioned above, you can just do `this.modifyMana(event.ability, MANA_MODIFIERS[abilityId].value)` here which is a bit cleaner I think.

src/parser/jobs/drk/modules/GCDs.js
AzariahAeroborn: Ah, I aped the way RDM\Dualcast did it.  I can adjust to a module-global const.

src/parser/jobs/drk/modules/ResourceSimulator.js
AzariahAeroborn: One part me doing a minimal change with the refactor tool in my IDE, and one part C&S doesn't always give the 1200 mana gain, so it didn't "fit" within the MANA_MODIFIERS object the way I was looking at it.  I'm ok with changing to using the MODIFIERS object if that's the preferred approach.

src/parser/jobs/drk/modules/ResourceSimulator.js
AzariahAeroborn: Didn't trip it to do individual lines on my testing.  Let me know which style is preferred.

ackwell: Numbers as literal values in objects shouldn't trip the linter - the magic number rule is to prevent unnamed numbers, and the object key acts as a name.

src/parser/jobs/drk/index.js
```diff
@@ -25,13 +25,18 @@ export default {
 	},
 	contributors: [
 		{user: CONTRIBUTORS.ACRI, role: ROLES.THEORYCRAFT},
-		{user: CONTRIBUTORS.AZARIAH, role: ROLES.DEVELOPER},
+		{user: CONTRIBUTORS.AZARIAH, role: ROLES.MAINTAINER},
```
ackwell: image

src/parser/jobs/drk/modules/Buffs.js
```diff
@@ -65,124 +68,144 @@ export default class Buffs extends Module {
...
-	static _parseEventStack(stack) {
-		let out = 0
-		let previous_entry = undefined
-		while (stack.length !== 0) {
-			const entry = stack.shift()
-			if (!entry.active && previous_entry !== undefined && previous_entry.active) {
-				out += entry.timestamp - previous_entry.timestamp
-			}
-			previous_entry = entry
+	// noinspection JSMethodCanBeStatic
```
ackwell: Hm? Is that an intellij thing?

src/parser/jobs/drk/modules/OGCDDowntime.js
```diff
@@ -0,0 +1,14 @@
+import ACTIONS from 'data/ACTIONS'
+import CooldownDowntime from 'parser/core/modules/CooldownDowntime'
+
+export default class OGCDDowntime extends CooldownDowntime {
+	// eslint-disable-next-line no-magic-numbers
```
ackwell: 'aite that bothers me that lint is triggering on this.
google would suggest eslint doesn't support class props like this which is fucking annoying.

src/parser/jobs/drk/modules/ResourceSimulator.js
```diff
@@ -39,8 +42,10 @@ const RESOURCE_STATUS_EFFECTS = {
 	[STATUSES.ANOTHER_VICTIM.id]: {
 		duration: 15000,
 		type: 'debuff',
+		// eslint-disable-next-line no-magic-numbers
 		expire_mana: MAX_MANA * 0.2,
 		expire_blood: 0,
+		// eslint-disable-next-line no-magic-numbers
 		activate_mana: MAX_MANA * 0.3,
```
ackwell: arguably these are a bit magic but the context makes it less so.

PR #346, opened 2019-02-20
killing things is now less deadly for your log

rexatlanti:
(Got it? Got it? :D??)
Previously a bug counted a killing blow on entities as a player death in gauge. Not anymore now

### DIFF
--- a/src/parser/jobs/blm/modules/Gauge.js
+++ b/src/parser/jobs/blm/modules/Gauge.js
@@ -132,7 +132,6 @@ export default class Gauge extends Module {
 	}
 
 	onAstralUmbralTimeout(event) {
-		console.log('timeout', this.parser.formatTimestamp(event.timestamp))
 		this._astralFireStacks = 0
 		this._umbralIceStacks = 0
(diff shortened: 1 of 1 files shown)


ayyaruq:
:blobfacepalm:

src/parser/jobs/blm/modules/Gauge.js
```diff
@@ -104,21 +104,20 @@ export default class Gauge extends Module {
 
 			this.updateStackTimers(event)
 
-			if (!this.parser.byPlayer(event) || !GAUGE_EVENTS.includes(event.type)) {
-				continue
+			if (!GAUGE_EVENTS.includes(event.type)) { continue }
+			if (this.parser.byPlayer(event)) {
+				switch (event.type) {
```
ackwell: For the sake of fixing the bug I'll let it pass, but this is a bit ugly.

You can do stuff like

```js
switch (event.type) {
case 'cast':
	this._onCast(event)
	// fall through
case 'begincast':
case 'damage':
	break;
}
```

PR #347, opened 2019-02-21
fixed a bug with a lot of help that made all F3s instant after a f3p …

rexatlanti:
…and added f3 times into core


src/parser/core/modules/Weaving.js
```diff
@@ -145,6 +146,7 @@ export default class Weaving extends Module {
 
 	// Basic weave check. For job-specific weave concerns, subclass Weaving and override this method. Make sure it's included under the same module key to override the base implementation.
 	isBadWeave(weave, maxWeaves) {
+
```
MikeMatrix: Nitpick: maybe get rid of this line change

PR #357, opened 2019-02-27
Make <Rotation/> great again

MikeMatrix:


src/components/ui/DbLink.js
```diff
@@ -70,9 +75,9 @@ class TooltipBase extends React.PureComponent {
 			</span>
 		}
 
-		const link = <span>
-			{showIcon && <img src={baseUrl + data.icon} alt="" className={styles.image}/>}
-			{showName && <span className={styles.link}>{children || data.name}</span>}
+		const link = <span className={className}>
+			{showIcon && <img src={baseUrl + data.icon} alt="" className={_.compact([styles.image, iconClassName]).join(' ')}/>}
```
ackwell: We've got `classnames` in deps.

src/components/ui/Rotation.module.css
```diff
@@ -1,7 +1,24 @@
-.action {
-	height: 40px;
+.container {
+	--ogcd-size: 30px;
```
ackwell: Doesn't look like you're using the cascade to override the var, probably better to use `@value` instead. It'll get compiled out :blobmorning:

src/components/ui/DbLink.js
```diff
@@ -57,10 +58,14 @@ class TooltipBase extends React.PureComponent {
 			showTooltip,
 			showName,
 			name,
+
+			className,
```
ackwell: Mmh... I'm not a _huge_ fan of passing classes through components. Sort of ties the components together.

The margins can be done in the wrapper, the sizes... not sure. Possibly sanest as a prop though.

MikeMatrix: Personally not a huge fan of introducing a fuck ton of wrapper divs for things, but it does seem more sane

src/components/ui/Rotation.tsx
```diff
@@ -37,17 +38,20 @@ export default class Rotation extends Component<RotationProps> {
...
-				const linkClassName = [styles.link]
-				if (!action.onGcd) {
-					linkClassName.push(styles.ogcd)
-				}
+				const linkClassName = [
+					styles.link,
+					{[styles.ogcd]: !action.onGcd},
+				]
+
+				const iconSize = action.onGcd ? styles.gcdSize : styles.ogcdSize
```
ackwell: You may have forgotten about `@value` but i sure as fuck had forgotten about using it in js from the css import.

MikeMatrix: :yousmart:

PR #358, opened 2019-02-28
Babel TypeScript 2: The Typening.

ackwell:
Do not merge. This will delet mike's stuff. Also everything else. Bad idea.
Will look into this evening.


ackwell:
Shelving.

Ignoring the stupid stuff you need to write to bypass some webpack warnings, babel (understandably) is unable to handle `design:type` decorator metadata - rendering the `@dependency` decorator non-functional.

Issue #353, opened 2019-02-27
Dark knight: blood Gauge tracking graph


szjiajin:
As it is important for DK to align bloodspiller with raid buffs, I think it would be very helpful to add a blood gauge tracking graph like the "Gauge Usage” graph for warrior.
I don't know if I am capable of helping but if you need anything please tell me.

AzariahAeroborn:
Working on getting this implemented, should have a PR ready to go some time this weekend.  Thanks for the feedback.

ackwell:
Resolved in #360

PR #360, opened 2019-03-02
DRK updates

AzariahAeroborn:
- Fix omission of Carve & Spit from tracked oGCDs
- Bug fix (workaround) for logs with two Darkside applybuff events without a removebuff in between.  Assume Darkside is on continuously from the first applybuff event (ignore the "duplicate" event)
- Add Blood gauge graph per #353

### DIFF
--- a/src/parser/jobs/drk/modules/ResourceSimulator.js
+++ b/src/parser/jobs/drk/modules/ResourceSimulator.js
@@ -3,7 +3,7 @@ import React, {Fragment} from 'react'
 import {ActionLink} from 'components/ui/DbLink'
 import ACTIONS from 'data/ACTIONS'
 import STATUSES from 'data/STATUSES'
-import Module from 'parser/core/Module'
+import Module, {DISPLAY_MODE} from 'parser/core/Module'
 import {Suggestion, SEVERITY} from 'parser/core/modules/Suggestions'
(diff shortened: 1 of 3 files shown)


src/parser/jobs/drk/modules/ResourceSimulator.js
```diff
@@ -121,6 +124,7 @@ const COMBO_GRIT_GENERATORS = {
 export default class Resources extends Module {
 	static handle = 'resourcesim'
 	static title = 'Resource Analyzer'
+	static collapsible = false
```
ackwell: This doesn't do anything - it was quickly replaced with `static displayMode` (`Module.ts:14`). If it looks good as is, though, probably don't need to change the mode.

AzariahAeroborn: Ah, I was wondering why it didn't work, I figured I had something screwed up with my local branch along the same lines as why the GCD display wasn't right the other day.  

I'd rather mark this module as non-collapsible, since otherwise it collapses right at the end of the graph and doesn't make it very obvious that there's anything more available.  I might cut the spenders/generators panels in a future rework of the module, but as long as they're there, I think they need to display by default.

ackwell: Fair

PR #365, opened 2019-03-12
add update time of CN 4.4 edition

Bluefissure:
Is this the right way to push a PR to this project?
Known little about typescript ^ ^


MikeMatrix:
Looks good, that would be January 15th, correct?

PR #366, opened 2019-03-13
backfill update times of CN edition

Bluefissure:
It seems patch 4.05 is merged to 4.06 so the update times are same


MikeMatrix:
Might be some slightly undefined behavior with same dates? Maybe just add 1 second to `4.6` so the order is truly defined, but maybe @ackwell can give some more insight.

ackwell:
Non-global editions support sparse patch dates.
If two patches share the same date, the ***newest*** should be dated.
I. E. In this case, 4.05 and 4.06 share a date, so it only needs to be defined for 4.06.

That said, there's no issue in them having the same date.

PR #373, opened 2019-03-19
Small suggestion fix

rexatlanti:
Imaging being 95+ and getting anal about a suggestion intended for noobs 👀 I mean... fixed the severity, since you might not actually lose anything from doing it. If it happens too often, it's def something to watch out. If it costed you an F4, it removes the double suggestion


src/parser/jobs/blm/modules/RotationWatchdog.js
```diff
@@ -178,13 +179,13 @@ export default class RotationWatchdog extends Module {
...
 		//Suggestions for ending UI in T3
-		if (this._UIEndingInT3) {
+		if (this._UIEndingInT3 && !this._missedF4sCauseEndingInT3) {
 			this.suggestions.add(new Suggestion({
 				icon: ACTIONS.THUNDER_III.icon,
 				content: <Trans id="blm.rotation-watchdog.suggestions.ui-ending-in-t3.content">
 					Avoid ending your Umbral Ice with a non-proc <ActionLink {...ACTIONS.THUNDER_III}/>. This can lead to MP issues and fewer <ActionLink {...ACTIONS.FIRE_IV}/> casts under Astral Fire.
 				</Trans>,
-				severity: SEVERITY.MEDIUM,
+				severity: (this._UIEndingInT3 > MAXALLOWEDT3FAILS ? SEVERITY.MEDIUM : SEVERITY.MINOR),
```
ackwell: Might be worth switching to `TieredSuggestion` with this change

PR #387, opened 2019-05-30
BLM Gauge changes & action definition updates

AkairyuGestalter:
Action definitions preliminary until we have real IDs and stuff, but at least calling modules can reference them...

### DIFF
--- a/src/data/ACTIONS/BLM.js
+++ b/src/data/ACTIONS/BLM.js
@@ -39,7 +39,6 @@ export default {
 		name: 'Umbral Soul',
 		icon: 'https://xivapi.com/i/002000/002653.png', // Needs updating once known, still Freeze's icon
 		onGcd: true,
-		castTime: 2.0,
 	},
 	FIRE_I: {
 		id: 141,
--- a/src/data/ACTIONS/BLM.js
+++ b/src/data/ACTIONS/BLM.js
@@ -143,16 +143,10 @@ export default {
 		icon: 'https://xivapi.com/i/000000/000466.png',
 		cooldown: 8,
 	},
-	CONVERT: { // This looks to be going away in ShB, unsure if it's just a name update to Manafont or a completely new action
+	MANAFONT: { // For now, I'm going to assume SE just re-named the action. Updatable if wrong.
 		id: 158,
-		name: 'Convert',
-		icon: 'https://xivapi.com/i/002000/002651.png',
-		cooldown: 180,
-	},
-	MANAFONT: {
-		id: 999999999158, // Needs updated ID
 		name: 'Manafont',
-		icon: 'https://xivapi.com/i/002000/002651.png', // May updated Icon, this is the old Convert Icon, unless it gets re-used
+		icon: 'https://xivapi.com/i/002000/002651.png',
 		cooldown: 180,
 	},
 	MANAWARD: {
@@ -177,7 +171,7 @@ export default {
 		id: 3574,
 		name: 'Sharpcast',
 		icon: 'https://xivapi.com/i/002000/002657.png',
-		cooldown: 60,
+		cooldown: 30,
 	},
 	ENOCHIAN: {
 		id: 3575,
--- a/src/data/ACTIONS/BLM.js
+++ b/src/data/ACTIONS/BLM.js
@@ -34,6 +34,13 @@ export default {
 		onGcd: true,
 		castTime: 3.0,
 	},
+	UMBRAL_SOUL: {
+		id: 999999999159, // ID once known
+		name: 'Umbral Soul',
+		icon: 'https://xivapi.com/i/002000/002653.png', // Needs updating once known, still Freeze's icon
+		onGcd: true,
+		castTime: 2.0,
+	},
 	FIRE_I: {
 		id: 141,
 		name: 'Fire',
@@ -68,6 +75,13 @@ export default {
 		icon: 'https://xivapi.com/i/002000/002652.png',
 		onGcd: true,
 		castTime: 4.0,
+	},
+	DESPAIR: {
+		id: 999999999162, // ID once known
+		name: 'Flare',
+		icon: 'https://xivapi.com/i/002000/002652.png', // Update once this is known, still Flare's icon.
+		onGcd: true,
+		castTime: 3.0,
 	},
 	THUNDER: {
 		id: 144,
@@ -110,6 +124,12 @@ export default {
 		onGcd: true,
 		castTime: 2.5,
 	},
+	XENOGLOSSY: {
+		id: 9999999997422, // ID once known
+		name: 'Xenoglossy',
+		icon: 'https://xivapi.com/i/002000/002664.png', // Needs update once known, still Foul's icon
+		onGcd: true,
+	},
 	SLEEP: {
 		id: 145,
 		name: 'Sleep',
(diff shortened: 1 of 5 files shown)


src/parser/jobs/blm/modules/Gauge.js
```diff
@@ -154,19 +156,21 @@ export default class Gauge extends Module {
...
 	onConsumePolyglot() {
-		if (!this._hasPolyglot && this._overwrittenFoul > 0) {
+		if (!this._polyglotStacks > 0 && this._overwrittenPolyglot > 0) {
 			// Safety to catch ordering issues where Foul is used late enough to trigger our overwrite check but happens before Poly actually overwrites
-			this._overwrittenFoul--
+			this._overwrittenPolyglot--
 		}
-		this._hasPolyglot = false
+		this._polyglotStacks--
+		this._polyglotStacks = Math.max(this._polyglotStacks, 0)
```
ackwell: would generally prefer this be in one line, a-la `this.foo = Math.max(this.foo - 1, 0)`

src/parser/jobs/blm/modules/Gauge.js
```diff
@@ -245,7 +256,7 @@ export default class Gauge extends Module {
 		this._enochianDownTimer.stop = 0
 	}
 
-	_renderLostFouls(time) {
+	_renderLostPolyglots(time) {
```
ackwell: render?

src/data/ACTIONS/BLM.js
```diff
@@ -123,9 +143,9 @@ export default {
 		icon: 'https://xivapi.com/i/000000/000466.png',
 		cooldown: 8,
 	},
-	CONVERT: {
+	MANAFONT: { // For now, I'm going to assume SE just re-named the action. Updatable if wrong.
```
ackwell: Keep the comment until confirmed with final data, but looks like convert's ID hasn't changed in the update. Should be safe.

src/data/ACTIONS/BLM.js
```diff
@@ -34,6 +34,13 @@ export default {
 		onGcd: true,
 		castTime: 3.0,
 	},
+	UMBRAL_SOUL: {
+		id: 999999999159, // ID once known
+		name: 'Umbral Soul',
+		icon: 'https://xivapi.com/i/002000/002653.png', // Needs updating once known, still Freeze's icon
+		onGcd: true,
+		castTime: 2.0,
```
rexatlanti: US is on GCD without a cast time. See https://imgur.com/CDUU3Nk

src/parser/jobs/blm/modules/Gauge.js
```diff
@@ -154,19 +156,21 @@ export default class Gauge extends Module {
...
 
 	onGainPolyglot() {
-		if (this._hasPolyglot) {
-			this._overwrittenFoul++
+		this._polyglotStacks++
+		if (this._polyglotStacks > MAX_POLYGLOT_STACKS) {
+			this._overwrittenPolyglot++
 		}
-		this._hasPolyglot = true
+		this._polyglotStacks = Math.max(this._polyglotStacks, MAX_POLYGLOT_STACKS)
```
rexatlanti: It needs a a math.min. If you get a 3rd stack but your MAX_POLYGLOT_STACKS = 2 then it says 3 here instead of 2.

src/data/ACTIONS/BLM.js
rexatlanti: US is on GCD without a cast time. See https://imgur.com/CDUU3Nk

src/parser/jobs/blm/modules/Gauge.js
```diff
@@ -344,28 +362,28 @@ export default class Gauge extends Module {
...
 
-		if (this._lostFoul) {
+		if (this._lostPolyglot) {
 			this.suggestions.add(new Suggestion({
-				icon: ACTIONS.FOUL.icon,
-				content: <Trans id="blm.gauge.suggestions.lost-foul.content">
-					You lost <ActionLink {...ACTIONS.FOUL}/> due to dropped <ActionLink {...ACTIONS.ENOCHIAN}/>. <ActionLink {...ACTIONS.FOUL}/> is your strongest GCD, so always maximize its casts.
+				icon: ACTIONS.XENOGLOSSY.icon,
+				content: <Trans id="blm.gauge.suggestions.lost-polyglot.content">
+					You lost <ActionLink {...ACTIONS.XENOGLOSSY}/> due to dropped <ActionLink {...ACTIONS.ENOCHIAN}/>. <ActionLink {...ACTIONS.XENOGLOSSY}/> is your strongest GCD, so always maximize its casts.
```
rexatlanti: We should probably have this worded more general as "Polyglot actions" instead of Xeno. Both will be used depending on aoe or sinle target, and I don't wanna have people think they should only use one of them all the time.

`You lost <ActionLink {...ACTIONS.XENOGLOSSY}/> or <ActionLink {...ACTIONS.FOUL}/> due to dropped...`

Might be an idea as well.

src/parser/jobs/blm/modules/Gauge.js
```diff
@@ -344,28 +362,28 @@ export default class Gauge extends Module {
...
 
-		if (this._overwrittenFoul) {
+		if (this._overwrittenPolyglot) {
 			this.suggestions.add(new Suggestion({
-				icon: ACTIONS.FOUL.icon,
-				content: <Trans id="blm.gauge.suggestions.overwritten-foul.content">
-					You overwrote <ActionLink {...ACTIONS.FOUL}/> due to not casting it every 30s. <ActionLink {...ACTIONS.FOUL}/> is your strongest GCD, so always maximize its casts.
+				icon: ACTIONS.XENOGLOSSY.icon,
+				content: <Trans id="blm.gauge.suggestions.overwritten-polyglot.content">
+					You overwrote <ActionLink {...ACTIONS.XENOGLOSSY}/> due to not casting it for 30s after gaining a second Polyglot. <ActionLink {...ACTIONS.XENOGLOSSY}/> is your strongest GCD, so always maximize its casts.
```
rexatlanti: Same as with "lostPolyglot"

src/parser/jobs/blm/modules/Gauge.js
AkairyuGestalter: Fine by me, done.

src/parser/jobs/blm/modules/Gauge.js
AkairyuGestalter: Didn't really look at what the function did, since I was just changing the name now that there's more than one Polyglot spender, but you have a good point. Changed to be even more accurate (_countLostPolyglots)

src/data/ACTIONS/BLM.js
AkairyuGestalter: Yeah, I tried to leave comments where I wasn't completely sure about things. I'll definitely comb back through the definitions once we know IDs and stuff to adjust it if it's wrong.

src/parser/jobs/blm/modules/Gauge.js
AkairyuGestalter: I am a dumb. Fixed.

src/data/ACTIONS/BLM.js
AkairyuGestalter: So it is, I definitely read that tooltip wrong. Thanks for the save, fixed.

src/parser/jobs/blm/modules/Gauge.js
AkairyuGestalter: Redid the wording to try to get at the "Spend something and don't drop or overwrite stacks" point without pushing one or the other (besides switching the icon for the suggestion to be the Xenoglossy one instead). Lemme know if you've got other ideas.

src/parser/jobs/blm/modules/Gauge.js
AkairyuGestalter: See above :)

PR #389, opened 2019-06-03
Implement Ninki resource graph.

rachellhea:
Feature request from Kaede Akamatsu.

Local testing done. Mostly a copy-paste job from the DRK Blood implementation by Mike.

### DIFF
--- a/src/parser/jobs/nin/modules/Ninki.js
+++ b/src/parser/jobs/nin/modules/Ninki.js
@@ -1,3 +1,4 @@
+import {t} from '@lingui/macro'
 import {Trans, Plural} from '@lingui/react'
 import React, {Fragment} from 'react'
 
@@ -25,7 +26,7 @@ const NINKI_SPENDERS = {
 
 export default class Ninki extends Module {
 	static handle = 'ninki'
-	static title = 'Ninki Timeline'
+	static title = t('nin.ninki.title')`Ninki Timeline`
 	static displayMode = DISPLAY_MODE.FULL
 	static dependencies = [
 		'cooldowns',
@@ -33,9 +34,7 @@ export default class Ninki extends Module {
 	]
 
 	_ninki = 0
-	_history = {
-		ninki: [],
-	}
+	_ninkiHistory = []
 	_wasteBySource = {
 		[ACTIONS.MUG.id]: 0,
 		[ACTIONS.ATTACK.id]: 0,
@@ -86,7 +85,7 @@ export default class Ninki extends Module {
 
 	_pushToHistory() {
 		const timestamp = this.parser.currentTimestamp - this.parser.fight.start_time
-		this._history.ninki.push({t: timestamp, y: this._ninki})
+		this._ninkiHistory.push({t: timestamp, y: this._ninki})
 	}
 
 	_onComplete() {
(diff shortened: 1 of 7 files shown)


src/parser/jobs/nin/modules/Ninki.js
```diff
@@ -22,12 +25,17 @@ const NINKI_SPENDERS = {
 
 export default class Ninki extends Module {
 	static handle = 'ninki'
+	static title = 'Ninki Timeline'
```
Toastdeib: This should get i18n'd before we merge it.

src/parser/jobs/nin/modules/Ninki.js
```diff
@@ -22,12 +25,17 @@ const NINKI_SPENDERS = {
...
+	static displayMode = DISPLAY_MODE.FULL
 	static dependencies = [
 		'cooldowns',
 		'suggestions',
 	]
 
 	_ninki = 0
+	_history = {
```
Toastdeib: Minor, but I see no reason not to just have the `_history` variable be an array instead of an object containing an array. It makes sense for RDM (since `_history` tracks both Black and White Mana), but NIN's only got the one resource.

src/parser/jobs/nin/modules/Ninki.js
```diff
@@ -104,4 +121,26 @@ export default class Ninki extends Module {
 			</Trans>,
 		}))
 	}
+
+	output() {
+		const _ninkiColor = Color(JOBS.NINJA.colour)
```
Toastdeib: Super minor, but this should just be `ninkiColor`, no `_`. The underscore is convention for private instance variables, not local ones.

src/parser/jobs/nin/modules/Ninki.js
```diff
@@ -104,4 +121,26 @@ export default class Ninki extends Module {
...
+
+		/* eslint-disable no-magic-numbers */
+		const chartdata = {
+			datasets: [
+				{
+					label: 'Ninki',
```
Toastdeib: I feel like this should be i18n'd as well, but I'm not actually sure how, so I'm fine brushing that one under the carpet unless we have another chart somewhere that does it to use as an example. RDM didn't, I can say that much (I based my version on it).

rachellhea: I can't find one that does this either, so I'm going to skip it for now.

rachellhea:
Local testing result

src/parser/jobs/nin/modules/Ninki.js
rachellhea: All above fixed in 95923f4.

src/parser/jobs/nin/modules/Ninki.js
```diff
@@ -104,4 +120,26 @@ export default class Ninki extends Module {
...
+					steppedLine: true,
+					data: this._ninkiHistory,
+					backgroundColor: ninkiColor.fade(0.8),
+					borderColor: ninkiColor.fade(0.5),
+				},
+			],
+		}
+		/* eslint-enable no-magic-numbers */
+
+		return <Fragment>
```
ackwell: for future ref, you can write fragments as `<> ... </>` instead of `<Fragment> ... </Fragment>`

PR #398, opened 2019-06-22
Rewrite InnerRelease module in TypeScript.

rachellhea:
There are a mix of bug fixes and overall UI adjustments here. Notably, Upheaval and Onslaught stuff doesn't show up on the suggestions in the current build.

I also adjusted the UI to use the RotationTable to be more in line with the Paladin user experience. Dark Knight should probably do the same if it doesn't already (I didn't check).

Some Before and After:

| Before | After |
| --- | --- |
| image | image |
| image | image |


MikeMatrix:
Looking good.
Only glanced over it so far, but seems to be fine.
The only thing that I've seen so far, would be potentially replacing the `Suggestion`s and conditions to show them with `TieredSuggestion`s instead.
It's pretty much just a minor nit-pick really though. I'll give it a better look later.

rachellhea:
> The only thing that I've seen so far, would be potentially replacing the `Suggestion`s and conditions to show them with `TieredSuggestion`s instead.

With how dependent WAR is on proper Inner Release to do damage, I'm hesitant to tier these, especially since Minor issues are hidden by default. Missing a GCD under IR is especially crippling. But that's more philosophical.

src/parser/jobs/war/modules/InnerRelease.tsx
```diff
@@ -0,0 +1,252 @@
...
+
+	get gaugeDumps(): number {
+		return this.rotation
+			.filter(e => GAUGE_DUMPS.includes(e.ability.guid))
+			.length
+	}
+
+	get gcds(): number {
+		return this.rotation
+			.map(e => getAction(e.ability.guid) as TODO)
```
ackwell: Remind me to swap this to `getDataBy` when merging to shb

PR #399, opened 2019-06-27
Shb blm despair thunder

AkairyuGestalter:
Some preliminary work for Despair and Thunder stuff for Shadowbringers. More to come later.

### DIFF
--- a/src/parser/jobs/blm/modules/RotationWatchdog.js
+++ b/src/parser/jobs/blm/modules/RotationWatchdog.js
@@ -37,6 +37,7 @@ export default class RotationWatchdog extends Module {
 	static displayOrder = DISPLAY_ORDER.ROTATION
 
 	static dependencies = [
+		'checklist',
 		'suggestions',
 		'gauge', // eslint-disable-line xivanalysis/no-unused-dependencies
 		'invuln',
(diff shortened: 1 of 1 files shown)


src/parser/jobs/blm/modules/RotationWatchdog.js
```diff
@@ -244,6 +269,20 @@ export default class RotationWatchdog extends Module {
 				</Trans>,
 			}))
 		}
+
+		this.checklist.add(new Rule({
+			name: <Trans id="blm.rotation-watchdog.checklist.dots.name">Keep your <StatusLink {...STATUSES.THUNDER_III} /> DoT up</Trans>,
+			description: <Trans id="blm.rotation-watchdog.checklist.dots.description">
+				Your <StatusLink {...STATUSES.THUNDER_III} /> DoT contributes significantly to your overall damage, both on its own, and from additional <StatusLink {...STATUSES.THUNDERCLOUD} /> procs. Try to keep the DoT applied.
```
rexatlanti: I'm not 100% happy with the wording here but dunno something better... basically I don't want people to spam t3ps because that's a pps loss. You always try to maximize your dot uptime with the least amount of thunder casts neccessary. To me the suggestion sounds a bit like "casting lots of thunder = good" when it has to be more nuanced

ackwell:
No idea why github thought auto-closing this was a good idea. @AkairyuGestalter if you re-target onto `master` (it's not letting me do it), should be able to re-open.

PR #402, opened 2019-06-27
Initial pass for 5.0 DRG

Toastdeib:
Still got few TODOs, mainly updating the rotation watchdog module and fixing the whole Raiden Thrust thing (how do combos), so those'll be coming along in the next few days.


Toastdeib:
The deed is done.

PR #403, opened 2019-06-28
First pass at GNB support.

rachellhea:
Still need icons from XIV API for statuses and actions, and also need status IDs. Waiting on dataminers for those.

Unfortunately hard to test this since we don't have logs for GNB yet (obviously). One potential bug that I noted in a self-review: I think spamming Solid Barrel would trick the Ammo module into thinking you had a ton of ammo, but only the combo'd variant gets the ammo increase.


rachellhea:
Quick pre-maint update: acchan got me the status IDs, so made those fixes rq.

src/parser/jobs/gnb/index.tsx
```diff
@@ -4,21 +4,26 @@ import CONTRIBUTORS, {ROLES} from 'data/CONTRIBUTORS'
...
 export default new Meta({
 	modules: () => import('./modules' /* webpackChunkName: "jobs-gnb" */),
 
 	Description: () => <TransMarkdown source={description}/>,
-	// supportedPatches: {
-	// 	from: '✖',
-	// 	to: '✖',
-	// }
+
+	supportedPatches: {
```
ackwell: Probably isn't quite ready to be marked as supported - leaving this commented out will still let people use the job, but it'll be quite explicitly marked as not ready.

If you're happy with the existing logic and feel it _is_ ready to be marked as supported, should clean up the description so we're not dropping TODOs on users for a supported job.

src/parser/jobs/gnb/modules/Ammo.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+const AMMO_SPENDERS = {
+	[ACTIONS.GNASHING_FANG.id]: 1,
+	[ACTIONS.BURST_STRIKE.id]: 1,
+	[ACTIONS.FATED_CIRCLE.id]: 1,
+}
+
+const MAX_AMMO = 2
+
+class AmmoState {
```
ackwell: Given how it's being used, this should probs be an `interface`.
I'm also not seeing any paths where `time` or `ammo` could be `null` at a glance (apologies if I missed something) - so you can probs reduce the type on both of those to just `number`. As an `interface`, you wouldn't need the initialiser, either.

Off-topic for the above, I'll _generally_ prefer `foo?: type` over `foo: type | null`. There's absolutely a bunch of stuff in the existing codebase that uses `null` instead of `undefined`, but I'm generally moving away from that to a consistent `undefined`.

src/parser/jobs/gnb/modules/Ammo.tsx
```diff
@@ -0,0 +1,149 @@
...
+	time: number | null = null
+	ammo: number | null = null
+}
+
+// TODO: Check for bad Burst Strikes (i.e. when at 1 ammo + Bloodfest is on CD + Gnashing is off CD)
```
ackwell: just a suggestion: might be easier for other devs if analytical TODOs were maintained in a job-level todo file. RDM has a pretty good example.

src/parser/jobs/gnb/modules/Ammo.tsx
```diff
@@ -0,0 +1,149 @@
...
+		this.ammo = Math.max(this.ammo - AMMO_SPENDERS[event.ability.guid], 0)
+		this.pushToHistory()
+	}
+
+	private onDeath() {
+		this.dumpRemainingResources()
+	}
+
+	private dumpRemainingResources() {
+		this.leftoverAmmo = this.ammo
```
ackwell: This'll always be overwritten by the `dumpRemainingResources` call in `onComplete`. Might be worth adding an extra key to `wasteBySource` for the "death/end of fight" case, and adding to that.

src/parser/jobs/gnb/modules/Ammo.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+		const totalWaste = Object.keys(this.wasteBySource)
+			.map(Number)
+			.reduce((sum, source) => sum + this.wasteBySource[source], 0)
+			+ this.leftoverAmmo
+
+		this.checklist.add(new Rule({
+			name: t('gnb.ammo.waste')`Cartridge Waste`,
+			description: <Trans id="gnb.ammo.waste.content">
+				Wasted cartridges or dying with cartridges loaded is a potency loss of about X per cartridge wasted.
```
ackwell: Placeholder~

Can probably word it without the potency to avoid maintaining the text (text changes will invalidate the translations), or if you'd like to keep it, compute it as a variable beforehand such that the translation can just treat it as a black box.

src/parser/jobs/gnb/modules/Ammo.tsx
```diff
@@ -0,0 +1,149 @@
...
+	output() {
+		const ammoColor = Color(JOBS.GUNBREAKER.colour)
+
+		/* eslint-disable no-magic-numbers */
+		const chartdata = {
+			datasets: [
+				{
+					label: 'Cartridges',
+					steppedLine: true,
+					data: this.ammoHistory,
```
ackwell: does this work? it's going straight into chartjs which _iirc_ requires a particular format.

src/parser/jobs/gnb/modules/Cooldowns.tsx
```diff
@@ -0,0 +1,43 @@
+import ACTIONS from 'data/ACTIONS'
+import CoreCooldowns from 'parser/core/modules/Cooldowns'
+
+export default class Cooldowns extends CoreCooldowns {
+	static cooldownOrder = [
+		// Stance
+		ACTIONS.ROYAL_GUARD.id,
```
ackwell: hm... not sure how well this'll show up. given how tank stances work now, i should probably have a chat with you tank peeps about how best to display stances (if it's worth it at all? idk).

src/parser/jobs/gnb/modules/OGCDDowntime.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+export default class AbilityDowntime extends CooldownDowntime {
+	allowedDowntime = 2500
+	trackedCds = [
+		ACTIONS.DANGER_ZONE.id,
+		ACTIONS.BLASTING_ZONE.id,
+		ACTIONS.ROUGH_DIVIDE.id,
+		ACTIONS.BOW_SHOCK.id,
+	]
+	// account for charge system? big question mark if this even works
```
ackwell: probably need to sync up with @rexatlanti about that; most jobs have charges somewhere so it'll need to be tracked in some manner.

src/parser/jobs/gnb/modules/NoMercy.tsx
```diff
@@ -0,0 +1,261 @@
...
+
+	// Don't check for correct Continuations; that will be covered by the Continuation module.
+	// Don't check for correctness on the Gnashing Fang combo; that's covered by the built-in Combo tracker.
+}
+
+const NO_MERCY_BUFF_DURATION = 20000 // in milliseconds
+
+class NoMercyState {
+	start: number
+	end: number | null = null
```
ackwell:
```suggestion
	end?: number
```

src/parser/jobs/gnb/modules/Continuation.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+	private ripBuffCounter = 0
+	private tearBuffCounter = 0
+	private gougeBuffCounter = 0
+	private ripCounter = 0
+	private tearCounter = 0
+	private gougeCounter = 0
+
+	protected init() {
+		this.addHook('cast', {by: 'player'}, this.onTrigger)
```
ackwell: Given you're summing the values at the end, you can probably simplify this handling such that it only hooks casts it needs to know about. Something _vaguely_ along the lines of...

```js
ACTIONS = [RIP, TEAR, GOUGE]
STATUSES = [READY_TO_RIP, READY_TO_TEAR, READY_TO_GOUGE]

this.addHook('cast', {by: 'player', abilityId: ACTIONS}, () => this.actions++)
this.addHook('applybuff', {by: 'player', abilityId: STATUSES}, () => this.buffs++)
```

(also saves the switch)

rachellhea:
Big update after going through testing routines. Should be mostly clean now aside from style. Gonna address those comments next.

src/parser/jobs/gnb/modules/Ammo.tsx
rachellhea: Fixed this, turns out it did *not* work.

src/parser/jobs/gnb/index.tsx
rachellhea: Fixed this up until we're ready to mark it as supported. Also updated the description.

src/parser/jobs/gnb/modules/Ammo.tsx
rachellhea: This is something that I think needs more dev time; I'll add it to the list of things to get done.

src/parser/jobs/gnb/NOTES.md
```diff
@@ -0,0 +1,17 @@
+# GNB Feature Suggestions
```
ackwell: 😩 👌

PR #405, opened 2019-07-06
We double weaving now, bois (also fixed a smol Ninki bug I introduced)

Toastdeib:


Toastdeib:
Changelog added.

Toastdeib:
Also added a tiny bugfix for the Kassatsu module because why not (please don't yell at me).

src/parser/jobs/nin/modules/Weaving.js
```diff
@@ -33,7 +35,7 @@ export default class Weaving extends CoreWeaving {
 		let checkState = STATE.NORMAL
 		let tcjCount = 0
 		let ninjutsuCounted = false
-		let dwadDupe = false
+		let dreams = 0
```
ackwell: don't let your dreams be dreams

PR #406, opened 2019-07-06
BLM ShB Action ID and Rotation Tweaks 

mdhausman:
- Update new BLM actions with IDs and icons, remove TODO-style comments
- Update Gauge; Despair does not consume Umbral Hearts
- Update expected fire 4 after Manafont from 2 to 1
    - Despair -> Manafont -> Fire 4 -> Despair is now preferred

### DIFF
--- a/src/data/ACTIONS/BLM.js
+++ b/src/data/ACTIONS/BLM.js
@@ -35,9 +35,9 @@ export default {
 		castTime: 3.0,
 	},
 	UMBRAL_SOUL: {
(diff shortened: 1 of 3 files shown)


mdhausman:
Oh heck, this was already done in #399 I think. Once that gets merged this can just be scrapped I think.

mdhausman:
I hadn't noticed #399 before writing this, but upon double checking the code in there I don't think these conflict.

rexatlanti:
lööks good, thank you :3

PR #407, opened 2019-07-06
Warrior IR copy cleanup

AzariahAeroborn:
Make missedGaugeDumps (incorrect GCD in IR window) more descriptive, clarify when Decimate is better than Fell Cleave


AzariahAeroborn:
While I was stealing this module for DRK's Delirium window, I noticed that the check for incorrect GCDs in an IR window could result in confusing recommendations (if a user missed 3 GCDs and used 1 non-gauge GCD, the recommendation for incorrect GCDs would say 4 incorrect GCDs used).  I also updated the copy for using Decimate vs Fell Cleave so it doesn't imply Decimate is better against 2 targets.

NamelessFaceless:
lgtm, I'll finish updating the module to SHB later in the week.

PR #412, opened 2019-07-08
5.0 WHM Support

kkragenbrink:
Todo:
- [X] Updated cast time for Cure and Medica II.
- [X] Added support for Dia, Glare, Afflatus Misery, Afflatus Solace, Afflatus Rapture, Temperance
- [X] Updated DoTs to check Dia uptime, and removed Aero uptime checking.
- [x] Update Gauge.js for new Lilies system.
- [x] Verify Afflatus Misery usage.


src/parser/jobs/whm/modules/Lilies.js
```diff
@@ -0,0 +1,191 @@
...
+	 * Because lilies are generated every 30 seconds, and we don't want to
+	 * be constantly checking, we run a while loop every time the event is
+	 * more recent than when we should have gained a lily to catch up on our
+	 * lily generation.
+	 *
+	 * @param {*} event
+	 */
+	_calculateLilies(event) {
+		if (this._lastGained === 0) { this._lastGained = this.parser.fight.start_time }
+		while (event.timestamp > this._lastGained + LILY_INTERVAL) {
```
ackwell: Hm - while I doubt the loop will execute more than 0-1 times per event, I'm not particularly comfortable with it nonetheless.

Should be possible to calculate directly from the delta, something like

```js
// untested, batteries not included. manufacturer provides no warranty for suitability in a production environment
const delta = event.timestamp - this._lastGained
const newLilies = Math.floor(delta / LILY_INTERVAL)
this._lastGained += newLilies * LILY_INTERVAL

// modify _lilies/_totalLilies/_unusedLilies here etc
```

src/parser/jobs/whm/modules/DoTs.js
```diff
@@ -85,47 +77,29 @@ export default class DoTs extends Module {
...
-				why: <Trans id="whm.dots.suggestion.clip-a2.why">
-					{this.parser.formatDuration(this._clip[STATUSES.AERO_II.id])} of {STATUSES.AERO_II.name} lost to early refreshes.
-				</Trans>,
-			}))
-		}
-		if (this._clip[STATUSES.AERO_III.id] > DOT_CLIPPING_THRESHOLD) {
-			const isMinor = this._clip[STATUSES.AERO_III.id] <= CLIP_MAX_MINOR[STATUSES.AERO_III.id]
-			const isMedium = this._clip[STATUSES.AERO_III.id] <= CLIP_MAX_MEDIUM[STATUSES.AERO_III.id]
+		if (this._clip[STATUSES.DIA.id] > DOT_CLIPPING_THRESHOLD) {
+			const isMinor = this._clip[STATUSES.DIA.id] <= CLIP_MAX_MINOR[STATUSES.DIA.id]
```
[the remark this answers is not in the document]
kkragenbrink: will do

src/parser/jobs/whm/index.js
```diff
@@ -26,23 +25,35 @@ export default new Meta({
...
 			<Message.Content>
 				<Trans id="whm.about.description.invulnwarning">Currently, for all healers, boss invulnerability checking doesn't function properly. This results in inaccuracy for many time-related functions such as the Always be casting checklist item or DoT uptime.</Trans>
 			</Message.Content>
 		</Message>
 	</>,
-	// supportedPatches: {
-	// 	from: '4.05',
-	// 	to: '4.5',
-	// },
+	supportedPatches: {
```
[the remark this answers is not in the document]
kkragenbrink: As supported as it can be once I get through your changes.  WHM hasn't changed a lot from 4.5 except the addition of the lilies, and I've added lilies to our tracking and updated everything else.  I absolutely would like to consider ways to deepen the support, but the important stuff is all here.

src/parser/jobs/whm/modules/DoTs.js
```diff
@@ -85,47 +77,29 @@ export default class DoTs extends Module {
...
+			const isMedium = this._clip[STATUSES.DIA.id] <= CLIP_MAX_MEDIUM[STATUSES.DIA.id]
 			this.suggestions.add(new Suggestion({
-				icon: ACTIONS.AERO_III.icon,
-				content: <Trans id="whm.dots.suggestion.clip-a3.content">
-					Avoid refreshing DoTs significantly before their expiration, this will allow you to cast more Stone IV.
+				icon: ACTIONS.DIA.icon,
+				content: <Trans id="whm.dots.suggestion.clip-dia.content">
+					Avoid refreshing DoTs significantly before their expiration, this will allow you to cast more Glare. (Note: We do not yet consider using Dia for initial damage on the move)
```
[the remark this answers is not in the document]
kkragenbrink: The note was already there in the pre-5.0 stuff, but I'll remove it at least in all the languages I speak.  I can't do Japanese, unfortunately.

src/parser/jobs/whm/modules/Lilies.js
kkragenbrink: fixed

src/parser/jobs/whm/modules/Lilies.js
```diff
@@ -0,0 +1,182 @@
...
+		const delta = event.timestamp - this._lastGained
+		const newLilies = Math.floor(delta / LILY_INTERVAL)
+		this._lastGained += newLilies * LILY_INTERVAL
+
+		this._lilies = Math.min(MAX_LILIES, this._lilies + newLilies)
+		this._totalLilies += newLilies
+		if (this._lilies + newLilies > MAX_LILIES) {
+			this._unusedLilies += this._lilies + newLilies - MAX_LILIES
+		}
+	}
```
AzariahAeroborn: I believe this incorrect handling for lilies.  Based on my observations, and testing against a dummy, the charge timer does not run at all while you are at 3 lilies.  Once you hit 3 lilies, no timer is running until you use a lily, at which point the timer starts at 0 and takes 30 seconds to gain you another lily.

kkragenbrink: I will double check when I get home. If that's true then this is a more prominent issue than I thought. Doesn't change much though since what I am predicting is how many you could have had.

kkragenbrink: @AzariahAeroborn You are correct.  Once at 3 lilies, the timer stops counting entirely.  I still need to track it the way I am (to know how many you _could_ have generated), but I also need to track how many you _actually_ generated, as well, for the purpose of watching your blood lilies.  Thanks for pointing out the discrepency. I'll try and get that fixed tomorrow.

kkragenbrink: @AzariahAeroborn I actually wound up not tracking either nearly so closely.  Since capping isn't a major concern, it's very easy to just figure out whether you capped by lily consumption vs. blood lily consumption. At this point, I'm no longer tracking lily generation at all, as I don't think it's necessary.  The end result is we have a minor suggestion about lily capping, a checklist for blood lily spends (since they're always a DPS gain when you have already generated them), and a suggestion when you waste blood lily generation by spending lilies while you have a blood lily.  This should be a more accurate reflection of what "good" play is like.

locale/en/messages.json
```diff
@@ -666,23 +666,28 @@
...
+  "whm.lily-blood.checklist.content": "Use <0/> when it is available. It is a major source of your DPS.",
+  "whm.lily-blood.requirement.name": "Blood Lilies Generated",
+  "whm.lily-blood.suggestion.content": "Use <0/> to avoid wasting blood lily growth.",
+  "whm.lily-blood.suggestion.why": "{0, plural, one {# blood lily} other {# blood lilies}} did not bloom due to early lily use.",
   "whm.overheal.requirement.all": "Overheal (all sources)",
   "whm.overheal.requirement.hot": "Overheal (HoT)",
   "whm.overheal.requirement.nonhot": "Overheal (non-HoT)",
   "whm.overheal.rule.description": "Avoid wasting heals by healing for more than required to fill a target's HP bar. While some overheal is inevitable, overheal only serves to generate more enmity, for no gain. Being efficient with healing additionally helps with your MP management.",
-  "whm.overheal.rule.name": "Avoid overheal",
+	"whm.overheal.rule.name": "Avoid overheal",
```
ackwell: Hm... I have _no_ idea why this is indenting differently. Weird. Not an issue I guess, the build will automatically override this anyway.

src/parser/jobs/whm/modules/DoTs.js
```diff
@@ -5,31 +5,26 @@ import ACTIONS from 'data/ACTIONS'
...
-//aero 2 clips are less severe as it may be used for the initial damage when moving
-
-const CLIP_MAX_MINOR = {
-	[STATUSES.AERO_II.id]: 30000,
-	[STATUSES.AERO_III.id]: 9000, //less than 3 gcds wasted
-}
-const CLIP_MAX_MEDIUM = {
-	[STATUSES.AERO_II.id]: 60000,
-	[STATUSES.AERO_III.id]: 24000,
+const SUGGESTION_TIERS = {
```
ackwell: much nicer; thanks for that!

src/parser/jobs/whm/modules/Lilies.js
```diff
@@ -0,0 +1,141 @@
...
+			}))
+		}
+	}
+
+	checkBloodLilies() {
+		this._bloodLiliesWasted = Math.floor(this._liliesWasted / BLOOD_LILY_BLOOM)
+		const bloodWarnTarget = (this._bloodLiliesGenerated - UNUSED_BLOOD_LILIES_WARN) / this._bloodLiliesGenerated * 100
+		const bloodFailTarget = (this._bloodLiliesGenerated - UNUSED_BLOOD_LILIES_FAIL) / this._bloodLiliesGenerated * 100
+
+		this.checklist.add(new TieredRule({
```
ackwell: Hm, sorry if I missed this the first time - this checklist entry and the suggestion below it essentially cover the same ground.

I _think_ the metric makes more sense as a suggestion than a checklist (takes long enough to generate a misery cast that the %ge will be quite chunky).

src/parser/jobs/whm/modules/Lilies.js
```diff
@@ -0,0 +1,141 @@
...
+					<Trans id="whm.lily-cap.suggestion.why">
+						{<Plural value={this._unused} one="# lily" other="# lilies" />} went unused.
+					</Trans>
+				</Fragment>,
+			}))
+		}
+	}
+
+	checkBloodLilies() {
+		this._bloodLiliesWasted = Math.floor(this._liliesWasted / BLOOD_LILY_BLOOM)
```
ackwell: Might wanna `+ 1` this if they've got a blood bloom at the end of the fight that they haven't spent.

src/parser/jobs/whm/modules/Lilies.js
kkragenbrink: They were slightly different. One was blood lilies spent, the other was misspending lilies when you have a blood lily... but point taken. They're similar enough that one covers the other for the most part.

src/parser/jobs/whm/modules/Lilies.js
```diff
@@ -0,0 +1,112 @@
...
+					<Trans id="whm.lily-cap.suggestion.why">
+						{<Plural value={this._unused} one="# lily" other="# lilies" />} went unused.
+					</Trans>
+				</Fragment>,
+			}))
+		}
+	}
+
+	checkBloodLilies() {
+		this._bloodLiliesWasted = Math.floor(this._liliesWasted / BLOOD_LILY_BLOOM) + (this._blooming === BLOOD_LILY_BLOOM)
```
ackwell: I am low-key disappointed that JS lets you do this. But nice.

PR #414, opened 2019-07-08
Shb blm despair thunder

AkairyuGestalter:
Retrying because Github do

### DIFF
--- a/src/parser/jobs/blm/modules/RotationWatchdog.js
+++ b/src/parser/jobs/blm/modules/RotationWatchdog.js
@@ -343,7 +343,7 @@ export default class RotationWatchdog extends Module {
 				//Only display rotations with more than 3 casts since less is normally weird shit with Transpose
 				//Also throw out rotations with no Fire spells
 				const fire3Count = this._rotation.casts.filter(cast => cast.ability.guid === ACTIONS.FIRE_III.id).length
-				const fireCount = [fire3Count, fire1Count, fire4Count, despairCount].reduce((accumulator, currentValue) => accumulator + currentValue, 0)
+				const fireCount = fire3Count + fire1Count + fire4Count + despairCount
 				if (fireCount === 0) {
 					this._rotationsWithoutFire++
 				}
--- a/src/parser/jobs/blm/modules/RotationWatchdog.js
+++ b/src/parser/jobs/blm/modules/RotationWatchdog.js
@@ -37,6 +37,7 @@ export default class RotationWatchdog extends Module {
 	static displayOrder = DISPLAY_ORDER.ROTATION
 
 	static dependencies = [
+		'checklist',
 		'suggestions',
 		'gauge', // eslint-disable-line xivanalysis/no-unused-dependencies
 		'invuln',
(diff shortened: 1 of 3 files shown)


src/parser/jobs/blm/modules/RotationWatchdog.js
```diff
@@ -212,16 +223,30 @@ export default class RotationWatchdog extends Module {
...
+
+		//Suggestion for hard T3s under AF. Should only have one per cycle
+		if (this._extraT3s) {
 			this.suggestions.add(new Suggestion({
 				icon: ACTIONS.THUNDER_III_FALSE.icon,
 				content: <Trans id="blm.rotation-watchdog.suggestions.wrong-t3.content">
-					Never hard cast a <ActionLink {...ACTIONS.THUNDER_III}/> in your Astral Fire phase, since that costs MP which could be used for more <ActionLink {...ACTIONS.FIRE_IV}/>s.
+					Don't hard cast more than one <ActionLink {...ACTIONS.THUNDER_III}/> in your Astral Fire phase, since that costs MP which could be used for more <ActionLink {...ACTIONS.FIRE_IV}/>s.
 				</Trans>,
 				severity: SEVERITY.MAJOR,
```
ackwell: Should this always be major? Seems a bit harsh.

src/parser/jobs/blm/modules/RotationWatchdog.js
```diff
@@ -300,7 +343,7 @@ export default class RotationWatchdog extends Module {
 				//Only display rotations with more than 3 casts since less is normally weird shit with Transpose
 				//Also throw out rotations with no Fire spells
 				const fire3Count = this._rotation.casts.filter(cast => cast.ability.guid === ACTIONS.FIRE_III.id).length
-				const fireCount = [fire3Count, fire1Count, fire4Count].reduce((accumulator, currentValue) => accumulator + currentValue, 0)
+				const fireCount = [fire3Count, fire1Count, fire4Count, despairCount].reduce((accumulator, currentValue) => accumulator + currentValue, 0)
```
ackwell: could probably just add them together lol

src/parser/jobs/blm/modules/RotationWatchdog.js
AkairyuGestalter: Hardcasting more than one T3 in a fire cycle means you're going to lose out on either a Fire 4 or a Despair due to MP issues, which is the same situation you'd be in prior to ShB for hardcasting any at all. Since we were dinging Major then, I think it's still appropriate (also with the sharpcast change, you have so many opportunities for Thunderclouds now, you REALLY shouldn't be hardcasting it twice in fire)

src/parser/jobs/blm/modules/RotationWatchdog.js
AkairyuGestalter: .-. indeed. fix't

src/parser/jobs/blm/modules/RotationWatchdog.js
rexatlanti: There is no reaason to hardcast T3 in af twice since your dot is long enough to last into UI. It was a major before and should be major now. Dots are 5% of our damage, F4s are about 75%

PR #415, opened 2019-07-09
5.0 Bard Support

Tycho860:
Currently, if you click on Bard, it displays a loading screen indefinitely, as an error occurs due to the removal of Piercing Resistance Down.  I went in and fixed the issue, and then decided to add in all the changes that have been made in Shadowbringers.

- Added Shadowbite, Burst Shot, and Apex Arrow
- Removed Foe Requiem and Misery's End
- Changed potentcies to their new values.
- Removed Critical Up and Straight Shot buff
- Removed the various Troubadour buffs and added the new single buff.
- Updated Nature's Minne's duration to it's new value (30 to 15).
- Removed References to Disembowl granting Piercing Resistance Down
- Updated Barrage Module to reflect the changes made in Shadowbringers
- Updated Barrage Module to warn user if it detects no Barrages cast.
- Disabled the Straight Shot module as it no longer applies.

On the Barrage module changes in specific, because Refulgent Arrow has been changed to replace Straight Shot, Barrage now gives a proc to use Refulgent Arrow, and that Empyreal Arrow is no longer a weapon skill, I updated it to reflect these changes.  I added checking for a straighter shot proc, and then when a barrage is casts, it contains whether or not straighter shot was already active when used.  If it was, then you lost a cast of Refulgent Arrow, which is never recommended.  I then added a weighting to it, added it to the checklist, and added a warning to the individual barrage that missed it saying what happened.

Also, I noticed a todo to add something to tell the user if there is no barrages casts, so I went and added something to do that.  Although, it might be able to be implemented better.


Toastdeib:
I... may be part of the reason why the BRD module exploded, and I apologize for that - I removed `STATUSES.PIERCING_RESISTANCE_DOWN` as part of my DRG changes since I didn't realize that it was being referenced here.

src/data/STATUSES/BRD.js
```diff
@@ -47,11 +48,6 @@ export default {
 		// Duration depends on trait
 		duration: 18,
 	},
-	FOE_REQUIEM_DEBUFF: {
```
ackwell: ;-;7

src/parser/jobs/brd/modules/index.js
```diff
@@ -15,7 +14,6 @@ export default [
 	OGCDDowntime,
 	PitchPerfect,
 	SongUptime,
-	StraightShot,
```
ackwell: Is SS gonna be making a come-back soon? If not, can probs nuke the file for it too.

Tycho860: It's not, I'll get rid of it.  Didn't want to remove it without talking about it first.

PR #416, opened 2019-07-10
Bugfixing pass for GNB - No Mercy rush and Solid Barrel combo cartridge generation.

rachellhea:
* Added the Gunbreaker color code from FFLogs.
* Fixed the bug where uncombo'd Solid Barrels and Demon Slaughters would contribute to Cartridge generation.
* Initialized the cartridge history graph to 0 on log start.
* Added a nifty panel showing a breakdown of waste by source event (including death).
* Excluded rushed No Mercy windows from missed GCD calculations. Only doing this for GCDs because missed casts suggests looking at the casts themselves to see if using them outside No Mercy constitutes to an overall gain. Open to discussion on this.

Screen Shot 2019-07-10 at 12 21 55 PM


src/parser/jobs/gnb/modules/Ammo.tsx
```diff
@@ -144,10 +208,35 @@ export default class Ammo extends Module {
...
 		}
+
+		const chartOptions = {
+			scales : {
+				yAxes: [{
+					ticks: {
+						beginAtZero: true,
+						min: 0,
+						max: 2,
+						callback: ((value: number) => {
```
ackwell: would be good to revisit this and see if you can get it working with the built in step handling

PR #419, opened 2019-07-12
Data update for 5.0 PLD.

rachellhea:
No actual logic changes yet, aside from completely purging the Oath module since it no longer matters.


src/parser/jobs/pld/modules/OGCDDowntime.tsx
```diff
@@ -6,5 +6,6 @@ export default class OGCDDowntime extends CooldownDowntime {
 	trackedCds = [
 		ACTIONS.SPIRITS_WITHIN.id,
 		ACTIONS.CIRCLE_OF_SCORN.id,
+		ACTIONS.INTERVENE.id,
```
ackwell: Is intervene something you actually want to use on CD?

rachellhea: Nope. But idk how charges are being implemented. The rule of thumb is: "Use 1 charge if you gain a GCD or an Intervene. Otherwise dump 2 into every FoF."

ackwell: Fukkin. Was confusing intervene with intervention, hence the comment.

Works for now, can test how it jives with real stuff.

PR #420, opened 2019-07-12
Initial changes for 5.0 Scholar

kkragenbrink:


src/parser/jobs/sch/modules/DISPLAY_ORDER.js
```diff
@@ -1,7 +1,6 @@
 export default {
 	AETHERFLOW: 51,
 	DOTS: 52,
-	SHADOW_FLARE: 53,
 	WEAVING: 54,
 	CLERIC_STANCE: 99,
```
ackwell: can probs delete CS's DO too

PR #424, opened 2019-07-16
New Snapshots Module for Bard

Ririan:
The Snapshots module is very straight forward.  Because recommending optimal snapshotting can be situational, it is instead left to the user to decide if they made a good snapshot.

What this module does is simple, it takes in cast events of either of the DoTs or Iron Jaws, puts those into an array, and then outputs what statuses were snapshotted when they were cast formatted as a table.

Along side that, I updated some stats to match the current max level, removed and changed a few crit buffs, and changed the Perfect Pitch module from stating things related to Repertoire proccing off of crits.

With this set of changes I have also set Bard to a supported job on patch 5.01 as it now offers a good level of job specific analysis that is also relevant to Shadowbringers.  While more analysis may be added later, what is here now is plenty to have list this as supported.


src/parser/jobs/brd/modules/Snapshots.js
```diff
@@ -0,0 +1,82 @@
...
+				</Table.Cell>
+				{snapshotCell}
+			</Table.Row>
+		})
+
+		// Output is a Table, where every row after the header contains individual snapshots
+		return <Table>
+			<Table.Header>
+				<Table.Row key="header">
+					<Table.HeaderCell>Time</Table.HeaderCell>
```
Toastdeib: This (and the two below it) should be i18n'd.

src/parser/jobs/brd/modules/AdditionalStats.js
```diff
@@ -8,19 +8,15 @@ import STATUSES from 'data/STATUSES'
 import math from 'mathjsCustom'
 
 // Relevant crit buffs
+// TODO: Add DNC crit buffs when the statues gets added
```
Toastdeib: This is more a comment for the file overall, but - aren't crits no longer a factor for BRD procs? Unless they changed it back, I'm pretty sure Repertoire stacks build on flat DoT ticks now, not crits specifically. It's probably outside the scope of this PR (especially if we want to keep them small), but you can probably cut this file down a lot to remove all the wacky math for determining crit rate.

src/parser/jobs/brd/modules/Snapshots.js
```diff
@@ -0,0 +1,82 @@
...
+							//35px is a purely arbitrary value that I think looks nice.  The default icon size for statuses are too small.
+							return <StatusLink  key={id} showName={false} iconSize="35px" {...getDataBy(STATUSES, 'id', Number(id))}/>
+						}
+					})
+				}
+			</Table.Cell>
+
+			return <Table.Row key={snapshotEvent.timestamp}>
+				<Table.Cell>
+					{this.util.formatTimestamp(snapshotEvent.timestamp)}
```
Toastdeib: Minor, but you can also do `this.parser.formatTimestamp()` and drop the `util` dependency.

src/parser/jobs/brd/modules/Snapshots.js
```diff
@@ -0,0 +1,82 @@
...
+
+		this.addHook('cast', {
+			by: 'player',
+			abilityId: [ACTIONS.CAUSTIC_BITE.id, ACTIONS.STORMBITE.id, ACTIONS.IRON_JAWS.id],
+		}, this._onSnapshot)
+
+	}
+
+	_onSnapshot(event) {
+		this._snapshotEvents.push(event)
```
Toastdeib: Minor, but this can be in-lined in the constructor if you want with a lambda:
```js
this.addHook('cast', {...}, event => this._snapshotEvents.push(event))
```

src/parser/jobs/brd/modules/AdditionalStats.js
Ririan: This is true, I have avoided changing anything there because I had no reason too and I didn't want to mess it up without fully understanding it's purpose first.  I'll handle this in a different pr though.

src/parser/jobs/brd/modules/Snapshots.js
Ririan: Gotcha, will push that change right away.

ackwell:
Needs an extract, but I'll handle that on master for now. Will be adding a CI check for it.

PR #426, opened 2019-07-16
Added a Reassemble usage module for MCH

Toastdeib:
Apparently other folks haven't been doing their `yarn extract`s, because I got a lotta shit that isn't mine in here. :eyes:


src/parser/jobs/mch/modules/Reassemble.js
```diff
@@ -0,0 +1,84 @@
...
+import React from 'react'
+
+import {ActionLink} from 'components/ui/DbLink'
+import {TieredSuggestion, SEVERITY} from 'parser/core/modules/Suggestions'
+import ACTIONS from 'data/ACTIONS'
+import STATUSES from 'data/STATUSES'
+import Module from 'parser/core/Module'
+
+// The values in this const will be added to the bad use count as they're cast (hence Drill/Anchor being 0)
+const REASSEMBLE_GCDS = {
```
ackwell: Any particular drive behind the comprehensive mapping or would a whitelist work? Doubt it'll be changing any time in the future, but the less tweaking required the better

Toastdeib: Main reason was to reduce conditionals. If I did a straight whitelist (i.e. all GCDs that aren't Drill/Air Anchor), I wouldn't have the list to filter on for the hook, so I'd have to do a `getDataBy(...).onGcd` check and also a "skip if it's one of these two" check. This way I only have to check for the status, the filter and 1/0 mapping handle the rest of the logic for me.

PR #427, opened 2019-07-17
Dancer action and status definitions

AkairyuGestalter:


src/data/ACTIONS/DNC.js
```diff
@@ -1,3 +1,236 @@
...
+		icon: 'https://xivapi.com/i/003000/003454.png',
+		name: 'Standard Step',
+		onGcd: false, // it is but it isn't...
+		cooldown: 30,
+	},
+	TECHNICAL_STEP: {
+		id: 15998,
+		icon: 'https://xivapi.com/i/003000/003473.png',
+		name: 'Technical Step',
+		onGcd: false, // it is but it isn't...
```
ackwell: Ah, I follow what you were talking about on discord now. My bad.

When I get some time, I'll check the game data to see how these are represented, it's likely something we can take into account.

PR #429, opened 2019-07-17
Cleanup of the F4 logic and new Suggestion

rexatlanti:
I hope the PR isn't too big. The new F4 logic should be more sane than the old one. Furthermore a new suggestion has been added to not use manafont before Despair. It looks like my `yarn extract` kept erroring so I didn't add that file to the PR ;w;


src/parser/jobs/blm/modules/RotationWatchdog.js
```diff
@@ -210,6 +204,20 @@ export default class RotationWatchdog extends Module {
 				</Trans>,
 			}))
 		}
+		// Suggestion to not use manafont before Despair
+		if (this._mfBeforeDespair) {
+			this.suggestions.add(new TieredSuggestion({
+				icon: ACTIONS.MANAFONT.icon,
+				content: <Trans id="blm.rotation-watchdog.suggestions.mf-before-despair.content">
+					Using <ActionLink {...ACTIONS.MANAFONT} /> before <ActionLink {...ACTIONS.DESPAIR} /> leads to less <ActionLink {...ACTIONS.DESPAIR} />s than possible being cast. Try to avoid that since <ActionLink {...ACTIONS.DESPAIR} /> is stronger than <ActionLink {...ACTIONS.FIRE_IV} />.
```
AkairyuGestalter: Suuuuuper minor nitpick, but it's more grammatically correct to use 'fewer' instead of 'less' since it's a countable thing we're talking about. Other than that, it's good.

rexatlanti: me not good england

src/parser/jobs/blm/modules/RotationWatchdog.js
```diff
@@ -292,27 +300,42 @@ export default class RotationWatchdog extends Module {
...
 			const despairCount = this._rotation.casts.filter(cast => cast.ability.guid === ACTIONS.DESPAIR.id).length
 			const hasManafont = this._rotation.casts.filter(cast => cast.ability.guid === ACTIONS.MANAFONT.id).length > 0
+			const lastEvent = this._rotation.casts[this._rotation.casts.length-1]
 
 			const hardT3Count = this._rotation.casts.filter(cast => cast.ability.overrideAction).filter(cast => cast.ability.overrideAction.id === ACTIONS.THUNDER_III_FALSE.id).length
 			if (hardT3Count > 1) {
 				this._extraT3s++
 			}
+			//check whether manafont was used before despair
+			if (hasManafont) {
```
ackwell: lots of nesting here; could be condensed into one conditional

PR #434, opened 2019-07-19
BLM Dropped Sharpcasts Module

AkairyuGestalter:
Sharpcast module based on Procs to track buff windows, and count times where the buff fell off without being used.

Also extracted i18n since I was adding suggestion strings. Caught some other stuff that hadn't been extracted yet apparently.


src/parser/jobs/blm/modules/Sharpcast.js
```diff
@@ -0,0 +1,136 @@
...
+				type: 'background',
+				start: window.start - fightStart,
+				end: window.stop - fightStart,
+				group: groupId,
+				content: <img src={STATUSES.SHARPCAST.icon} alt={STATUSES.SHARPCAST.name}/>,
+			}))
+		})
+
+		// Suggestions to use sharpcasts that wore off.
+		if (this._droppedSharpcasts > 0) {
```
Toastdeib: Minor, but you don't need to bother with this check on tiered suggestions; it's handled implicitly.

src/parser/jobs/blm/modules/Sharpcast.js
```diff
@@ -0,0 +1,136 @@
...
+import {Trans, Plural} from '@lingui/react'
+import {StatusLink} from 'components/ui/DbLink'
+
+const SHARPCAST_DURATION_MILLIS = STATUSES.SHARPCAST.duration * 1000
+
+const SHARPCAST_CONSUMER_IDS = [
+	ACTIONS.FIRE_I.id,
+	ACTIONS.THUNDER_III.id,
+	ACTIONS.THUNDER_IV.id,
+	ACTIONS.SCATHE.id,
```
Toastdeib: Do you think it would be worth tracking how many Sharpcasts are consumed by each individual action? I'm no BLM main, but I'm wondering if it would be worth adding a suggestion that yells at people for wasting it on Scathe (unless that's not a waste anymore).

src/parser/jobs/blm/modules/Sharpcast.js
```diff
@@ -0,0 +1,136 @@
...
+		this.addHook('cast', {by: 'player', abilityId: SHARPCAST_CONSUMER_IDS}, this._onCast)
+		this.addHook('death', {to: 'player'}, this._onDeath)
+		this.addHook('complete', this._onComplete)
+	}
+
+	_onRemoveSharpcast(event) {
+		this._stopAndSave(event.timestamp)
+	}
+
+	_onGainSharpcast(event) {
```
Toastdeib: It may be simpler and cleaner to use `this.combatants.selected.hasStatus()` instead of tracking buff windows manually; you can look at [Kassatsu.js](https://github.com/xivanalysis/xivanalysis/blob/master/src/parser/jobs/nin/modules/Kassatsu.js#L41) for an example.

AkairyuGestalter: I'm using the buff window data to show it on the timeline, so I'm going to need to keep it anyways. If I wasn't trying to do that, I agree, your idea would be cleaner.

src/parser/jobs/blm/modules/Sharpcast.js
AkairyuGestalter: It's definitely a waste to use Sharpcast on it, but there is something to be said for not yelling at them if Sharpcast comes off CD, gets popped, then gets consumed on a Scathe as the boss goes invuln, and the CD is back up by the time the invuln period is over. They would've wasted the Sharpcast in that case, so I don't want to ding them for using it on a normally-suboptimal action that was actually the optimal one in the niche case.

src/parser/jobs/blm/modules/Sharpcast.js
AkairyuGestalter: Yeah, I noticed that while working on the dnc module earlier, I'll pull those

src/parser/jobs/blm/modules/Sharpcast.js
```diff
@@ -0,0 +1,134 @@
...
+	ACTIONS.THUNDER_III.id,
+	ACTIONS.THUNDER_IV.id,
+	ACTIONS.SCATHE.id,
+]
+
+export default class Sharpcast extends Module {
+	static handle = 'sharpcast'
+	static dependencies = [
+		'timeline',
+		'procs', // eslint-disable-line @xivanalysis/no-unused-dependencies
```
ackwell: Any reason you need procs? Doesn't look like you're using it, and there's no normalisers running.
If you're relying on execution order for something outside normalisers, I'd rather it be a side effect of a public api access.

src/parser/jobs/blm/modules/Sharpcast.js
```diff
@@ -0,0 +1,134 @@
...
+		}
+	}
+
+	// Consolidate old onCast functions into one central function
+	_onCast(event) {
+		const actionId = event.ability.guid
+
+		const action = getDataBy(ACTIONS, 'id', actionId)
+		if (action && action.onGcd) {
+			this._tryConsumeSharpcast(event)
```
ackwell: seems like a bit of a redundant split here, but not gonna block on it

ackwell:
Oh; don't forget to add a changelog entry, too!

src/parser/jobs/blm/modules/Sharpcast.js
AkairyuGestalter: I did that to make sure the proc timeline group had already been set up before this module tries adding anything to it. I'm not 100% it's actually needed, I just put it there to be safe.

AkairyuGestalter: Is an API now

src/parser/jobs/blm/modules/Procs.js
```diff
@@ -81,6 +81,21 @@ export default class Procs extends Module {
 		this.timeline.addGroup(this._group) // Group for showing procs on the timeline
 	}
 
+	getGroupIdForStatus(status) {
```
ackwell: 👍

PR #436, opened 2019-07-20
Fix issues with DRK multi-hit and resource modules

AzariahAeroborn:


AzariahAeroborn:
I intentionally have not updated the changelog yet.  I'm working on one more change that will change "no resource gain on miss/invuln" handling to use the hit type enum rather than checking for damage = 0, but wanted to split that into a separate pull request because it will impact core.  I will update the changelog to include all three changes with that pull.

src/parser/jobs/drk/modules/ResourceSimulator.js
```diff
@@ -179,6 +179,12 @@ export default class Resources extends Module {
 	}
 
 	checkBloodOvercap(actionBloodChange) {
+		if (actionBloodChange < 0) {
+			// Spender
+			if (this.combatants.selected.hasStatus(STATUSES.DELIRIUM.id)) {
+				actionBloodChange = 0
```
ackwell: _Generally_ avoid modifying function args, pretend they're `const`. In this particular case, the `if` below checks against this value, so you can probably afford to

```suggestion
				return
```

to signify there's no consumption

src/parser/jobs/drk/modules/ResourceSimulator.js
```diff
@@ -222,27 +228,34 @@ export default class Resources extends Module {
...
 			const actionInfo = RESOURCE_GENERATORS[abilityId]
-			if (!actionInfo.requiresCombo || event.type === 'combo') {
+			if ((!actionInfo.requiresCombo && event.type !== 'combo') || event.type === 'combo') {
 			// Only gain resources if the action does not require a combo or is in a valid combo
 				actionBloodGain += RESOURCE_GENERATORS[abilityId].blood
 				actionMPGain += RESOURCE_GENERATORS[abilityId].mp
 			}
 		}
 
+		console.log(`Net Blood Change: ${actionBloodGain}`)
```
ackwell: logging~

AzariahAeroborn: Whoops.  Removed.

src/parser/jobs/drk/modules/ResourceSimulator.js
AzariahAeroborn: Change updated, also consolidated the double-IF into a single and statement.

PR #437, opened 2019-07-20
Update DRK resource tracker to use HitType instead of damage for determining failed hits

AzariahAeroborn:


AzariahAeroborn:
This pull should be handled after #436 
_sigh_ I hate git magic.  Give me good ol' Black Magic any day.

src/fflogs.ts
```diff
@@ -107,6 +107,23 @@ export enum AbilityType {
 	MAGICAL_DIRECT = 1024,
 }
 
+export enum HitType {
```
ackwell: 👍

src/parser/jobs/drk/modules/ResourceSimulator.js
```diff
@@ -218,31 +225,37 @@ export default class Resources extends Module {
 	}
 
 	_onEvent(event) {
-		if (event.amount === 0) {
-			// Misses or attacks against invuln targets do not trigger resource gain
-			return
-		}
+		const _actionHit = (event.HitType !== HitType.MISS && event.HitType !== HitType.IMMUNE)
```
ackwell:
```suggestion
		const _actionHit = (event.hitType !== HitType.MISS && event.hitType !== HitType.DODGE)
```

I _think_ `IMMUNE` is a wow or eso thing - unless you've seen ff logs with it?

AzariahAeroborn: https://www.fflogs.com/v1/report/events/damage-done/z8HnpaXZMkWGdvj3?start=37735832&end=38171812&sourceid=362

https://www.fflogs.com/reports/z8HnpaXZMkWGdvj3#fight=49&type=damage-done&source=362&view=events&ability=7519

Based on results of Omega M/F and hitting the "wrong" target with Contre Sixte, it appears that Invulnerable is being mapped to the Immune hit type.

AzariahAeroborn: Would it make more sense to extend the 'damage' and 'aoedamage' event types to have a SuccessfulHit (or UnSuccessfulHit) property, and put a const array of "hit failed" types in core?  I think that would save a lot of potential duplication by consumers, but I'm not sure on where I would need to add that logic to the base 'damage' event.

AzariahAeroborn:
I added an additional change to this PR - minor bug fix for the way I was zeroing blood usage for spenders under Delirium (by zeroing after calculating base spend plus Blood Weapon effects, I was netting to +0 for Bloodspiller under Delirium + Blood Weapon, when it should be a net of +10).  I also moved the MP / Dark Arts handling out to the same point for clarity and so that I don't fiddle with actionMPChange within the checkMPOvercap method.  

I added it to this PR to avoid yet another merge conflict (especially since there already is one between this PR and #436... _sigh_).  Let me know if you want me to pull this back and chop things up better.

src/parser/jobs/drk/modules/ResourceSimulator.js
ackwell: >Based on results of Omega M/F and hitting the "wrong" target with Contre Sixte, it appears that Invulnerable is being mapped to the Immune hit type.

Huh, you're right. I must've totally missed that in my log scouring when I wrote that initial mapping, thanks for the catch.

src/parser/jobs/drk/index.js
```diff
@@ -24,5 +24,10 @@ export default new Meta({
 			Changes: () => <>Initial release of 5.0 support for Dark Knight</>,
 			contributors: [CONTRIBUTORS.AZARIAH],
 		},
+		{
```
ackwell: only just noticed (muh bad), but while it doesn't _matter_ per se (cus changelog is sorted), generally add changelogs to the top.

src/parser/jobs/drk/modules/ResourceSimulator.js
```diff
@@ -218,25 +213,41 @@ export default class Resources extends Module {
...
 		const abilityId = event.ability.guid
 		let actionBloodGain = 0
 		let actionMPGain = 0
+
 		if (RESOURCE_SPENDERS.hasOwnProperty(abilityId)) {
-			actionBloodGain += RESOURCE_SPENDERS[abilityId].mp
-			actionMPGain += RESOURCE_SPENDERS[abilityId].mp
+			if (RESOURCE_SPENDERS[abilityId].blood < 0 && this.combatants.selected.hasStatus(STATUSES.DELIRIUM.id)) {
+				// Blood spender under delirium - no change
+				actionBloodGain += 0
```
ackwell: I only just re-read this and realised you're `+= 0` which is basically a noop. If it significantly improves the readability to you I'm not gonna block on it, but it seems a little redundant.

ackwell:
Erk, conflict - took a look myself but idek DRK so i'll let someone who knows what they're doing sort it out, soz :notlikemeow:

AzariahAeroborn:
Okies, got the merge conflict resolved - stemmed from the logging I left in ResourceSimulator on accident and cleaned up in #436 but not here.  Also fixed one more cast event hook I was missing :notlikemeow:

src/parser/jobs/drk/modules/ResourceSimulator.js
AzariahAeroborn: Noop was intentionally added for readability, since I can't return early out of the event there (using Bloodspiller under Delirium + Blood Weapon needs to trigger both that conditional plus the conditional for "using a GCD under Blood Weapon" further down, for a net of actionBloodGain = 10)

PR #438, opened 2019-07-21
Dance performance metrics

AkairyuGestalter:
Suggestions to:
1. Not miss finishes (unless invuln)
2. Enough moves to trigger the correct finish.
3. No extra moves in a performance.

Checklist rule for maintaining Standard Finish uptime. Probably needs better logic for giving a pass during downtime (if we do at all), but should be good for a first pass.


src/parser/jobs/dnc/modules/DirtyDancing.tsx
```diff
@@ -0,0 +1,298 @@
...
+	dirty: boolean = false
+	missed: boolean = false
+	footloose: boolean = false
+
+	public get error(): boolean {
+		return this.dirty || this.missed || this.footloose
+	}
+
+	public get expectedFinishId(): number {
+		const actualFinish = _.last(this.rotation)
```
ackwell: would it make more sense to base the expected finish on the start move?

src/parser/jobs/dnc/modules/DirtyDancing.tsx
```diff
@@ -0,0 +1,298 @@
...
+		let expectedFinish = -1
+		if (actualFinish) {
```
ackwell: nit; can invert some of these and shortcut with `return` to flatten it out a bunch

src/parser/jobs/dnc/modules/DirtyDancing.tsx
```diff
@@ -0,0 +1,298 @@
...
+			dance.rotation.push(event)
+		} else {
+			dance = this.addDanceToHistory(event)
+		}
+		dance.dancing = false
+	}
+
+	private resolveDance(event: DamageEvent) {
+		const dance = this.lastDance
+		if (dance && !dance.resolved) {
```
ackwell: likewise; can shortcut out here so you don't need to lock the rest inside the block

src/parser/jobs/dnc/modules/DirtyDancing.tsx
```diff
@@ -0,0 +1,298 @@
...
+				dance.dirty = true
+			}
+			// If the finisher didn't hit anything, and something could've been, ding it.
+			// Don't gripe if the boss is invuln, there is use-case for finishing during the downtime
+			if (event.amount === 0 && !this.invuln.isInvulnerable('all', finisher.timestamp)) {
+				dance.missed = true
+			}
+			// Dancer messed up if more step actions were recorded than we expected
+			const actualCount = dance.rotation.filter(step => DANCE_MOVES.includes(step.ability.guid)).length
+			let expectedCount = 0
```
ackwell: this isn't doing much; can just `const` is and set it with the value on the next line.

src/parser/jobs/dnc/modules/DirtyDancing.tsx
```diff
@@ -0,0 +1,298 @@
...
+	ACTIONS.TRIPLE_TECHNICAL_FINISH.id,
+	ACTIONS.QUADRUPLE_TECHNICAL_FINISH.id,
+]
+
+const FINISHES = [
+	...STANDARD_FINISHES,
+	...TECHNICAL_FINISHES,
+]
+
+const EXPECTED_DANCE_MOVE_COUNT = {
```
ackwell: If you add a `-1: 0` to this, it'll catch the wat-case.

src/parser/jobs/dnc/modules/DirtyDancing.tsx
AkairyuGestalter: The problem with that is the correct opener begins with a Double Standard Finish that was set up pre-pull, so we can't go off the starter in that case. Given that, I figured it'd be easier to just always infer the proper finish from the one we actually got.

src/parser/jobs/dnc/modules/DirtyDancing.tsx
AkairyuGestalter: I can't think of a satisfactory way to flatten this one that a) keeps the .includes checks from firing if actualFinish is undefined, and b) doesn't have a bunch of extra conditionals to make sure we only return a real value if the actual finisher is contained in TECHNICAL/STANDARD_FINISHES. That might be 4 hours of sleep talking though so I'll look at it again tomorrow unless this gets merged first.

src/parser/jobs/dnc/modules/DirtyDancing.tsx
ackwell: Good point; hadn't thought of that. Makes sense this way around then.

src/parser/jobs/dnc/modules/DirtyDancing.tsx
ackwell: Can fix it later, but something like this usually works:

```ts
if (!actualFinish) {
	return -1
}

if (tech) {
	return quad_finish
}

if (standard) {
	return double_finish
}
```

PR #441, opened 2019-07-21
Bard Sidewinder module

Ririan:
This sidewinder module checks for casts of Sidewinder and Shadowbite on targets without both DoTs applied and warns the user through a suggestion, and generates a list with timestamps of when either Sidewinder or Shadowbite was cast, and with what DoT (or no DoT) it was cast on.

The module is very straight forward, and rather simple.   Also included are some small fixes and I added a change log.


src/parser/jobs/brd/modules/Sidewinder.js
```diff
@@ -0,0 +1,179 @@
...
+			}
+			return false
+		})
+	}
+
+	_onComplete() {
+		if (!this._badCasts.length) {
+			return
+		}
+		if (this._amountOfBadSidewinders) {
```
ackwell: Don't need to wrap `TieredSuggestion` in a conditional, it'll automatically hide if it doesn't meet the lowest tier.

src/parser/jobs/brd/modules/Sidewinder.js
```diff
@@ -0,0 +1,179 @@
...
+				Only use <ActionLink {...ACTIONS.SIDEWINDER}/> when you have both <ActionLink {...ACTIONS.CAUSTIC_BITE}/> and <ActionLink {...ACTIONS.STORMBITE}/> active on the target. Remember that a DoT doesn't apply as soon as you cast it, so you have to wait for it to apply before casting <ActionLink showIcon={false} {...ACTIONS.SIDEWINDER}/>.
+					</Trans>
+				</>,
+				tiers: {
+					5: SEVERITY.MAJOR,
+					3: SEVERITY.MEDIUM,
+					1: SEVERITY.MINOR,
+				},
+				value: this._amountOfBadSidewinders,
+				why: <>
```
ackwell: Fragments aren't needed in this case, the `Trans` is the only child. Can replace with parentheses.

src/parser/jobs/brd/modules/Sidewinder.js
```diff
@@ -0,0 +1,179 @@
...
+				this._amountOfBadSidewinders++
+			} else {
+				this._amountOfBadShadowbites++
+			}
+
+			this._badCasts.push({abilityId: abilityId, dotsApplied: enemyDots.length, appliedDot: enemyDots[0], timestamp: event.timestamp})
+		}
+	}
+
+	_onApplyDot(event) {
```
ackwell: The core `enemies` module tracks statuses already, you can use `enemies.getEntity(<actorId>)` to get an object with info, then if there _is_ info, `enemy.hasStatus` can be used to check for statuses. While not _quite_ the same as what you're tracking, MCH's Wildfire module does something similar if you'd like an example.

src/parser/jobs/brd/modules/Sidewinder.js
```diff
@@ -0,0 +1,179 @@
...
+				value: this._amountOfBadSidewinders,
+				why: <>
+					<Trans id="brd.sidewinder.suggestion.bad-shadowbite.reason">
+						{this._amountOfBadSidewinders} casts of <ActionLink showIcon={false} {...ACTIONS.SIDEWINDER}/> without both DoTs applied.
+					</Trans>
+				</>,
+			}))
+		}
+
+		if (this._amountOfBadShadowbites) {
```
ackwell: Probably worth combining these two suggestions, as shadowbite and sidewinder follow the same logic. There's room for a suggestion to not use shadowbite on a single target, though!

src/parser/jobs/brd/modules/Snapshots.js
```diff
@@ -42,7 +44,8 @@ export default class Snapshots extends Module {
 			const snapshotCell = <Table.Cell>
 				{
 					Object.keys(snapshot.statuses).map(id => {
-						if (snapshot.statuses[id].isActive) {
+						//To avoid showing statuses we do not currently know of, as that will cause an infinite loading circle
```
ackwell: Not a bad idea; might be worth showing an "X unknown" next to the icons so we can find out when we're missing statuses we should know about - or are they usually boss mechanics and such?

Ririan: When I looked into it, I couldn't find an icon or any information about them on xivapi.  I'd have to assume that those statuses are boss mechanics or something like that.

src/parser/jobs/brd/modules/Sidewinder.js
Ririan: I didn't know that, that simplifies things a lot.  Thanks!

Ririan:
These commits should fix the things you pointed out. Let me know if there is anything else that needs changing.

ackwell:
There's those expected conflicts

Ririan:
What started out as me just fixing the change log conflict ended with me going down a huge rabbit hole.  Now, it will properly detect how many dots were applied to the target Shadowbite was cast on.

Also, it now alerts the user of the amount of casts of Shadowbite on a single target, the amount of potency lost from each mistake, and some other small fixes and touch ups here and there.

src/parser/jobs/brd/modules/Sidewinder.js
```diff
@@ -108,30 +207,121 @@ export default class Sidewinder extends Module {
...
-					</List.Content>
-				</List.Item>
+			let singleTargetItem
+			let notBothDotsItem
+
+			if (!cast.hasBothDots) {
+				if (cast.dotsApplied) {
+					notBothDotsItem =
+							<Trans id="brd.sidewinder.list.one-dot">
+								<ActionLink {...ability}/> was cast with only one DoT applied to the target {cast.targetsHit > 1 ? 'and hit ' + cast.targetsHit + ' enemies' : ''}
```
ackwell: Likewise, can't be translated. Best way to check is to double-check the extracted message files - this particular message comes through as `<0/> was cast with only one DoT applied to the target {0}` - so that `{0}` is gobbling up the EN text.

src/parser/jobs/brd/modules/Sidewinder.js
```diff
@@ -108,30 +207,121 @@ export default class Sidewinder extends Module {
...
+			totalPotencyLost += cast.missedPotency
+
 			return <List.Item key={cast.timestamp}>
-				<List.Content>
-					<Trans id="brd.sidewinder.list.no-dots">
-						{this._createTimelineButton(cast.timestamp)} <ActionLink {...ability}/> was cast with no DoTs applied.
-					</Trans>
+				{this._createTimelineButton(cast.timestamp)}
+				<List.Content verticalAlign="middle">
+					{notBothDotsItem}{singleTargetItem}. <Label horizontal size="small" color="red" pointing="left"><Trans id="brd.sidewinder.list.missed-potency"><Icon name="arrow down"/>Lost {this._formatDamageNumber(cast.missedPotency)} potency</Trans></Label>
```
ackwell: Try to keep all text that's part of a single statement/phrase in one translation ID. Context/grammar/wording/symbols change between different languages, so what works in EN often will break in others. That trailing `.` likewise may be incorrect in some other languages.

src/parser/jobs/brd/modules/Sidewinder.js
```diff
@@ -108,30 +207,121 @@ export default class Sidewinder extends Module {
...
 				</List.Content>
 			</List.Item>
 		})
 		// Output is a List, where every item is an incorrect cast
-		return <List divided relaxed>
+		return <>
+		<List divided relaxed>
 			{items}
 		</List>
+		<Message warning attached="bottom"><Trans id="brd.sidewinder.total-mistakes">{this._badCasts.length} mistake{this._badCasts.length !== 1 ? 's' : ''} lost a total of {this._formatDamageNumber(totalPotencyLost)} potency over the course of the fight</Trans></Message>
```
ackwell: Needs a plural rule

src/parser/jobs/brd/modules/Sidewinder.js
```diff
@@ -108,30 +207,121 @@ export default class Sidewinder extends Module {
...
+
+	_formatDamageNumber(damage) {
+		const truncDamage = Math.trunc(damage)
+		return truncDamage.toLocaleString()
+	}
+
+	_getRawDamage(event) {
+		let fixedMultiplier = event.debugMultiplier
+
+		// AND ALSO FOR RANGED TRAIT, BECAUSE APPARENTLY IT'S PHYSICAL DAMAGE ONLY REEEEEEEEEE
```
ackwell: yeeeeeeeeeeeeeeeeeep

src/parser/jobs/brd/modules/Sidewinder.js
```diff
@@ -108,30 +207,121 @@ export default class Sidewinder extends Module {
...
+			rawDamage = Math.trunc(rawDamage / this.additionalStats.critMod)
+		}
+
+		if (event.directHit) {
+			rawDamage = Math.trunc(rawDamage / DHIT_MOD)
+		}
+		return rawDamage
+	}
+
+	_checkForSingleTargetShadowbites() {
```
ackwell: Not going to block on it; but for a function prefixed with "`check`", there's a _lot_ of logic and side-effects going on here.

Ririan:
Thank you for the link to the documentation, and for the tips.  Hopefully this fixes everything you mentioned.

PR #442, opened 2019-07-21
DRG correctness updates

Toastdeib:
Official DRG support! This should fix up all remaining loose ends from StB that weren't updated for ShB yet. Most of it's small stuff, the only interesting bit is the logic change in Buffs.js to check for the GCD cast right _before_ Lance Charge/Dragon Sight rather than the first one after them.


src/parser/jobs/drg/modules/Buffs.js
```diff
@@ -190,19 +189,36 @@ export default class Buffs extends Module {
...
 			},
-			value: badlyTimedLcs,
+			value: badLanceCharges,
 			why: <Trans id="drg.buffs.suggestions.bad-lcs.why">
-				{badlyTimedLcs} of your Lance Charge windows started on a non-optimal GCD.
+				{badLanceCharges} of your Lance Charge windows started right after a standard combo finisher.
+			</Trans>,
+		}))
+
+		// I'm not going to say how close I came to naming this variable badDragons
```
ackwell: I'm almost but not quite entirely disappointed in you. And I'm not going to say in which direction.

src/parser/jobs/drg/modules/Buffs.js
```diff
@@ -190,19 +189,36 @@ export default class Buffs extends Module {
...
+		const badDragonSights = this._buffWindows[STATUSES.RIGHT_EYE.id].history.filter(window => window.casts.length > 0 && window.isBad).length
+		this.suggestions.add(new TieredSuggestion({
```
ackwell: This is pretty similar to the suggestion above. Might be worth consolidating them?

PR #443, opened 2019-07-21
NIN correctness updates

Toastdeib:
Official NIN support! Much like #442, this should tidy up all the leftover jank from StB that no longer applies. Biggest changes in this PR are in the Ninki module, since I realized I wasn't account for Bunshin or the fact that it only generates on successfully comboed actions now (although interestingly enough, that _doesn't_ apply to the clone - using uncomboed GCDs with Bunshin up is still worth 4 Ninki).


src/parser/jobs/nin/modules/Kassatsu.js
```diff
@@ -110,9 +82,9 @@ export default class Kassatsu extends Module {
 			this.suggestions.add(new Suggestion({
 				icon: ACTIONS.DOTON.icon,
 				content: <Trans id="nin.kassatsu.suggestions.doton.content">
-					Avoid using <ActionLink {...ACTIONS.DOTON}/> under <ActionLink {...ACTIONS.KASSATSU}/> unless using it against multiple targets that you expect to die slowly. If the targets will die in under 15 seconds, <ActionLink {...ACTIONS.KATON}/> is the better AoE option.
+					Avoid using <ActionLink {...ACTIONS.DOTON}/> under <ActionLink {...ACTIONS.KASSATSU}/> even in AoE situations, as it has a lower potency than <ActionLink {...ACTIONS.GOKA_MEKKYAKU}/> even if every tick hits.
```
ackwell: :oof:

PR #445, opened 2019-07-22
Shadowmonkers prep

ayyaruq:
This PR just gets rid of deprecated monk stuff, suggestions etc are still incorrect so not adding changelog (unless you think necessary?) or bumping supported version quite yet. As far as I can tell, nothing breaks.


src/parser/jobs/mnk/modules/Speedmod.js
```diff
@@ -7,6 +7,7 @@ const GREASED_LIGHTNING_STATUSES = [
 	STATUSES.GREASED_LIGHTNING_I.id,
```
ackwell: missing the rename here by the looks. probs worth doing a search+replace across the system
```suggestion
	STATUSES.GREASED_LIGHTNING.id,
```

PR #446, opened 2019-07-22
Fixed the first Raging Strikes not being tracked in the OGCD Tracker if cast pre-pull

Ririan:
Currently, if a Bard casts Raging Strikes pre-pull, the OGCD tracker will erroneously state that they have missed a potential cast.  This is due to the fact that the OGCD tracker only checks for cast events, and while the parser already handles casting statuses pre-pull, it only fabricates an applybuff.

To fix this, I added a normalizer in bards OGCD tracker to check to see if the timestamp of the first Raging Strikes buff apply comes before the first cast of Raging Strikes.  If this is the case, then it will fabricate a new Raging Strikes event with the timestamp of the first Raging Strikes apply buff event.

Also, my avatar is currently not being loaded, so I fixed that too.


src/parser/jobs/brd/modules/OGCDDowntime.js
```diff
@@ -18,4 +19,37 @@ export default class OGCDDowntime extends CooldownDowntime {
...
+				}
+				// Once we have gotten this information, no need to waste time
+				if (firstRagingStrikesCast && firstRagingStrikesApply) {
+					break
+				}
+			}
+		}
+
+		if (firstRagingStrikesCast > firstRagingStrikesApply) {
+			this.parser.fabricateEvent({
```
ackwell: `fabricateEvent` only works "properly" within the standard parser lifecycle. in a normaliser, you'll need to `splice` the event into the array

PR #451, opened 2019-07-23
2dance2furious: technical drift

AkairyuGestalter:
Basic step CD drift checking & OGCDDowntime module


src/parser/jobs/dnc/modules/DirtyDancing.tsx
```diff
@@ -248,6 +271,32 @@ export default class DirtyDancing extends Module {
 				}),
 			],
 		}))
+
+		if (this.totalDrift[ACTIONS.STANDARD_STEP.id]/STEP_COOLDOWN_MILLIS[ACTIONS.STANDARD_STEP.id] > 1) {
```
ackwell: Hm. Not sure how i feel about this in general. Stats are good, but I don't think they should be used as suggestions. Suggestions and checklist are for highlighting issues, stats are to provide job-specific _statistics_.

AkairyuGestalter: Fair, I'd originally done that since I wasn't wholly confident in the calculation, but after looking at some more logs with this code, I'm happier with it, and we can always tinker with it later if needed. Done.

PR #452, opened 2019-07-23
Automaton Queen usage module for MCH

Toastdeib:
_Yass queen~_
Also flags MCH as supported. \o/


src/data/ACTIONS/MCH.js
```diff
@@ -254,17 +254,20 @@ export default {
 		id: 17206,
 		name: 'Roller Dash',
 		icon: 'https://xivapi.com/i/003000/003505.png',
+		onGcd: true, // These aren't technically on -our- GCD but it makes them look nicer in the rotation display
```
ackwell: hacky as fuck but shouldn't harm too much I don't think

PR #453, opened 2019-07-23
Kill the Triple module

rexatlanti:
Small overhauls across the board in terms of OGCD usage and "Times where you did nothing". Also removed the Triple module after talking with Laqi. It's really just not useful enough to be shown.


.gitignore
```diff
@@ -32,3 +32,5 @@ test.html
 *.swp
 .vscode/taskmarks.json
 .idea/
+
+.vscode/ipch/aac57bbcfe5f904f/mmap_address.bin
```
ackwell: wat

src/parser/jobs/blm/modules/NotCasting.js
```diff
@@ -85,18 +96,19 @@ export default class NotCasting extends Module {
 	}
 
 	_onComplete(event) {
+		const gcdLength = this.gcd.getEstimate(false)
```
ackwell: sure you don't want to be bounding this?

PR #454, opened 2019-07-23
Override Weaving for PLD to support instant-cast spells under Requiescat.

rachellhea:
Title.

Sample log with double weaves under Requiescat: https://www.fflogs.com/reports/rF8NmGvLgxzp9JqA#fight=9&type=damage-done&view=events&source=5&sourcebuffs=-1000076&ability=16459

```
00:04:02.721 | Damelia Lhea Holy Spirit Innocence *18266*
00:04:03.456 | Damelia Lhea Spirits Within Innocence 8418
00:04:04.183 | Damelia Lhea Intervene Innocence 4654
00:04:05.216 | Damelia Lhea Holy Spirit Innocence 15661
```

### DIFF
--- a/src/parser/jobs/pld/modules/Weaving.tsx
+++ b/src/parser/jobs/pld/modules/Weaving.tsx
@@ -1,0 +1,29 @@
+import ACTIONS from 'data/ACTIONS'
+import STATUSES from 'data/STATUSES'
+import {dependency} from 'parser/core/Module'
+import Combatants from 'parser/core/modules/Combatants'
+import CoreWeaving from 'parser/core/modules/Weaving'
+
+const SPELLS = [
+	ACTIONS.HOLY_SPIRIT.id,
+	ACTIONS.HOLY_CIRCLE.id,
+	ACTIONS.CONFITEOR.id,
+	ACTIONS.CLEMENCY.id,
+]
+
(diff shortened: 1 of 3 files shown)


src/parser/jobs/pld/modules/Weaving.tsx
```diff
@@ -0,0 +1,43 @@
...
+const SPELLS = [
+	ACTIONS.HOLY_SPIRIT.id,
+	ACTIONS.HOLY_CIRCLE.id,
+	ACTIONS.CONFITEOR.id,
+	ACTIONS.CLEMENCY.id,
+]
+
+export default class Weaving extends CoreWeaving {
+
+	private isRequiescatActive: boolean = false
```
Toastdeib: To avoid redoing the work of tracking the buff window, you can replace this and the constructor hooks with the `combatants` dependency, and then change the `this.isRequiescatActive` check in the function below to `this.combatants.selected.hasStatus(STATUSES.REQUIESCAT.id)`. See [SAM's Combos.ts](https://github.com/xivanalysis/xivanalysis/blob/master/src/parser/jobs/sam/modules/Combos.ts) for a Typescript example.

rachellhea: Should probably update this in a few other places, but good to know for this review.

PR #457, opened 2019-07-24
Meikyo Drift Check + Use check

Rjb99:
I've added in the cavemen tech drift check for Meikyo, this is the same code I plan to use for tsubame as they literally do the exact same thing, so once this gets reviewed I'll just copy and adjust the ids for the other skill.

I've commented out some lines that I planned to have for downtime forgiveness on the drift calcs, but they kept flagging incorrectly. So if you have an better method I might be able to use, that would be very helpful.


src/parser/jobs/sam/modules/Meikyo.js
```diff
@@ -57,39 +85,81 @@ export default class Meikyo extends Module {
...
-		if (this._badMeikyoCasts > 0) {
-			this.suggestions.add(new TieredSuggestion({
-				icon: ACTIONS.MEIKYO_SHISUI.icon,
-				content: <Trans id= "sam.meikyo.suggestions.badmeikyo.content">
+		//invuln check time for drift
+
+		const INVULN = (this.invuln.getInvulnerableUptime()/1000)
+		//const FORGIVEN_DRIFT = Math.floor(INVULN / MEIKYO_COOLDOWN) //this will calculate the amount of full recasts during a fight where a person could not use the skill.
+		const FIGHT_DURATION = ((this.parser.fightDuration/1000) - INVULN)
+		const EXPECTED_MEIKYO = Math.floor(FIGHT_DURATION/ MEIKYO_COOLDOWN)
```
ackwell: likely a total non-issue; but in a fight precisely 10m long, a 1m CD can be used 11 times, rather than just 10 - so this might be undershooting a tad.

Rjb99: Is this a on the minute mark problem or just always + 1 because things start off cooldown?

src/parser/jobs/sam/modules/Meikyo.js
```diff
@@ -57,39 +85,81 @@ export default class Meikyo extends Module {
 
 	_onComplete() {
 
-		if (this._badMeikyoCasts > 0) {
-			this.suggestions.add(new TieredSuggestion({
-				icon: ACTIONS.MEIKYO_SHISUI.icon,
-				content: <Trans id= "sam.meikyo.suggestions.badmeikyo.content">
+		//invuln check time for drift
+
+		const INVULN = (this.invuln.getInvulnerableUptime()/1000)
```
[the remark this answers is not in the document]
Rjb99: downtime seems to also throw the results, my test parse missed 7 meikyo's but with  the downtime calculated and applied, it has them missing only 5. as far as I'm aware there is no downtime in e2, nor did the person die during the 7:28 fight duration.
Test parse: https://www.fflogs.com/reports/f2g9hFAv1TwXYLPR#fight=2&type=deaths&source=158

I'll delete this for now, and if minute+ fight transitions become a thing again, I'll look back into it. Worst case it's something I have to hard code an exception.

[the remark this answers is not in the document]
ackwell: the log you linked was e1, which has that fucklong animation crap in the middle - though there's also a lot of estimated downtime which is probably wrong 'cus they're not pressing their buttons

hm

src/parser/jobs/sam/modules/Meikyo.js
```diff
@@ -57,39 +82,74 @@ export default class Meikyo extends Module {
...
                                         You missed <Plural value={this._missedMeikyoCasts} one="# GCD" other= "# GCDs" /> under the effect of <ActionLink {...ACTIONS.MEIKYO_SHISUI}/>.
-				</Trans>,
+			</Trans>,
+		}))
+
+		if (this._totalMeikyoBuffs === 0) {
+			this.suggestions.add(new Suggestion({
+				icon: ACTIONS.MEIKYO_SHISUI.icon,
+				content: <Trans id="sam.meikyo.suggestions.deleteyourself.content"> <ActionLink {...ACTIONS.MEIKYO_SHISUI}/> is a tool used to speed up your sen generation and pump out more <ActionLink {...ACTIONS.MIDARE_SETSUGEKKA}/> over the course of the fight. Not using this skill will serverly lower your damage output. </Trans>,
+				severity: SEVERITY.MAJOR, //I wish I could set this as morbid
```
ackwell: >I wish I could set this as morbid

_never_ 👺

PR #460, opened 2019-07-26
Dnc feather estimation

AkairyuGestalter:
Avg probability/expected value-based calculation for current feathers. It looks ok-ish, and until we can be more precise about actual feather amounts, the suggestions for DNC gauges should be more lenient in their tiers.


src/parser/jobs/dnc/modules/DNCGauges.module.css
```diff
@@ -0,0 +1,11 @@
+.helpText {
```
ackwell: i should really make this an option so peeps don't need css to do it

src/parser/jobs/dnc/modules/FeatherGauge.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+export default class FeatherGauge extends Module {
+	static handle = 'feathergauge'
+	static title = t('dnc.feather-gauge.title')`Feather Gauge`
+
+	@dependency private suggestions!: Suggestions
+
+	private feathersConsumed = 0
+	private avgGenerated = 0
+	private history: any[] = []
```
ackwell: Is this `any` avoidable? Ideally it should never be used.

...wonder if that's something i can add to linting...

src/parser/jobs/dnc/modules/FeatherGauge.tsx
```diff
@@ -0,0 +1,116 @@
...
+		this.addHook('complete', this.onComplete)
+	}
+	private onCastGenerator() {
+		this.avgGenerated += FEATHER_GENERATION_CHANCE
+		this.setFeather(this.currentFeathers + FEATHER_GENERATION_CHANCE)
+	}
+	private onConsumeFeather() {
+		this.feathersConsumed++
+		// If we consumed a feather when we think we don't have one, clearly we do, so update the history to reflect that
+		// TODO: count how many feathers have been spent since the last possible generation event and endure the history for
```
ackwell: While not quite the same, SAM's kenki module does something similar to this to re-adjust the minimum kenki count on spender usage (keeps the possible range on the graph from spiraling out). Might be able to re-use some of that logic?

PR #461, opened 2019-07-26
BugFix - RDM Enchanted Moulinet Cost

Cephid:
Fixed Enchanted Moulinet to cost 20 White|Black instead of the old 30.  Also added a retroactive (for 5.0) Change Log specifically so I don't forget when I mark it fully supported.

### DIFF
--- a/src/parser/jobs/rdm/index.js
+++ b/src/parser/jobs/rdm/index.js
@@ -32,5 +32,15 @@ export default new Meta({
 	],
 
 	changelog: [
+		{
+			date: new Date('2019-07-14'),
+			Changes: () => <>Initial update of existing logic for 5.0 (no new stuff)</>,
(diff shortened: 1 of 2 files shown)


src/parser/jobs/rdm/index.js
```diff
@@ -32,5 +32,15 @@ export default new Meta({
 	],
 
 	changelog: [
+		{
+			date: new Date('2019-07-14'),
+			Changes: () => <>Initial update of existing logic for 5.0 (no new stuff)</>,
+			contributors: [CONTRIBUTORS.LEYLIA],
+		},
+		{
```
ackwell: Can you flip these for me next time you add to the changelog? Once they get a bit long, it'll be easier to add to the top than scroll to the bottom to add there (plus it's in-line with the UI, which shows recent first).

PR #463, opened 2019-07-26
Esprit module baseline implementation

AkairyuGestalter:
Probability stuff again whee
Included the CSS file shared with feather gauge again in case this gets reviewed and merge to prod first


src/parser/jobs/dnc/modules/Esprit.tsx
```diff
@@ -0,0 +1,189 @@
...
+			}
+		}
+		this.avgGenerated += generatedAmt
+		if (generatedAmt > 0) {
+			this.setEsprit(this.currentEsprit + generatedAmt)
+		}
+	}
+	private onConsumeEsprit() {
+		this.espritConsumed++
+		if (this.currentEsprit < SABER_DANCE_COST) {
```
ackwell: This is really close to kenki honestly - might be able to shore up the logic with some of that back-tracking magic. Then again maybe there's _too_ much unkown to esprit for it to be of any use.

PR #465, opened 2019-07-27
DRK: Blood Weapon tracking, fix bug with Darkside initial application

AzariahAeroborn:


src/parser/jobs/drk/modules/BloodWeapon.tsx
```diff
@@ -0,0 +1,146 @@
...
+			lastDelirium.end = event.timestamp
+		}
+	}
+
+	private onComplete() {
+		const missedGcds = this.bloodWeaponWindows
+			.reduce((sum, bloodWeaponWindow) => sum + Math.max(0, EXPECTED_CONSTANTS.GCD - bloodWeaponWindow.gcds), 0)
+
+		// missed GCDs
+		if (missedGcds > 0) {
```
ackwell: don't need the conditional around the tiered suggestion

PR #472, opened 2019-07-28
Final update for basic samurai support

Rjb99:
Alright, here's the final basic PR for SAM. just added the same things to Tsubame as meikyo and adjusted the display of the drift to be whole seconds, not megadecimal levels of accuracy. Also updated a wrong id for an action and made it so SAM is officially supported.


src/parser/jobs/sam/modules/Tsubame.js
```diff
@@ -1,12 +1,14 @@
...
-import {Trans} from '@lingui/react'
+import {Trans, Plural} from '@lingui/react'
 import React from 'react'
 import {ActionLink} from 'components/ui/DbLink'
 import ACTIONS from 'data/ACTIONS'
 //import STATUSES from 'data/STATUSES'
 import Module from 'parser/core/Module'
 import {TieredSuggestion, Suggestion, SEVERITY} from 'parser/core/modules/Suggestions'
 
+const TSUBAME_COOLDOWN = 60
```
ackwell: Can this be pulled from `ACTIONS.TSUBAME_GAESHI.cooldown`?

PR #474, opened 2019-07-28
Added Phoenix pet

KelosPelmand:
This pull should be handled after #473


KelosPelmand:
I did consider that initially.  I ended up making a new file because they were more different than ww->sf and am->r.  Phoenix also wants to check for player casts during that time (4 FoF, 4 BoP).

If we do combine it, I see a few things we need to consider:

- Do we want to keep two separate lists or intersperse the summons (so row 1 is a bahamut ("8 ww, 2 am"), row 2 is a phoenix ("4 FOF, 4 BOP, 8 SF, 2 R") and so on)?  Separating is a bit more work in the output function, but should be managable.
- How do we handle the Message? If we separate, we could continue with two different message blocks, if we do not, it will need rewritten.

ackwell:
The message more or less applies to both, with some tweaks. I see no reason to duplicate the info.

I'm not _heavily_ opinionated on splitting vs combining, but displaying them combined chronologically makes more sense to me. As far as demis go, you're essentially alternating between the two, and they have very similar "rules" as one another.

src/parser/jobs/smn/modules/Ruin2.js
```diff
@@ -94,7 +95,7 @@ export default class Ruin2 extends Module {
 			this._all.push(event)
 			// Explicitly setting the ogcd tracker to true while bahamut is out,
 			// we don't want to fault people for using R2 for WWs during bahamut.
-			this._ogcdUsed = this.gauge.bahamutSummoned()
+			this._ogcdUsed = (this.pets.getCurrentPet() === PETS.DEMI_BAHAMUT.id)
```
ackwell: Probably won't have a huge effect given popular bahamut rotations, but this _probably_ isn't worth being as lenient about any more.

KelosPelmand: My understanding is that a few instant Ruins are used at the end of Bahamut to ensure all 8 WW, even though you don't have anything to weave.  Depending on how many Further Ruin stacks you have, these may be Ruin 2 or Ruin 4.

PR #475, opened 2019-07-28
Added oGCD tracking for SMN.

KelosPelmand:
This pull should be handled after #474


src/data/ACTIONS/SMN.js
```diff
@@ -31,13 +42,15 @@ export default {
...
 		cooldown: 30,
+		cooldownGroup: SMN_COOLDOWN_GROUP.ENERGY,
 	},
 
 	ENKINDLE: {
 		id: 184,
 		name: 'Enkindle',
 		icon: 'https://xivapi.com/i/002000/002677.png',
 		cooldown: 120,
+		cooldownGroup: SMN_COOLDOWN_GROUP.ENKINDLE, //this skill isn't techically usable anymore, but would be part of this group
```
ackwell: If the skill isn't usable, probably better off removing the data - there isn't a reason to keep it sitting around.

src/data/ACTIONS/ACN.js
```diff
@@ -25,6 +25,7 @@ export default {
 		name: 'Summon',
 		icon: 'https://xivapi.com/i/000000/000516.png',
 		cooldown: 10,
+		cooldownGroup: 180, //SMN_COOLDOWN_GROUP.SUMMON
```
ackwell: are you able to import+use the commented reference?

src/parser/jobs/smn/modules/GeneralCDDowntime.js
```diff
@@ -0,0 +1,23 @@
+import ACTIONS from 'data/ACTIONS'
+import CooldownDowntime from 'parser/core/modules/CooldownDowntime'
+
+export default class GeneralCDDowntime extends CooldownDowntime {
+	//Time that Jump deems ok for a OGCD to be down : ^)
```
ackwell: jump does smn stuff? :woah:

src/parser/core/modules/CooldownDowntime.js
```diff
@@ -43,27 +43,30 @@ export default class CooldownDowntime extends Module {
...
-						id,
-						encounterLength - timeOnCooldown + allowedDowntime,
-						firstUseOffset,
-						encounterLength
-					),
-				})
-			)
+		const groups = {}
+		this.trackedCds.forEach(id  => {
+			const groupId = getDataBy(ACTIONS, 'id', id).cooldownGroup || 0
```
ackwell: This has the theoretical potential to error, if the ID is ever not found. Probably _ok_ in this situation, but avoidable.

I'd rather use something like `'unknown'` or `'ungrouped'` for the fallback, though - `0` is a bit of a magic value.

src/parser/core/modules/CooldownDowntime.js
```diff
@@ -43,27 +43,30 @@ export default class CooldownDowntime extends Module {
...
+		for (const key in groups) {
+			if (key === '0') {
+				//handle non-grouped cooldowns
+				groups[key].map(id => {
+					OGCDRequirements.push(this.checkCooldown(id, <ActionLink {...getDataBy(ACTIONS, 'id', id)} />, encounterLength))
+				})
+			} else {
+				//handle cooldown group
+				const displayName = []
+				groups[key].forEach(id => {
```
ackwell: Can simplify this to just a `.map`

src/parser/jobs/smn/modules/GeneralCDDowntime.js
KelosPelmand: I copy/pasted that file from RDM and adjusted the skill list.  I just left the note as what little "where this number came from" as was there.

PR #476, opened 2019-07-28
Updated checks around Aetherflow generators and spenders

KelosPelmand:
This pull should be handled after #475


src/parser/jobs/smn/modules/Aetherflow.js
```diff
@@ -88,10 +94,16 @@ export default class Aetherflow extends Module {
 		this._badDotReqCasts[actionId][statusesMissing]++
 	}
 
-	_onPainflareDamage(event) {
-		// Only fault single target PFs _outside_ rushes.
-		if (event.hits.length <= 1 && !this.gauge.isRushing()) {
-			this._badPainflares.push(event)
+	_onAoeDamage(event) {
+		if (event.ability.guid === ACTIONS.PAINFLARE.id) {
```
ackwell: would be good to move the breakpoints into config up the top, potentially. having separate conditionals for each is a bit ehh.

if that's not feasible, can you move the nested conditionals out to the outer statements pl0x

src/parser/jobs/smn/modules/Aetherflow.js
```diff
@@ -135,6 +147,22 @@ export default class Aetherflow extends Module {
...
+
+		// Energy Siphon suggestion, I want to say < 3 is medium because 3 on two targets is greater than a full Energy Drain (or 2 casts on 1 target)
+		const numBadEnergySiphons = this._badEnergySiphons.length
+		this.suggestions.add(new TieredSuggestion({
+			icon: ACTIONS.ENERGY_SIPHON.icon,
+			content: <Trans id="smn.aetherflow.suggestions.energysiphon.content">
+				Avoid casting <ActionLink {...ACTIONS.ENERGY_SIPHON}/> on less than three targets, as it deals less damage than <ActionLink {...ACTIONS.ENERGY_DRAIN}/> per hit.
+			</Trans>,
+			why: <Trans id="smn.aetherflow.suggestions.energysiphon.why">
+				{numBadEnergySiphons} single- or two-target
```
ackwell: wording nit; would probably flow better to say "X casts? of ES hit less than 3 targets"

also needs updating to say ES instead of PF

src/parser/jobs/smn/modules/Gauge.js
```diff
@@ -32,10 +30,7 @@ export default class Gauge extends Module {
 	// -----
 	// Properties
 	// -----
-	// I'm assuming they're starting with max.
-	// TODO: Check this in some manner maybeeeeee?
-	_aetherflow = MAX_AETHERFLOW
-	_aethertrailAttunement = 0
+	_aetherflow = 0
```
ackwell: welp, that's one less assumption we need to deal with i guess

src/parser/jobs/smn/modules/Gauge.js
```diff
@@ -97,19 +92,14 @@ export default class Gauge extends Module {
 		}
 
 		if (AETHER_ACTIONS.includes(abilityId)) {
-			// Aether actions convert flow into trail
-			// TODO: Check for using flow when none (logic issue)
-			this._aetherflow --
-			this._aethertrailAttunement ++
+			if (this._aetherflow > 0) {
```
ackwell: conditionals don't need to be nested

might be worth raising a broken log if they're using aetherflow when we think we're at 0?

src/parser/jobs/smn/modules/Aetherflow.js
KelosPelmand: Agreed.

As you may have guessed from the end of the why string saying Painflare, I just copied the Painflare message and tweaked it to account for the 3 target minimum instead of 2.

KelosPelmand:
circle ci seems to have run the test for this but not reported back correctly.  If I click through to Details, it looks like the test passed.

ackwell:
No idea what borked with CI, sorry for the delay!

PR #477, opened 2019-07-28
Now check for DoTs on Ruin spells like Fester does.

KelosPelmand:
This pull should be handled after #476


src/parser/jobs/smn/modules/MissingDoTs.js
```diff
@@ -0,0 +1,135 @@
+import {t} from '@lingui/macro'
```
ackwell: good idea moving this out into its own module

src/parser/jobs/smn/modules/MissingDoTs.js
```diff
@@ -0,0 +1,135 @@
...
+	static handle = 'missingdots'
+	static title = t('smn.missingdots.title')`Missing DoTs`
+	static dependencies = [
+		'enemies',
+		'suggestions',
+		'timeline',
+	]
+
+	_badDotReqCasts = {
+		[ACTIONS.FESTER.id]: {
```
ackwell: would _somewhat_ rather the action -> potency config was up top in a const - then the entire shape of this can be generated at runtime.

doing that would also let you re-use that config for the hook filter in the constructor

src/parser/jobs/smn/modules/MissingDoTs.js
```diff
@@ -0,0 +1,135 @@
...
+		this.addHook('complete', this._onComplete)
+		this.addHook('cast', {
+			by: 'player',
+			abilityId: [ACTIONS.FESTER.id, ACTIONS.RUIN_II.id, ACTIONS.RUIN_III.id, ACTIONS.RUIN_IV.id],
+		}, this._onDotReqCast)
+	}
+
+	_onComplete() {
+		// Suggestion for fester/ruin casts without both dots
+		// TODO: add exception for initial cast (and possibly first cast after return) where Ruins are being cast before Tri-D
```
ackwell: mm, that could be an interesting one to try and account for

src/parser/jobs/smn/modules/MissingDoTs.js
```diff
@@ -0,0 +1,135 @@
...
+		}
+		this._currentMissingDotWindow.casts.push(event)
+
+		// Add to the appropriate key
+		// Just tracking flat count for now. Expand to events if need info (for the timeline, yes pls)
+		this._badDotReqCasts[actionId].badCastCounts[statusesMissing - 1]++
+	}
+
+	output() {
+		return <RotationTable data={this._missingDotWindows
```
ackwell: interesting idea inverting the usage of the rotation table - how's it look with this config? interested to see how it handles it

src/parser/jobs/smn/modules/MissingDoTs.js
KelosPelmand: the pre-cast could probably be filtered out by timestamp < 2s or some suitably small number.  Guides seem to be suggesting clipping to cast Tri-D before the next Ruin 2, so you wouldn't need to catch that one.

Following that reasoning, you would probably cast Tri-D on return before the first post-return GCD.  If Tri-D isn't available, you should be doing B3/M3 hardcasts first, so that case wouldn't be an issue.

There may also be some issues around trance resets (boss returns, dots expired, trance ready, Tri-D still on cooldown), but i don't know what the expected action will be there, so i didn't try to code for it.

src/parser/jobs/smn/modules/MissingDoTs.js
KelosPelmand: depends on how bad you are being.  In the parse I was testing against, I had several one or two skill lines.  in that regard, it's similar to the broken combo table for other classes, except there's not a "Reason" column.

PR #481, opened 2019-07-28
Watch while i break the build then fix it again

ackwell:
too damn right i'm gonna

hey acc check how fucking atrocious my bash is


ackwell:
fixed the lint warnings too

PR #482, opened 2019-07-28
Move failed attacks to core, update combos to break on failed attacks

AzariahAeroborn:


AzariahAeroborn:
When I went back to double-check that updating combos fixed the "bloodspiller not generating blood" I had initially after refactoring the successfulHit logic, I observed that the signature of the ComboEvent interface had been changed by these changes, since it's extending off of aoedamage now instead of off cast.  Updated the inheritance accordingly.  

The only consumer of ComboEvent outside of the combos module is GNB's Ammo module, which is not effected by the additional properties in the updated signature.

src/parser/core/modules/Combos.tsx
```diff
@@ -18,16 +18,30 @@ const GCD_TIMEOUT_MILLIS = 15000
...
 }
 
+interface Hit {
+	id: number
+	instance: number
+	times: number
+	amount: number
+}
+
+export interface AoeEvent extends AbilityEvent {
```
ackwell: be good to add a note here and in the aoe module that this should be over there

AzariahAeroborn: Updates made.

AzariahAeroborn:
Not sure what order you're going to want to resolve these in (or if the changes are simple enough that auto-merge will be able to handle them), but this PR now will have merge conflicts with #486 (RotationTable) and #490 (SAM combos)

PR #484, opened 2019-07-28
Add i18n tags to Barrage module

Ririan:
Currently, there are nothing related to i18n in the Barrage module outside of the title of it, so I went and fixed that.


src/parser/jobs/brd/modules/Barrage.js
```diff
@@ -419,10 +421,17 @@ export default class Barrage extends Module {
 		// Severity of a panel is determined by the highest severity of the issues in it described
 		const severity = tuples.length ? this._severitySelector(tuples.map(t => t.severity)) : SUCCESS
 
-		// Default panel title
-		const defaultTitle = <>
-			{this.util.formatTimestamp(barrage.timestamp)} - {ACTIONS.BARRAGE.name} used{!barrage.isDropped && <> on <ActionLink {...barrage.skillBarraged} /></>}
-		</>
+		// Panel title
+		let defaultTitle = <></>
```
ackwell: I'm not going to block on it, but generally use `const` where possible.

can write the below with a ternary quite succinctly

```js
const defaultTitle = barrage.isDropped
	? <Trans>stuff</Trans>
	: <Trans>other stuff<Trans/>
```

PR #491, opened 2019-07-30
DNC supported for 5.0+

AkairyuGestalter:
A batch of code cleanup, adding proper about messages, and marking DNC as supported


src/parser/jobs/dnc/modules/CommonData.tsx
```diff
@@ -0,0 +1,34 @@
+import ACTIONS from 'data/ACTIONS'
```
ackwell: not a module; would be good to move it outside the modules folder

src/parser/jobs/dnc/modules/Esprit.tsx
```diff
@@ -78,20 +61,20 @@ export default class EspritGauge extends Module {
 	private potentialOvercap = 0
 	private espritConsumed = 0
 	private avgGenerated = 0
-	private history: any[] = []
+	private history: Array<{t: number, y: number, isGenerator: boolean}> = [{t: 0, y: 0, isGenerator: false}]
```
ackwell: 👍 

if you find yourself needing that type elsewhere, you can move it into an interface - but doesn't seem necessary in this case.

src/parser/jobs/dnc/modules/FeatherGauge.tsx
```diff
@@ -35,55 +38,64 @@ export default class FeatherGauge extends Module {
...
+
 		// If we consumed a feather when we think we don't have one, clearly we do, so update the history to reflect that
-		// TODO: count how many feathers have been spent since the last possible generation event and endure the history for
-		// those n events is correct.
 		if (this.currentFeathers < 1) {
-			const prevHistory = this.history.pop()
-			prevHistory.y = 1
-			this.history.push(prevHistory)
+			// Add the underrun amount to all events back to the previous spender so the graph shows we had enough to spend
+			const lastGeneratorIndex = _.findLastIndex(this.history, event => event.isGenerator)
```
ackwell: [Array.lastIndexOf](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/lastIndexOf) 😉 

no biggie though, lodash is great

src/parser/jobs/dnc/modules/FeatherGauge.tsx
```diff
@@ -35,55 +38,64 @@ export default class FeatherGauge extends Module {
...
+			const underrun = 1 - this.currentFeathers
+			for (let i = lastGeneratorIndex; i < this.history.length; i++) {
+				this.history[i].y += underrun
+			}
+
+			// Find the last spender event prior to the generator event found above and linearly smooth the graph between the two events
```
ackwell: Having a partial-feather makes my eye twitch a bit, but I can see the logic - we don't truly know, so partial is probably more sane than trying to attribute it sanely.

Still... egh

AkairyuGestalter: RNG gonna RNG ¯\_(ツ)_/¯ I suppose I could remove all of the half-step entries before displaying the graph if we really want to, but then I'd have to enhance the backtracking logic to ensure every spend event is integer, as well as the generator event preceding it, which could get hairy. Rather leave it for now and do a separate PR to further enhance the graph later...

src/parser/jobs/dnc/modules/FeatherGauge.tsx
AkairyuGestalter: couldn't figure out a simple way to use lastIndexOf to find the index where a given property of the objects in the array is a value (true in this case). All of the solutions I found were needlessly complicated, so I went with lodash instead since it has what I wanted out of the box.

src/parser/jobs/dnc/modules/CommonData.tsx
AkairyuGestalter: np, that's a fair comment. moved it up a folder level

src/parser/jobs/dnc/modules/FeatherGauge.tsx
ackwell: Yeah that's fine. It sort of makes sense to show the partial gains anyway, as we _are_ only guessing.

PR #493, opened 2019-07-30
5.05 AST Patch changes

suShirow:
Should ideally be tested when the servers go up. Will need to source other's logs.


suShirow:
uh oh, abort, abort

Fank:
nani?

suShirow:
False alarm, at first I thought it'd be an easy fix, but then the logs looked REALLY MESSED UP, but then I realized my past self was wise and included this neat var ,`CARD_GRANTING_ABILITIES` that I oughta updated

src/parser/jobs/ast/modules/ArcanaTracking/ArcanaTracking.tsx
```diff
@@ -84,6 +85,7 @@ export default class ArcanaTracking extends Module {
 	static displayOrder = DISPLAY_ORDER.ARCANA_TRACKING
 
 	@dependency private precastStatus!: PrecastStatus
+	@dependency private cooldowns!: Cooldowns
```
kkragenbrink: This does not seem to be used.  Is it intended to be?

Otherwise the PR still looks fine to me.

PR #495, opened 2019-07-30
First pass for Warrior 5.0 Support

NamelessFaceless:
Still got few TODOs, and I'm not gonna mark the job as supported for a week or two as I work on it - But it's a start. Sorry for the bunch of dumb commits, feel free to squash them or whatever.

Severities can be taken a look at, but I think they're good where they're at.


src/parser/jobs/war/modules/InnerRelease.tsx
```diff
@@ -182,6 +198,32 @@ export default class InnerRelease extends Module {
 				severity: SEVERITY.MEDIUM,
 			}))
 		}
+
+		if (innerChaosUsed > 0) {
```
ackwell: Might be worth combining these two, as the abilities essentially share their concepts.

Can do later though.

PR #497, opened 2019-07-30
5.05 Update and Gauge Bugfix

Cephid:
Modified Gauge to use new successfulHit instead of event.amount === 0 for overkill issues.  Modified Enchanted Reprise cost from 10 to 5.


src/parser/jobs/rdm/modules/Gauge.js
```diff
@@ -90,7 +90,7 @@ class GaugeAction {
 		const abilityId = event.ability.guid
 		const {white, black} = MANA_GAIN[abilityId] || {}
 		if (white || black) {
-			if (event.amount === 0) {
+			if (!event.successfulHit.successfulHit) {
```
Toastdeib: Shouldn't this just be `!event.successfulHit` instead of the doubled version, or am I misunderstanding how the AoE module changes work? Also conflicts.

AzariahAeroborn: Sorry, this wound up as a side conversation because I saw it at work and just posted in #caster-development, and never reposted here. 

I originally had a bug (fixed in #501) that meant the doubled .successfulHit was necessary.  Per our conversation on Discord, when I fixed my bug with the AoE module changes, I also updated RDM/Gauge.js to implement the corrected interface, hence now the merge conflict.

PR #504, opened 2019-07-30
Return Energy Drain to its rightful place

nonowazu:
The aetherflow module now counts energy drain as an applicable cast (again)


Toastdeib:
One small sanity check - did you verify that the action ID for Energy Drain is the same for both SMN and SCH? I don't trust SE not to add a separate SCH-specific one with a different ID, and I'm not in a position to check a log myself right now.

PR #507, opened 2019-08-01
Add Ruin II to SCH

feffes:
SCH has a separate Ruin II ability from ACN with its own id. This commit simply adds that ability to the SCH job actions.

Compare:
ACN: https://garlandtools.org/db/#action/172
SCH: https://garlandtools.org/db/#action/17870

### DIFF
--- a/src/data/ACTIONS/ACN.js
+++ b/src/data/ACTIONS/ACN.js
@@ -60,7 +60,7 @@ export default {
 		cooldownGroup: SMN_COOLDOWN_GROUP.SUMMON,
 	},
 
-	RUIN_II: {
+	SMN_RUIN_II: {
 		id: 172,
 		name: 'Ruin II',
 		icon: 'https://xivapi.com/i/000000/000502.png',
--- a/src/data/ACTIONS/SCH.js
+++ b/src/data/ACTIONS/SCH.js
@@ -27,6 +27,13 @@ export default {
 		castTime: 2.5,
 	},
 
+	SCH_RUIN_II: {
+		id: 17870,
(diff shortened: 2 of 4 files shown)


KelosPelmand:
adding it this way breaks the SMN parsing, since it has a skill with the same id.  We probably need to do SMN_RUIN_II and SCH_RUIN_II like Aetherpact does.

feffes:
> adding it this way breaks the SMN parsing, since it has a skill with the same id. We probably need to do SMN_RUIN_II and SCH_RUIN_II like Aetherpact does.
I'm not sure if changing the "SMN" ruin II is the best idea since it's still technically the same spell as the base class. I would propose that we either keep it as the "default" ruin II or that we prepend "ACN" to it instead.

KelosPelmand:
I think any of those could work, and it's a matter of how likely someone is to accidentally mess something up later depending on how we do it.

If we leave acn/smn r2 as "RUIN_II", a future SCH contributor could more easily type it by accident without realizing there's another RUIN_II to use instead (the advantage of this approach is that the SMN files don't have to change, at least for this pr).  "ACN_RUIN_II" is at least a slight tip off that something is weird, since most skill aren't class prefixed.  "SMN_RUIN_II" is obviously wrong in a SCH file.

I'm not strongly attached to any particular approach.

ackwell:
At the end of the day, we don't support low-level parses - especially ACN-tier low.

Only reason `ACN.js` exists at _all_ was to avoid duplicating shared actions - it makes the most sense to rename to `SMN_RUIN_II`, as that's actually what it'll be used for.

PR #512, opened 2019-08-03
DNC Technical Finish Windows

AkairyuGestalter:
Tracking Technical Finish stuff as a buff window
Devilment suggestions here now
Groundwork for letting the player know they should/shouldn't have let RW drop
Adjusted the procs suggestion wording in the meantime
Also set up a defined display order for the modules


src/parser/jobs/dnc/modules/DISPLAY_ORDER.ts
```diff
@@ -0,0 +1,6 @@
+export default {
```
ackwell: technically not a module, makes more sense in the parent dir

src/parser/jobs/dnc/modules/DirtyDancing.tsx
```diff
@@ -354,7 +327,7 @@ export default class DirtyDancing extends Module {
 		}
 	}
 
-	private getNotesIcon(ruleFailed: boolean): any {
+	private getNotesIcon(ruleFailed: boolean): TODO {
```
ackwell: Should be able to drop the return type annotation - pretty sure it'll be a `JSX.Element`

src/parser/jobs/dnc/modules/Technicalities.tsx
```diff
@@ -0,0 +1,262 @@
...
+import Module, {dependency} from 'parser/core/Module'
+import Combatants from 'parser/core/modules/Combatants'
+import {AoeEvent} from 'parser/core/modules/Combos'
+import Suggestions, {SEVERITY, TieredSuggestion} from 'parser/core/modules/Suggestions'
+import Timeline from 'parser/core/modules/Timeline'
+
+import DISPLAY_ORDER from './DISPLAY_ORDER'
+import FeatherGauge from './FeatherGauge'
+
+// Harsher than the default since you'll only have 4-5 total windows anyways
```
ackwell: Can skip straight to medium if you think it's warranted

src/parser/jobs/dnc/modules/Technicalities.tsx
```diff
@@ -0,0 +1,262 @@
...
+	timelyDevilment: boolean = true
+	poolingProblem: boolean = false
+
+	constructor(start: number) {
+		this.start = start
+	}
+}
+
+export default class Technicalities extends Module {
+	static handle = 'technicalities'
```
ackwell: quite an apt name given how much fiddling around this has taken :meowlul:

src/parser/jobs/dnc/modules/Technicalities.tsx
```diff
@@ -0,0 +1,262 @@
...
+			return
+		}
+
+		if (this.isWindowOkToClose(lastWindow, event)) {
+			lastWindow.end = event.timestamp
+
+			// Check to see if this window could've had more feathers due to possible pooling problems
+			if (this.feathers.feathersSpentInRange(lastWindow.start, lastWindow.end) < FEATHER_THRESHHOLD) {
+				const previousWindow = this.history[this.history.length-2]
+				const feathersBeforeWindow = this.feathers.feathersSpentInRange((previousWindow ? previousWindow.start : this.parser.fight.start_time)
```
ackwell: Should this be `previousWindow.end`? potentially going to pick up spent feathers in the previous window's duration as-is, i _think_.

AkairyuGestalter: Hurrr, yes, you're right. Don't code while MSQlette-ing, kids. I'll fix that tonight.

src/parser/jobs/dnc/modules/Technicalities.tsx
AkairyuGestalter: Still wanted to leave the possibility of a herp or fat-finger or something and not be too mean about it. They get one slip-up though, after that it's Med+

PR #517, opened 2019-08-04
New core module for buff windows, refactor DRK & WAR to use shared module

AzariahAeroborn:
This needs to be handled after #514


src/parser/core/modules/BuffWindow.tsx
```diff
@@ -0,0 +1,279 @@
...
+		return this.rotation
+			.filter(e => actionsById.includes(e.ability.guid))
+			.length
+	}
+}
+
+interface BuffWindowExpectedGCDs {
+	expectedPerWindow: number,
+	suggestionContent: TODO,
+	severityTiers: object,
```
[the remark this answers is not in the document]
AzariahAeroborn: Updated type - Suggestions module is currently JS, so added an interface here to emulate what Suggestions is expecting, as TS types aren't available from the module. (also for suggestionContent)

AzariahAeroborn:
I think it's better to keep the output fairly generalized and consistent, rather than requiring the implementing child modules to roll their own output logic.  A middle ground might be to increase the abstraction for calculating expected GCD or tracked cooldown amounts, so implementing modules can override the expected counts on a per-window basis as needed, added that abstraction approach.

AzariahAeroborn:
I'm going to make one change to this tomorrow after work, to change from hooking `cast` to hooking `applybuff` for the start of the buff window; hooking cast is causing problems for Blood Weapon windows at the moment, since meta is to pre-pull BW.

src/data/getDataBy.ts
```diff
@@ -1,5 +1,30 @@
...
+	id: number
+	name: string
+	icon: string
+	onGcd?: boolean
+	breaksCombo?: boolean
+	combo?: ActionComboData
+	castTime?: number
+	cooldown?: number
+	autoAttack?: boolean
+}
```
ackwell: purely because action data's been heavily over-used as meta information, can i just get a

```ts
[key: string]: unknown
```

added to the end of this action interface?

It'll permit people to use custom keys in the action data without editing core, but the `unknown` will force them to typecast it before usage (so it's easier to see mistakes when reviewing)

src/data/getDataBy.ts
```diff
@@ -1,5 +1,30 @@
 import {ensureArray} from 'utilities'
 
+interface ActionComboData {
```
ackwell: For consistency with existing data, these interfaces shouldn't have the `Data` suffix.

---

Esp. given precendent (thanks!) with the `AoE.d.ts`, could these be moved into the action and status folders so as to passively type the values? I've not actually tested to see if it'd work.

src/parser/core/modules/AoE.d.ts
```diff
@@ -0,0 +1,17 @@
+import {AbilityEvent} from 'fflogs'
```
ackwell: Pull in a merge when you come through, should remove all the #514 stuff

src/parser/core/modules/BuffWindow.tsx
```diff
@@ -0,0 +1,329 @@
...
+
+	protected init() {
+		this.addHook('cast', {by: 'player'}, this.onCast)
+		this.addHook('applybuff', {by: 'player'}, this.onApplyBuff)
+		this.addHook('removebuff', {by: 'player'}, this.onRemoveBuff)
+		this.addHook('complete', this.onComplete)
+	}
+
+	private onCast(event: CastEvent) {
+		const action: ActionData | undefined = getDataBy(ACTIONS, 'id', event.ability.guid)
```
ackwell: if `ACTIONS` is typed as mentioned previously, _hopefully_ this won't need an explicit type on it

src/parser/core/modules/BuffWindow.tsx
```diff
@@ -0,0 +1,329 @@
...
+			return
+		}
+
+		if (this.activeBuffWindow) {
+			this.activeBuffWindow.rotation.push(event)
+		}
+	}
+
+	private onApplyBuff(event: BuffEvent) {
+		if (!this.buffStatus || event.ability.guid !== this.buffStatus.id) {
```
ackwell: Unless it's expected that the buff status will be changing on a regular basis, this id check should be in the `init` as a filter param.

src/parser/core/modules/BuffWindow.tsx
```diff
@@ -0,0 +1,329 @@
...
+}
+
+interface BuffWindowTrackedCooldown {
+	action: ActionData
+	expectedPerWindow: number,
+	suggestionContent: JSX.Element | string
+	severityTiers: SeverityTiers
+}
+
+export default class BuffWindowModule extends Module {
```
ackwell: For new modules, esp. those in TS, I wanna start preferencing named exports (`export class` rather than `export default class`).

---

Given the below required `buffAction` and such, would it make sense for this to be an `abstract class`? That way you can enforce that subclasses provide it (unless they're JS) - the compiler will refuse to complete until the abstracts are implemented.

src/parser/core/modules/BuffWindow.tsx
```diff
@@ -0,0 +1,329 @@
...
+
+	// Override point for class-specific rushing logic per BuffWindow - default no effect
+	// Return a positive number to increase expected GCDs for this window, and a negative number to decrease
+	protected changeExpectedGCDsClassLogic(buffWindow: BuffWindowState): number {
+		return 0
+	}
+
+	// Rushing handling for end of fight rushing.  Can be overridden if class rules for end of fight rushing vary
+	protected reduceExpectedGCDsEndOfFight(buffWindow: BuffWindowState): number {
+		if ( this.buffStatus && this.buffStatus.duration ) {
```
ackwell: can be flattened 😉

src/parser/core/modules/BuffWindow.tsx
```diff
@@ -0,0 +1,329 @@
...
+			}
+		}
+
+		// Default: no rushing reduction
+		return 0
+	}
+
+	// For consumers that have tracked cooldowns that expect the same number of that cooldown per window, this will use the
+	//   expectedPerWindow property on that cooldown as the baseline.  This method can be overridden if the logic of
+	//   expected tracked cooldowns per window is variable
```
ackwell: For comments about a method or prop, _especially_ those used by other files, consider using TSDoc comments (`/** ... */`), as they'll show up in intellisense without needing to check the source of the imported file.

src/parser/core/modules/BuffWindow.tsx
```diff
@@ -0,0 +1,329 @@
...
+				content: this.requiredGCDs.suggestionContent,
+				tiers: this.requiredGCDs.severityTiers,
+				value: invalidGCDs,
+				why: <Trans id="core.buffwindow.suggestions.badgcd.why">
+					{invalidGCDs} incorrect <Plural value={invalidGCDs} one="GCD was" other="GCDs were" /> used during {this.buffAction.name} windows.
+				</Trans>,
+			}))
+		}
+
+		if ( this.trackedCooldowns ) {
```
ackwell: If you initialise these as `protected foo: Foo[] = []` rather than `protected foo?: Foo[]`, it'll remain an optional property, but you'll be able to skip the existence checks as it always has _some_ value.

src/parser/core/modules/BuffWindow.tsx
```diff
@@ -0,0 +1,329 @@
...
+			this.trackedCooldowns.forEach(cooldown => {
+				if ( this.buffAction ) {
+					const missedCooldowns = this.buffWindows
+						.reduce((sum, curWindow) => sum + Math.max(0, cooldown.expectedPerWindow - curWindow.getActionCountByIds([cooldown.action.id])), 0)
+
+					this.suggestions.add(new TieredSuggestion({
```
ackwell: By doing this in a loop, there's a real chance of generating a _massive_ number of suggestions.

I'd _generally_ prefer these to be grouped as a single thing - "You missed x important skills" rather than "You missed y foo" + "You missed z bar" + ...

src/parser/core/modules/BuffWindow.tsx
AzariahAeroborn: I guess I don't have enough TS-foo to know how to make sure ACTIONS is typed in a way that getDataBy is returning the correct type.  Without it, I'm getting a TS error at line 102: "Property 'autoAttack' does not exist on type {insert all the 609 different types that "exist" for current ACTIONS objects}".  

Aka, I seem to be choking on the fact that not every ACTION has the autoAttack property defined, and just adding a .hasOwnProperty check doesn't seem to solve the issue.

src/parser/core/modules/BuffWindow.tsx
AzariahAeroborn: I implemented this way to duplicate the way Warrior's IR is currently doing it.  Not opposed to changing, but would need some thought as to how that's going to impact the output for IR.

src/parser/core/modules/BuffWindow.tsx
AzariahAeroborn: _Especially_ if the class is an abstract class to enforce buffAction & buffStatus properties, I can't put the ID check in the init: https://github.com/Microsoft/TypeScript/issues/15247

I didn't have much luck with putting the ID check in the init even after I moved away from an abstract class/props, so I left it this way.

src/parser/core/modules/BuffWindow.tsx
AzariahAeroborn: See also the notes below - I had fiddled both ways with this being an abstract class... changed it to non-abstract to try to fix putting the filters on buffStatus in the init() and then never changed it back when filtering on buffStatus still didn't work in the parent init.  Will change back to abstract since I didn't gain anything by moving away from abstract.

src/parser/core/modules/BuffWindow.tsx
ackwell: Bleh i keep forgetting that

src/parser/core/modules/BuffWindow.tsx
ackwell: 🤢 i had a poke around with the typings and while i _get_ what's going on, it's going to take more time to fix than I have time for immediately, primarily for the statuses.

feel free to move the types back to where they were originally or just delete them and TODO these.

src/parser/core/modules/BuffWindow.tsx
ackwell: Ping me when that's resolved, i remember seeing a convo about it on discord.

AzariahAeroborn: Per discussion with Skye, went ahead and set trackedCooldowns / trackedBadCooldowns to collapse to a single suggestion regardless of number of cooldowns being tracked.

I also updated the naming to consistently use "Action" instead of "Cooldown", since potentially there may be implementers that want to track GCDs as well as oGCDs (e.g. Fight or Flight x Goring Blade)

src/parser/core/modules/BuffWindow.tsx
AzariahAeroborn: Is it going to cause a problem to have ActionData / StatusData in the .d.ts files I have currently in the data directories?  Or can we leave them there (since I do agree they make more sense there rather than glommed onto getDataBy) and if future-you fixes the typings, then this module can be updated accordingly?

src/parser/core/modules/BuffWindow.tsx
```diff
@@ -55,13 +55,18 @@ interface BuffWindowRequiredGCDs {
 	severityTiers: SeverityTiers
 }
 
-interface BuffWindowTrackedCooldown {
-	action: Action
-	expectedPerWindow: number,
+interface BuffWindowTrackedActions {
+	actions: BuffWindowTrackedAction[]
+	iconAction: Action
```
ackwell: Given it's only being used for the icon, makes more sense to just `icon: string` this, and let consumers pick if they want to source that from an action or not.

src/parser/core/modules/BuffWindow.tsx
ackwell: `.d.ts` for something that doesn't actually have an equivalent js file is pretty jank, but we can get away with it for now. It'll be easy enough to clean up when they get typed properly.

PR #518, opened 2019-08-04
Fix aetherflow counts

### LINKED ISSUE #489: Aetherflow stacks count as negative when recitation is used
Apologies if this has already been fixed.

When looking at the aetherflow section for SCH (an [example](https://xivanalysis.com/analyse/4qNYywkKXMQxv19C/1/4/) here), it looks like recitation use messes with the counts for stacks, giving me -1 stacks wasted.

I'm not super familiar with typescript, but I could try to take a stab at it if someone wants to work with me on a PR.

nonowazu:
This addresses a bug with recitation messing with aetherflow counts. The code obviously isn't perfect, but the extra structure should open the door to monitoring recitation uses in future updates.

Fixes #489

### DIFF
--- a/src/parser/jobs/sch/modules/Aetherflow.js
+++ b/src/parser/jobs/sch/modules/Aetherflow.js
@@ -5,6 +5,7 @@ import {Trans} from '@lingui/react'
 import {ActionLink} from 'components/ui/DbLink'
 import {getDataBy} from 'data'
 import ACTIONS from 'data/ACTIONS'
+import STATUSES from 'data/STATUSES'
 import Module from 'parser/core/Module'
 import {Rule, Requirement} from 'parser/core/modules/Checklist'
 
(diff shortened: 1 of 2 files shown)


src/parser/jobs/sch/modules/Aetherflow.js
```diff
@@ -51,11 +73,21 @@ export default class Aetherflow extends Module {
...
+		this._uses.push({timestamp: ts, debit: 1, id: [id]})
+	}
+
 	_onCast(event) {
 		const abilityId = event.ability.guid
 
 		if (AETHERFLOW_CD_ACTIONS.includes(abilityId)) {
-			this._totalAetherflowCasts++
+			// should be the standard case
+			if (!this._recitationActive) {
```
ackwell: nit; these could be part of the same conditional

PR #521, opened 2019-08-04
Fix issue with single stack use

nonowazu:
This should not affect the majority use case of xivanalysis for SCH (instanced raids, etc), but it could look bad if it's content like dungeons or when they have a parse where they started with aetherflow stacks.

For some examples of where this currently breaks, see the following:
* [striking dummy test](https://xivanalysis.com/analyse/M6bkczhpNrAR2Pqd/3/7/)
* [old content where AF could be used outside of combat](https://xivanalysis.com/analyse/nV4z6WXbqjy9vHd1/219/5888/)

Won't help with old content, but I'd rather cover the edge case for current content anyway.

### DIFF
--- a/src/parser/jobs/sch/modules/Aetherflow.js
+++ b/src/parser/jobs/sch/modules/Aetherflow.js
@@ -136,7 +136,7 @@ export default class Aetherflow extends Module {
 				</Table.Row>
 			</Table.Header>
 			<Table.Body>
-				{[].concat(aetherflows, dissipations, ...uses)
+				{[].concat(aetherflows, dissipations, uses)
 					.sort((a, b) => a.timestamp - b.timestamp)
 					.reduce((prev, curr) => {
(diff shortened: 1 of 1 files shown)


src/parser/jobs/sch/modules/Aetherflow.js
```diff
@@ -189,6 +189,13 @@ export default class Aetherflow extends Module {
...
 							totalWasted += wasted
 						}
+
+						if (!Array.isArray(timestamp)) {
+							// I mean, they should be doing more than one AF cast per stack
+							// but who am I to judge?
+							timestamp = [timestamp]
+						}
+
 						return <Table.Row key={timestamp}>
```
ackwell: might be soem warnings on this, not sure if react is super happy about getting an array as a key. might be ok i guess.

will def re-render every time though.

PR #525, opened 2019-08-06
Core+DNC allowable dropped combos

AkairyuGestalter:
Core Combos updated to include overrideable function that allows subclasses to flag broken combos as allowable.
DNC combos module to select an appropriate icon, and allow Flourish to drop if both Standard and Technical Steps are used during Cascade's combo window.

Note that this will be affected by #514 since it's currently referencing the AoeEvent from Core Combos.

### DIFF
--- a/src/parser/core/modules/Combos.tsx
+++ b/src/parser/core/modules/Combos.tsx
@@ -236,19 +236,23 @@ export default class Combos extends Module {
 		}))
 	}
 
+	/**
+	 * To be overridden by subclasses. This is called in _onComplete() and passed two arrays of event objects - one for events that
+	 * broke combos, and one for combo GCDs used outside of combos. Subclassing modules can add job-specific suggestions based on
+	 * what particular actions were misused and when in the fight.
+	 * The overriding module should return true if the default suggestion is not wanted
+	 */
 	addJobSpecificSuggestions(comboBreakers: AoeEvent[], uncomboedGcds: AoeEvent[]) {
-		// To be overridden by subclasses. This is called in _onComplete() and passed two arrays of event objects - one for events that
-		// broke combos, and one for combo GCDs used outside of combos. Subclassing modules can add job-specific suggestions based on
-		// what particular actions were misused and when in the fight.
-		// The overriding module should return true if the default suggestion is not wanted
 		return false
 	}
 
+	/**
+	 * To be overridden by subclasses. This is called in recordBrokenCombo, and receives the event triggering the broken combo,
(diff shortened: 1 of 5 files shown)


src/parser/core/modules/Combos.tsx
```diff
@@ -242,6 +244,14 @@ export default class Combos extends Module {
 		return false
 	}
 
+	isAllowableComboBreak(event: AoeEvent, context: AoeEvent[]): boolean {
+		// To be overridden by subclasses. This is called in recordBrokenCombo, and receives the event triggering the broken combo,
```
ackwell: Info should be a TSDoc comment above the function definition so that intellisense and stuff will see it, i.e.

```ts
/**
 * To be overridden...
 */
isAllowableComboBreak(...) {}
```

src/parser/jobs/dnc/modules/Combos.ts
```diff
@@ -0,0 +1,25 @@
...
+	// Override statics
+	static suggestionIcon = ACTIONS.CASCADE.icon
+
+	@dependency private dancing!: DirtyDancing
+
+	// Override check for allowable breaks. If two dances were started during the initial context's GCD timeout window,
+	// (ie, both Standard and Technical were danced), then we'll allow it
+	isAllowableComboBreak(event: AoeEvent, context: AoeEvent[]): boolean {
+		// Shouldn't ever be the case, but protect against weird shit
+		if (context.length !== 1) {
```
ackwell: geeeenerally avoid explicit numbers on arrays and similar if it's avoidable - `if (content.length < 1)`, and `dancesInRange(...) >= 2`. It won't nessicarily get hit, but it means it'll be handled sanely if it does.

src/parser/core/modules/Combos.tsx
AkairyuGestalter: Can do. Want me to change that for the addJobSpecificSuggestions comment while I'm at it?

src/parser/jobs/dnc/modules/Combos.ts
AkairyuGestalter: I was intentional about the explicit length check. Since DNC's combo is a two-step, the only valid context we should be getting is length === 1 so anything else is probably bad data (might even be worth flagging as a broken log...?). Its primary purpose though is to make sure the context[0].timestamp references later don't bomb if something odd _does_ sneak through.

src/parser/core/modules/Combos.tsx
AkairyuGestalter: Nvm, I did it anyways :)

PR #527, opened 2019-08-07
Track Chain Stratagem

nonowazu:
A simple addition to track the SCH raid buff :)

If you need me to update all the files in `locale`, I'd be happy to add stubs for those as well.
Also, I'm unsure if I'm being doubly redundantly redundant by having the Chain Stratagem ActionLink in the description.


Toastdeib:
You'll need to update locale, yes, but not manually - those files are all auto-generated. Just run `yarn extract`, it'll take care of the rest, and then you can commit what it spits out.

nonowazu:
Whoops! I've corrected the issue. Thanks for letting me know!

src/parser/jobs/sch/modules/ChainStrat.js
```diff
@@ -0,0 +1,18 @@
...
+
+const CHAIN_STRAT_START_OFFSET = 10000
+const CHAIN_STRAT_OFFSET = 7500
+
+export default class ChainStrat extends CooldownDowntime {
+	trackedCds = [ACTIONS.CHAIN_STRATAGEM.id]
+	allowedDowntime = CHAIN_STRAT_OFFSET
+	firstUseOffset = CHAIN_STRAT_START_OFFSET
+	description = <Trans id="sch.chainstrat.cooldown.description">
+		Try to use <ActionLink {...ACTIONS.CHAIN_STRATAGEM}/> on cooldown throughout the fight, particularly when your team has other buffs up for maximum effect.\
```
suShirow: There's an extra slash at the end of this string

nonowazu: Well, that was dumb of me. The next commit should fix that.

nonowazu:
I swear, I'll get the hang of this commit -> `yarn extract` cycle. One more change inbound 😢

suShirow:
> I swear, I'll get the hang of this commit -> `yarn extract` cycle. One more change inbound 😢

Haha, I've been on this ride before too

zorkian:
LGTM as a Scholar. For a "viewing one single character analysis", CS should be on cooldown basically 100% of the time. I'm not much of a JS guy but the code seems reasonable too.

ackwell:
Ah fukkit; this is missing a changelog entry

can you add a changelog entry for this with the next sch update pl0x

PR #528, opened 2019-08-07
Added a new module for tracking bad True Thrust uses

Toastdeib:
What it says on the tin. Pretty simple module, should yell at people for using True Thrust anywhere in a normal rotation (start of fight and coming out of downtime are forgiven).

### DIFF
--- a/src/parser/jobs/drg/modules/RaidenThrust.js
+++ b/src/parser/jobs/drg/modules/RaidenThrust.js
@@ -1,0 +1,62 @@
+import {Trans, Plural} from '@lingui/react'
+import React from 'react'
+
+import {ActionLink} from 'components/ui/DbLink'
+import ACTIONS from 'data/ACTIONS'
+import Module from 'parser/core/Module'
+import {TieredSuggestion, SEVERITY} from 'parser/core/modules/Suggestions'
+
+export default class RaidenThrust extends Module {
+	static handle = 'raiden'
+	static dependencies = [
+		'downtime',
+		'suggestions',
+	]
+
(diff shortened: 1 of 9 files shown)


src/parser/jobs/drg/modules/RaidenThrust.js
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	_onTrueThrustCast(event) {
+		const downtime = this.downtime.getDowntime(this._lastOpenerCast, event.timestamp)
+		this._lastOpenerCast = event.timestamp
+		if (downtime > 0) {
+			// We're coming out of downtime, so TT is okay
+			// Note: This -does- make the assumption that the downtime was long enough to break the combo, so it may forgive the rare case
+			// where the player reset their combo after a short (like, 10s or less) downtime without needing to, but that should be fine
+			this._trueThrustOkay = true
```
AkairyuGestalter: Assuming getDowntime gives you a timespan of how much downtime there was in the range, could you compare downtime >= GCD_TIMEOUT_MILLIS (See core Combos, it's not exported atm but I could change that on #525) to determine if the downtime was long enough to force a combo break?

Toastdeib: I did consider that, but it wasn't exported and it honestly didn't feel necessary. If the downtime wasn't long enough to break the combo and they still TT'd, it means they broke the combo manually and should be dinged regardless.

ackwell: drop a note when you're happy to get this merged 👍

PR #531, opened 2019-08-08
SMN - pet death when caster dies, filter pet actions for checking which pet is out

AzariahAeroborn:


src/parser/jobs/smn/modules/Pets.js
```diff
@@ -134,7 +136,6 @@ export default class Pets extends Module {
 
 	_onEvent(event) {
 		if (
-			(this._lastPet || this.parser.byPlayerPet(event)) &&
```
ackwell: Ah; this should probably be left as-is, with no filter (as much as I hate to say that). It needs a bit of a rework, but should be safe in the short term - it needs to fire as often as possible so it can swap the pet from demi -> normal as close to the correct time as it can.

This is actually a perfect use-case for the "timestamp hook" I was dreaming up for core gauge - I'll add a note to revisit this once I've got that running.

AzariahAeroborn: Actually, that isn’t a necessary change, since the swap back logic defines the end of the Demi’s life as 20000ms (per the defined Demi life span constant) after the summon event, unless onPetDeath fires earlier.  Under normal circumstances, the pet swap event gets synthesized “earlier” than the next event that triggers the swap back logic. 

If you compare the results with and without your adjustment commit, you should see exactly the same Demi summon lengths.

PR #539, opened 2019-08-09
DNC: Show procs on timeline

AkairyuGestalter:
Added timeline display for proc buffs, based on BLM
Also adjusting isAllowableComboBreak, tech windows be that way


src/parser/jobs/dnc/modules/Procs.js
```diff
@@ -29,28 +40,86 @@ export default class Procs extends Module {
 		[STATUSES.FLOURISHING_WINDMILL.id]: 0,
 	}
 
+	_buffWindows = {
```
ackwell: consider generating initial states lazily so you don't need to write them all out one by one.

src/parser/jobs/dnc/modules/Procs.js
```diff
@@ -92,6 +161,41 @@ export default class Procs extends Module {
...
+			})
+		})
 	}
 
+	getGroupIdForStatus(status) {
+		const groupId = 'procbuffs-' + status.id
+
+		// Make sure a timeline group exists for this buff
+		if (!this._group.nestedGroups.includes(groupId)) {
+			this.timeline.addGroup(new Group({
```
ackwell: might get some value from `ItemGroup`s for this

PR #541, opened 2019-08-10
Add fairy gauge and open SCH for business

nonowazu:
This is actually my first 'full' module, so I do appreciate any feedback that you might have on this.

Changes (in no particular order):

* Module to track fairy gauge usage - there is also some groundwork to also track fairy uptime, but since the pet can't die unless the player dies, I think this a future enhancement
* Update and add some SCH/Fairy actions and statuses - some were to promote gauge tracking, the others were for potential future tracking
* Update changelog (because I forgot last time!!)
* Add support for patches 5.0 - 5.05

I did at least remember to `yarn extract` this time, but I fully expect that I've forgotten something, or that there's potential issues with my js.


src/parser/jobs/sch/modules/FaerieGauge.js
```diff
@@ -0,0 +1,173 @@
...
+	// both spenders and generators consume the same amount
+	_onGaugeSpend() {
+		this._gauge -= GAUGE_GAIN_AMOUNT
+		this._updateHistory()
+	}
+
+	_onGaugeGenerate() {
+		if (this.combatants.selected.hasStatus(STATUSES.DISSIPATION.id) || ! this._fairyOut) {
+			// can't generate a guage without the fairy, so bail out
+			return
```
AzariahAeroborn: Should this track as waste for usage with no fairy out?  Especially when they're not under Dissipation, and just haven't resummoned the fairy after death, though I imagine there are cases where it's fine to burn off AF stacks while under Dissipation.

nonowazu: Perhaps track in a separate variable? `this._noFairyWaste`, perhaps? I can see some situations where you come back from death and absolutely need the aetherflow and moves for immediate health and MP?

AzariahAeroborn: Per our discussion on Discord, I’m ok with not tracking “waste” due to not having a fairy out, and just having a future recommendation if someone doesn’t resummon the fairy promptly after a death.

nonowazu: Yeah, I agree. Seriously, though - thank you for the suggestions and input!

src/parser/jobs/sch/modules/FaerieGauge.js
```diff
@@ -0,0 +1,188 @@
...
+		}
+
+		this._gauge += GAUGE_GAIN_AMOUNT
+		// figure out if we've capped and add to waste
+		if (this._gauge > GAUGE_MAX) {
+			this._gauge = GAUGE_MAX
+			this._waste += GAUGE_GAIN_AMOUNT
```
ackwell: should be safe given how fairy gauge works, but generally you'd want to mark the discrete amount overcapped as waste (mainly on classes that might have 95/100 then get a +10)

PR #542, opened 2019-08-10
Completed SMN changes for Shadowbringers

KelosPelmand:


src/parser/jobs/smn/modules/Pets.js
```diff
@@ -190,49 +246,59 @@ export default class Pets extends Module {
...
 
-		// Disabling ifrit check post-4.4 due to changes made to RS
-		if (
-			this.parser.patch.before('4.4') &&
-			numCasters === 1 && mostUsedPet !== PETS.IFRIT_EGI.id
-		) {
+		// Since Ifrit's EA2 and Enkindle hit AoE, we can sometimes tell that
+		// it was used in an AoE situation, but we can't tell any more often than
+		// when those particular skills are used.  Therefore, just list this as Medium
+		// rather than trying to determine a percentage like we did with Garuda.
```
ackwell: I'm a _little_ wary of jumping straight to `MEDIUM` for this - though I can see a level of logic in it, given the CDs on his AoEs are relatively lengthy compared to the summon CDs and such.

src/parser/jobs/smn/modules/Pets.js
```diff
@@ -190,49 +246,59 @@ export default class Pets extends Module {
...
+		if (this._ifritMultiHits > 0) {
 			this.suggestions.add(new Suggestion({
 				icon: ACTIONS.SUMMON_III.icon,
 				severity: SEVERITY.MEDIUM,
-				content: <Trans id="smn.pets.suggestions.prefer-ifrit.content">
-					You should be primarily using Ifrit-Egi when there are no other casters in the party - Ifrit's raw damage and <ActionLink {...ACTIONS.RADIANT_SHIELD}/> provide more than Garuda can bring to the table in these scenarios.
+				content: <Trans id="smn.pets.suggestions.ifrit-aoe.content">
+					Ifrit-Egi should only be used for single target.
```
ackwell: Same story as Gary w/r/t content brevity.

src/parser/jobs/smn/modules/Pets.js
```diff
@@ -190,49 +246,59 @@ export default class Pets extends Module {
...
 				</Trans>,
-				why: <Trans id="smn.pets.suggestions.prefer-ifrit.why">
-					{this.getPetUptimePercent(mostUsedPet)}% {this.getPetName(mostUsedPet)} uptime, Ifrit-Egi preferred.
+				why: <Trans id="smn.pets.suggestions.ifrit-aoe.why">
+					Ifrit-Egi hit multiple targets with {this._ifritMultiHits} uses of <ActionLink {...ACTIONS.FLAMING_CRUSH} /> or <ActionLink {...ACTIONS.INFERNO} />.
```
ackwell: Could do with a plural on the `uses`

src/parser/jobs/smn/modules/Ruin2.js
```diff
@@ -109,29 +116,32 @@ export default class Ruin2 extends Module {
 
 	_onComplete() {
 		const potLossPerR2 = RUIN3_POT - RUIN2_POT
-		const issues = this._issues.length
-		const warnings = this._warnings.length
+		// It would be a waste to not use any stacks of Further Ruin left after the
+		// last weave of the fight, so do not warn about them.
+		const issues = this._issues.filter(cast => cast.ability.guid === ACTIONS.SMN_RUIN_II.id || cast.timestamp < this._lastOgcd.timestamp).length
```
ackwell: nitpicking; could move that filter fn out so you don't need to define the same logic twice

src/parser/jobs/smn/modules/Ruin4.tsx
```diff
@@ -0,0 +1,137 @@
...
+			if (fightTimeRemaining > END_OF_FIGHT_LEEWAY) {
+				this.playerSkillCount++
+			}
+		} else if (event.ability.guid === ACTIONS.ASSAULT_I_EARTHEN_ARMOR.id) {
+			this.earthenArmorCount++
+		}
+	}
+
+	private onPetCast(event: CastEvent) {
+		if (FURTHER_RUIN_PET_ACTIONS.includes(event.ability.guid) &&
```
ackwell: this action filter can be moved into the hook filter, can also invert the conditional to avoid the indentation

src/parser/jobs/smn/modules/Ruin4.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+	protected init() {
+		this.addHook('cast', {by: 'player'}, this.onPlayerCast)
+		this.addHook('cast', {by: 'pet'}, this.onPetCast)
+		this.addHook('death', {to: 'player'}, this.onDeath)
+		this.addHook('complete', this.onComplete)
+	}
+
+	private onPlayerCast(event: CastEvent) {
+		if (event.ability.guid === ACTIONS.RUIN_IV.id) {
```
ackwell: not blocking; but each of these conditional blocks could be implemented as hook filters

src/parser/jobs/smn/modules/Ruin4.tsx
```diff
@@ -0,0 +1,137 @@
...
+			}
+		}
+	}
+
+	private onDeath() {
+		this.currentStackCount = 0
+	}
+
+	private onComplete() {
+		if (this.overage > 0) {
```
ackwell: same story re: automatic hiding from the tiered suggestions

src/parser/jobs/smn/modules/Ruin4.tsx
```diff
@@ -0,0 +1,137 @@
...
+				</Trans>,
+				tiers: SEVERITY_STACK_COUNT,
+				value: this.overage,
+				why: <Trans id="smn.ruin-iv.overage.why">
+					{this.overage} Further Ruin <Plural value={this.overage} one="stack was" other="stacks were"/> lost.
+				</Trans>,
+			}))
+		}
+
+		let numberLost = this.earthenArmorCount
```
ackwell: I'm not _entirely_ sure how comfortable I am considering Earthen Armor as a total loss, but can leave it for now. Would be good to go over these sort of details with Nem once we're both back from holidays and such.

src/parser/jobs/smn/modules/Ruin4.tsx
```diff
@@ -0,0 +1,137 @@
...
+	private onDeath() {
+		this.currentStackCount = 0
+	}
+
+	private onComplete() {
+		if (this.overage > 0) {
+			this.suggestions.add(new TieredSuggestion({
+				icon: STATUSES.FURTHER_RUIN.icon,
+				content: <Trans id="smn.ruin-iv.overage.content">
+					You used Egi Assault skills in a way that overcapped your stacks of <ActionLink {...STATUSES.FURTHER_RUIN}/>.
```
ackwell: Try to avoid telling the user what they _did_, tell them what they _should do_. (goes for the other suggestions below, too)

src/parser/jobs/smn/modules/Pets.js
KelosPelmand: I wasn't quite sure how to level that either.  On one hand, its possible that there were only multiple targets for a GCD or two around those usages.  On the other hand, every single pet auto attack could theoretically be with multiple targets available and you can't really tell.  Given the primary usage of the site is raids, which have limited multi-target periods, I'd also be fine just marking this minor.  I'm not sure what tiers I'd use if we wanted to tier it.

ackwell: I think it's safe to leave as-is for now - just keep an eye out for feedback, see if there's complaints which might warrant changing severities around.

KelosPelmand:
This should have all the requested changes now.

PR #543, opened 2019-08-11
Fix War's Gauge Overcapping because of Inner Chaos's cost was ignored

Noirvent:
Probably was it just an oversight during the first pass for 5.0, but Inner Chaos's rage cost is not being substracted from the gauge, resulting in a major warning for overcapping the gauge by several hundreds.

I also re-added Steel Cyclone, Fell Cleave, Decimate and Inner Beast to the INFURIATE_REDUCERS since they still reduce it, along Chaotic Cyclone and Inner Chaos (although the tooltip for improved infuriate doesn't mention the last two, they do reduce the cooldown).

There's also Mythril Tempest who should be added to the RAGE_GENERATORS list for 20 gauge, but I didn't do it because it's a combo bonus and I didnt check how that was handled yet

As this is the first time ever I'm doing a PR, don't hesistate to tell me if I'm doing it wrong


KelosPelmand:
It appears you should be able to just add Mythril Tempest to the generator list, since all the others on that list also only generate in a combo.  Looking at the hooks, it is hooking the combo hook (which i assume sends events for every correctly combo'd gcd) and filtering to just the gcds in the generators array.

NamelessFaceless:
`Inner Beast` also turns into `Fell Cleave´ due to the Mastery trait. Having it on the list is irrelevant.

src/parser/jobs/war/modules/Gauge.js
```diff
@@ -27,10 +27,16 @@ const RAGE_SPENDERS = {
...
 	[ACTIONS.ONSLAUGHT.id]: 20,
+	[ACTIONS.CHAOTIC_CYCLONE.id]:50,
+	[ACTIONS.INNER_CHAOS.id]:50,
 }
 
 // Actions that reduce Infuriate
 const INFURIATE_REDUCERS = [
+	ACTIONS.FELL_CLEAVE.id,
+	ACTIONS.INNER_BEAST.id,
+	ACTIONS.STEEL_CYCLONE.id,
```
NamelessFaceless: As mentioned before, irelevant to have Inner Beast and Steel Cyclone. We don't support logs that would have these in, since a trait upgrades them to Fell Cleave and Decimate, respectively. Otherwise, it looks good.

PR #545, opened 2019-08-11
Initial core gauge implementation

ackwell:
This implements the initial core gauge module (not that there's much of it), as well as a simple abstract gauge module, and an implementation for the standard stack/numeric/counter type gauge.

No timer gauge for this PR, it'll require some changes elsewhere in core that need to be done first.

Likewise, no automation at this point - just setting out the framework, can chip common behavior into it as it becomes apparent.

Also includes a port of WAR's beast gauge to the core gauge system as a testing bed.

Also includes a minor tweak to how `init` is called so that it has proper access to class properties.


ackwell:
zoom zoom

PR #547, opened 2019-08-11
WIP: Adjust CooldownDowntime calculation method

AzariahAeroborn:
The current calculation method for oGCD cooldown downtime results in several unexpected behaviors around fights that are just longer than an even number of cooldown usages, especially when the player is really bad at using a particular tracked cooldown.  

Currently, the CooldownDowntime module calculates as:
- Determine max possible uses based on fight duration - a 60 second cooldown in a 10:02 fight will have 11 possible uses max
- Determine how long the skill was off cooldown between each use, and reduce those "downtime"s by defined allowable amounts or time the boss was invulnerable while skill was off cooldown
- Calculate how many skills we think they used by taking max possible and subtracting (downtime/cooldown), with an adjustment for end of fight leftover time.

The specific way the calculation is being handled means that a 60 second cooldown in a 10:02 fight will calculate 10/11 uses if one use was missed, even though only 9 uses were actually recorded (leading to "bad" percentages), and the calculated number of uses can go negative if the skill is never used and there's the right combination of fightDuration compared to skillCooldown with the calling module providing no defined firstUseOffset delay.

Proposed change is two parts:
- Synthesize cast events for pre-pull statuses, so we can count those uses.  This resolves the primary blocker to counting actual uses, for skills like Acceleration which are routinely used pre-pull.  
- Change to counting actual uses, and adjusting the computation for max uses to account for allowable drift between uses.  In order to simplify the calculation and presentation for this drift calculation, change the allocation of "invulnerable time" to the cooldown usage _after_ the one that came off cooldown before the boss went invulnerable.

The only other consumer of the getTimeOnCooldown method is the SCH Aetherflow module, which is not passing a true value to considerInvulnTime, so will not be effected by the change to the invulnerable time calculations.

For purposes of WIP, I've only added a cross-reference ID between the status and the action for RDM's Acceleration.  Assuming these changes are determined to be viable overall, I'll go through the other data files and add appropriate cross-reference IDs where necessary.

### DIFF
--- a/src/data/STATUSES/RDM.js
+++ b/src/data/STATUSES/RDM.js
@@ -20,6 +20,7 @@ export default {
 		id: 1238,
 		name: 'Acceleration',
 		icon: 'https://xivapi.com/i/013000/013405.png',
+		actionId: 7518,
 	},
 	EMBOLDEN_PHYSICAL: {//Note that this is the Embolden that boosts Physical Damage - what other people receive from RDM
 		id: 1297,
--- a/src/parser/core/modules/CooldownDowntime.js
+++ b/src/parser/core/modules/CooldownDowntime.js
@@ -39,10 +39,7 @@ export default class CooldownDowntime extends Module {
 		this.addHook('complete', this._onComplete)
 	}
 
-	_onComplete(event) {
-		const endTime = event.timestamp
-		const startTime = this.parser.fight.start_time
-		const encounterLength = endTime - startTime
+	_onComplete() {
 		const OGCDRequirements = []
 
 		const groups = {}
@@ -55,7 +52,7 @@ export default class CooldownDowntime extends Module {
 			if (key === UNGROUPED_COOLDOWN_ID) {
 				//handle non-grouped cooldowns
 				groups[key].map(id => {
-					OGCDRequirements.push(this.checkCooldown(id, <ActionLink {...getDataBy(ACTIONS, 'id', id)} />, encounterLength))
+					OGCDRequirements.push(this.checkCooldown(id, <ActionLink {...getDataBy(ACTIONS, 'id', id)} />))
 				})
 			} else {
 				//handle cooldown group
@@ -65,7 +62,7 @@ export default class CooldownDowntime extends Module {
 						<ActionLink {...getDataBy(ACTIONS, 'id', id)} />
 					</>
 				})
-				OGCDRequirements.push(this.checkCooldown(groups[key][0], displayName, encounterLength))
+				OGCDRequirements.push(this.checkCooldown(groups[key][0], displayName))
 			}
 		}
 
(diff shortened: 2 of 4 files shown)


KelosPelmand:
I think the described algorithm will give "missed usages" in some cases it should not.  For example:

Consider the same 60s cd with the same 10:02 fight duration, but there is a downtime from 1:50 to 2:10.  If the player uses the skill at 0 and 1:05 and 2:10 and on CD after that, the described algorithm sounds like it would:
* Mark the 5s the skill was off cd after the first usage (1:00-1:05), see that it is not in an invuln period, and mark as "lost time" 
* See the 5s the skill was off cd after the second usage (2:05-2:10), see that it is in an invuln period, and mark it as "ok"

The correct behavior in this case would be to mark both 5s intervals as "ok" since that drift does not contribute to losing a usage in this case.

I think the goal overall should be to find a way to calculate number of expected usages, and just count actual usages from there (synthing precasts as needed).  I have an idea of an algorithm to calculate expected usages, but I haven't yet reached the point I'm confident it counts correctly in all cases.

AzariahAeroborn:
Even presuming the implementing module didn't allow for any first use delay or between-uses allowed drift, the scenario you described would result in 10 possible uses being calculated:

- Fight length: 602 seconds
- Allowable drift due to module definitions: 0
- Allowable drift due to boss invulnerability: 5 seconds
- Adjusted fight length: 597 seconds
- Expected uses: Math.ceil(597 seconds / 60 seconds) = 10

Whereas, by contrast, the current implementation would come up with 10/11 possible uses:
- Fight length: 602 seconds
- Expected uses: Math.ceil(602 seconds / 60 seconds) = 11
- Skill downtime between first and second use: 5 seconds
- Skill downtime between second and third use: 0 seconds (due to boss invulnerability)
- Lost uses due to downtime: Math.floor(5 seconds / 60 seconds) = 0
- Lost uses due to downtime versus fight remainder: (5 seconds > 2 seconds) ? 1 : 0 = 1
- Total lost uses: 1
- Percentage: 10 / 11

KelosPelmand:
Hmm, that is true.  The problem does come up if the fight extends to 10:07

* Fight length: 607 seconds
* Allowable drift due to module definitions: 0
* Allowable drift due to boss invulnerability: 5 seconds
* Adjusted fight length: 602 seconds
* Expected uses: Math.ceil(602 seconds / 60 seconds) = 11

In reality though, the earliest possible time for the 11th use is 10:10, which means there are only really 10 uses possible.

This doesn't even get into allowable drift between uses.  Once that comes into play, things get weird, since you can have situations, such as the 10:02 fight, 60s cd, but now 2s allowed drift per use, where 10 uses is "100%" because the drift was allowed, even though a player that doesn't drift can get 11.

Personally, I'd be ok with the module just finding the max possible uses with maximum allowed drift on each use and just displaying any parse with that many or more uses as 100%

AzariahAeroborn:
Closing this in favor of splitting the changes into two parts, per Discord discussion

PR #548, opened 2019-08-12
Movement Skills Module

Cephid:
New Module for CaC and Displacement|Engagement usage relative to Manafication uses.  Also moved overcapped mana from manafication to its own set of statistics at TC request.  The final math does give quite a bit of grace back to the user on %'s due to the fact that a fight can end before you can get another manafication.  This PR marks RDM as a supported class for 5.05.


src/parser/jobs/rdm/index.js
```diff
@@ -43,8 +48,8 @@ export default new Meta({
 			contributors: [CONTRIBUTORS.LEYLIA],
 		},
 		{
-			date: new Date('2019-07-26'),
-			Changes: () => <>Fix an issue with Moulinet Mana costs</>,
+			date: new Date('2019-07-14'),
+			Changes: () => <>Initial update of existing logic for 5.0 (no new stuff)</>,
```
ackwell: looks like this got kludged somehow

src/parser/jobs/rdm/modules/Gauge.js
```diff
@@ -314,6 +341,28 @@ export default class Gauge extends Module {
 					<Trans id="rdm.gauge.suggestions.black-mana-invuln-why">You lost {this._blackManaLostToInvulnerable} Black Mana due to misses or spells that targeted an invulnerable target</Trans>
 				</Fragment>,
 			}))
+
+			this.statistics.add(new SimpleStatistic({
```
ackwell: I'd be inclined to merge these into a single statistic, given their related nature. Not necessarily add the values together, just display both in the same box.

src/parser/jobs/rdm/modules/MovementSkills.js
```diff
@@ -0,0 +1,205 @@
...
+import {getDataBy} from 'data'
+
+const CACS_PER_MANAFICATION = 3
+const DISPS_PER_MANAFICATION = 3
+const ENGAGEMENTS_PER_MANAFICATION = 4
+//This is due to the fact that manafication happens after your first cac/disp
+const GRACE_FOR_PULL = 2
+
+export default class MovementSkills extends Module {
+	static handle = 'movementskills'
```
ackwell: ubernit; `movementSkills`

src/parser/jobs/rdm/modules/MovementSkills.js
```diff
@@ -0,0 +1,205 @@
...
+		cac: 0,
+		disp: 0,
+		engagement: 0,
+		events: [],
+	}
+
+	constructor(...args) {
+		super(...args)
+		//Default Target to hit
+		this.target = 95
```
ackwell: can move this and the description to the checklist so it's not polluting the class interface

src/parser/jobs/rdm/modules/MovementSkills.js
```diff
@@ -0,0 +1,205 @@
...
+
+export default class MovementSkills extends Module {
+	static handle = 'movementskills'
+	static title = t('rdm.movementskills.title')`Movement Skills`
+	static dependencies = [
+		'checklist',
+	]
+
+	_history = []
+	_last_manafic = {
```
ackwell: nit; `_lastManafic`

src/parser/jobs/rdm/modules/MovementSkills.js
```diff
@@ -0,0 +1,205 @@
...
+			this._history.push(this._last_manafic)
+			//setup the new instance, since we're unlikely to end the fight on manafication push
+			this._last_manafic = {
+				timestamp: event.timestamp,
+				cac: 0,
+				disp: 0,
+				engagement: 0,
+				events: [],
+			}
+			break
```
ackwell: if you `return` here, you can move the `events.push` outside the switch so it's not duped everywhere

src/parser/jobs/rdm/modules/MovementSkills.js
```diff
@@ -0,0 +1,205 @@
...
+	_onComplete() {
+		let cacs = 0
+		let disp = 0
+		let engagement = 0
+		let manafics = 0
+		const requirements = []
+		this._history.push(this._last_manafic)
+
+		//Parse out the final numbers
+		this._history.map(manafic => {
```
ackwell: I'd rather do this with a `.reduce` than a `.map`, personally.

src/parser/jobs/rdm/modules/MovementSkills.js
```diff
@@ -0,0 +1,205 @@
...
+		}
+
+		return new Requirement({
+			name: <ActionLink {...getDataBy(ACTIONS, 'id', id)} />,
+			percent: finalValue,
+		})
+	}
+
+	_checkDisp(disps, engagements, dispID, manafics) {
+		let noFail = 0
```
ackwell: looks like this should be a boolean. given the `if` case below only sets the value, can probably just move the `if` conditional to the right side of the assignment.

src/parser/jobs/rdm/modules/MovementSkills.js
Cephid: I only want to capture disp, engage, and cac and absolutely no other events; wouldn't moving it outside the switch cause me to capture every single event?

ackwell: If you only want those, should probably filter the hook as such - sorry for missing that.

can also utilise the default case to return, but that's pretty ugly.

would be good if you could avoid the conditionals, too

ackwell:
Ah; looks like this was off `master`.

Make sure you rebase your master onto `upstream`, as the histories will now have diverged - you'll get weird commit histories if you don't.

PR #550, opened 2019-08-13
Fixing some Devilment usage check bugs

AkairyuGestalter:
Any given window will now only give either the early or the late suggestion, not both. Also fixed a bug with pre-pull Devilments where the 'first use' was later than the cooldown but was still being marked as ok to happen pre-window.

### DIFF
--- a/src/parser/jobs/dnc/modules/Technicalities.tsx
+++ b/src/parser/jobs/dnc/modules/Technicalities.tsx
@@ -33,6 +33,7 @@ const WINDOW_STATUSES = [
 
 const FEATHER_THRESHHOLD = 3
 const POST_WINDOW_GRACE_PERIOD_MILLIS = 500
+const DEVILMENT_COOLDOWN_MILLIS = ACTIONS.DEVILMENT.cooldown * 1000
 
 class TechnicalWindow {
 	start: number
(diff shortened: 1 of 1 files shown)


src/parser/jobs/dnc/modules/Technicalities.tsx
```diff
@@ -159,24 +171,25 @@ export default class Technicalities extends Module {
 	private handleDevilment(lastWindow: TechnicalWindow | undefined) {
 		// Don't ding if this is the first Devilment, depending on which job the Dancer is partnered with, it may
 		// be appropriate to use Devilment early. In all other cases, Devilment should be used during Technical Finish
-		if (!this.combatants.selected.hasStatus(STATUSES.TECHNICAL_FINISH.id) && this.firstDevilment) {
+		if (!this.combatants.selected.hasStatus(STATUSES.TECHNICAL_FINISH.id) && (this.lastDevilmentTimestamp < 0 ||
+			// If the first use we detect is after the cooldown, assume they popped it pre-pull and this 'first'
+			// Use is actually also bad
+			this.parser.currentTimestamp >= DEVILMENT_COOLDOWN_MILLIS)) {
```
ackwell: not gonna block the pr on it (looks like it's going to need a lint rule), but please pref. breaking long boolean conditionals onto a new line a-la

```js
if (
	foo ||
	bar
) {
	// ...
}
```

just keeps it consistent with the rest of the code style

PR #556, opened 2019-08-15
Update Error printer to add a refresh button for NotFound errors

AzariahAeroborn:
This should give users a better error message and a way to try to self-fix NotFound errors that occur due to live logging reports being cached from an earlier version.


src/components/ErrorBoundary.js
```diff
@@ -15,6 +18,27 @@ class ErrorBoundary extends Component {
...
+	refreshFights = () => {
+		const {reportStore} = this.context
+		reportStore.refreshReport()
+	}
+
+	renderRefreshButton = (error) => {
+		if (error instanceof NotFoundError) {
+			return <Header>
+				<div className={styles.refreshheader + ' pull-right'}>
+					<span className={styles.refresh} onClick={this.refreshFights}>
```
ackwell: Doesn't need doing _now_, but this would be a prime candidate to pull out into a shared component (inc. the click handling)

src/components/ErrorBoundary.js
```diff
@@ -15,6 +18,27 @@ class ErrorBoundary extends Component {
...
 
+	refreshFights = () => {
+		const {reportStore} = this.context
+		reportStore.refreshReport()
+	}
+
+	renderRefreshButton = (error) => {
+		if (error instanceof NotFoundError) {
+			return <Header>
+				<div className={styles.refreshheader + ' pull-right'}>
```
ackwell: nit; `classNames(styles.refreshHeader, 'pull-right')` (will need an import)

while it realistically doesn't do much more than join the classes in this use-case, it's "safer" to leave that logic to something else.

src/components/ErrorBoundary.js
```diff
@@ -15,6 +18,27 @@ class ErrorBoundary extends Component {
 
 	static contextType = StoreContext
 
+	refreshFights = () => {
+		const {reportStore} = this.context
+		reportStore.refreshReport()
+	}
+
+	renderRefreshButton = (error) => {
+		if (error instanceof NotFoundError) {
```
ackwell: Should probably invert this conditional. Needs silly wrapping `(!(...))`, but better than all that indentation.

src/errors.js
```diff
@@ -50,7 +50,7 @@ export class NotFoundError extends GlobalError {
 		}, options))
 	}
 	get detail() {
-		return `No ${this.type} was found with the specified ID.`
+		return `No ${this.type} was found with the specified ID.  If this report was generated by live logging, it may be cached -- try pressing Refresh to reload it.`
```
ackwell:
```suggestion
		return `No ${this.type} was found with the specified ID. If this report was generated by live logging, it may have been cached - try pressing Refresh to reload it.`
```

src/components/ErrorBoundary.module.css
```diff
@@ -0,0 +1,8 @@
+.refresh {
+	font-size: 1rem;
+	cursor: pointer;
+}
+
+.refreshheader {
```
ackwell:
```suggestion
.refreshHeader {
```

PR #557, opened 2019-08-15
BRD Replaced repeated code with shared functions

Ririan:
While working on multiple modules at the same time, I ended up having to reuse the same code in multiple places so my PRs wouldn't become locked behind others.  Seeing as how all of my PRs have been merged, this is a pass through to DRY this all up.

Namely, functions for creating a timestamp button, formatting a damage number, and calculating the approximate potency have all been moved into the Util module. Pitch Perfect and Sidewinder modules have been changed to reflect this.

As a part of this, I have converted the Util module to typescript for practice.


src/parser/jobs/brd/modules/Util.tsx
```diff
@@ -0,0 +1,155 @@
...
+	name: string
+	icon: string
+	duration?: number
+	amount?: number
+}
+interface DowntimeInfo {
+	start: number
+	end: number
+}
+interface AdditionalDamageEvent extends DamageEvent {
```
ayyaruq: I have to do this in a MNK module and it's really dumb. There's actually a normaliser for this but the typing is busted so it essentially doesn't apply properly. See `src/parser/core/modules/HitType.js`.

src/parser/jobs/brd/modules/Util.tsx
```diff
@@ -0,0 +1,155 @@
...
+	static handle = 'util'
+
+	@dependency private combatants!: Combatants
+	@dependency private downtime!: Downtime
+	@dependency private enemies!: Enemies
+	@dependency private invuln!: Invuln
+	@dependency private timeline!: Timeline
+	@dependency private additionalStats!: AdditionalStats
+
+	hasBuff(status: StatusEffect) {
```
ayyaruq: If there are returning stuff, probably a good idea to type the returns.
```suggestion
	hasBuff(status: StatusEffect): boolean {
```

src/parser/jobs/brd/modules/Util.tsx
```diff
@@ -0,0 +1,155 @@
...
+const DHIT_MOD = 1.25
+
+const TRAIT_STRENGTH = 0.20
+
+interface StatusEffect {
+	id: number
+	name: string
+	icon: string
+	duration?: number
+	amount?: number
```
ayyaruq: I can't see `amount` being used anywhere from StatusEffect. Core has a status type (from `src/data/STATUSES/STATUSES.d.ts`) which can be imported with `import {Status} from 'src/data/STATUSES'`. It's otherwise an identical interface.

src/parser/jobs/brd/modules/Util.tsx
ackwell: This shouldn't be required any more (fixed it yesterday!). If you merge latest upstream, `DamageEvent` should include both these fields already.

PR #560, opened 2019-08-15
Remove Scholar Ruin II Weaving Suggestion

panicstevenson:
Remove the Scholar suggestion regarding Ruin II weaving as it is no
longer accurate with the changes made to the job in Shadowbringers.

The module is detrimental to players trying to analyze their fights
currently because it overwrites the existing Weaving module that
otherwise explains poor weaving, and gives misinformation that it is
suboptimal to use Ruin II in a fight unless moving.

Signed-off-by: Panic Stevenson <Panic.Stevenson@gmail.com>


nonowazu:
I'm not sure removing the module outright would be a good change. Weaving using Ruin II and Biolysis is very good for keeping up DPS – removing the module outright would be bad. It would be better to revert it to a default weaving suggestion.

panicstevenson:
Removing the module allows Scholar to inherit the default weaving suggestion instead of using its custom one:

image

The issue with Biolysis is resolved with https://github.com/xivanalysis/xivanalysis/pull/559.

nonowazu:
I'm sorry, I did go back to edit my message a bit – I'm very sorry if it was missed. I'm currently discussing with a few people from the balance to figure this out, but there may still need to be custom logic for `isBadWeave` if biolysis is indeed introducing a delay.

If you'd like to hop on the discord channel, I'd love to discuss this with you so we don't duplicate efforts.

panicstevenson:
Discussed the changes with @nonowazu offline.

Closing in favor of https://github.com/xivanalysis/xivanalysis/pull/562.

PR #561, opened 2019-08-16
5.05 SAM support

Rjb99:
sorry for the delay, Vacation plus eating glue on my end.

What is PR does is:
-remove sen overwrite because hagakure exists now to fix sen alignment
-add hagakure back.
-changed the oGCD tracking for Meikyo/Tsubame to the core's oGCD tracker instead of my caveman one
-fixed some incorrect data in the ACTIONS file.


src/data/ACTIONS/SAM.js
```diff
@@ -287,7 +277,37 @@ export default {
...
+		cooldown: 60,
+	},
+
+	KAESHI_HIGANBANA: {
+		id: 16484,
+		name: 'Kaeshi: Higanbana',
+		icon: 'https://xivapi.com/i/003000/003181.png',
+		onGcd: true,
+		cooldown: 60,
+		cooldownGroup: SAM_COOLDOWN_GROUP.TSUBAME,
```
AzariahAeroborn: Does this need gcdRecast: 2.5 to avoid funky GCD uptime calculations? (Same with the other two below)

src/parser/jobs/sam/modules/OGCDDowntime.js
```diff
@@ -0,0 +1,14 @@
+import ACTIONS from 'data/ACTIONS'
+import CooldownDowntime from 'parser/core/modules/CooldownDowntime'
+
+export default class OGCDDowntime extends CooldownDowntime {
+	//Time that samurais have deemed ok for a OGCD to be down
+	// eslint-disable-next-line no-magic-numbers
+	allowedDowntime = 5000
+	trackedCds = [
```
AzariahAeroborn: You need to add a firstUseOffset as well - the CooldownDowntime module doesn't consider allowedDowntime as delay from pull to first use, so pulls that are just longer than an even multiple of a cooldown will have funky expected uses if you don't specify a firstUseOffset.

src/data/ACTIONS/SAM.js
Rjb99: not sure why that was reverted. but I've unreverted it

src/parser/jobs/sam/modules/OGCDDowntime.js
```diff
@@ -1,6 +1,9 @@
 import ACTIONS from 'data/ACTIONS'
 import CooldownDowntime from 'parser/core/modules/CooldownDowntime'
-
+//meikyo is used after the 7th gcd
+const FIRSTUSEOFFSETMEIKYO = 1750
```
AzariahAeroborn: Should this be 17500?

src/parser/jobs/sam/index.js
```diff
@@ -25,14 +25,25 @@ export default new Meta({
 
 	supportedPatches: {
 		from: '5.0',
```
ayyaruq: With the changes to Meikyou and Tsubame, logic is a bit diff now. The addition of Hagakure doesn't change anything on old logs but the other 2 skills do. I think it's probably a good idea to change the `from` version to `5.05` as well now, just to avoid bad suggestions on cooldown timing in 5.0 logs.

Rjb99: Adjusted to not support 5.0

src/parser/jobs/sam/modules/Sen.tsx
```diff
@@ -65,28 +74,27 @@ export default class Sen extends Module {
...
-		this.allowoverwrite = true
-
-	}
-
 	private remove() {
 		this.sen = _.mapValues(this.sen, () => false)
 	}
 
+	private onHagakure() {
+	// work out how many sen are currently active
```
ackwell: intendation died a death

PR #562, opened 2019-08-16
Fix weave suggestions

nonowazu:
PR#559 will have to be merged first, or people will freak out at how many correct weaves are being flagged.

There might be an opportunity to check for additional bad weaves later, but first priority is making sure that the suggestion is correct.


src/parser/jobs/sch/modules/Weaving.js
```diff
@@ -17,75 +16,32 @@ const WEAVING_SEVERITY = {
...
-			why:  <Trans id="sch.weaving.suggestion.why">
-				<Plural value={this._badWeaves.length} one="# instance" other="# instances"/> instances of weaving with Ruin II while not moving.
+			content: <Trans id="sch.weaving.content">
+				Try to use <ActionLink {...ACTIONS.SCH_RUIN_II}/> and <ActionLink {...ACTIONS.BIOLYSIS}/> to weave your actions, and avoid weaving more actions than you have time for in a single GCD window.
+				Doing so will delay your next GCD, reducing possible uptime. Check the <a href="javascript:void(0);" onClick={() => this.parser.scrollTo(this.constructor.handle)}><NormalisedMessage message={this.constructor.title}/></a> module below for more detailed analysis.
 			</Trans>,
+			why: <Plural
+				id="core.weaving.why"
+				value={badWeaves.length}
+				_1="# instance of incorrect weaving"
```
suShirow: Nitpick (in the style of ack) but he mentioned to prefer `one` over `_1` since it makes more sense in the context of `other`.

src/parser/jobs/sch/modules/Weaving.js
```diff
@@ -17,75 +16,32 @@ const WEAVING_SEVERITY = {
...
+			this._saveIfBad()
 		}
 
-		this._pos = this.combatants.selected.resources
-	}
-
-	// Now this, we want to overwrite
-	_onComplete() {
+		// Few triples is medium, any more is major
+		const badWeaves = this._badWeaves
```
ayyaruq: Any reason for copying the object like this? If you were filtering or something I get it but this is just a raw copy. Probably super minor in terms of performance overhead but yeah. Just a nit. You use the value only twice, and for the same thing, so maybe just assign like in this suggestion:

```suggestion
		const badWeaveCount = this._badWeaves.length
```

nonowazu: No reason, actually. I have no problems changing it. One sec, and I'll update the PR.

ayyaruq:
@nonowazu did you forget to commit changes to the non-English languages with that last `yarn extract`?

nonowazu:
@ayyaruq That's pretty interesting. `yarn extract` only yielded the English locales...
I usually `git stage locale/` for that, so that threw me for a loop as to why I'd miss that.

ayyaruq:
Weird, I just `git add locale/`, lmk when it's good (assuming it correctly changed those locale files, maybe it didn't need to? I'll merge when ready since Sushi approvesd already.

nonowazu:
I did double-check on my end by re-running `yarn extract` and then `git stage`, so it looks like there's nothing to add on that end.

PR #563, opened 2019-08-16
Update Warrior Actions to Match 5.05 Job Guide

panicstevenson:
Update Warrior actions to match values listed in the 5.05 job guide.

 - Remove Unchained.
 - Add Shake It Off and Holmgang to Warrior statuses.
 - Add and update Warrior status durations.
 - Add and update Warrior action potencies.
 - Update Marauder and Warrior action cooldowns.

Signed-off-by: Panic Stevenson <panic.stevenson@gmail.com>

### DIFF
--- a/src/data/STATUSES/WAR.js
+++ b/src/data/STATUSES/WAR.js
@@ -34,14 +34,14 @@ export default {
 	},
 
 	SHAKE_IT_OFF: {
-		id: 7388,
+		id: 1457,
 		name: 'Shake It Off',
 		icon: 'https://xivapi.com/i/002000/002563.png',
 		duration: 15,
 	},
 
 	HOLMGANG: {
-		id: 43,
+		id: 409,
 		name: 'Holmgang',
 		icon: 'https://xivapi.com/i/000000/000266.png',
 		duration: 6,
--- a/src/data/STATUSES/WAR.js
+++ b/src/data/STATUSES/WAR.js
@@ -10,24 +10,21 @@ export default {
 		id: 87,
 		name: 'Thrill Of Battle',
 		icon: 'https://xivapi.com/i/010000/010254.png',
+		duration: 10,
 	},
 
 	RAW_INTUITION: {
 		id: 735,
 		name: 'Raw Intuition',
 		icon: 'https://xivapi.com/i/012000/012555.png',
+		duration: 5,
 	},
 
 	VENGEANCE: {
 		id: 89,
 		name: 'Vengeance',
 		icon: 'https://xivapi.com/i/010000/010256.png',
-	},
-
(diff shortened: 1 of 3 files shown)


src/data/STATUSES/WAR.js
```diff
@@ -36,21 +33,38 @@ export default {
 		icon: 'https://xivapi.com/i/012000/012551.png',
 	},
 
+	SHAKE_IT_OFF: {
```
ayyaruq: Shake is technically a party buff, but then again a bunch of non-offensive similar skills (eg Mantra) are in job-specific statuses so derp.

src/data/STATUSES/WAR.js
```diff
@@ -36,21 +33,38 @@ export default {
...
 	},
 
 	NASCENT_CHAOS: {
 		id: 1897,
 		name: 'Nascent Chaos',
 		icon: 'https://xivapi.com/i/012000/012560.png',
+		duration: 30,
 	},
 
 	NASCENT_FLASH: {
```
ayyaruq: If adding Flash and including party defensive buffs like Shake, might be worth having Glint as well and putting them all in the Shared statuses or something.

panicstevenson: Does that make more sense to bundle with this change or to have as a separate change, possibly including things like Mantra?

ayyaruq: Probably something to talk about in Discord with the other devs and @ackwell for his thoughts on defensive stuff in the first place. I don't think any tank jobs have modules for this so whether it belongs in shared, or somewhere else, is best discussed with at least the other maintainers.

Toastdeib: My two cents - this is fine where it is. Every other job I've seen that has buff statuses (even full party buffs, like Litany, Brotherhood, etc.) put them in their respective job's STATUSES file. The SHARED file is more for non-job-specific things, like role actions, Weakness/Brink of Death, etc.

Toastdeib: @ayyaruq Unless you have particularly strong feelings, I'm going to go ahead and merge this. The data all looks fine (double-checked it against garland), and I think if we want to move party buffs we should do that in a separate restructuring PR, since this follows current convention.

ayyaruq: @toastdeib I'm fine with it, the above was more a question of where stuff goes and handling Nascent Glint (the buff given to the target of Nascent Flash). The number changes are all good.

Toastdeib: I'd say Nascent Glint should also go in here if and when we add it, but we can burn that bridge when we get to it. :blobmorning:

PR #565, opened 2019-08-16
Fix warnings for biolysis

nonowazu:
Severity was in ms, and the maxClip value was in seconds. This at least causes the warnings to show up.


suShirow:
This happened to me in the ast dot module too haha

src/parser/jobs/sch/modules/DoTs.js
```diff
@@ -45,17 +52,16 @@ export default class DoTs extends CoreDoTs {
...
 			content: <Trans id="sch.dots.suggestions.clipping.content">
 				Avoid refreshing Biolysis significantly before its expiration, except when at the end of the fight. Unnecessary refreshes use up your mana more than necessary, and may cause you to go out of mana.
 			</Trans>,
-			tiers: CLIPPING_SEVERITY,
-			value: maxClip,
+			tiers: SEVERITIES.CLIPPING,
+			value: clipPerMinute,
 			why: <Trans id="sch.dots.suggestions.clipping.why">
-				{this.parser.formatDuration(clip[STATUSES.BIOLYSIS.id])} of <StatusLink {...STATUSES.BIOLYSIS}/> lost to early refreshes.
+				An average of {this.parser.formatDuration(this.getClippingAmount(STATUSES.BIOLYSIS.id) * 1000)} of <StatusLink {...STATUSES.BIOLYSIS}/> clipped every minute, for a total of {this.parser.formatDuration(clip[STATUSES.BIOLYSIS.id])} lost to early refreshes.
```
Toastdeib: No need to call `this.getClippingAmount()` again here; you should just reuse the `clipPerMinute` you defined earlier and used for the `value` field.

src/parser/jobs/sch/modules/DoTs.js
```diff
@@ -5,19 +5,25 @@ import {ActionLink, StatusLink} from 'components/ui/DbLink'
...
+// In seconds
+const SEVERITIES = {
+	CLIPPING: {
+		6: SEVERITY.MINOR,
+		9: SEVERITY.MEDIUM,
+		12: SEVERITY.MAJOR,
+	},
+	UPTIME: {
+		84: TARGET.WARN,
+		91: TARGET.SUCCESS,
```
Toastdeib: I'm not a SCH by any stretch of the imagination, this feels a bit low to me for a DoT uptime target, especially one that isn't a comboed DoT. Am I thinking too much like a DPS main here?

src/parser/jobs/sch/modules/DoTs.js
```diff
@@ -5,19 +5,25 @@ import {ActionLink, StatusLink} from 'components/ui/DbLink'
...
+	},
 }
 
 export default class DoTs extends CoreDoTs {
-	static handle = 'dots'
+	static handle = 'Biolysis'
```
Toastdeib: Very nitpicky, but our handles are usually `lowerCamelCase` by convention, so this should be changed to `biolysis`.

src/parser/jobs/sch/modules/DoTs.js
nonowazu: Well, for example, here's the uptime for some 99th and 100th percentiles in E1S/E2S:

* E1S
  * https://xivanalysis.com/analyse/1aAJLzymwHnpjBYF/2/3/
  * https://xivanalysis.com/analyse/hx1dtYCm4TMqA87G/7/8/
  * https://xivanalysis.com/analyse/NBR4GmL2PwxFch9Y/9/3/
* E2S
  * https://xivanalysis.com/analyse/ACq4jtrxGpJW92M3/11/3/
  * https://xivanalysis.com/analyse/G4XPKWgNnCZdDrh1/5/6/
  * https://xivanalysis.com/analyse/xpD6T3yFvrBnHLcA/14/6/

These seem to range from 80-odd to 96%. I mean, part of it could be the dps main mentality, but 95% already pretty hard to reach for even the highest damage SCHs on fflogs. Maybe change success to 95%, and warn from 85% and up?

src/parser/jobs/sch/modules/DoTs.js
nonowazu: Alright, got the requested fix in – thanks for the catch!

src/parser/jobs/sch/modules/DoTs.js
nonowazu: After some soul searching, I've decided to up the number. Given some uptime metrics I'm seeing from others, might be worth amending the text to talk about uptime detection with healers, but I see no reason that should block on the current percentages.

src/parser/jobs/sch/modules/DoTs.js
nonowazu: Not a nitpick if it's a convention – fix'd :)

PR #566, opened 2019-08-17
Swiftcast tracking

nonowazu:
This needs a lot of work, and frankly I think a lot of input and work would be required before this should be accepted. If this isn't interesting for core I could retool for only SCH use, but I think this sort of tracking could be useful for casters/healers in general.

In particular, my concerns are:
1. If it's core, it should be written in typescript
2. This probably needs a method to override that would allow devs and maintainers to determine if it's a 'good' or 'bad' swiftcast usage
3. Since this draws off `BuffWindowModule`, the output (which I've mostly copied just for the sake of having something to show for it) is aimed at rotation, not an individual cast during a window.

### DIFF
--- a/src/parser/core/modules/Swiftcast.js
+++ b/src/parser/core/modules/Swiftcast.js
@@ -1,0 +1,88 @@
+import React from 'react'
+import {t} from '@lingui/macro'
+import {Trans} from '@lingui/react'
+import {RotationTable} from 'components/ui/RotationTable'
+import {ActionLink} from 'components/ui/DbLink'
+
+import {BuffWindowModule} from 'parser/core/modules/BuffWindow'
+import ACTIONS from 'data/ACTIONS'
+import STATUSES from 'data/STATUSES'
+import {SEVERITY} from 'parser/core/modules/Suggestions'
+import {getDataBy} from 'data/getDataBy'
+
+export default class Swiftcast extends BuffWindowModule {
+	static handle = 'swiftcast'
+	static title = t('core.swiftcast.title')`Swiftcasts`
+
+	buffAction = ACTIONS.SWIFTCAST
+	buffStatus = STATUSES.SWIFTCAST
+
+	expectedGCDs = {
+		expectedPerWindow: 1,
+		suggestionContent: <Trans id="core.swiftcast.suggestions.missedgcd.content">
+			Try to use at least one move during every <ActionLink {...ACTIONS.SWIFTCAST} />
+		</Trans>,
+		severityTiers: {
+			1: SEVERITY.MAJOR,
+		},
+	}
+
+	// override to only use gcd casts
+	onCast(event) {
+		const action = getDataBy(ACTIONS, 'id', event.ability.guid)
+
+		if (!action || action.autoAttack || !action.castTime) {
+			return
+		}
+
+		if (this.activeBuffWindow) {
+			this.activeBuffWindow.rotation.push(event)
+		}
+	}
+
+	// sooper hacky...
+	output() {
+		const rotationData = this.buffWindows
+			.map(buffWindow => {
+				const windowStart = buffWindow.start - this.parser.fight.start_time
+				const windowEnd = (buffWindow.end != null ? buffWindow.end : buffWindow.start) - this.parser.fight.start_time
+				const targetsData = {}
+
+				if (this.expectedGCDs) {
+					targetsData.missedgcd = {
+						actual: buffWindow.gcds,
+						expected: this.getBuffWindowExpectedGCDs(buffWindow),
+					}
+				}
+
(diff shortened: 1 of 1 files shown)


src/parser/core/modules/Swiftcast.js
```diff
@@ -0,0 +1,88 @@
...
+			return
+		}
+
+		if (this.activeBuffWindow) {
+			this.activeBuffWindow.rotation.push(event)
+		}
+	}
+
+	// sooper hacky...
+	output() {
```
AzariahAeroborn: It seems weird to me to override this just to chop off the headers of the rotation table.  Especially given that you're still potentially allowing for requiredGCDs or trackedActions... what identifies those columns to the viewer without the header row?  Is this something that could just work with the base class output() function?

nonowazu: This actually does keep the headers -- it just removes the required GCDs column. I'm definitely thinking that overriding `output()` is a good solution because the original function is geared at showing how many moves happened during a buff window instead of the singular move that was used with swiftcast. I like the idea that there's a "go to point in timeline" button, and the move that was used with swiftcast - the 'required number of moves' seems hard to use in this context.

AzariahAeroborn: The default behavior of the BuffWindow module is to only show any given column if the associated class property is overridden, so that it is not null.  I would expect that just using the default output() behavior given the rest of this class, you should see a table that has: timestamp with goTo link, column headed "GCDs" with the actual/expected GCDs used, and the rotation used.

nonowazu: Ah, yes, that is definitely the default output. But as a side effect of overriding `onCast` to ignore actions without cast time makes the "GCDs" column bug out a bit and report actual as 1, and expected as 0.

I fully admit that BuffWindow isn't really made for this purpose, but maybe another table format might be more useful anyway?

AzariahAeroborn: Ah, the 1/0 is being caused by reduceExpectedGCDsEndOfFight().  Since swiftcast pops early when used, you can just override that to always return 0 and that will fix that issue.

nonowazu: Oh, cool. I'll un-override output and put in a patch for that, then. Still, what about a better way to change the column 'Rotation' to 'Ability Used', or something like that?

nonowazu: I pushed up those changes, and things definitely look better for the one edge case I was experiencing.

panicstevenson:
Was there any interest in the module showing what you weaved subsequently after your cast? Swiftcast is great if you're moving, weaving, or casting a spell that takes longer than normal.

I couldn't say that I understand the level of difficulty here, so I wouldn't say it's necessary for this feature to ship, but it would definitely help in showing the whole picture in my opinion.

src/parser/core/modules/Swiftcast.js
```diff
@@ -0,0 +1,45 @@
...
+
+import {BuffWindowModule} from 'parser/core/modules/BuffWindow'
+import ACTIONS from 'data/ACTIONS'
+import STATUSES from 'data/STATUSES'
+import {SEVERITY} from 'parser/core/modules/Suggestions'
+import {getDataBy} from 'data/getDataBy'
+
+export default class Swiftcast extends BuffWindowModule {
+	static handle = 'swiftcast'
+	static title = t('core.swiftcast.title')`Swiftcasts`
```
panicstevenson: Might be more appropriate to call this `Swiftcast` or `Swiftcast Usage` to better match the other modules featured for other jobs.

nonowazu: That's definitely more accurate. I'll get this fix up as part of my next commit.

nonowazu:
> Was there any interest in the module showing what you weaved subsequently after your cast? Swiftcast is great if you're moving, weaving, or casting a spell that takes longer than normal.
> 
> I couldn't say that I understand the level of difficulty here, so I wouldn't say it's necessary for this feature to ship, but it would definitely help in showing the whole picture in my opinion.

I'm not sure I heard interest in that _per se_; the specific request ([here](https://discordapp.com/channels/441414116914233364/470049935077933076/610515990853451787) if anyone is interested) was:

> something I’ve wanted for all healers is just a track that quickly shows you what you used swiftcast on

However, maybe we could give at least some factoids like "were they moving?". For weaves, the timeline button could take them there and they should be able to see what they weaved with it. Maybe move some of that tracking logic to the weaving module (detect swiftcast used for a weave) and show there?

nonowazu:
Given some conversations with Azariah and some review, I think it might be wiser to close this PR as the code for swiftcast tracking is too trivial to be a core module at this time.

PR #567, opened 2019-08-17
WIP: Fix 6SS definition to use gcdRecast instead of cooldown

ayyaruq:
This is slightly different to how I thought it was, blame Square. I'll redo it correctly and push later.


ayyaruq:
This is actually a GCD, not a weaved cooldown. The cooldown and gcdRecast are the same so change isn't required. May need some adjustments or an override of core weaving since it seems to run off the speedmod GCD rather than current GCD.

PR #568, opened 2019-08-17
Crit the Boot.

ayyaruq:
100% blame @AkairyuGestalter for the module name. I'm disappointed I couldn't fit "Crit the Boot" or "steppies" into the text anywhere so suggestions welcome.


src/parser/jobs/mnk/modules/Steppies.tsx
```diff
@@ -0,0 +1,122 @@
...
+	private steppies: Boot[] = []
+
+	protected init(): void {
+		this.addHook('cast', {by: 'player', abilityId: ACTIONS.BOOTSHINE.id}, this.onCast)
+		this.addHook('damage', {by: 'player', abilityId: ACTIONS.BOOTSHINE.id}, this.onStep)
+		this.addHook('complete', this.onComplete)
+	}
+
+	private onCast(event: CastEvent): void {
+		this.currentBoot = new Boot(this.combatants.selected.hasStatus(STATUSES.LEADEN_FIST.id), event.timestamp)
```
Toastdeib: Any particular reason to have this in its own separate hook instead of just creating the `Boot` object in the `damage` hook?

src/parser/jobs/mnk/modules/Steppies.tsx
```diff
@@ -0,0 +1,122 @@
...
+import Module, {dependency} from 'parser/core/Module'
+import Checklist, {Requirement, Rule} from 'parser/core/modules/Checklist'
+import Combatants from 'parser/core/modules/Combatants'
+import Suggestions, {SEVERITY, TieredSuggestion} from 'parser/core/modules/Suggestions'
+
+import DISPLAY_ORDER from './DISPLAY_ORDER'
+
+const LEAD_BOOT_POTENCY = 300
+
+// Typically a player with lag can derp positional in opener, this usually happens 2 times per fight.
```
Toastdeib: I don't know shit about optimal MNK play so feel free to ignore this one, but if 1-2 misses can happen due to lag, these severities strike me as a bit, well, severe.

src/parser/jobs/mnk/modules/Steppies.tsx
ayyaruq: We need to know buff state when the cast happens, but then hit state on damage.

src/parser/jobs/mnk/modules/Steppies.tsx
ayyaruq: From discord but, tl;dr gonna bump majors a bit since it's a very easy issue for users to fix and a _big_ DPS loss, but leaving mediums alone since even with lag you can just pre-pull True North or Riddle of Earth. 5.0 fights have enough raid damage to make RoE easy to use, plus TN has charges now.

src/parser/jobs/mnk/modules/Steppies.tsx
ayyaruq: As discussed in Discord.

PR #569, opened 2019-08-17
Add a getter for the CounterGauge current value

ayyaruq:
Greased Lightning gauge needs to know the current value. Might be worth having a getter on the max too but I don't think any other jobs need that, and I can work around it with the constants I use to set it in the first place.


src/parser/core/modules/Gauge/CounterGauge.ts
```diff
@@ -37,11 +37,15 @@ export class CounterGauge extends AbstractGauge {
 
 	private history: CounterHistory[] = []
 
+	public get value(): number {
```
ackwell:
```suggestion
	get value(): number {
```

(ts stuff is public by default)

PR #572, opened 2019-08-17
Add Main Attribute to Job Data Structure

panicstevenson:
Add attribute enumeration and add each job's main damage modifier to
their respective job data object entry.

Tanks use STR as their main damage modifier as per the [Tank Mastery](https://user-images.githubusercontent.com/8849608/63216433-eb5a7e00-c0e9-11e9-8578-280b944cda32.png) trait added in Shadowbringers.

Useful change for adding specific potion icons to the Potions module.

Interested in better ways to implement this change.

### DIFF
--- a/src/data/JOBS.ts
+++ b/src/data/JOBS.ts
@@ -16,6 +16,14 @@ const colors = {
 	misc: '#767676',
 }
 
+export enum Attribute {
+	STRENGTH,
+	DEXTERITY,
+	VITALITY,
+	INTELLIGENCE,
+	MIND,
+}
+
 const roleData = {
 	TANK: {
 		id: 1,
@@ -69,6 +77,7 @@ export interface Job {
 	icon: string
 	colour: string
 	role: Role['id']
+	attribute: Attribute
 }
 
 // Yeah I know there's lots of repetition but they're all different apis and endpoints and shit and I don't wanna pull it apart later to fix a desync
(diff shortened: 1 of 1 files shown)


src/data/JOBS.ts
```diff
@@ -16,6 +16,14 @@ const colors = {
 	misc: '#767676',
 }
 
+export const STATS = {
```
ackwell: This file is TS, so we can lean on the magic of `enum`s for this.

```ts
export enum Stat { // note singular
	STR,
	DEX,
	// ...
}
```

That way the job interface can include `mainStat: Stat`, which is way more restrictive.

I'm 50/50 on `STR, DEX, ...` and `STRENGTH, DEXTERITY, ...` - maybe erring towards the latter.

panicstevenson: Thanks for the input! I don't really use either JavaScript or TypeScript, so I didn't know if they'd play nicely exporting between the two, but it looks like it just passes over the integer value and works well.

I've updated the object literal to match the enum you have above, with the stat long names in place of their three letter abbreviations.

Any opinion on `player.mainStat` versus `player.stat` in the object passed around?

panicstevenson: Actually, I think they're called Attributes in game; but I'm away from a computer right now to check. Perhaps that's better.

panicstevenson:
Closing after some discussion on Discord about the future state of the Potions module.

PR #575, opened 2019-08-19
Core timer gauge

ackwell:
A WIP implementation of a timer gauge for the core gauge module. Due to its requirements, there's a fair chunk of core parser/module changes as well.

**Do not merge.** This is in an incomplete state, and contains issues (outlined below) that will need to be resolved before it's good to go.

Currently, this PR:
* Renames `addHook` to `addEventHook`, and `removeHook` to `removeEventHook` (old names are aliased for compatibility).
* Adds `addTimestampHook` and `removeTimestampHook`.

It will also need:
* The ability to hook the expiration of the gauge.

Current issue(s):
* This implementation handles timestamps at the `Module` level, which means that during the execution of timestamp hooks, `parser.currentTimestamp` is _incorrect_.
* Timestamp hooks require access to `Module`'s `addTimestampHook`, which gauge classes don't currently have access to.

As a timeline to explain the first point, assume we have a timestamp hook set on the timestamp `50`:

```
Event: {type: 'foo', timestamp: 0}
	- parser updates current timestamp to 0
	- module checks timestamp hooks, finds nothing

Event: {type: 'foo', timestamp: 100}
	- parser updates current timestamp to 100
	- module checks timestamp hooks, finds the one for 50
		- execute hook for 50: parser's currentTimestamp will be 100 here, which is wrong.
```

The obvious "fix" for this is to do the long-planned move of hook logic out of module, to a parser-level system - however that comes with further issues, as _currently_, the dependency error cascade relies on hooks being handled at the module level to disable erroring modules.


ackwell:
itt stale closes a WIP
itt stale can get fucked

ackwell:
I think the major issues outlined above have been addressed now, both with patches that should be removed once event handling is moved above the module level.

* The inaccuracy with `parser.currentTimestamp` has been resolved by temporarily overriding the parser's internal timestamp while handling timestamp hooks, then restoring it. Please don't do this anywhere else lmao.
* Timestamp hooking functions are being manually passed down to the timestamp gauge object - not pretty, but it follows the pattern from the abstract gauge for now, so it can be resolved at the same time hopefully.

ackwell:
I think this is at a point it's good to go, now. Alongside fixing some issues that came to light while I was running a test implementation, I've added

- The ability to pause the timer
- Graph generation (that ended up being like 1/3 of the timer code)

There's a few edge cases around graph generation for pause/resume that are imperfect, but they have no effect on the data itself - and should be obvious if they crop up.

PR #579, opened 2019-08-20
Paladin/Gunbreaker buff downtime tracking

qaphla:
This PR adds the PLD self-buffs Fight or Flight and Requiescat, the GNB self-buff No Mercy, and the GNB resource-generator Bloodfest to the list of oGCDs that xivanalysis tracks the downtime for.

### DIFF
--- a/src/parser/jobs/gnb/modules/OGCDDowntime.tsx
+++ b/src/parser/jobs/gnb/modules/OGCDDowntime.tsx
@@ -1,9 +1,21 @@
 import ACTIONS from 'data/ACTIONS'
 import CooldownDowntime from 'parser/core/modules/CooldownDowntime'
 
+/* This offset allows for four weaponskills to be used before No Mercy in the
+ * opener, as is currently recommended. Should be updated if openers change to
+ * have later No Mercy.
+ */
+const FIRST_USE_OFFSET_NO_MERCY = 10000
+
 export default class AbilityDowntime extends CooldownDowntime {
 	allowedDowntime = 2500
+
+	firstUseOffsetPerOgcd = {
+		[ACTIONS.NO_MERCY.id]: FIRST_USE_OFFSET_NO_MERCY,
+	}
+
 	trackedCds = [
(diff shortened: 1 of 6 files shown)


src/parser/jobs/pld/modules/OGCDDowntime.tsx
```diff
@@ -1,9 +1,28 @@
 import ACTIONS from 'data/ACTIONS'
 import CooldownDowntime from 'parser/core/modules/CooldownDowntime'
 
+const ALLOWEDDOWNTIMEFOF = 4000
+const ALLOWEDDOWNTIMEREQ = 4000
+
+const FIRSTUSEOFFSETFOF = 24000
```
rachellhea: Probably worth sticking a comment here describing why we allow first use offset to be so big.

src/parser/jobs/gnb/modules/OGCDDowntime.tsx
```diff
@@ -1,9 +1,20 @@
 import ACTIONS from 'data/ACTIONS'
 import CooldownDowntime from 'parser/core/modules/CooldownDowntime'
 
+const FIRSTUSEOFFSETNOMERCY = 10000
```
Toastdeib: These should be split with underscores for readability (e.g. `FIRST_USE_OFFSET_NO_MERCY` and `FIRST_USE_OFFSET_BLOODFEST`. Same thing goes for the consts in the PLD module as well.

src/parser/jobs/pld/modules/OGCDDowntime.tsx
qaphla: Should be resolved in current version --- I also added brief explanations for the GNB numbers.

src/parser/jobs/gnb/modules/OGCDDowntime.tsx
qaphla: Fixed in current version.

PR #582, opened 2019-08-23
Basic RoF support: oGCD usage tracking, active Fist checker, RotationTable update.

rachellhea:
Sample image of suggestions:

Screen Shot 2019-08-23 at 1 15 36 PM

And the output table:

Screen Shot 2019-08-23 at 1 16 07 PM


src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -86,90 +121,140 @@ export default class RiddleOfFire extends Module {
...
+					header: <ActionLink showName={false} {...ACTIONS.SHOULDER_TACKLE}/>,
+					accessor: 'shoulderTackle',
 				},
+			]}
+			data={this.history
+				.map(riddle => ({
+					start: riddle.start - this.parser.fight.start_time,
+					end: riddle.end != null ?
+						riddle.end - this.parser.fight.start_time
+						: riddle.start - this.parser.fight.start_time,
```
ayyaruq: If it's null we use the start time? I might be misunderstanding, did you mean the fight end? I used to have a thing to close Rushing RoFs which should take care of this anyway, probably just needs some sanity checking on end being undefined.

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -32,11 +35,36 @@ class Riddle {
...
+
+	get elixirField() {
+		return this.casts.filter(event => event.ability.guid === ACTIONS.ELIXIR_FIELD.id).length
+	}
+
+	get tackles() {
+		return this.casts.filter(event => event.ability.guid === ACTIONS.SHOULDER_TACKLE.id).length
+	}
+
+	public gcdsInFist(fistId: number) {
```
ayyaruq: To align with other modules, maybe rename to `gcdsByFist`.

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -1,29 +1,32 @@
 import {t} from '@lingui/macro'
 import {Plural, Trans} from '@lingui/react'
+import _ from 'lodash'
```
ayyaruq: I don't see lodash being used anywhere here?

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -86,90 +121,140 @@ export default class RiddleOfFire extends Module {
...
+			data={this.history
+				.map(riddle => ({
+					start: riddle.start - this.parser.fight.start_time,
+					end: riddle.end != null ?
+						riddle.end - this.parser.fight.start_time
+						: riddle.start - this.parser.fight.start_time,
+					targetsData: {
+						gcds: {
+							actual: riddle.gcds,
+							expected: EXPECTED_GCDS,
```
ayyaruq: Making expected GCDs a field in the Riddle the same way AoE uses targets might be a good idea. This is so we can show the correct expectation for opener and rushing Riddles and avoid false-flagging every opener. Depends on either using the new core gauge, or seeing if non-RoW was in use. Could probably do it in the stopAndSave function.

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -86,90 +121,140 @@ export default class RiddleOfFire extends Module {
 			this.stopAndSave()
 		}
 
-		// Aggregate GCDs under each RoF
-		const missedGcds = this.history
+		// The opener Riddle is special (10 GCDs with some FoF uptime), so keep track of it.
+		const openerRiddle = this.history[0]
```
ayyaruq: This is making an assumption the user is doing a proper opener.

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
rachellhea: Yeah, this is just a sanity check. We do the same thing on Requiescat. Basically, the end of the window is either:

- Specified specifically, in which case we should use it (but we have to subtract the start time to get the relative end).
- Not specified, in which case it should be the end of the fight.

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
rachellhea: Right, but our ultimate goal is to not punish people who are doing proper openers.

rachellhea: Now that I think about it, the core issue here is detecting RoF sequences that start at sub-GL4. Should be doable with the new gauge module stuff.

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -32,11 +33,37 @@ class Riddle {
 	end?: number
 	active: boolean = false
 	rushing: boolean = false
+	gcdsInEachFist: { [fistId: number]: number } = {
+		[FISTLESS]: 0,
+		[STATUSES.FISTS_OF_EARTH.id]: 0,
+		[STATUSES.FISTS_OF_FIRE.id]: 0,
+		[STATUSES.FISTS_OF_WIND.id]: 0,
+	}
+	expectedGcds: number = 11 // Baseline GL4 RoFs are 11 GCDs
```
ackwell: minor; should this be initialised to `EXPECTED_GCDS`?

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -32,11 +33,37 @@ class Riddle {
...
 	}
+
+	get gcds() {
+		return this.casts.filter(event => {
+			const action = getDataBy(ACTIONS, 'id', event.ability.guid) as TODO
+			return action && action.onGcd
+		}).length
+	}
+
+	get elixirField() {
```
ackwell: ubernit;
```suggestion
	get elixirFields() {
```

---

i really feel that this repetitive accessor pattern is getting a bit out of hand, but there's prior art so it's ok for now. gonna need to figure out a way to clean them up

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -70,9 +99,14 @@ export default class RiddleOfFire extends Module {
 			return
 		}
 
-		// we only care about actual skills
-		if (this.riddle && this.riddle.active && action.onGcd) {
+		// MNK mentors want the oGCDs :angryeyes:
```
ackwell: rip

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -86,90 +120,122 @@ export default class RiddleOfFire extends Module {
...
+		// want a minor suggestion to track Riddles that only had 1 Tackle
+		const riddlesWithOneTackle = nonRushedRiddles.filter(riddle => riddle.tackles === 1).length
+		const riddlesWithZeroTackles = nonRushedRiddles.filter(riddle => riddle.tackles === 0).length
+		const droppedExpectedOgcds = droppedElixirFields + riddlesWithZeroTackles
 
 		this.suggestions.add(new TieredSuggestion({
 			icon: ACTIONS.RIDDLE_OF_FIRE.icon,
 			content: <Trans id="mnk.rof.suggestions.gcd.content">
-				Aim to hit {POSSIBLE_GCDS} GCDs into each <StatusLink {...STATUSES.RIDDLE_OF_FIRE} />.
+				Aim to hit {EXPECTED_GCDS} GCDs during each <StatusLink {...STATUSES.RIDDLE_OF_FIRE} />.
```
ackwell: mediumish; I feel these could probably do with a bit of expanding with the content, they're a bit terse and repetitive.

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -86,90 +120,122 @@ export default class RiddleOfFire extends Module {
...
-	}
-
-	private getGcdCount(riddle: Riddle): number {
-		return riddle.casts.filter(cast => {
-			const action = getDataBy(ACTIONS, 'id', cast.ability.guid) as TODO
-			return action && action.onGcd
-		}).length
-	}
+			// Check for any GCDs spent outside of Fists of Wind, which means the expected GCDs will be 9
+			if (this.riddle.gcdsByFist(FISTLESS)
```
ackwell: nit;

```js
// pref
if (
	foo &&
	bar
) {
	// ...
}

// over
if (foo &&
	bar) {
	// ...
}

ackwell:
Bumping so stale bot doesn't get too eager.

@ayyaruq do you have any intention of taking up this?

ayyaruq:
Yes, once the gauge stuff is done I'm inheriting the branch from Lhea and taking over the PR.

ayyaruq:
Superceded by #723, commits cherry-picked

PR #583, opened 2019-08-23
Basic PB support: number of GCDs and window logging.

rachellhea:
Big fan of this new core module tbh

Screen Shot 2019-08-23 at 1 47 35 PM
Screen Shot 2019-08-23 at 1 47 48 PM


src/parser/jobs/mnk/modules/PerfectBalance.tsx
```diff
@@ -0,0 +1,45 @@
...
+	trackedBadActions = {
+		icon: ACTIONS.TWIN_SNAKES.icon,
+		actions: [
+			{
+				action: ACTIONS.TWIN_SNAKES,
+				expectedPerWindow: 0,
+			},
+		],
+		suggestionContent: <Trans id="mnk.pb.suggestions.trackedBadActions.content">
+			Using <ActionLink {...ACTIONS.TWIN_SNAKES} /> or inside of <ActionLink {...ACTIONS.PERFECT_BALANCE} /> should be avoided under most circumstances. Try to refresh it before entering the window.
```
AzariahAeroborn: It looks like you have a stray or after the Twin Snakes action link.
```suggestion
			Using <ActionLink {...ACTIONS.TWIN_SNAKES} /> inside of <ActionLink {...ACTIONS.PERFECT_BALANCE} /> should be avoided under most circumstances. Try to refresh it before entering the window.
```

src/parser/jobs/mnk/modules/PerfectBalance.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+export default class PerfectBalance extends BuffWindowModule {
+	static handle = 'pb'
+	static title = t('mnk.pb.title')`Perfect Balance Windows`
+
+	buffAction: Action = ACTIONS.PERFECT_BALANCE
+	buffStatus: Status = STATUSES.PERFECT_BALANCE
+
+	expectedGCDs = {
+		expectedPerWindow: 6,
```
ayyaruq: This kinda depends on what Fist the player is in though, and if they're in stack recovery. I think it's ok for now but once the MNK gauge is done will wanna rethink this I suppose. AFAIK BuffWindowModule has some extra methods for customising this logic so should be ok. More of a note than a thing that needs to change though.

src/parser/jobs/mnk/modules/PerfectBalance.tsx
```diff
@@ -0,0 +1,45 @@
...
+		suggestionContent: <Trans id="mnk.pb.suggestions.missedgcd.content">
+			Try to land 6 GCDs during every <ActionLink {...ACTIONS.PERFECT_BALANCE} /> window.  If you cannot do this with full uptime and no clipping, consider adjusting your gearset for more Skill Speed.
+		</Trans>,
+		severityTiers: {
+			1: SEVERITY.MEDIUM,
```
ayyaruq: Many users (especially in Japan) use a different opener that doesn't rely on PB DK+Boot spam. Users with heavy latency will also struggle on this. I think 1 as minor and 2 as medium works, 3 is def major because wtf.

src/parser/jobs/mnk/modules/PerfectBalance.tsx
```diff
@@ -0,0 +1,45 @@
...
+			3: SEVERITY.MAJOR,
+		},
+	}
+
+	trackedBadActions = {
+		icon: ACTIONS.TWIN_SNAKES.icon,
+		actions: [
```
ayyaruq: Should probably add Form Shift and Six-Sided Star to this. Possibly also Four-Point Fury. We have separate suggestions atm as part of the Greased Lightning module but I'm planning to gut that and only use it for Speedmod. Not sure how the icon would best be handled on this though.

rachellhea: 1. Six-Sided Star should be caught by missing the expected number of GCDs, unless it's the last one which can probably just be hand-waved as "we assume you know what you're doing and if you don't an actual human is gonna nitpick you."
2. Agree with Form Shift.
3. I'll follow up with Tiff and Aya re:recovery scenarios.

src/parser/jobs/mnk/modules/PerfectBalance.tsx
rachellhea: Agreeable.

src/parser/jobs/mnk/modules/PerfectBalance.tsx
rachellhea: It does, but my discussion with Tiff implied to me that maximum Wind uptime is desirable.

What we could do is a set a baseline of 5, since that is always possible even in recovery, and then let 6th be a bonus that has to be checked by a human?

src/parser/jobs/mnk/modules/PerfectBalance.tsx
rachellhea: Tiff agrees that Twin Snakes shouldn't be penalized.

src/parser/jobs/mnk/modules/PerfectBalance.tsx
ayyaruq: I think that's a good idea. There's no reason to get less than 5, it's just deciding if 5 is acceptable or not depending on stance/stacks. We already have checks for maximum wind but dunno if we should use PB expected GCDs to reinforce it.

BuffWindowModule exposes `changeExpectedGCDsClassLogic(buffWindow: BuffWindowState): number` which can check the current Fists or GL and then return the amount to increment/decrement the expectedPerWindow value. Not sure if better to use stance or stacks as that metric though.

src/parser/jobs/mnk/modules/PerfectBalance.tsx
ayyaruq: The latency was more of a "use PB in the last 3rd of your GCD if you're aiming for 6" kinda thing. I guess what I'm looking for is "always hit 5, but don't necessarily penalise 5 if the user couldn't absolutely get 6." It might be easier to use the matchHighest instead of default matchLowest in the severity, like how RoF does (did?).

src/parser/jobs/mnk/modules/PerfectBalance.tsx
rachellhea: I think realistically we need full stack support and lookback to support mandating 6. I'd rather not dig down that rabbit hole as of now.

PR #584, opened 2019-08-23
Permit triple weave when summoning phoenix

ackwell:
title

gonna sleep now, feel free to merge and shit if it looks 'aite

also includes a rudimentary `.d.ts` for core weaving because eh


suShirow:
Will wait for Nem's greenlight before merging, given the discussion above

PR #590, opened 2019-08-24
Update MNK AoE for 5.0 skills

ayyaruq:
WIP for updating AoE, mostly so @AzariahAeroborn and @KelosPelmand can take a look and see what we do and how core Aoe could work.

**Not intended to be merged.** This is just MNK/melee inspiration for #615.


suShirow:
For the extract, you can try to revert your local modifications to `/locale`, pull from xiva master, then extract again. Works for me whenever i have that issue.

PR #592, opened 2019-08-24
Shadowmonkers weaving

ayyaruq:
Looks like CoreWeaving doesn't actually call the maxWeaves param so constant var it is. Not sure about this conditional but tslint complained about potential undefined so here we are. GitHub getting weird with indentation, lmk if there's a nicer way to do this.


suShirow:
You have an extra closing curly bracket in there, acc.

src/parser/jobs/mnk/modules/Weaving.ts
```diff
@@ -0,0 +1,17 @@
...
+		if (weave.leadingGcdEvent &&
+			weave.leadingGcdEvent.ability &&
+			weave.leadingGcdEvent.ability.guid === ACTIONS.SIX_SIDED_STAR.id) {
+				return weave.weaves.length > SSS_MAX_WEAVES
+			}
+
+		return super.isBadWeave(weave)
+		}
+	}
+}
```
suShirow: Keel dis, den relint

suShirow:
It's probably fine at the moment to leave it a const var. It ain't broke. GJ

PR #594, opened 2019-08-24
Extend PrepullStatus to synthesize cast events

AzariahAeroborn:
Define links in job STATUS files to indicate which action generates a given status, then update PrepullStatus to synthesize actions for those specified actions if it determines that it needs to synthesize an applyBuff event for the related status.  

Also includes a minor adjustment to Cooldowns module needed to truncate the cooldown of the synthesized pre-pull usage if necessary, since we don't know for certain how far pre-pull the skill was used.  This change prevents overlapping on the timeline if the second usage occurs sooner into the fight than it could have if the skill was first used at exactly 0 (the synthesized time).

### DIFF
--- a/src/data/ACTIONS/AST.js
+++ b/src/data/ACTIONS/AST.js
@@ -193,7 +193,6 @@ export default {
 		name: 'Play',
 		icon: 'https://xivapi.com/i/003000/003102.png',
 		cooldown: 0,
-		statusesApplied: [STATUSES.THE_BALANCE, STATUSES.THE_BOLE, STATUSES.THE_ARROW, STATUSES.THE_SPEAR, STATUSES.THE_EWER, STATUSES.SPIRE, STATUSES.LORD_OF_CROWNS, STATUSES.LADY_OF_CROWNS],
 	},
 
 	REDRAW: {
@@ -278,6 +277,7 @@ export default {
 		name: 'The Balance',
 		icon: 'https://xivapi.com/i/003000/003110.png',
 		cooldown: 0,
+		statusesApplied: [STATUSES.THE_BALANCE],
 	},
 
 	THE_BOLE: {
@@ -285,6 +285,7 @@ export default {
 		name: 'The Bole',
 		icon: 'https://xivapi.com/i/003000/003111.png',
 		cooldown: 0,
+		statusesApplied: [STATUSES.THE_BOLE],
 	},
 
 	THE_ARROW: {
@@ -292,6 +293,7 @@ export default {
 		name: 'The Arrow',
 		icon: 'https://xivapi.com/i/003000/003112.png',
 		cooldown: 0,
+		statusesApplied: [STATUSES.THE_ARROW],
 	},
 
 	THE_SPEAR: {
@@ -299,6 +301,7 @@ export default {
 		name: 'The Spear',
 		icon: 'https://xivapi.com/i/003000/003113.png',
 		cooldown: 0,
+		statusesApplied: [STATUSES.THE_SPEAR],
 	},
 
 	THE_EWER: {
@@ -306,6 +309,7 @@ export default {
 		name: 'The Ewer',
 		icon: 'https://xivapi.com/i/003000/003114.png',
 		cooldown: 0,
+		statusesApplied: [STATUSES.THE_EWER],
 	},
 
 	THE_SPIRE: {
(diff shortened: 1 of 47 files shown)


AzariahAeroborn:
Inverted the STATUSES<-->ACTIONS linkage to be actions indicating which statuses they apply, per conversation on Discord.  Added an empty statusesApplied property to the unknown action so that getDataBy will search that property.

src/parser/core/modules/Cooldowns.js
```diff
@@ -188,7 +188,12 @@ export default class Cooldowns extends Module {
 		// If there's a current object, move it into the history
 		// TODO: handle errors on CD overlap
 		if (cd.current) {
-			cd.history.push(cd.current)
+			if (cd.current.timestamp < this.parser.fight.start_time && cd.length * 1000 > this.parser.currentTimestamp) {
```
suShirow:
```suggestion
			if (cd.current.timestamp < this.parser.fight.start_time && action.cooldown * 1000 > this.parser.currentTimestamp - cd.current.timestamp) {
```
This is bugged. Suggestion should fix it.

suShirow:
Note: Upon merge this PR will break AST modules: `Sect.tsx`, `ArcanaTracking.tsx` but I'll put in a PR to make way for this soon enough.

src/data/ACTIONS/AST.js
```diff
@@ -129,27 +134,31 @@ export default {
...
 		cooldown: 120,
+		statusesApplied: [STATUSES.SYNASTRY_SELF, STATUSES.SYNASTRY],
 	},
 
 	COLLECTIVE_UNCONSCIOUS: {
 		id: 3613,
 		name: 'Collective Unconscious',
 		icon: 'https://xivapi.com/i/003000/003140.png',
 		cooldown: 90,
+		statusesApplied: [STATUSES.COLLECTIVE_UNCONSCIOUS_SHIELD, STATUSES.COLLECTIVE_UNCONSCIOUS],
```
suShirow:
```suggestion
statusesApplied: [STATUSES.COLLECTIVE_UNCONSCIOUS_SHIELD, STATUSES.COLLECTIVE_UNCONSCIOUS, STATUSES.WHEEL_OF_FORTUNE]
```

src/parser/core/modules/Cooldowns.js
AzariahAeroborn: I went with a slightly different approach, to be more explicit that I'm checking against fight duration for the first post-pull use, but should be the same result.

src/data/ACTIONS/AST.js
AzariahAeroborn: Added, thanks

suShirow:
Only gonna merge this later tonight so I have time to work out the switch on ast modules o/

src/parser/core/modules/PrecastStatus.js
```diff
@@ -41,6 +45,23 @@ export default class PrecastStatus extends Module {
...
 
+				// Determine if this buff comes from a known action, fab a cast event
+				const statusInfo = getDataBy(STATUSES, 'id', event.ability.guid)
+				if (statusInfo) {
+					const actionInfo = getDataBy(ACTIONS, 'statusesApplied', statusInfo)
+					// Action found - push it into _combatantActions array if not already there and synthesize a cast event
+					// If action has already been synthesized (and pushed into _combatantActions array), do nothing
+					if (actionInfo && this._combatantActions.indexOf(actionInfo.id) === -1) {
+						this._combatantActions.push(actionInfo.id)
+						const castEvent = event
```
suShirow: I found a pretty big bug with the precastStatus logic in `PrecastStatus.js`. 

First, I realized that the `statusesApplied` weren't being set on the right actions - Astro "Play" isn't actually the action that grants arcana buffs, it's the arcanas themselves like `ACTIONS.THE_BALANCE`

So when I fixed that, I noticed that 
1. prepull cast events were duplicated 
2. removebuff events for the action were having the wrong guid
3. It wouldn't get picked up by the main module loop

Upon closer inspection, my suspicion is that this line here is modifying the `removebuff` event directly rather than creating a clone. I'm not exactly sure if just doing {...event} will work though so I'll keep testing and get back to you.

suShirow: Actually, this is all it took, changing the contents from this line 56-61:
```
						events.splice(0, 0, {
							...event,
							ability: {
								...event.ability,
								name: actionInfo.name,
								abilityIcon: actionInfo.icon.replace('https://xivapi.com/i/', '').replace('/', '-'),
								guid: actionInfo.id,
							},
							timestamp: startTime - 2,
							type: 'cast',
						})
					}
```

Issue #597, opened 2019-08-25
Add a WHM Swiftcast tracker


kkragenbrink:
See #587 for an example.

PR #604, opened 2019-08-25
Add Swiftcast tracking module for WHM (#597)

nivthefox:


src/parser/jobs/whm/modules/Swiftcast.js
```diff
@@ -0,0 +1,56 @@
...
+import {ActionLink} from 'components/ui/DbLink'
+
+const MISSED_SWIFTCASTS_SEVERITIES = {
+	1: SEVERITY.MAJOR,
+}
+
+export default class Swiftcast extends BuffWindowModule {
+	static displayOrder = DISPLAY_ORDER.SWIFTCAST
+	static handle = 'swiftcast'
+	static title = t('whm.swiftcast.title')`Swiftcast Actions`
```
nonowazu: Just a minor nit, and I only really noticed once it went live on SCH, but we might want to change this to something like 'Swiftcast Usage'

With the table column header, you see 'Swiftcast Usage' in two places really close to one another.

Perhaps change the column header instead? I dunno, just offering my two cents.

suShirow: Swiftcast Usage for the title is more in line with the language in other modules around xiva imo

PR #608, opened 2019-08-30
Updated SMN missing DoT check for 5.08.

KelosPelmand:
No longer checks for festerruins on new parses, but still does on old parses. 
(Until the patch is officially listed, it should treat all parses as new.)


ayyaruq:
Looks good to me but should prob also get a look from someone more familiar with SMN. Might be able to make those assignments cleaner on the conditional somehow but yolo.

PR #611, opened 2019-08-30
AST: simple CO & Divination tracking

suShirow:
Decided to isolate Divination in OGCDDowntime. 
Neutral Sect and CU for another day :blobsweat:

CO diverges into two different copies on the suggestion depending on Sect. The logic is the same (use it dammit it's free).


suShirow:
This needs a change log

suShirow:
Changelog added. Modified core/modules/CooldownDowntime.js to allow changing the title of the checklist entry. Defaults to the standard "Use your OGCDs" title.

suShirow:
Oh I got it, thanks for pointing that out!

PR #612, opened 2019-08-30
Update Twin Snakes handling for new action and rotational preferences

ayyaruq:
This PR is probably best read by the first commit only, everything else was just renaming shit.

The logic in this is a bit funky so I'll do my best to explain. For users in GL1-3, the expectation is to keep to a 6 GCD cycle. We want to flag early refreshes to avoid potency loss, but we need to handle how this works with Four-Point Fury. Since FPF adds 10s it's still fine to get 6 GCDs into it, provided GL stacks are up, so that'll be a future TODO once the Gauge module is working properly.

For GL4, many but not all players do 2 cycles of True Strike before refreshing TS, with the expectation that TS drops for itself, but not the DK or Boot before it. The module doesn't currently check GL or downtime, so that's again future TODO work.

As with my other 5.0 MNK stuff, no changelog until Gauge is done for which I have a full log prepared.


src/parser/jobs/mnk/modules/TwinSnakes.tsx
```diff
@@ -0,0 +1,226 @@
...
+
+		this.suggestions.add(new TieredSuggestion({
+			icon: ACTIONS.TWIN_SNAKES.icon,
+			content: <Trans id="mnk.twinsnakes.suggestions.early.content">
+				Avoid refreshing <ActionLink {...ACTIONS.TWIN_SNAKES} /> signficantly before its expiration as you're losing uses of the higher potency <ActionLink {...ACTIONS.TRUE_STRIKE} />.
+			</Trans>,
+			tiers: {
+				1: SEVERITY.MEDIUM,
+				4: SEVERITY.MAJOR,
+			},
```
rachellhea: I'd say 1 minor, 3 medium, 5 major

src/parser/jobs/mnk/modules/TwinSnakes.tsx
```diff
@@ -0,0 +1,226 @@
...
+import Invulnerability from 'parser/core/modules/Invulnerability'
+import Suggestions, {SEVERITY, Suggestion, TieredSuggestion} from 'parser/core/modules/Suggestions'
+
+import DISPLAY_ORDER from './DISPLAY_ORDER'
+
+const TWIN_SNAKES_CYCLE_BUFFER = 3000
+const TWIN_SNAKES_CYCLE_LENGTH = 6
+
+class TwinState {
+	casts: CastEvent[]
```
rachellhea:
```suggestion
	casts: CastEvent[] = []
```

src/parser/jobs/mnk/modules/TwinSnakes.tsx
```diff
@@ -0,0 +1,226 @@
...
+	start: number
+	end?: number
+
+	constructor(timestamp: number) {
+		this.start = timestamp
+		this.casts = []
```
rachellhea:
```suggestion
```

src/parser/jobs/mnk/modules/TwinSnakes.tsx
```diff
@@ -0,0 +1,226 @@
...
+	}
+
+	// Mainly here in case we care about oGCDs being unbuffed later
+	public get gcds(): number {
```
rachellhea:
```suggestion
	get gcds(): number {
```

Typescript getters are public by default.

src/parser/jobs/mnk/modules/TwinSnakes.tsx
```diff
@@ -0,0 +1,226 @@
...
+	private onComplete() {
+		// Close off the last window
+		this.stopAndSave(this.parser.fight.end_time)
+
+		// Calculate derped potency to early refreshes
+		const lostTruePotency = this.earlySnakes * (ACTIONS.TRUE_STRIKE.potency - ACTIONS.TWIN_SNAKES.potency)
+
+		this.checklist.add(new Rule({
+			name: <Trans id="mnk.twinsnakes.checklist.name">Keep Twin Snakes up</Trans>,
+			description: <Trans id="mnk.twinsnakes.checklist.description">Twin Snakes is an easy 10% buff to your DPS across the board.</Trans>,
```
rachellhea:
```suggestion
			description: <Trans id="mnk.twinsnakes.checklist.description">Twin Snakes is an easy 10% buff to your DPS.</Trans>,
```

src/parser/jobs/mnk/modules/TwinSnakes.tsx
```diff
@@ -0,0 +1,226 @@
...
+	private onDrop(event: BuffEvent): void {
+		this.stopAndSave(event.timestamp)
+	}
+
+	private onComplete() {
+		// Close off the last window
+		this.stopAndSave(this.parser.fight.end_time)
+
+		// Calculate derped potency to early refreshes
+		const lostTruePotency = this.earlySnakes * (ACTIONS.TRUE_STRIKE.potency - ACTIONS.TWIN_SNAKES.potency)
```
rachellhea: Is this necessary? We can just make a note of how many times they refreshed early, and I think that's easier to parse.

src/parser/jobs/mnk/modules/TwinSnakes.tsx
```diff
@@ -0,0 +1,226 @@
...
+			value: this.lateSnakes,
+			why: <Trans id="mnk.twinsnakes.suggestions.late.why">
+				<Plural value={this.lateSnakes} one="# GCD was" other="# GCDs were" /> lost to delayed refreshing.
+			</Trans>,
+		}))
+
+		this.suggestions.add(new TieredSuggestion({
+			icon: ACTIONS.FOUR_POINT_FURY.icon,
+			content: <Trans id="mnk.twinsnakes.suggestions.toocalm.content">
+				Try to get <StatusLink {...STATUSES.TWIN_SNAKES} /> up before using <ActionLink {...ACTIONS.FOUR_POINT_FURY} /> to take advantage of it's free refresh.
```
rachellhea:
```suggestion
				Try to get <StatusLink {...STATUSES.TWIN_SNAKES} /> up before using <ActionLink {...ACTIONS.FOUR_POINT_FURY} /> to take advantage of its free refresh.
```

Grammar correction.

src/parser/jobs/mnk/modules/TwinSnakes.tsx
ayyaruq: ack mentioned this but I like being explicit, same for a lot of my returns being typed even when it's easy to infer.

src/parser/jobs/mnk/modules/TwinSnakes.tsx
ayyaruq: I think it's useful, just because of how hard MNK hits so being able to quantify potency is good. The actual suggestion includes how many times an early refresh was done on top of the potency.

src/parser/jobs/mnk/modules/TwinSnakes.tsx
ayyaruq: Any particular reasoning? It's a decent potency loss even just once, and 4 times is enough to noticeably effect DPS output. I'm fine with changing it, I think I chose 4 for major because that's the point where you've lost a full True Strike worth of damage.

rachellhea: Fair enough.

src/parser/jobs/mnk/modules/TwinSnakes.tsx
```diff
@@ -0,0 +1,225 @@
...
+
+	private onCast(event: CastEvent): void {
+		const action = getDataBy(ACTIONS, 'id', event.ability.guid) as TODO
+
+		if (!action) {
+			return
+		}
+
+		// Only include GCDs
+		if (action.onGcd) {
```
suShirow: I think you should be able to flatten this function by adding another conditional to the above return thing
```
if(!action || !action.onGcd) {
return
}
```

src/parser/jobs/mnk/modules/TwinSnakes.tsx
```diff
@@ -0,0 +1,225 @@
...
+			value: this.earlySnakes,
+			why: <Trans id="mnk.twinsnakes.suggestions.early.why">
+				{lostTruePotency} potency lost to <Plural value={this.earlySnakes} one="# early refresh" other="# early refreshes" />.
+			</Trans>,
+		}))
+
+		this.suggestions.add(new TieredSuggestion({
+			icon: ACTIONS.TWIN_SNAKES.icon,
+			content: <Trans id="mnk.twinsnakes.suggestions.late.content">
+				Refresh <ActionLink {...ACTIONS.TWIN_SNAKES} /> on every third combo is a potency gain but only when you don't drop the buff for the GCD before it.
```
suShirow: grammar: Refreshing

src/parser/jobs/mnk/modules/TwinSnakes.tsx
```diff
@@ -0,0 +1,225 @@
...
+			},
+			value: this.lateSnakes,
+			why: <Trans id="mnk.twinsnakes.suggestions.late.why">
+				<Plural value={this.lateSnakes} one="# GCD was" other="# GCDs were" /> lost to delayed refreshing.
+			</Trans>,
+		}))
+
+		this.suggestions.add(new TieredSuggestion({
+			icon: ACTIONS.FOUR_POINT_FURY.icon,
+			content: <Trans id="mnk.twinsnakes.suggestions.toocalm.content">
```
suShirow: toocalm 🙄

ayyaruq: For when you don't have enough Fury!

PR #614, opened 2019-08-30
Support 5.0-5.08 properly for MNK forms

ayyaruq:
This ended up bigger than I thought, PoorForm check is deleted to make way for #583. Ternary on timeout to handle 5.0 launch for non-global zones without breaking compat for 5.05 and later. The rest is mostly wording updates since we no longer recommend TK and how FS gets used is a little different now.


src/parser/jobs/mnk/modules/Forms.tsx
```diff
@@ -115,41 +112,30 @@ export default class Forms extends Module {
...
-					Avoid using <ActionLink {...ACTIONS.DRAGON_KICK}/> and <ActionLink {...ACTIONS.ARM_OF_THE_DESTROYER}/> outside of <StatusLink {...STATUSES.OPO_OPO_FORM}/>. Their special effects only activate when in the correct form and <ActionLink {...ACTIONS.BOOTSHINE} /> has equal or higher potency depending on crits.
-				</Trans>,
-				why: <Trans id="mnk.forms.suggestions.formless.why">
-					<Plural value={this.formless} one="# combo-starter was" other="# combo-starters were" />  used Formlessly, cancelling this special effects.
-				</Trans>,
-			}))
-		}
+		this.suggestions.add(new TieredSuggestion({
+			icon: ACTIONS.FORM_SHIFT.icon,
+			severity: SEVERITY.MEDIUM,
```
suShirow: I don't think severity should be specified as this is TieredSuggestion; it gets overwritten by `tiers`
Anyway medium is the default severity :p Sorry for the nitpick

ayyaruq: It's a leftover, good catch!

PR #615, opened 2019-08-30
Added core module to check usages of AoE skills

KelosPelmand:
This module checks for usages of AoE skills that have single target alternatives, such as BLMs Xenoglossy or NINs Hellfrog Medium, to ensure they hit the minimum number of targets.  It should not be used to check skills that can hit AoE but do not have alternatives, such as DRGs Dragonfire Dive or SMNs Deathflare, as it is correct to use those skills on single targets.

For the AoE combos of melee DPS and Tanks, define a trackedAbility with the first abilities of the AoE and single target combos.  This is to provide leeway to finishing AoE combos when the number of targets may drop below the minimum during the combo.

The icon of the first tracked ability will be used as the icon for the suggestion.


ayyaruq:
Mostly looks good to me, one thing that stands out is the icon stuff using the first skill in the suggestion. This is prob fine for most jobs, but for MNK because we have different numbers of min hits and can AoE any action in our combo cycle I'd want to be able to show separate suggestions for each skill.

KelosPelmand:
It is certainly possible to have a separate suggestion per skill.  My reasoning for a single suggestion was that we are checking for and suggesting the same thing for each of them.  The number of minimum targets may be different per skill, but the underlying guidance is the same for all of them (use ST when it would do more damage).

I had also thought about listing the minimum targets as a column in the table, but decided against it, partially due to MNK's variable number of targets for AotD.  If we want to add that column, it would be possible to add a second override for generating that column's content, similar to the one already provided during the checking phase.

ayyaruq:
I had a think about it and suggestions comes down to, is it a mentor reviewing an analysis and wanting summary feedback for a player, or a player doing self-analysis and wanting to see the conditions for each thing.

Since MNK AoE is somewhat more variable, it makes sense for us to show multiple suggestions in the latter case, but not the former. We can also mix and match our AoE with normal skills, the value of a suggestion per-skill is higher, like in the case of mobs dying mid-combo, do you cancel combo, or do you keep going? For MNK we don't need to cancel *or* continue it. Having per-skill suggestions make it more obvious where to switch and optimise. This is not the case for any other job as far as I know. This might be best just to ask users in Discord what they prefer. The one thing I do want to absolutely see is, for a list of AoE actions, with a single suggestion, the icon should be the first one that got messed up, not the first in the list.

For the table, I'm not sure if it's of that much value to MNK, but it could be, just need a way to cleanly represent when they missed min-targets on buffed vs unbuffed.

ackwell:
I guess this is a good example of what I was talking about in `#dev-ui-ux` earlier - the actual logic around suggestions for this sounds somewhat nuanced, and in need of discussion/work - but the core _data_ of AoEs and how many things they hit doesn't really rely on it.

Would it be worth cutting out suggestion stuff for first-pass to get the info out there at all, then look into deriving suggestions from that info in a second changeset?

Initial worry on doing that would be a potentially bloated output without some job-specific rules, though.

KelosPelmand:
Core aoe now ready to merge.  Suggestion icon is now a property instead of using icon of first checked ability.

src/parser/core/modules/BuffWindow.tsx
```diff
@@ -38,7 +38,7 @@ export class BuffWindowState {
 	}
 }
 
-interface SeverityTiers {
+export interface SeverityTiers {
```
ackwell: Not a fan of AoE importing sevs from buffwindows, they should be stand-alone modules, esp. as they have no other dependency on each other.

Technically the correct solution is to add a type def for suggestions, but that's out of scope here. Cleaner to just add a dupe interface privately to your module.

src/parser/jobs/smn/modules/MultiHitSkills.js
```diff
@@ -1,65 +1,37 @@
...
-}
-
-export default class MultiHit extends Module {
-	static handle = 'multihit'
-	static title = t('smn.multihit.title')`Multi-Hit`
+export default class MultiHit extends AoEUsages {
 	static dependencies = [
 		'gauge',
-		'suggestions',
+		//need to include base class dependency or it will not be pulled in
```
ackwell: This is brittle, relies on anyone changing the base module to check that this is also updated.

Using old-style js modules like this, it's usually handled by spreading the parent's dependencies in:

```js
static dependencies = [
	...ParentModule.dependencies,
	'newDependency',
]
```

Having said that, I'm not _entirely_ sure if that will work when the base class is using `@dependency` - a quick check should make it clear. If it _doesn't_ work, adding `@dependency gauge` should work fine in JS, and will work.

src/parser/core/modules/AoEUsages.tsx
```diff
@@ -0,0 +1,139 @@
...
+	/**
+	 * Implementing modules MAY override the severity tiers for incorrect AoE usages.
+	 */
+	protected severity: SeverityTiers = {
+		1: SEVERITY.MINOR,
+		4: SEVERITY.MEDIUM,
+		7: SEVERITY.MAJOR,
+	}
+
+	private badUsages: Map<number, number> = new Map<number, number>()
```
ackwell: nit; ts is pretty magic, you can just `private badUsages = new Map<number, number>()` and it'll type it for you

src/parser/core/modules/AoEUsages.tsx
```diff
@@ -0,0 +1,139 @@
...
+		}
+	}
+
+	private onComplete() {
+		const totalBadUsages = Array.from(this.badUsages.values()).reduce((acc, cur) => acc + cur, 0)
+
+		this.suggestions.add(new TieredSuggestion({
+			icon: this.suggestionIcon,
+			content: <Trans id="core.aoeusages.suggestion.content">
+				Use single target abilities unless there are enough targets for the AoE alternatives to do more damage.
```
ackwell: I wonder if this can be fleshed out more. Feels a bit terse, perhaps.

src/parser/core/modules/BuffWindow.tsx
KelosPelmand: Duping the interface feels like it should be a last resort.  In other languages I work in, I'd just pull SeverityTiers into it's own file and import it into both BuffWindow and my new files.  Would that be suitable in ts here?

src/parser/jobs/smn/modules/MultiHitSkills.js
KelosPelmand: I was not aware of this syntax for this, I'll change it in the next go.

src/parser/core/modules/AoEUsages.tsx
KelosPelmand: File under: things TS/JS let you do that other languages don't.  Most other languages I'm used to don't allow type inference for things other than local variables.  Will shorten in the next go

src/parser/core/modules/AoEUsages.tsx
KelosPelmand: I am open for alternate ideas.  I was aiming for the shortest thing that still says everything it needs to.  I don't think listing target counts here would make much sense, given that they can vary by skill or situation.

src/parser/core/modules/BuffWindow.tsx
ackwell: ehhh... sorta. Like, yeah, it's technically the right solution, as part of a `.d.ts` file, but that then overrides inferred types from the JS file completely, so you'd need to type _all_ of suggestions, not just the interface. It's on The List:tm: to do.

If you really don't want to dupe it, can make a `SeverityTiers.ts`, but it'll play havok given the rest of the module is JS.

src/parser/core/modules/AoEUsages.tsx
ackwell: I wanna say to avoid trying to be short in suggestion `content`. Will agree about listing target counts, though.

This is a total off-the-top-of-my-head attempt, but something like "Avoid using AoE actions when an alternative would do more damage. Many AoE actions share resources and/or cooldowns with a single-target action."

That's not a great example at all, but the point is to try and explain why the suggestion is a suggestion, as well as the path to fix it.

The `why` property itself is more "why this suggestion was triggered", rather than "why this suggestion is a suggestion".

KelosPelmand: I'm still trying to come up with something for the content, but i wanted to clarify about the `why`. Are you making a general statement about `why` properties or saying that this suggestion does not currently have a good `why`?

ackwell: Just general while I was thinking about it

KelosPelmand: how about:

> Avoid using AoE abilities when they would do less damage than an alternative single-target ability that shares a resource cost or cooldown.

for the suggestion content and changing the table description to:

> Some AoE abilities share resource costs or cooldowns with a single-target ability.  If the AoE skill cannot hit enough targets, the single-target ability will do more total damage and should be used instead.

ackwell: Better. I get the sense a bit of that detail could be moved from the table desc to the suggestion but whatever, same same.

PR #628, opened 2019-09-04
Fix normalisation constant for sub-gcd casts

EmoniArkewright:
Time between GCD starts is used for estimation, so excluding instant casts we need the larger of gcd recast and cast time to normalise with.
Currently sub-gcd casts are overestimated by a factor of 2.5/cast time


ayyaruq:
This looks fine to me, might just get someone more versed in the GCD module normaliser and caster stuff to double check too.

PR #633, opened 2019-09-07
Backwards compatibility for FFLogs calculated events change

AzariahAeroborn:
Hook both calculated damage/heal events and "base" damage/heal events.  If we see a calculated event (which occurs at same timestamp as corresponding cast event and will always be before the damage/heal event), stop listening for base events, since this is a new log.


AzariahAeroborn:
Per discussion with ack on discord, changed approach to be more normaliser based, new fflogsEvent normaliser will allow modules to hook damageEventName or healEventName from that normaliser to get the correct event name depending on if log is before or after calculated events.

PR #636, opened 2019-09-09
Statuses in timeline

Airex:


Folder.DotSettings.user
```diff
@@ -0,0 +1,4 @@
+﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
```
Toastdeib: I didn't notice it on my first pass, but this looks like an auto-generated file from whatever IDE you're using and probably shouldn't be included in the commit.

Airex:
I tried to play with localization and stuck with passing t() objects into VIS component as application fails. Have to find a way to resolve localization in runtime  and pass translated values. Also, we have not localized action and statuses names  regarding to game locale settings.

src/parser/core/modules/Statuses.js
```diff
@@ -0,0 +1,180 @@
...
+
+		if (prev) {
+			prev.end = event.timestamp - this.parser.fight.start_time
+		}
+	}
+
+	_addStatus(event, status) {
+		if (this._statuses.some(it => {
+			const diff = Math.abs(event.timestamp - this.parser.fight.start_time - it.start)
+			return it.status.id === status.id && diff <= 2 * 1000
```
nivthefox: `2 * 1000` should be best represented as a constant somewhere, so that it's obvious what this value means.

src/parser/core/modules/Statuses.js
```diff
@@ -0,0 +1,182 @@
...
+
+		})
+	}
+
+	_lookForColor(status) {
+		const setting = Object.values(STATUS_EFFECT_TYPES)
+			.find(value => {
+				return value.ids.some(id => id === status.id)
+			})
+		return setting && setting.settings && setting.settings.color || ''
```
ackwell: nit; pref `undefined` for a null value

src/data/STATUSES/index.js
```diff
@@ -67,8 +67,76 @@ function correctIdsToMatchLogs (obj) {
...
 			status.id = status.id + STATUS_ID_OFFSET
 		}
+		if (status.stackWith) {
+			status.stackWith += STATUS_ID_OFFSET
+		}
 	})
 	return obj
 }
 
+export const STATUS_EFFECT_TYPES = {
```
ackwell: blocking;

I'm well on-record as not being particularly on-board with classifying statuses - but if people are _determined_ for it to be a thing, then I think it's going to be necessary to strike up a conversation about _how_ we do it - and I don't think this current structure is going to be it.

Given it (looks?) pretty unused currently, can we nuke the structure + usage of it in the short term, and re-visit adding it in a future PR? Would help reduce the surface of this PR, too - as it's pretty damn big.

src/parser/core/modules/Statuses.js
```diff
@@ -0,0 +1,182 @@
...
+
+		this.addHook('complete', this._onComplete)
+		this.addHook('applybuff', {by: 'player'}, this._onApply)
+		this.addHook('applydebuff', {by: 'player'}, this._onApply)
+		this.addHook('refreshdebuff', {by: 'player'}, this._onRefresh)
+		this.addHook('refreshbuff', {by: 'player'}, this._onRefresh)
+		this.addHook('removebuff', {by: 'player'}, this._onRemove)
+		this.addHook('removedebuff', {by: 'player'}, this._onRemove)
+
+		this.addHook('applybuff', {by: 'pet'}, this._onApply)
```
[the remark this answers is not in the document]
Airex: isn't better to provide ability to assign array to by prop, by: ['player', 'pet']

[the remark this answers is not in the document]
ackwell: If this becomes a recurring issue, sure. Outside scope of this PR, however.

Airex:
So, I have added ability to define Statuses module per job. In this case each status can be configured with additional options, at this moment it just provides possibility to map it to other status so they can stack at same row, but it can be good place to provide color settings for each status as well.

src/parser/core/modules/Timeline/Timeline.module.css
```diff
@@ -17,6 +17,27 @@
...
+
+:global .vis-item.vis-background .vis-item-content {
+	width: 100% !important;
+}
+
+:global .vis-range .vis-item-content {
+	padding: 0;
+}
+
+:global .vis-range .progress-fl img {
```
ackwell: medium; doesn't look like `.progress-fl` is being used anywhere now.

src/data/STATUSES/index.js
```diff
@@ -67,6 +67,9 @@ function correctIdsToMatchLogs (obj) {
 		} else {
 			status.id = status.id + STATUS_ID_OFFSET
 		}
+		if (status.stackWith) {
```
ackwell: medium; `stackWith` is unused now, so this is redundant.

src/parser/core/modules/Statuses.js
```diff
@@ -0,0 +1,174 @@
...
+
+	_statuses = {}
+	_groups = {}
+	_statusToActionMap = {}
+	_actionToMergeNameMap = {}
+
+	constructor(...args) {
+		super(...args)
+
+		const ids = [this.parser.player.id, ...this.parser.player.pets]
```
ackwell: medium; is this intentional? `player.pets` is an array of pet objects, not IDs. If you want just actor IDs, it'd be `[this.parser.player.id, ...this.parser.player.pets.map(pet => pet.id)]`

src/parser/core/modules/Statuses.js
```diff
@@ -0,0 +1,174 @@
...
+				})
+			}
+		})
+
+		// Map statuses to actions
+		Object.values(ACTIONS).forEach(ac => {
+			if (ac.statusesApplied) {
+				ac.statusesApplied.forEach(st => {
+					if (st) {
+						this._statusToActionMap[st.id] = ac.id
```
ackwell: medium; It looks like the only place the value of the map is used, you're pulling up the action data with `getDataBy`. Object references are cheap in JS, you'd be better off just setting the entire action object as the value in the map here, and skipping the `getDataBy` call later.

Airex:
> * It looks like you might have some auto-formatting config running in your editor, potentially? (i.e. changes in Cooldowns.js) - try to avoid running autoformatting that isn't encoded in the eslint/tslint config in the repo, as it just generates diff kludge.
Yeah, resharper in VS 2017. makes life easier, but must be configured for this project separatelly.

src/parser/core/modules/Statuses.js
Airex: done

src/parser/core/modules/Statuses.js
Airex: done

src/data/STATUSES/index.js
Airex: done

src/parser/core/modules/Timeline/Timeline.module.css
Airex: done

src/parser/core/modules/Statuses.js
```diff
@@ -47,7 +47,7 @@ export default class Statuses extends Module {
 			if (ac.statusesApplied) {
 				ac.statusesApplied.forEach(st => {
 					if (st) {
-						this._statusToActionMap[st.id] = ac.id
+						this._statusToActionMap[st.id] = getDataBy(ACTIONS, 'id', ac.id)
```
ackwell:
```suggestion
						this._statusToActionMap[st.id] = ac
```

Already got the action, don't need to look it up.

PR #639, opened 2019-09-11
Riddle of Earth module

ayyaruq:
May wanna review from specific commit but tl;dr I just nuked the GL normaliser outright because Speedmod isn't working properly anyway and it'd be better to just use the stack counters once gauge is merged. This also removes from old RoE suggestions since there are better ways to save stacks than RoE now, it's mostly just defence and memeing positionals.


src/parser/jobs/mnk/modules/RiddleOfEarth.tsx
```diff
@@ -0,0 +1,84 @@
...
+	start: number
+	end?: number
+
+	constructor(start: number) {
+		this.start = start
+	}
+}
+
+export default class RiddleOfEarth extends Module {
+	static handle = 'riddleofearth'
```
ackwell: nit: `riddleOfEarth`

src/parser/jobs/mnk/modules/RiddleOfEarth.tsx
```diff
@@ -0,0 +1,84 @@
...
+		this.addHook('applybuff', {to: 'player', abilityId: STATUSES.RIDDLE_OF_EARTH.id}, this.onGain)
+		this.addHook('applybuff', {to: 'player', abilityId: STATUSES.EARTHS_REPLY.id}, this.onReply)
+		this.addHook('removebuff', {to: 'player', abilityId: STATUSES.RIDDLE_OF_EARTH.id}, this.onDrop)
+
+		this.addHook('complete', this.onComplete)
+	}
+
+	private onGain(event: BuffEvent): void {
+		this.riddle = new Riddle(event.timestamp)
+		this.riddle.active = true
```
ackwell: It doesn't look like `active` is really doing much. You're setting it to `true` in the only place `Riddle` is constructed, and it isn't set to `false` until it's in the history. Worth nuking?

ayyaruq: It's mainly there to make sure we only push an open window at the end, instead of accidentally duping I guess. Maybe I'm being overly cautious?

ayyaruq:
Speedmod killed for now since it didn't work anyway and I removed the old GL module in favour of new RoE. I plan to maybe re-add it once gauge is merged, if it even works then (might not, who knows).

src/parser/jobs/mnk/modules/index.ts
```diff
@@ -17,12 +16,11 @@ export default [
 	Demolish,
 	Fists,
 	Forms,
-	GreasedLightning,
 	MnkAoE,
 	OGCDDowntime,
 	PerfectBalance,
+	RiddleOfEarth,
 	RiddleOfFire,
-	Speedmod,
```
ackwell: I can see the delete of `GreasedLightning`, but not `Speedmod` - is it sitting around unused atm?

src/parser/jobs/mnk/index.tsx
```diff
@@ -21,25 +21,56 @@ export default new Meta({
 		<Message warning icon>
 			<Icon name="warning sign"/>
 			<Message.Content>
-				<Trans id="mnk.about.description.warning"><b>The module is still a work in progress</b> and may occasionally give you bad feedback. If you notice any issues, or have any questions or feedback, please drop by our Discord channel!</Trans>
+				<Trans id="mnk.about.description.warning">
```
ackwell: Consider using markdown for this, given you're using some formatting tags.

src/parser/jobs/mnk/modules/index.ts
ayyaruq: I kept the file in for now, since I intend to re-add Speedmod eventually. The way it is atm (operating off statuses) didn't really fly when the statuses don't exist anymore, and it was inaccurate anyway. I tried to get it to use a scalar set by the specific gauge state but it was kinda derpy so I just disabled it for now to avoid confusing users.

src/parser/jobs/mnk/index.tsx
ayyaruq: I feex

PR #641, opened 2019-09-11
I feex Steppies

ayyaruq:
This won't actually fix it, no of merging yet. Calculated events need to be typed and added to fflogs defs.


ayyaruq:
Fixed by ack

PR #646, opened 2019-09-16
SAM support expansion/fine tuning

Rjb99:
1) meikyo has been converted to core.

2) temporary AoE check is in place until I have time to look at and adjust for core (this thing has been in my repo for months half finished)

3)reduced suggestions for third eye to all minor minus not using based on SAM feedback.


src/parser/jobs/sam/modules/AoeChecker.js
```diff
@@ -0,0 +1,136 @@
...
+
+					this._badKenkiCasts++
+				}
+			}
+
+		} else if (GAIN_AT_3.has(action)) {
+
+			//Step 2: Check break point
+
+			if (event.hits.length < 3) {
```
suShirow: Since lint complains this is a magic number, maybe convert to big const so that everyone doesn't have to see the warning pop up everytime :p

src/parser/jobs/sam/modules/Meikyo.tsx
```diff
@@ -0,0 +1,63 @@
...
+			GCDs used during <ActionLink {...ACTIONS.MEIKYO_SHISUI}/> should be limited to sen building skills. </Trans>,
+		severityTiers: {
+			1: SEVERITY.MAJOR,
+		},
+
+	}
+
+// I have NO clue what I'm doing, I just hope this works for overriding considerAction
+
+considerAction(action: Action) {
```
suShirow: This works fine. 
Nit: You can probably skip the else and just do 
```
if (IGNORE_THIS.has(action.id)) {
			return false
		} 
super(action)
```
But same difference

src/parser/jobs/sam/modules/Meikyo.tsx
```diff
@@ -0,0 +1,63 @@
+import {t, Trans} from '@lingui/macro'
```
suShirow: Something went weird with the format indenting for this file x_x

src/parser/jobs/sam/modules/Meikyo.tsx
suShirow: or even just 
```
if (IGNORE_THIS.has(action.id)) {
			return false
		} 
return true
```

src/parser/jobs/sam/modules/AoeChecker.js
```diff
@@ -0,0 +1,136 @@
...
+                        Using aoe moves when you don't reach the target threshold for it to be a gain is a loss.
+			</Trans>,
+			tiers: {
+				1: SEVERITY.MINOR,
+				3: SEVERITY.MEDIUM,
+				5: SEVERITY.MAJOR,
+			},
+			value: this._badAoeGCDs,
+			why: <Trans id = "sam.aoechecker.suggestions.badgcd.why">
+                                        You misused kenki moves <Plural value ={this._badAoeGCDs} one="# time" other="# times" /> during the fight.
```
suShirow: Seems like the wording here isn't updated to reflect that this one is triggered by bad GCDs.

PR #649, opened 2019-09-20
Add interrupted cast tracking for SCH

nonowazu:
By request in the SCH section of Balance, add tracking for dropped casts.


src/parser/jobs/sch/modules/Interrupts.js
```diff
@@ -0,0 +1,112 @@
...
+	static handle = 'interrupts'
+	static title = t('sch.interrupts.title')`Interrupted Casts`
+
+	static dependencies = [
+		'timeline',
+		'gcd',
+		'suggestions',
+	]
+
+	_currentCast = null
```
ackwell: nit; pref. `undefined` over `null`

It's _stupid_ that JS even has both, but TS makes it _much_ nicer to deal with `undefined` than it does `null` (`foo?: bar` vs `foo: bar | null`).

src/parser/jobs/sch/modules/Interrupts.js
```diff
@@ -0,0 +1,112 @@
...
+	_onCast(event) {
+		const guid = event.ability.guid
+		// if the thing they started casting doesn't match up with what they cast, that's an interrupted cast
+		if (this._currentCast && guid !== this._currentCast.ability.guid) {
+			this._pushDropCasts(event)
+		}
+		this._currentCast = null
+	}
+
+	_pushDropCasts(currentEvent) {
```
ackwell: nit; `_pushDroppedCast`

src/parser/jobs/sch/modules/Interrupts.js
```diff
@@ -0,0 +1,112 @@
...
+			timestamp: this._currentCast.timestamp,
+		})
+	}
+
+	_onComplete() {
+		this.suggestions.add(new TieredSuggestion({
+			icon: ACTIONS.INTERJECT.icon,
+			tiers: INTERRUPT_SEVERITY,
+			value: this._droppedCasts.length,
+			content: <Trans id="sch.interrupts.suggestion.content">If you can, try to preposition yourself so you don't have to move during mechanics as much as possible. Utilizing slidecasting will lower the need to use <ActionLink {...ACTIONS.SCH_RUIN_II}/> to instantly relocate or interrupt your current Broil III cast</Trans>,
```
ackwell: Yeah, I feel this & the `why` could do with a bit of work, but that can be done at a later stage.

---

nit; avoid adding spaces around substitutions in jsx, looks a bit jank.

src/parser/jobs/sch/modules/Interrupts.js
```diff
@@ -0,0 +1,112 @@
...
+					<Table.HeaderCell>
+						<strong><Trans id="sch.interrupts.table.cast">Cast</Trans></strong>
+					</Table.HeaderCell>
+				</Table.Row>
+			</Table.Header>
+			<Table.Body>
+				{
+					this._droppedCasts.map((cast) =>
+						<Table.Row key={cast.timestamp}>
+							<Table.Cell textAlign="center">
```
ackwell: medium; This cell seems to be really crunching up, which is making each row taller than it need be. Possibly due to the `collapsing` on the header?

Fuck I hate semantic aaaa

PR #651, opened 2019-09-20
Added old actions to the sim and combo validity for unheated shots

Toastdeib:
Small update so old synced content doesn't display logs as broken. This is only a resource sim/combo action update. I'm not bothering to add analysis for Rook Autoturret usage; the Automaton Queen section of the parse will still come up empty even though Battery expenditure is calculated properly.


suShirow:
I thought we don't support old <80 content (._.)

PR #652, opened 2019-09-21
Fixed incorrect DWaD snapshot warnings

Toastdeib:


src/parser/jobs/nin/modules/TrickAttackWindow.js
```diff
@@ -24,8 +24,8 @@ export default class TrickAttackWindow extends Module {
 		super(...args)
 		this.addHook('cast', {by: 'player', abilityId: ACTIONS.DREAM_WITHIN_A_DREAM.id}, () => this._dwadCast = true)
 		this.addHook('combo', {by: 'player', abilityId: ACTIONS.ARMOR_CRUSH.id}, this._onArmorCrush)
-		this.addHook('damage', {by: 'player', abilityId: ACTIONS.DREAM_WITHIN_A_DREAM.id}, this._onDwadHit)
-		this.addHook('damage', {by: 'player', abilityId: ACTIONS.ASSASSINATE.id}, this._onAssassinate)
+		this.addHook('calculateddamage', {by: 'player', abilityId: ACTIONS.DREAM_WITHIN_A_DREAM.id}, this._onDwadHit)
+		this.addHook('calculateddamage', {by: 'player', abilityId: ACTIONS.ASSASSINATE.id}, this._onAssassinate)
```
AzariahAeroborn: This will not be backwards compatible with logs uploaded before the 5.08 event change. If you look at the way Entities.js or AoE.js in core are hooking for damage events, that should be compatible with both log formats.

Toastdeib: Mm, fair point. I'll peek at what we're doing elsewhere while silently cursing fflogs.

PR #654, opened 2019-09-24
AST primitive synastry tracking

suShirow:
Based off a suggestion I got from someone awhile back. Could evolve conceptually to an interesting tracking module, either for Synastry usage or ST heal usage, or both. 

Was just a primordial soup sitting in my branches, decided to go ahead and PR it in since it's not without value in current state.

Currently counts how many ST GCD Heals were used while Synastry was available, spits out minor suggestion.


src/parser/jobs/ast/modules/Synastry.tsx
```diff
@@ -0,0 +1,77 @@
...
+		const held = event.timestamp - this.lastUse - (ACTIONS.SYNASTRY.cooldown * 1000)
+		if (held > 0) {
+			this.totalHeld += held
+		}
+		// update the last use
+		this.lastUse = event.timestamp
+	}
+
+	onComplete() {
+		if (this.nonSynastryHeals) {
```
Toastdeib: No need for the `if` here; `TieredSuggestion`s handle it implicitly.

src/parser/jobs/ast/modules/Synastry.tsx
```diff
@@ -0,0 +1,77 @@
...
+		this.nonSynastryHeals++
+	}
+
+	private onCast(event: CastEvent) {
+		this.uses++
+		if (this.lastUse === 0) { this.lastUse = this.parser.fight.start_time }
+
+		const held = event.timestamp - this.lastUse - (ACTIONS.SYNASTRY.cooldown * 1000)
+		if (held > 0) {
+			this.totalHeld += held
```
Toastdeib: This doesn't appear to be getting used; were you going to include a suggestion for it? Also if yes, would probably be worth accounting for fight downtime to avoid excessive penalization.

src/parser/jobs/ast/modules/Synastry.tsx
```diff
@@ -0,0 +1,77 @@
...
+	protected init() {
+		this.addHook('cast', {abilityId: ACTIONS.SYNASTRY.id, by: 'player'}, this.onCast)
+		this.addHook('cast', {abilityId: GCD_ST_HEAL, by: 'player'}, this.onSingleTargetHealCast)
+		this.addHook('complete', this.onComplete)
+	}
+
+	private onSingleTargetHealCast(event: CastEvent) {
+		// Ignore if Synastry is still on CD or we already have it up
+		if (this.combatants.selected.hasStatus(STATUSES.SYNASTRY_SELF.id)
+		|| this.lastUse + (ACTIONS.SYNASTRY.cooldown * 1000) > event.timestamp) {
```
Toastdeib: Not a huge issue since you're using `lastUse` elsewhere to track the hold time, but it's worth noting that the core Cooldowns module tracks this for you and you can simplify the manual tracking into `this.cooldowns.getCooldownRemaining(ACTIONS.SYNASTRY.id) > 0`.

PR #655, opened 2019-09-24
Capture broken log triggers on sentry

ackwell:
As per request. Keep ya fingers on the revert buzzer for this one, I've got no insight on how often these fire, and we might flood sentry.


src/parser/core/Parser.tsx
```diff
@@ -268,7 +268,7 @@ class Parser {
 				Sentry.withScope(scope => {
 					scope.setTags(tags)
 					scope.setExtras(extra)
-					Sentry.captureException(extra)
+					Sentry.captureException(error)
```
Toastdeib: lolrip

PR #659, opened 2019-09-28
Improved Firebird Trance triple-weave check

KelosPelmand:
Original implementation allowed any triple weave that included a FBT.  Now only a triple weave after an instant GCD with FBT as the second weave is allowed.  The following will now be considered bad weaves:

* 3 weaves, Phoenix 3rd, after R2/4
* 3 weaves, Phoenix 2nd, after hardcast R3
* 3 weaves, Phoenix 1st, after R2/4

### DIFF
--- a/src/parser/jobs/smn/modules/Weaving.ts
+++ b/src/parser/jobs/smn/modules/Weaving.ts
@@ -1,15 +1,39 @@
 import ACTIONS from 'data/ACTIONS'
+import {CastEvent} from 'fflogs'
+import {dependency} from 'parser/core/Module'
+import CastTime from 'parser/core/modules/CastTime'
 import CoreWeaving, {WeaveInfo} from 'parser/core/modules/Weaving'
 
 const PERMITTED_PHOENIX_WEAVES = 3
 
 export class Weaving extends CoreWeaving {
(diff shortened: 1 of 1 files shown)


src/parser/jobs/smn/modules/Weaving.ts
```diff
@@ -1,15 +1,39 @@
...
+		// weave after an instant cast skill.  (If it is the first weave, the triple
+		// is not needed. If it is the third weave, the gain from triple weaving
+		// does not occur.)
 		const hasPhoenixSummon = weave.weaves
 			.some(event => event.ability.guid === ACTIONS.FIREBIRD_TRANCE.id)
-		if (hasPhoenixSummon) {
-			return super.isBadWeave(weave, PERMITTED_PHOENIX_WEAVES)
+		if (hasPhoenixSummon && weave.weaves.length === PERMITTED_PHOENIX_WEAVES) {
+			// Need to make sure we don't allow a triple weave after a hardcast
+			if (!maxWeaves) {
```
ackwell: this block of code here smells like core needs some tweaks to expose its own maxWeaves logic, duping up on that is a bit nasty

PR #663, opened 2019-10-05
fix: Use cast time when tracking shadowbite dmg

donb1991:
Fixing issue with Shadowbite hit detection. Was using the dmg timestamp which is going to be little different for each enemy hit. 

Fix I implemented was just use the last cast timestamp instead of the dmg event timestamp. I tried to write a test but ended up with quite a bit of set up for the events.

Example of log with issue
https://xivanalysis.com/analyse/BdkGL9728rXnWFgN/1/13/

### DIFF
--- a/src/parser/jobs/brd/modules/Sidewinder.js
+++ b/src/parser/jobs/brd/modules/Sidewinder.js
@@ -46,6 +46,7 @@ export default class Sidewinder extends Module {
 	_singleTargetShadowbitesPotencyLoss = 0
 
 	_badCasts = []
+	_lastShadowbiteCast = 0
 	_shadowbiteDamageTimestamps = new Map()
 
 	constructor(...args) {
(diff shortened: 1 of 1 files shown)


src/parser/jobs/brd/modules/Sidewinder.js
```diff
@@ -56,6 +57,11 @@ export default class Sidewinder extends Module {
 			abilityId: ACTIONS.SIDEWINDER.id,
 		}, this._onSidewinderCast)
 
+		this.addHook('cast', {
+			by: 'player',
+			abilityId: ACTIONS.SHADOWBITE.id,
+		}, this._onShadowbiteCast)
+
```
AzariahAeroborn: A simpler way to get the same result, and not have to track last Shadowbite cast time, would be to use the FFLogsEvents compatibilty shim instead, like core entities does: https://github.com/xivanalysis/xivanalysis/blob/8035dedd6b4361ac4fabee31a687c116cd3d9cf8/src/parser/core/modules/Entities.js#L119

This would have two advantages:
- Won't mysteriously break if the cast events are missing or if their timing changes relative to damage events
- Doesn't require hooking two sets of actions for each Shadowbite cast, so easier to maintain going forward.

donb1991: Thanks for the feedback. I went ahead and use the fflog event. Let me know if you want me to squash the commits into one or not.

PR #670, opened 2019-10-07
WIP: new invuln and AE

rexatlanti:
This PR will use the new targetability updates to track downtime better.
Current issues:
circular dependency between Invuln -> AE -> Entities -> Invuln
Possible solution is to remove `getStatusUptime` from `Entities` and and put it into `Status`. Since that is the sole `invuln` dependency in the whole module. Cons: Every statusUptime use needs to depend on Status then instead of Entities.


ackwell:
Note to self; disable stalebot.

ackwell:
🎉

PR #672, opened 2019-10-08
Add korean patch date for 5.0

Ra-Workspace:


ackwell:
Ah, thanks for that.
Can you do me a favour and update the `Shadowbringers` patch date on the `stormblood` branch as well ([link to file](https://github.com/xivanalysis/xivanalysis/blob/stormblood/src/data/PATCHES.ts#L157))? Most patches don't need updating cross-branch, but xpacs need representation on all active branches.

Ra-Workspace:
I have done #673  😀

PR #683, opened 2019-10-23
BLM - Cycle optimization & downtime enhancements

AkairyuGestalter:
- Allow the Astral Fire cycle optimization where, with enough speed (Triple and/or Leylines generally required), the player can skip Blizzard 4 and Fire 1, and just do 5 Fire 4s and a Despair as their Astral Fire phase
- Stop dinging Enochian drops where it occurred during an UnableToAct window longer than the duration of the Astral Fire/Umbral Ice buff. I'm looking at you E1S.

### DIFF
--- a/src/parser/jobs/blm/modules/Gauge.js
+++ b/src/parser/jobs/blm/modules/Gauge.js
@@ -31,6 +31,7 @@ export default class Gauge extends Module {
 		'precastAction', // eslint-disable-line @xivanalysis/no-unused-dependencies
 		'suggestions',
 		'brokenLog',
+		'unableToAct',
 	]
 
 	_astralFireStacks = 0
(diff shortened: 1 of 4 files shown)


src/parser/jobs/blm/modules/Gauge.js
```diff
@@ -31,6 +31,7 @@ export default class Gauge extends Module {
 		'precastAction', // eslint-disable-line @xivanalysis/no-unused-dependencies
 		'suggestions',
 		'brokenLog',
+		'unableToAct',
```
ackwell: Any reason for the preference on `unableToAct`? `downtime` collates the data from both it and `invuln`, the latter of which is going to (eventually) be the one that reads downtime events

AkairyuGestalter: Umbral Soul can still be used to maintain AF/UI when the boss is invulnerable since it doesn't require a target, so we need to specifically look when when you couldn't do anything at all in order to say that dropping Enochian was ok.

ackwell: Ah, of course.

PR #684, opened 2019-10-24
Make time linear again

ackwell:
Or at least _less_ wibbly wobbly timey wimey.

Moved hooking logic out of `Module` into a new single-ish-ton `Dispatcher`. Filter QoL remains in modules to keep the dispatcher simple(r).


ackwell:
I've done a good comb over this, should be good to go. Aiming to merge tomorrowish.

PR #685, opened 2019-10-26
Mark WAR as supported through 5.08

qaphla:
This PR marks WAR as supported for 5.0-5.08 and updates the TODO list for WAR.

I haven't been able to dig up any bugs in the WAR modules that aren't core and equally present in other jobs that are marked as supported, and so it seems like there's no current reason to not mark WAR as supported. Some features (listed in the new TODO) would be nice to have, but again shouldn't be blocking of support.

### DIFF
--- a/src/parser/jobs/war/TODO.md
+++ b/src/parser/jobs/war/TODO.md
@@ -1,0 +1,15 @@
+# WAR TODOs
+
+## Known Bugs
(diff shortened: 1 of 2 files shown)


ayyaruq:
Did you talk to @NamelessFaceless about this?

Re Upheaval and IR, the core cooldowns and/or buffwindow module can probably handle those. Built in expected GCDs/CDs and drift/usage checks.

PR #691, opened 2019-10-27
Expand DRG Analysis

ebshimizu:
This is a large series of changes to the DRG module that adds a number of more technical analysis sections for DRG optimization. The modifications involve things about my own play that I was curious about, and follow most of the recommendations in [this DRG guide](https://docs.google.com/document/d/1ua5uewclVNvIaVJ44eX3F3-y3QLKcn2z8R8d9Vvj5B0/preview#heading=h.p8z7j5k8sxqk). I've added what I want to see, so I'd like to just kinda throw this over to the main repository to see if any of this is generally useful.

The modifications are as follows:
- New Module: Jumps. Tracks the number of jumps used and compares them against the theoretical maximum number of jumps in a fight. Reports a table consisting of how far your casts drifted and how much faster you'd have to be to meet the theoretical maximum. Sometimes this module will report extremely tight timings, as the fight might end right as a jump comes off cooldown. Added this report to the checklist and suggestions sections as well.
- New Module: Positionals. Tracks the number of Raiden Thrust procs you get during a fight, and reports your True North usage. Is able to determine how many extra procs you could get if you used True North when it was off cooldown. Added a suggestion for this module (it's currently very lenient).
- Adjustment to Module: Buffs. Computes the number of full personal buff (Lance Charge, Dragon Sight) windows you can get in the fight and reports if you could've fit more buffs in. Re-formatted the accordion into a table that includes recast drift. Allowed partial buff windows at the end of the fight (it's optimal to use there even if not getting full duration), and marked these windows with a *.
- Adjustment to Module: Blood/Life of the Dragon. Reformatted accordion into a table showing which personal buffs were active during Nastrond and Stardiver. 
- Experimental Adjustment: Life of the Dragon. An attempt was made to determine if a Life of the Dragon window should've been delayed to wait for buffs, according to the rules in [here (p. 17)](https://docs.google.com/document/d/1ua5uewclVNvIaVJ44eX3F3-y3QLKcn2z8R8d9Vvj5B0/preview#heading=h.p8z7j5k8sxqk). Whether or not this is possible depends heavily on the fight, and this only outputs a message here instead of in suggestions.

A note about formatting: I think my Prettier instance in VS code was maybe a bit aggressive about making formatting changes? Can certainly reformat if needed and/or split this into smaller PRs.


ebshimizu:
Makes sense, can do a reformatting pass this week.

Would it be easier for you guys if this PR were closed but the branch kept up, and a cleaned-up PR submitted instead? What I'll probably do is just copy over my changes in a clean copy of the modified files, and I'll do a pass on my new files to better match the existing style.

ackwell:
Doesn't really change much our side, do whatever is easiest for you within git. If you're able to split the feature branches up that'd be really great, though.

ebshimizu:
Ok, fixed up the modified files. Should be easier to diff those now. I'll fix up the new ones next.

It looks like I ended up doing some code re-use from one of the newly added files (a formatting utility to add a button to jump to a location in the timeline), so it could be split with a little work. I think I can split the LotD feature I marked as experimental off into a different branch though, as I'm the least confident about that one.

Toastdeib:
I'll look over this in the next couple of days, although I can't promise super quick turnaround time because I'm going to be pretty busy overhauling NIN to, well, work after 5.1 drops. Also as an aside, are you on our Discord server yet, @ebshimizu? If not you should jump on there, it's generally a good idea to discuss changes to existing modules (especially when they're this sizable) with the relevant job maintainer(s) before you go diving into it.

src/parser/jobs/drg/modules/Positionals.js
```diff
@@ -0,0 +1,244 @@
...
+				this._nextChargeIn += TRUE_NORTH_CD
+			} else {
+				// otherwise we're good, no recharge happening
+				this._nextChargeIn = null
+			}
+		}
+
+		// if we're over the max, uhhhhhhhhhh panic?
+		if (this._tnCharges > TRUE_NORTH_CHARGES) {
+			console.log('DRG Error: True north charges exceeded limit')
```
Toastdeib: This should be removed before it gets merged; we try to keep all `console.log()`s out of production. Same for the one a few lines down. I would recommend just bounding the charge count within [0,2] in this and the function below, rather than allowing them to get into weird states and logging.

src/parser/jobs/drg/modules/Positionals.js
```diff
@@ -0,0 +1,244 @@
...
+		'combatants',
+		'suggestions',
+	]
+
+	// tracking raiden thrust procs
+	_rtCombos = []
+	_currentCombo = null
+
+	// true north tracking
+	_tnCharges = 2
```
Toastdeib: I've been rolling ideas around for a while now for how to implement a core solution for charge tracking. For the time being, doing it manually here is fine, but down the line it may be something to refactor.

src/parser/jobs/drg/modules/index.js
```diff
@@ -6,6 +6,9 @@ import Debuffs from './Debuffs'
 import Procs from './Procs'
 //import RotationWatchdog from './RotationWatchdog'
 import Weaving from './Weaving'
+import Jumps from './Jumps'
+import Positionals from './Positionals'
```
Toastdeib: These should also be kept alphabetical, both in the import list and in the export array.

src/parser/jobs/drg/modules/Positionals.js
```diff
@@ -0,0 +1,244 @@
...
+					{combo.trueNorthCast2 ? <Icon color="green" name="check" /> : ''}
+				</Table.Cell>
+				<Table.Cell textAlign="center">
+					{this._checkIcon(combo.trueNorthCharges > 0, '')} ({combo.trueNorthCharges} <Plural	value={combo.trueNorthCharges} one="charge" other="charges"	/>)
+				</Table.Cell>
+			</Table.Row>
+		})
+	}
+
+	output() {
```
Toastdeib: I'm happy to debate this one, but I don't think adding an output table for Raiden Thrust will provide as much value as it does for the other modules where we currently do it, mostly because of the oversaturation of data. Raiden Thrust (ideally) gets used every rotation, which should be something like 23s long, give or take, so you're going to have 25-30 rows for an average fight. I'm going to pull down this branch later and see how it actually looks just in case that changes my mind, but it's worth considering the risk of overwhelming a user with too much information.

src/parser/jobs/drg/modules/BloodOfTheDragon.js
```diff
@@ -213,20 +233,95 @@ export default class BloodOfTheDragon extends Module {
...
+			if (a.event.timestamp > b.event.timestamp) {
+				return 1
+			}
+
+			return 0
+		})
+
+		const rows = casts.map(cast => {
+			const buffs = cast.buffs.map(id => {
+				return <StatusLink key={id}	showName={false} iconSize="35px" {...getDataBy(STATUSES, 'id', id)}	/>
```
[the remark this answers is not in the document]
ebshimizu: just browsing through but this one has a quick answer: the tabs inside strings was definitely leftover from my reformatting pass. I saw a few of them last night when I pushed that, but I'm sure I missed a few instances of this.

src/parser/jobs/drg/modules/Positionals.js
ebshimizu: I ended up debating this for a while myself. I put the full table there just to see how it looked, but it can be easily filtered down to missed procs only. It is kinda interesting to see when and where TN ended up being used over the course of the fight, but is indeed quite a bit of info.

src/parser/jobs/drg/modules/Jumps.js
```diff
@@ -0,0 +1,295 @@
...
+import Module from 'parser/core/Module'
+import ACTIONS from 'data/ACTIONS'
+import STATUSES from 'data/STATUSES'
+import {ActionLink, StatusLink} from 'components/ui/DbLink'
+import {Rule, Requirement} from 'parser/core/modules/Checklist'
+import {getDataBy} from 'data'
+import DISPLAY_ORDER from './DISPLAY_ORDER'
+import {TieredSuggestion, SEVERITY} from 'parser/core/modules/Suggestions'
+
+export default class Jumps extends Module {
```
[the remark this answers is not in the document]
ebshimizu: ok right I had a chance to quick test this just now, and gonna make some notes before I forget.

I did know about the module, but wasn't convinced it was 100% accurate, at least for [my own report](https://www.fflogs.com/reports/TZL6cVGXFwxDaRry/#fight=28). The OGCD module reports 100% usage for high jump, while I for sure miss at least 1 jump at the back half here (there's easily 42s of downtime). It's worth noting that the max jump calculation here only accounts for boss invulnerabilities right now, not downtime (single line change to swap to check if that's the difference). I'm in the discord now, so we can definitely chat more about this point there.

src/parser/jobs/drg/modules/BloodOfTheDragon.js
```diff
@@ -120,6 +126,12 @@ export default class BloodOfTheDragon extends Module {
 				duration: this._lifeDuration,
 				nastronds: [],
 				stardivers: [],
+				timeToNextBuff: {
+					[ACTIONS.LANCE_CHARGE.id]: this.cooldowns.getCooldownRemaining(ACTIONS.LANCE_CHARGE.id),
+					[ACTIONS.DRAGON_SIGHT.id]: this.cooldowns.getCooldownRemaining(ACTIONS.DRAGON_SIGHT.id),
+					[ACTIONS.BATTLE_LITANY.id]: this.cooldowns.getCooldownRemaining(ACTIONS.BATTLE_LITANY.id),
+				},
+				activeBuffs: this.jumps.getActiveDrgBuffs(),
```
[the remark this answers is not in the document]
ebshimizu: I ended up moving `getActiveDrgBuffs()` into, well, `Buffs` for now, along with the timeline button creation function.

ebshimizu:
splitting this into smaller PRs, going to go ahead and close this one.

PR #693, opened 2019-10-28
Small data changes to get DRG and MCH updated for 5.1

Toastdeib:
Figured I'd throw this one into the pot real fast before I work on the monstrosity that will be the NIN changes.


Toastdeib:
Mentioned it on Discord but just so it's on record - MCH has no modules that worry about AoE at all, let alone looking at potency, so it should be fine.

PR #700, opened 2019-11-01
[SAM] 5.1 update to pre-existing modules/data

Rjb99:
This PR updates the following:

Potency changes:
Yukikaze (400->360)
Jinpu & Shifu (340->320)
Shoha (fixed to 400)

Cooldown changes:
Shoha (set to 15)

Other changes:
Hagakure conversion ratio increased (5 -> 10 per sen)

Features to be added for 5.1 not in this PR in order to consider the PR supported for the patch:
-Shoha tracking with visual graph of stacks
-Adding ishikoten and Guren/Senei to OGCD tracker
-Adjusting Meikyo to include Iaijutsu again


suShirow:
The Shoha action didn't get a potency update but not sure if that matters

src/parser/jobs/sam/modules/Sen.tsx
```diff
@@ -37,7 +37,7 @@ const TSUBAME = [
 	ACTIONS.KAESHI_SETSUGEKKA.id,
 ]
 
-const KENKI_PER_SEN = 5
+const KENKI_PER_SEN = 10
```
ackwell: This will probably mean your support range starts at 5.1 - don't forget to update in the follow up!

PR #703, opened 2019-11-04
general updates to 5.1.  manafication fix

Cephid:
basic updates for 5.1.

### DIFF
--- a/src/data/ACTIONS/RDM.js
+++ b/src/data/ACTIONS/RDM.js
@@ -238,7 +238,7 @@ export default {
 		onGcd: true,
 		cooldown: 2.5,
 		castTime: 5,
-		potency: 300, //Increase Black by 11
+		potency: 350, //Increase Black by 11
 		breaksCombo: true,
 	},
 	VERTHUNDER_II: {
@@ -258,7 +258,7 @@ export default {
 		onGcd: true,
 		cooldown: 2.5,
 		castTime: 2,
-		potency: 350, //Increase Black by 9
+		potency: 300, //Increase Black by 9
 		breaksCombo: true,
 	},
 	VERAERO: {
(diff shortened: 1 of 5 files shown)


src/data/ACTIONS/RDM.js
```diff
@@ -258,7 +258,7 @@ export default {
 		onGcd: true,
 		cooldown: 2.5,
 		castTime: 2,
-		potency: 270, //Increase Black by 9
+		potency: 350, //Increase Black by 9
```
Toastdeib: Dunno if it matters for analysis, but for correctness - per garland, it looks like you got Fire/Stone and Aero/Thunder flipped, potency-wise (Fire/Stone should be 300, Aero/Thunder should be 350).

src/parser/jobs/rdm/modules/Gauge.js
```diff
@@ -247,7 +247,7 @@ export default class Gauge extends Module {
 			this._blackManaLostToInvulnerable += gaugeAction.mana.black.invulnLoss
 
 			this._whiteManaLostToManafication += gaugeAction.mana.white.manaficationLoss
-			this._blackManaWastedToManafication += gaugeAction.mana.black.manaficationLoss
+			this._blackManaLostToManafication += gaugeAction.mana.black.manaficationLoss
```
Toastdeib: RIP.

src/parser/jobs/rdm/index.js
```diff
@@ -21,7 +21,7 @@ export default new Meta({
 	</>,
 	supportedPatches: {
 		from: '5.05',
-		to: '5.08',
+		to: '5.1',
```
ackwell: are these changes gonna work on the old patch?

i know there's not much tooling for multiple patches worth of the DATA file, but it still has an effect.

Cephid: Acceleration uptime would be incorrect for older patches.  I didn't think about the from patch list.  I'll update it to 5.1 to 5.1

src/data/ACTIONS/RDM.js
Cephid: Indeed I did, I will correct it.

PR #707, opened 2019-11-05
Antman meditates on the GCD, feat. Samurai

ayyaruq:
Prob relevant to @Rjb99 too, this requires 5.1 so worth noting when we flag the jobs supported


Rjb99:
this is fine.

PR #708, opened 2019-11-05
5.1 NIN updates

Toastdeib:
Bit big, bit spicy, I apologize for that. But that's what happens when your job gets deleted and readded, I guess. ¯\\\_(ツ)_/¯

It's mostly small shit at least. The Ninjutsu and Ninki modules are the only ones that got significant changes.


src/data/ACTIONS/NIN.js
```diff
@@ -31,122 +31,199 @@ export default {
...
 
 	NINJUTSU: {
 		id: 2260,
 		name: 'Ninjutsu',
 		icon: 'https://xivapi.com/i/002000/002904.png',
-		onGcd: false,
+		onGcd: true,
 		cooldown: 20,
+		gcdRecast: 1.5,
 	},
```
KelosPelmand: nit: Ninjutsu (any of them) do not technically have a cooldown anymore. It was moved to the mudras.

src/data/ACTIONS/NIN.js
```diff
@@ -31,122 +31,199 @@ export default {
...
+		gcdRecast: 1.5,
+	},
+
+	FUMA_SHURIKEN_TCJ: {
+		id: 18873,
+		name: 'Fuma Shuriken',
+		icon: 'https://xivapi.com/i/002000/002907.png',
+		onGcd: true,
+		cooldown: 20,
+		gcdRecast: 1.5,
```
KelosPelmand: TCJ step 1 and 2 actually have 1.0 gcdRecast (surprised me when i checked it).  Also, like ninjutsu, technically no cooldown.

src/parser/jobs/nin/modules/Ninjutsu.js
```diff
@@ -8,71 +8,71 @@ import Module from 'parser/core/Module'
...
-		this.addHook('cast', {by: 'player', abilityId: ACTIONS.HYOTON.id}, this._onHyotonCast)
-		this.addHook('cast', {by: 'player', abilityId: ACTIONS.RABBIT_MEDIUM.id}, this._onRabbitCast)
-		this.addHook('cast', {by: 'player', abilityId: ACTIONS.DOTON.id}, this._onDotonCast)
+
+		this.addHook('cast', {by: 'player', abilityId: [ACTIONS.HYOTON.id, ACTIONS.HYOTON_TCJ.id]}, () => { this._hyotonCount++ })
+		this.addHook('cast', {by: 'player', abilityId: ACTIONS.RABBIT_MEDIUM.id}, () => { this._rabbitCount++ })
+		this.addHook('cast', {by: 'player', abilityId: [ACTIONS.DOTON.id, ACTIONS.DOTON_TCJ.id]}, this._onDotonCast)
 		this.addHook('aoedamage', {by: 'player', abilityId: STATUSES.DOTON.id}, this._onDotonDamage)
+		this.addHook('removebuff', {by: 'player', abilityId: STATUSES.DOTON.id}, this._finishDotonWindow)
 		this.addHook('complete', this._onComplete)
```
KelosPelmand: Could probably use a check that Katon or Gokka Mekkakyu actually hit multiple targets.

src/parser/jobs/nin/modules/Ninki.js
```diff
@@ -69,9 +81,9 @@ export default class Ninki extends Module {
...
 		super(...args)
-		this.addHook('cast', {by: 'player', abilityId: NINKI_GCDS}, event => this._addNinki(event, GCD_NINKI_GAIN))
-		this.addHook('combo', {by: 'player', abilityId: NINKI_COMBOS}, event => this._addNinki(event, GCD_NINKI_GAIN))
-		this.addHook('cast', {by: 'player', abilityId: NINKI_OGCDS}, event => this._addNinki(event, OGCD_NINKI_GAIN))
+		this.addHook('cast', {by: 'player', abilityId: NINKI_GCDS}, event => this._addNinki(event, GCD_NINKI_GAIN[event.ability.guid]))
+		this.addHook('combo', {by: 'player', abilityId: NINKI_COMBOS}, event => this._addNinki(event, GCD_NINKI_GAIN[event.ability.guid]))
+		this.addHook('cast', {by: 'player', abilityId: NINKI_OGCDS}, event => this._addNinki(event, OGCD_NINKI_GAIN[event.ability.guid]))
 		this.addHook('cast', {by: 'pet'}, event => this._addNinki(event, BUNSHIN_NINKI_GAIN))
 		this.addHook('cast', {by: 'player', abilityId: NINKI_SPENDERS}, this._onSpenderCast)
 		this.addHook('aoedamage', {by: 'player', abilityId: ACTIONS.HELLFROG_MEDIUM.id}, this._onHellfrogAoe)
```
KelosPelmand: Might be useful to check that Bunshin is cast over Bhava/Hellfrog when it is up.  (Not sure if there are cases where Bunshin should be held or not.)

src/parser/jobs/nin/modules/TrickAttackUsage.js
```diff
@@ -8,7 +8,7 @@ import Module from 'parser/core/Module'
 import {Suggestion, TieredSuggestion, SEVERITY} from 'parser/core/modules/Suggestions'
 
 const TA_COOLDOWN_MILLIS = ACTIONS.TRICK_ATTACK.cooldown * 1000
-const OPTIMAL_GCD_COUNT = 3 // Number of GCDs prior to the first TA in the opener
+const OPTIMAL_GCD_COUNT = 5 // Opener should be Suiton > AE combo > SE before Trick
```
KelosPelmand: Seems like this is a little looser of a check that it could be.  Opening with TCJ before Trick would mess this count up (besides being bad for other reasons).

src/parser/jobs/nin/modules/TrickAttackUsage.js
```diff
@@ -78,7 +80,7 @@ export default class TrickAttackUsage extends Module {
 			this.suggestions.add(new TieredSuggestion({
 				icon: ACTIONS.TRICK_ATTACK.icon,
 				content: <Trans id="nin.ta-usage.suggestions.opener.content">
-					Avoid unconventional timings for your first <ActionLink {...ACTIONS.TRICK_ATTACK}/> of the fight in order to line it up with all the other raid and personal buffs. In most openers, Trick Attack should be weaved in approximately 10 seconds into the fight.
+					Avoid unconventional timings for your first <ActionLink {...ACTIONS.TRICK_ATTACK}/> of the fight in order to line it up with all the other raid and personal buffs. In most openers, Trick Attack should be weaved in approximately 8-9 seconds into the fight.
```
KelosPelmand: The suggestion implies you are checking based on time in seconds, but you are actually checking gcd counts (also see abvoe comment).  Various things like gcd drag, using tcj, or using non-2-part ninjutsus will through the timings off.

src/data/ACTIONS/NIN.js
Toastdeib: Valid point - I'll correct the data.

src/data/ACTIONS/NIN.js
Toastdeib: Also valid; I'll correct the data here as well.

src/parser/jobs/nin/modules/Ninjutsu.js
Toastdeib: I'm going to put that on the "do it in a future PR list". I'm more concerned with just correcting what already exists for this one.

src/parser/jobs/nin/modules/Ninki.js
Toastdeib: Also putting this on the "do in the future" list for the same reasons as the Katon/Goka check you suggested.

src/parser/jobs/nin/modules/TrickAttackUsage.js
Toastdeib: While I don't love the implementation I did for this, I'm inclined to leave it as-is for now and revisit it later, since it's been this way since the initial implementation. Plus, 5th should be correct given the opener I was given/am using (Suiton > SE > GS > AE > SE before you weave Trick). Just checking the count instead of the exact GCDs used may result in false negatives, but I think it should be fine at least for now.

src/parser/jobs/nin/modules/TrickAttackUsage.js
Toastdeib: Same as above, going to leave this for now and revisit down the line. The reason I opted for time in the suggestion text is because it felt more understandable, particularly with regards to aligning it with other raid buffs that are going to at x point in the fight. It may be worth changing to show a GCD count instead, but I'd rather have that discussion outside the scope of this PR and just have this one be for correcting existing logic without changing it significantly (at least, outside of the places where significant changes are necessary).

Toastdeib:
To address your additional points:

> Kassatsu is not flagging for uses of Fuma or Raiton under Kassatsu

A fair point; I'm going to say it's outside the scope of this PR, as it's been inaccurate since the start of ShB (since Hyosho and Goka have been things this whole time), but it's going on my list.

> Refunding gcd uptime during TCJ is probably not needed any more.

Also a fair point; since Ninjutsu are now on the GCD, this is actually probably inflating the user's uptime. I'll remove GlobalCooldown.js entirely since it's no longer needed.

> There seems to be very little checking that cooldowns are actually used (core CooldownDowntime may be useful here)

Definitely outside the scope of this PR, but I'll add it to the end of my list.

> Shadow Fang isn't tracked at all (though the comment in index.js implies this is a known issue you are coming back to. the cooldown scaling with sks/huton will be troublesome)
> DWaD and Assassinate check that any usages that exist are in a TA, but don't flag if they are not used at all

I'm going to be redoing TrickAttackWindow.js as an extension of BuffWindow in a future PR, so that will cover both of these points plus Kassatsu (implicitly by checking for uses of Hyosho Ranryu).

> Kassatsu, Ninjutsu, and TCJ don't seem to check that you are not losing usages that I could see

At present they aren't. That said, Kassatsu will be covered by TrickAttackWindow.js after the revamp, TCJ will be easy enough to roll into a CooldownDowntime module in the future, and Ninjutsu... is a problem that I've wanted a good solution for for a while. And now it's more complicated because charges. But I'll work on it.

>Checks for what should appear in TA windows might need revisited (BuffWindow may be useful here)

See above.

> No check that all Bunshin stacks are spent when used.

Outside the scope of this PR, but it's going on the list.

> nit: Bhavacakra and Hellfrog technically share a cooldown and could be considered a cooldown group

I mean it's a negligible cooldown, but technically true. I'm going to gloss over it for now because I don't understand how cooldown groups work but that can go on my list too.

KelosPelmand:
updates look good.  

cooldown groups show up in two places.  one is Cooldowns.js for the timeline (old mudras were an example of how this worked).  the other is CooldownDowntime, which bahava/hellfrog wouldn't even be in.  (For this, you would edit the data files.  SMN egi assaults would be one example for this.)  I wouldn't be sad to see it never put in for these (hence it being a nit), since it would just combine the timeline lines together.

PR #711, opened 2019-11-08
Finished SMN 5.1 support

KelosPelmand:


src/parser/jobs/smn/modules/GeneralCDDowntime.js
```diff
@@ -1,9 +1,11 @@
 import ACTIONS from 'data/ACTIONS'
 import CooldownDowntime from 'parser/core/modules/CooldownDowntime'
 
-//Although the trance cooldown is 55s, it is better in many cases
-//to hold for a 60s alignment, so some downtime per use must be allowed.
-const TRANCE_ALLOWED_DOWNTIME = 5000
+//The trance cooldowns are both 55s.  However, in order to maintain overall
+//alignment with the 1 minute buff cycle, FBT should be held 10s per use
+//(for a 55+65 = 120s cycle).  Due to the way cooldown tracking works in
+//XIVA, both must have the same allowed downtime since they are in a cooldown group.
```
ackwell: sounds like an issue that needs to be resolved.

this is fine for now, but it'd be good to get the underlying issue resolved such that it's not required

src/parser/jobs/smn/modules/Ruin2.js
```diff
@@ -115,42 +102,23 @@ export default class Ruin2 extends Module {
...
+				Avoid casting <ActionLink {...ACTIONS.SMN_RUIN_II}/> if possible.
+				Prefer using your other instant casts for movement and weaving or Ruin III when movement and weaving are not required.
 			</Trans>,
-			why: <Trans id="smn.ruin-ii.suggestions.warnings.why">
-				{warnings * potLossPerR2} potency lost to {warnings} Ruin II or Ruin IV
-				<Plural value={warnings} one="cast" other="casts"/>
-				used only to move.
+			why: <Trans id="smn.ruin-ii.suggestions.avoid.why">
+				{totalCount} Ruin II <Plural value={totalCount} one="cast" other="casts"/> were performed.
+				({weavedCount} for weaving or triggering Wyrmwave, {moveOnlyCount} for only movement, and {wastedCount} for no benefit.)
```
ackwell: Not particularly on-board with this, it's an incredibly long `why`. If it can't be simplified, it's probably a sign that the single suggestion is relying on too many individual details.

In this case, I'd err towards just simplifying.

KelosPelmand: I've shortened this to just the total count.  The other counts were there to give extra data to let the user know "how bad" the ruin 2 casts were.  I had considered outputting it as a separate table below, but it doesn't feel like enough data for that.

src/parser/jobs/smn/modules/GeneralCDDowntime.js
KelosPelmand: Other options I considered around this:

- Under the proposed change to core cooldowns, I'd probably change to a 5s offset.  This would calculate as if they were used every 1 minute (which is the recommended timing for buff alignments).
- Not mark them as a group and give them both a "2 minute cooldown".  They aren't really a group in the typical sense (with normal groups, like RDM displacement/engagement, you can use only one of them all the time.), but they do trigger each other's cooldowns.  You would also need to give a firstUseOffset to Firebird Trance to account for needing to use Dreadwyrm Trance first.
- Don't use core cooldown logic, and make the custom usage calculation logic specifically for the trance to allow the 55/65 split.

ackwell: All sound like sane solutions to me, whichever is a go-ahead for you works.

PR #713, opened 2019-11-09
Rewrite CooldownDowntime to change how it computes actual and expected uses, update consumers as required

AzariahAeroborn:
Much thanks to @KelosPelmand for a ton of assistance with this, including most of the logic for calculateMaxUsages.

Goals of this effort:
- Change the computation to use actual counted uses (including pre-pull synthetic casts, if applicable)
- Change the expected uses computation to do a "look forward" comparison, starting at firstUseOffset for each skill and then moving forward by cooldown + allowedDowntime.  If an expected use occurs during detected downtime / invulnerable time, move it later until the boss is available to hit again.
- Display additional information, the actual number used and expected, in addition to the percentage for each skill

I intentionally did this as a "drop-in" replacement for the current module and updated all of the current consumers to work with the new signature, so that we don't wind up in an in-between state where some OGCD modules are updated and some aren't and it's even awkwarder to explain what's going on in support.

I have two main outstanding questions before this is really ready to merge:
1) Do the various variable names (especially for exposed properties) make sense?
2) In a long enough fight, or if a skill's allowedDowntime is relatively high compared to its cooldown, it becomes possible for a player to "exceed" the number of expected uses if they're not actually holding the skill as much as we allow for.  This is most notable with fairly short cooldowns.  Do we want to "clamp" this by increasing expected to match the number the player actually used, since they did actually accomplish the better-than-expected drift times?


src/parser/core/modules/CooldownDowntime.tsx
```diff
@@ -0,0 +1,264 @@
...
+import Checklist from './Checklist'
+import Downtime from './Downtime'
+
+const DEBUG_EXPECTED_CALCULATION = false && process.env.NODE_ENV !== 'production'
+
+interface CooldownReset {
+	/**
+	 * One or more skills that trigger a cooldown reset.
+	 */
+	actions: Action[],
```
ackwell: I'm guessing this is for stuff like WAR's gauge spenders dropping cd on infuriate? Being an array seems _kinda_ weird to me, but i guess that works.

src/parser/core/modules/CooldownDowntime.tsx
```diff
@@ -0,0 +1,264 @@
...
+	 */
+	// isAffectedBySpeed: boolean
+	/**
+	 * Any skills that deduct from charge time for this group.
+	 */
+	resetBy?: CooldownReset
+}
+
+export abstract class CooldownDowntime extends Module {
+	static handle = 'cooldowndowntime'
```
ackwell: niit

```suggestion
	static handle = 'cooldownDowntime'
```

src/parser/core/modules/CooldownDowntime.tsx
```diff
@@ -0,0 +1,264 @@
...
+	/**
+	 * Implementing modules MUST provide a list of tracked cooldowns
+	 */
+	protected abstract trackedCds: CooldownGroup[]
+
+	private usages = new Map<number, CastEvent[]>()
+	private resets = new Map<number, CastEvent[]>()
+
+	protected checklistName = <Trans id="core.cooldowndowntime.use-ogcd-cds">Use your OGCDs</Trans>
+	protected checklistDescription = <Trans id="core.cooldowndowntime.ogcd-cd-metric">Always make sure to use your OGCDs when they are up but don't clip them.</Trans>
```
ackwell: wording nits;
reads slightly better with a comma: "when they are up, but don't clip them."
possibly better again as "when they are available, but avoid clipping them."

src/parser/core/modules/CooldownDowntime.tsx
```diff
@@ -0,0 +1,264 @@
...
+		return true
+	}
+
+	protected init() {
+		const trackedIds = this.trackedCds.map(g => g.cooldowns)
+			.reduce((acc, cur) => acc.concat(cur))
+			.map(a => a.id)
+
+		const resetIds = this.trackedCds
+			.map(g => (g.resetBy === undefined) ? [] : g.resetBy.actions)
```
ackwell: If you merge a recent master in, you'll be able to simplify quite a few of these sort of expressions to something like (ts3.7)

```suggestion
			.map(group => group.resetBy?.actions ?? [])
```

same goes for a whole bunch of conditional checks below

src/parser/core/modules/CooldownDowntime.tsx
```diff
@@ -0,0 +1,264 @@
...
+	 * should not be counted as a usage of the cooldown.
+	 */
+	protected countUsage(event: CastEvent): boolean {
+		return true
+	}
+
+	protected init() {
+		const trackedIds = this.trackedCds.map(g => g.cooldowns)
+			.reduce((acc, cur) => acc.concat(cur))
+			.map(a => a.id)
```
ackwell: nit; I know there's lots of prior art floating around, but I'm trying to (slowly) cut down on single letter vars - esp. with the reduce in there, it took me a bit to realise this was an action. `action => action.id` is way more obvious with it's intent, despite the lack of beloved brevity.

src/parser/core/modules/CooldownDowntime.tsx
```diff
@@ -0,0 +1,264 @@
...
+			.map(g => (g.resetBy === undefined) ? [] : g.resetBy.actions)
+			.reduce((acc, cur) => acc.concat(cur))
+			.map(a => a.id)
+
+		this.addEventHook('cast', {by: 'player', abilityId: trackedIds}, this.onTrackedCast)
+		this.addEventHook('cast', {by: 'player', abilityId: resetIds}, this.onResetCast)
+		this.addEventHook('complete', this.onComplete)
+
+		this.trackedCds.forEach(g => {
+			this.usages.set(g.cooldowns[0].id, [])
```
ackwell: i realise it's a _little_ unintuitive, but JS sets/maps (not objects) keyed by "identity". Given that the `cooldowns` array is more or less guaranteed to be unique per group, and non-changing, you'd be able to forgo the `[0].id` jank and just use `group.cooldowns` (i think) (untested 😅)

image

src/parser/core/modules/CooldownDowntime.tsx
```diff
@@ -0,0 +1,264 @@
...
+			if (g.resetBy && g.resetBy.actions.find(a => a.id === event.ability.guid)) {
+				(this.resets.get(g.cooldowns[0].id) || [] ).push(event)
+			}
+		})
+	}
+
+	private onComplete() {
+		const cdRequirements = []
+		for (const cdGroup of this.trackedCds) {
+			const expected = this.calculateMaxUsages(cdGroup)
```
ackwell: So; correct me if I'm wrong here - the "forward look" mentioned in the PR description is being performed post-mortem? If so, I have no issue with it - the parse is already complete, so the "forward look" is more like a "backward look but not as backward as the last bit", which is more than fine.

src/parser/core/modules/CooldownDowntime.tsx
```diff
@@ -0,0 +1,264 @@
...
+
+		this.checklist.add(new Rule({
+			name: this.checklistName,
+			description: this.checklistDescription,
+			requirements: cdRequirements,
+			target: this.checklistTarget,
+		}))
+	}
+
+	private calculateMaxUsages(group: CooldownGroup): number {
```
ackwell: aaaaaaaaaaaaaaaaaaaaaaaaaaaaite my brain shut down when i got here, too late

---

note to akk, check the rest of the PR after this message

src/parser/core/modules/CooldownDowntime.tsx
AzariahAeroborn: Yeah, after the complete flag fires, we start at time 0 (or firstUseOffset) and then crawl forward through the fight to determine when the skill "should" have been used, to count those max usages.

src/parser/core/modules/CooldownDowntime.tsx
```diff
@@ -0,0 +1,264 @@
...
+	/**
+	 * One or more skills that trigger a cooldown reset.
+	 */
+	actions: Action[],
+	/**
+	 * The amount of time in ms that an action deducts from the remaining
+	 * cooldown of the affected cooldowns.
+	 */
+	refundAmount?: number,
+	fullReset?: boolean
```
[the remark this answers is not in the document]
AzariahAeroborn: Yeah, Kelos's original approach was to just use refundAmount and have full resets be the full cooldown as well... I went with the separate code path because I thought it was a little more clear to someone consuming CooldownDowntime, and because using the full cooldown length results in "wasted reset time" being incremented by some of the full reset time, which doesn't usually make sense for skills like Acceleration resetting RDM's move blocks.  Of course, we're not really doing anything sexy with wasted reset time yet, just counting it, so I'm fine with going to a single refundAmount on the interface.

src/parser/core/modules/CooldownDowntime.tsx
```diff
@@ -0,0 +1,264 @@
...
+					// TODO: time after window end before usage.  should it just be first use offset? depends on what else was delayed and state in rotation
+				}
+			}
+			this.debugLog('Total count for group ' + gRep.name + ' is ' + count + '. Total reset time lost is ' + this.parser.formatDuration(timeLost) + '.')
+
+			return count
+		}
+
+	private getDisplayTime(currentTime: number): string {
+			return this.parser.formatDuration(currentTime - this.parser.fight.start_time)
```
ackwell: this is basically a dupe of `this.parser.formatTimestamp(currentTime)`

---

indentation here and below got funky

src/parser/core/modules/CooldownDowntime.tsx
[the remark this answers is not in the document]
AzariahAeroborn: For now, I'm going to go ahead and go back to just using `refundAmount`, as no current consumers of CooldownDowntime have fullReset resetters anyway, and we're not actually doing anything with the wastage we're tracking.  If it becomes a future issue, we can tackle it then.

src/parser/core/modules/CooldownDowntime.tsx
```diff
@@ -0,0 +1,264 @@
...
+				currentTime += step
+				charges += 1
+
+				// apply resets that are found
+				while (gResets.length > 0 && gResets[0].timestamp < currentTime) {
+					const rs = gResets[0]
+					const previousTime = currentTime
+					if (fullReset) {
+						if (charges < maxCharges) {
+							// if not at max charges, the full reset adds a charge and resets the cooldown timer
```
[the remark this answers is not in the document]
AzariahAeroborn: Obsoleting this by removing fullResets altogether.  There were no modules using fullReset yet anyway (RDM does not use core CooldownDowntime for the move blocks)

src/parser/core/modules/CooldownDowntime.tsx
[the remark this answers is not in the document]
AzariahAeroborn: From looking through it, this kind of defies easy splitting, without switching to an object or some sort of class properties for the tracking variables within the function.  In my opinion, splitting it out would be more work than the benefit is worth.

PR #716, opened 2019-11-10
DRG: Life of the Dragon Formatting & Delay Prediction

ebshimizu:
* Buff Tracking: Tracks which personal buffs were active during Nastrond and Stardiver.
* Delay Prediction (Experimental): An attempt was made to determine if a Life of the Dragon window should've been delayed to wait for buffs, according to the rules in [here (p. 17)](https://docs.google.com/document/d/1ua5uewclVNvIaVJ44eX3F3-y3QLKcn2z8R8d9Vvj5B0/preview#heading=h.p8z7j5k8sxqk). Whether or not this is possible depends heavily on the fight, and this only outputs a message here instead of in suggestions as it's a super minor high-level optimization. 
* Formatting: Reformatted bullets into an accordion


src/parser/jobs/drg/modules/BloodOfTheDragon.js
```diff
@@ -54,6 +60,37 @@ export default class BloodOfTheDragon extends Module {
...
+		return active
+	}
+
+	createTimelineButton(timestamp) {
+		return <Button
+			circular
+			compact
+			icon="time"
+			size="small"
+			onClick={() => this.timeline.show(timestamp - this.parser.fight.start_time,	timestamp - this.parser.fight.start_time)}
```
Toastdeib: Pretty unimportant but you've still got a few of those errant tabs running around.

src/parser/jobs/drg/modules/BloodOfTheDragon.js
```diff
@@ -54,6 +60,37 @@ export default class BloodOfTheDragon extends Module {
 		this.addHook('complete', this._onComplete)
 	}
 
+	// duplicate code from other PRs
```
Toastdeib: I'm assuming this is going to get tidied up/consolidated once everything is merged?

src/parser/jobs/drg/modules/BloodOfTheDragon.js
```diff
@@ -213,20 +264,86 @@ export default class BloodOfTheDragon extends Module {
...
-					{this._lifeWindows.history.map(window => <li key={window.start}>
-						<strong>{this.parser.formatTimestamp(window.start)}</strong>:&nbsp;
-						<Trans id="drg.blood.windows.hits">
-							{this.parser.formatDuration(window.duration)} long, <Plural value={window.nastronds.length} one="# Nastrond" other="# Nastronds"/>, <Plural value={window.stardivers.length} one="# Stardiver" other="# Stardivers"/>
-						</Trans>
-					</li>)}
-				</ul>
+				<Message>
+					<Trans id="drg.blood.windows.preface">
+						Each of the bullets below represents a Life of the Dragon window, indicating when it started, how long it lasted, and how many window-restricted OGCDs it contained. Ideally, each window should contain a full three <ActionLink {...ACTIONS.NASTROND}/> casts and one <ActionLink {...ACTIONS.STARDIVER}/> cast.
```
Toastdeib: Might be worth rewording this from "Each of the bullets below..." now that it's a table and not bullets.

src/parser/jobs/drg/modules/BloodOfTheDragon.js
```diff
@@ -138,7 +181,11 @@ export default class BloodOfTheDragon extends Module {
 
 		if (!this._lifeWindows.current.nastronds.some(nastrond => nastrond.timestamp === event.timestamp)) {
 			// Dedupe Nastrond casts, since that can occasionally happen
-			this._lifeWindows.current.nastronds.push(event)
+			this._lifeWindows.current.nastronds.push({
+				...event,
```
Toastdeib: Certainly not necessary but since it looks like the only piece of data from the actual `event` blob you're using is the timestamp, it may actually be worth it to change `...event,` to `timestamp: event.timestamp,` and trim out all the excess so it's cleaner/more maintainable down the road. Same thing applies to the Stardiver bit ~20 lines down.

src/parser/jobs/drg/modules/BloodOfTheDragon.js
```diff
@@ -213,20 +264,86 @@ export default class BloodOfTheDragon extends Module {
...
+
+		for (const id in window.timeToNextBuff) {
+			buffsInDelayWindow[id] = window.timeToNextBuff[id] >= LOTD_BUFF_DELAY_MIN && window.timeToNextBuff[id] <= LOTD_BUFF_DELAY_MAX
+			couldBeDelayed = buffsInDelayWindow[id] || couldBeDelayed
+			canBeDelayed = window.timeToNextBuff[id] >= LOTD_BUFF_DELAY_MIN && canBeDelayed
+		}
+
+		const delayBuffs = Object.keys(buffsInDelayWindow).filter(id => buffsInDelayWindow[id]).map((id, idx) => {
+			const action = getDataBy(ACTIONS, 'id', parseInt(id))
+			return <Message.Item key={idx}><Trans id={`drg.blood.delay-buff.${id}`}><ActionLink {...action} /> in {this.parser.formatDuration(window.timeToNextBuff[id])}</Trans></Message.Item>
```
Toastdeib: I don't think you can do `<Trans>` IDs like this - it doesn't show up in the extract, at any rate. It should be fine to just use one string though (e.g. `<Trans id="drg.blood.delay-buff">`) - since the only part that's going to be translated is the word "in", you shouldn't need 3 different versions for the 3 different buffs.

src/parser/jobs/drg/modules/BloodOfTheDragon.js
ebshimizu: that's the plan, just marking it with a comment to remember which functions are duplicated.

src/parser/jobs/drg/modules/BloodOfTheDragon.js
ebshimizu: Hm yeah I dunno why I wanted to put the ID on that one. I'm not super familiar with how the `<Trans>` tags work and I think I did something similar in a different PR that I'll have to fix as well.

ebshimizu:
Formatting pass (really hoping all those tabs are gone lol). 

Picture of the re-formatted tables:
image

KelosPelmand:
My understanding was that LotD is now automatically 30s by level 80.  If so, do we need to list a duration with each window?

ebshimizu:
Yes, LotD is always 30s at level 80 so listing the duration is unnecessary.

Toastdeib:
I think this all looks good now, which means there's just one more thing that I forgot to mention in my earlier review pass - changelogs. You'll want to add yourself into the [contributors file](https://github.com/xivanalysis/xivanalysis/blob/master/src/data/CONTRIBUTORS.ts) (you should also add an avatar if you have one, ideally) and then update the [index.js](https://github.com/xivanalysis/xivanalysis/blob/master/src/parser/jobs/drg/index.js) file for DRG to add yourself to the contributor list and insert a new changelog entry (I've been inserting them at the top of the array but it sorts by date when displaying them anyway).

ebshimizu:
updated! Added myself as a dev for now. Would be happy to help out as a maintainer, but would need a bit more info on what to do between patches.

PR #717, opened 2019-11-10
DRG: Positional Tracking

ebshimizu:
New Module: Tracks the number of Raiden Thrust procs you get during a fight, and reports your True North usage. Is able to determine how many extra procs you could get if you used True North when it was off cooldown. Added a suggestion for this module (it's currently very lenient).

The charge tracking for this module is expected to be re-written once that capability is in the core modules.

At the moment, the entire positionals table is output. This can be filtered down to failed procs only.


src/parser/jobs/drg/modules/Positionals.js
```diff
@@ -0,0 +1,222 @@
...
+	}
+
+	_tnUsed(event) {
+		// remove charge
+		this._tnCharges -= 1
+
+		// if we're below 0, clamp
+		if (this._tnCharges < 0) {
+			this._tnCharges = 0
+		}
```
Toastdeib: If you want to shorten this, it can all be simplified down to `this._tnCharges = Math.max(this._tnCharges - 1, 0)`. That said, it might instead be worth throwing a broken log error if they somehow hit negative charges, because that should be physically impossible. Up to you really.

src/parser/jobs/drg/modules/Positionals.js
```diff
@@ -0,0 +1,222 @@
...
+	}
+
+	// called exactly when the timestamp resolves
+	_updateTnCharges() {
+		this._tnCharges += 1
+
+		if (this._tnCharges > TRUE_NORTH_CHARGES) {
+			// clamp to max
+			this._tnCharges = TRUE_NORTH_CHARGES
+		}
```
Toastdeib: Same thing here as in `_tnUsed` - you can simplify to `this._tnCharges = Math.min(this._tnCharges + 1, TRUE_NORTH_CHARGES)`, although the broken log may be less applicable here since charge regeneration is simulated, so it's more likely for this issue to be a result of buggy code than bad data.

src/parser/jobs/drg/modules/Positionals.js
```diff
@@ -0,0 +1,222 @@
...
+		this.suggestions.add(
+			new TieredSuggestion({
+				icon: ACTIONS.RAIDEN_THRUST.icon,
+				content: <Trans id="drg.positionals.suggestions.content">Performing the proper positionals for	<ActionLink {...ACTIONS.FANG_AND_CLAW} /> and <ActionLink {...ACTIONS.WHEELING_THRUST} /> will do more damage and	let you use <ActionLink {...ACTIONS.RAIDEN_THRUST} />. Try to land these positionals, and use <ActionLink {...ACTIONS.TRUE_NORTH} /> if	it's available.	</Trans>,
+				tiers: {
+					1: SEVERITY.MINOR,
+					10: SEVERITY.MEDIUM,
+				},
+				value: missed,
+				why: <Trans id="drg.positionals.suggestions.why">You missed {missed} <Plural value={missed} one="proc" other="procs" /> of Raiden Thrust.</Trans>,
```
Toastdeib: You should move the value into the `<Plural>` tag here with `#` as well:
```js
<Trans id="drg.positionals.suggestions.why">You missed <Plural value={missed} one="# proc" other="# procs" /> of Raiden Thrust.</Trans>
```
This also applies to any other `<Plural>` tags I may have missed. It makes it easier for localization, since the standard order of "[quantity] [noun]" doesn't necessarily apply in other languages.

src/parser/jobs/drg/modules/Positionals.js
```diff
@@ -0,0 +1,222 @@
...
+			content={this.parser.formatTimestamp(timestamp)}
+		/>
+	}
+	// end duplicate code
+
+	_onGCD(event) {
+		if (this._currentCombo) {
+			// did we use the proc? we might not due to invuln windows
+			if (event.ability.guid === ACTIONS.RAIDEN_THRUST.id) {
+				this._rtCombos.used = true
```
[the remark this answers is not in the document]
ebshimizu: This is checking if there's an active combo window open, so it's checking the existence of the object. I can explicitly set this to a check against `null`. The usage flag is set after checking to see if the window exists to see if we actually get a chance to use Raiden (might not due to invuln, but we still did the positional right).

src/parser/jobs/drg/modules/Positionals.js
ebshimizu: I'd be ok with throwing/showing an error. Is there a standard way to do that here, or should I just set a flag and display a message in the output section?

ackwell: You can depend on the [`brokenLog`](https://github.com/xivanalysis/xivanalysis/blob/master/src/parser/core/modules/BrokenLog.tsx) module, and use it's API to flag an issue. The rest is handled by core.

ebshimizu:
Did a pass on the formatting and dead code stuff.
Also now looking at the comment about checking `this._currentCombo.used`, I realize that I don't actually do anything with the info that we used/didn't use raiden thrust? This module is just concerned with whether or not you proc'd it, as the assumption is that you'd use it anyway with a correct rotation, so I think I can just remove that flag unless someone comes up with a reason for me to keep it in.

Also, since I didn't do it in the original post, some pictures for this one.

Full table: 
image
Filtered table: 
image

src/parser/jobs/drg/modules/Positionals.js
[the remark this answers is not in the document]
Toastdeib: I meant the `this._rtCombos.used = true` line. The `if` 3 lines up is fine, but `_rtCombos` is an array and doesn't have a `used` property on it.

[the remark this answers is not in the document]
ebshimizu: oh! yeah wow that should be causing problems, but it's not? I don't actually see much use in tracking if you landed RT so I'll fix that up.

ebshimizu:
oh uhhhhh yeah finally got around to marking those changes off as complete. did a bit of formatting, updated this branch to be even with the upstream, added a changelog entry. Also decided to go with the filtered table (not the full list of positionals).

src/parser/jobs/drg/modules/Positionals.js
```diff
@@ -0,0 +1,211 @@
...
+
+		return <Fragment>
+			<Message>
+				<Trans id="drg.positionals.analysis.message">Being at the rear when using <ActionLink {...ACTIONS.WHEELING_THRUST} /> or on the flank when using <ActionLink {...ACTIONS.FANG_AND_CLAW} /> will allow you to use <ActionLink {...ACTIONS.RAIDEN_THRUST} /> instead of <ActionLink {...ACTIONS.TRUE_THRUST} />. You should be trying to proc this ability as much as possible, relying on <ActionLink {...ACTIONS.TRUE_NORTH} /> in situations where you cannot reach the proper position. The table below displays missed positionals (if any), and whether or not <ActionLink {...ACTIONS.TRUE_NORTH} /> was available to use.</Trans>
+			</Message>
+			<Message info>
+				<Trans id="drg.positionals.analysis.missed"><Icon name="info" /> You missed <strong>{missed}</strong> of <strong>{this._rtCombos.length}</strong> possible <ActionLink {...ACTIONS.RAIDEN_THRUST} /> procs.
+					{missed > 0 ? (
+						<>
+							Of these missed procs, <strong>{this._rtCombos.filter(combo => !combo.success && combo.trueNorthCharges > 0).length}</strong> could be handled with <ActionLink {...ACTIONS.TRUE_NORTH} />
```
Toastdeib: You're going to need to do this string another way - it isn't getting picked up properly for translation. I'm not 100% sure, but I *think* turning it into two separate strings that you just list one after another should work. So something like:

```js
<Trans id="drg.positionals.analysis.missed"><Icon name="info" /> You missed <strong>{missed}</strong>...</Trans>
{missed > 0 ? (
    <Trans id="drg.positionals.analysis.truenorth">Of these missed procs...</Trans>
) : ('')}
```

Instead of embedding the second half of the string inside the first.

ackwell: According to DT ([link](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/lingui__react/Select.d.ts)), the plural tag accepts react nodes, so this can be handled in a single translation using something like

```
<Trans id="foo">Outer text<Plural value={missed} _0={<>Inner text<>}/>... </Trans<
```
Untested (on phone), but might be worth a shot.

ebshimizu: Opted for the split `Trans` tags, should be in the locale files now.

I did a bit of testing with `Plural` but it felt a little weird to use it to do conditional text rendering. It worked for `en`, but I'm not sure if it was in the right form in the locale files. Would need a bit more testing to see if that would work.

PR #718, opened 2019-11-10
DRG: Buffs Formatting & Max Uses Calc

ebshimizu:
* Feature: Computes the number of full personal buff (Lance Charge, Dragon Sight) windows you can get in the fight and reports if you could've fit more buffs in.
* Fix: Allowed partial buff windows at the end of the fight (it's optimal to use there even if not getting full duration), and marked these windows with a *.
* Formatting: Re-formatted the accordion into a table that includes recast drift.


ebshimizu:
note: I'm gonna close this one for a lil bit since I know a bit more about the code base and have to think about what to do with this. I'll re-open when it's ready. :)

PR #719, opened 2019-11-12
[5.1 SAM] Shoha tracker

Rjb99:
This PR adds the final piece to 5.1 support, a shoha tracker
this keeps track of all generated stacks and then compares them to uses to of shoha, 
it also bumps support to purely 5.1 due to meditation and shoha changes


src/parser/jobs/sam/modules/Shoha.tsx
```diff
@@ -0,0 +1,272 @@
...
+			{
+			to: 'player',
+			abilityId: STATUSES.MEDITATION.id,
+			}, this.onMeditateGain)
+
+		this.addHook('death', {to: 'player'}, this.onDeath)
+		this.addHook('complete', this.onComplete)
+	}
+
+	private onGenerator(event: CastEvent) {
```
ayyaruq: Indentation is who did this to you?

src/parser/jobs/sam/modules/Shoha.tsx
```diff
@@ -0,0 +1,272 @@
...
+		if (this.stacks === MAX_STACKS) {
+		const abilityId = event.ability.guid
+		const generatedStacks = GENERATORS[abilityId]
+
+		this.addGeneratedStackAndPush(generatedStacks, abilityId)
+	}
+	}
+
+		// This is exclusively for stacks gained via Meditate channel. and not any stacks gained from GCD casts
```
ayyaruq: I think your tab key is broken.

src/parser/jobs/sam/modules/Shoha.tsx
```diff
@@ -0,0 +1,272 @@
...
+
+	protected init() {
+		this.addHook('init', this.pushToHistory)
+		this.addHook(
+			'cast',
+			{
+				by: 'player',
+				abilityId: Object.keys(GENERATORS).map(Number),
+			},
+			this.onGenerator,
```
ayyaruq: Nit, onGenerate/onGain

src/parser/jobs/sam/modules/Shoha.tsx
```diff
@@ -0,0 +1,272 @@
...
+		)
+		this.addHook(
+			'cast',
+			{
+				by: 'player',
+				abilityId: Object.keys(SPENDERS).map(Number),
+			},
+			this.onSpender,
```
ayyaruq: Nit, onSpend

src/parser/jobs/sam/modules/Shoha.tsx
```diff
@@ -0,0 +1,272 @@
...
+
+			// just a safety net
+			if (this.stacks > MAX_STACKS) {
+				this.stacks = MAX_STACKS
+			}
+
+			this.pushToHistory()
+	}
+
+	private addGeneratedStackAndPush(generatedStacks: number, abilityId: number) {
```
ayyaruq: You sorta dupe this in Meditate anyway, but then only really use it once. Any reason not to just roll it into the normal generator method? I don't think it makes the function particularly unweildy.

src/parser/jobs/sam/modules/Shoha.tsx
```diff
@@ -0,0 +1,272 @@
...
+		this.stackHistory.push({t: timestamp, y: this.stacks})
+	}
+
+	private onComplete() {
+		this.dumpRemainingResources()
+
+		// final use amounts
+
+		const totalPossibleUses = Math.floor(this.totalGeneratedStacks/MAX_STACKS)
+		const totalUses = this.shohaUses
```
ayyaruq: Kinda useless assignment, the total possible too. You're not using them in the text or more than just the one value/target, so may as well just have the direct values there. Maybe useful for future additions tho.

src/parser/jobs/sam/modules/Shoha.tsx
```diff
@@ -0,0 +1,272 @@
...
+					{rows}
+				</tbody>
+			</table>
+		</Fragment>
+	}
+
+	private convertWasteEntryToRow(action: TODO) {
+		let actionName = action.name
+		if (action === ACTIONS.RAISE) {
+			actionName = 'Death'
```
ayyaruq: This will probably break i18n but I never found a good workaround for that when I did similar in MNK either tbh.

src/parser/jobs/sam/modules/Shoha.tsx
```diff
@@ -0,0 +1,272 @@
...
+			const waste = this.stacks - MAX_STACKS
+			this.wasteBySource[abilityId] += waste
+			this.stacks = MAX_STACKS
+		}
+
+		this.pushToHistory()
+	}
+
+	private onSpender(event: CastEvent) {
+		this.stacks = this.stacks - SPENDERS[event.ability.guid]
```
ayyaruq: If max is 3 and Shoha is 3, why not just reset it, or trying to catch broken logs (which you don't do in this PR)?

src/parser/jobs/sam/modules/Shoha.tsx
```diff
@@ -0,0 +1,272 @@
...
+	[ACTIONS.KAESHI_SETSUGEKKA.id]: 1,
+}
+
+const SPENDERS = {
+	[ACTIONS.SHOHA.id]: 3,
+}
+
+const MAX_STACKS = 3
+
+class StackState {
```
ayyaruq: You're not actually declaring this as a class instance anymore, probably better as an `interface`. Alternatively make history a real class and have a method to pull data out of it so you get time data for wastage etc, and can still generate the charts nicely.

src/parser/jobs/sam/modules/Shoha.tsx
Rjb99: I really should rename that to AddWaste or something. It started out as the go to adding stacks method, but now I've converted it to be used as a function to handle waste from the generators used at cap. 

I could probably just merge this into the Iaijustu cast code since the whole point of that code is just to check if we're already capped and then move into this function to add it to waste

src/parser/jobs/sam/modules/Shoha.tsx
Rjb99: I felt that seeing 43/44 meditation stacks wouldn't be as useful of a metric as the punishment from overcapping only comes from when you lose a 3n amount of stacks.

IE: 15/16 shoha casts means you missed 3 stacks, where as a 43/44 is just saying you had 2 extra stacks, 1 of which you wasted by overcapping it. which is literally a 0 dps loss

I could just directly use the shohaUses though for this.

src/parser/jobs/sam/modules/Shoha.tsx
Rjb99: I can change that. A lot of this code has some useless future proofing in case they add in ways to spend stacks without being at cap so I could just throw the new spenders/generators into the const groups for each. in hindsight it just makes the code look more chaotic.

src/parser/jobs/sam/modules/Shoha.tsx
ackwell: Pretty sure this isn't actually going to be doing anything.

`<ActionLink name={actionName} {...action}/>` can essentially be thought of as `{name: actionName, ...action}` - given `action` already _has_ a name, action's spread will override the earlier assignment.

Can technically use `<ActionLink {...action} name={actionName}/>` to do it the other way around, but that won't work either - the `name` is only used while it loads data off the API, at which point API data takes precedence.

If you _really_ want to show "death" instead of the action name for raise, you can do something like uh (and this is ugly)

```tsx
const actionChildren = action === ACTIONS.RAISE
	? <Trans id="some.id.here">Death</Trans>
	: undefined

return <>...<ActionLink {...action} children={actionChildren}/>...</>
```

src/parser/jobs/sam/modules/index.js
```diff
@@ -10,7 +10,7 @@ import Meikyo from './Meikyo'
 import ThirdEye from './ThirdEye'
 import Tsubame from './Tsubame'
 import OGCDDowntime from './OGCDDowntime'
-//import Shoha from './Shoha' It is not time yet my child, we've got bugs to fix first
+import Shoha from './Shoha' //It is not time yet my child, we've got bugs to fix first
```
ackwell: is it still not time yet? 🤔

src/parser/jobs/sam/modules/Shoha.tsx
ackwell: Eh, no issue with functions that are called once, esp. if they help split logic into more approachable blocks. Accurate names are 👌 though.

PR #723, opened 2019-11-16
Shadowmonkers of Fire

ayyaruq:
Updated form of #582 because what is pushing to other people's branches? Couldn't find an immediate need for gauge here even though I thought it needed it the whole time. I feel like we could use core buff windows here but we're not, and it's more effort to migrate to that right now than just address feedback and get this merged and supported.

Most of the work was by @rachellhea, so I've kept her commits as authored (cherry-picked), changelog too.


ayyaruq:
Linter wants a trailing comma added, but I'm fairly certain that's a syntax error as `Math.min` only takes 2 parameters.

PR #727, opened 2019-11-20
Add module debug logging fn

ackwell:
Inspired by #713

image

Color is derived from the module handle. I've _tried_ to make it readable on both light and dark themes. No promises for yellow though - try renaming your module? :^)

I slep now


ackwell:
Gonna dunk this.

@AzariahAeroborn if you could swap out to use this when you do a sweep over the CDR module, that'd be great - lemme know if you hit any snags.

PR #731, opened 2019-11-26
Layered data part 2: Actions boogaloo

ackwell:
This PR:
- Migrates actions to typescript
- Implements layered data for actions
- Introduces super hacky typing on the `ensureX` functions for both actions and statuses, but hey, it works
- Moves a few bits of data around to make the above easier.

Also in this PR:

As usual, pls2test if people have time.

`master` is 37.6% TS at time of writing. let's see how much this bumps it by~


ackwell:
What could possibly go wrong

ackwell:
image
EYYYYYYYYYYYYYYYY

PR #732, opened 2019-11-29
[Bugfix] Fix precast status for buffs that apply stacks

AzariahAeroborn:
Most noticeable for Acceleration (RDM) - when this was changed to a 3 stack buff in 5.1, we are now getting an `applybuff` event and an `applybuffstack` event with 2 stacks on the RDM's first cast of the pull.

See, for example: https://www.fflogs.com/reports/mphJdArR6vbTHLKj#fight=2&type=auras&source=1&view=events

This fix is just for RDM, but should be extensible to other similar jobs fairly easily.

### DIFF
--- a/src/parser/core/modules/PrecastStatus.js
+++ b/src/parser/core/modules/PrecastStatus.js
@@ -14,64 +14,87 @@ export default class PrecastStatus extends Module {
 
 	_combatantStatuses = {}
 	_combatantActions = []
+	_statusesToSynth = []
+	_actionsToSynth = []
+	_startTime = this.parser.fight.start_time
 
 	normalise(events) {
-		const startTime = this.parser.fight.start_time
-
 		for (let i = 0; i < events.length; i++) {
 			const event = events[i]
 			const targetId = event.targetID
 
+			const statusInfo = getDataBy(STATUSES, 'id', event.ability.guid)
+
 			this._combatantStatuses[targetId] = this._combatantStatuses[targetId] || []
 
-			if (event.type === 'applybuff') {
-				this._combatantStatuses[targetId].push(event.ability.guid)
+			if (event.type === 'applybuff' && !statusInfo.hasOwnProperty('stacksApplied')) {
+				// If status applies stacks, check applybuffstack for applying full stacks before considering this the first application of this status
+				this.markStatusAsTracked(statusInfo.id, targetId)
 			}
 
-			if (['removebuff', 'applybuffstack', 'removebuffstack', 'refreshbuff'].includes(event.type)) {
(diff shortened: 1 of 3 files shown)


src/parser/core/modules/PrecastStatus.js
```diff
@@ -14,64 +14,87 @@ export default class PrecastStatus extends Module {
...
+	_startTime = this.parser.fight.start_time
 
 	normalise(events) {
-		const startTime = this.parser.fight.start_time
-
 		for (let i = 0; i < events.length; i++) {
 			const event = events[i]
 			const targetId = event.targetID
 
+			const statusInfo = getDataBy(STATUSES, 'id', event.ability.guid)
```
ackwell: this could technically return `undefined` - probably safe to just `if (!statusInfo) { continue }` to prevent issues.

(checking `statusInfo.hasOwnProperty` if it's ever undefined will throw)

src/parser/core/modules/PrecastStatus.js
```diff
@@ -14,64 +14,87 @@ export default class PrecastStatus extends Module {
...
-				}
-
-				this._combatantStatuses[targetId].push(statusId)
+				this.fabricateStatusEvent(event, statusInfo)
+				this.markStatusAsTracked(statusInfo.id, targetId)
 			}
 		}
 
-		return events
+		const synthesizedEvents = this._actionsToSynth.concat(this._statusesToSynth)
```
ackwell: toooooooooootal nitpick;
```js
return [...this.actionsToSynth, ...this._statusesToSynth, ...events]
```

src/parser/core/modules/PrecastStatus.js
```diff
@@ -14,64 +14,87 @@ export default class PrecastStatus extends Module {
...
+	fabricateStatusEvent(event, statusInfo) {
+		// Fab an event and splice it in at the start of the fight
+		this._statusesToSynth.push({
+			// Can inherit most of the event data from the current one
+			...event,
+			// Override a few vals
+			timestamp: this._startTime - 1,
+			type: 'applybuff',
+		})
+
```
ackwell: for the sake of InsaneAccuracy:tm:, is it worth fabricating an `applybuffstack` event after the `applybuff` in the case we're fabbing a stacked event? would replicate fflog's events more closely (i.e. applying an accel mid-combat will generate `[applybuff(accel), applybuffstack(accel, 3)]`

AzariahAeroborn: Yeah, seems simple enough.  Dun (dun chicken)

PR #736, opened 2019-11-30
AST 5.1

suShirow:
This PR includes AST data for 5.1, and a little bit of backdata to test out the new layer system.
Card module supports legacy pre-5.1 system, let's hope I don't shoot myself in the foot for that.


suShirow:
Weird commits with Aza's AoE updates from September. It happened when I tried to test his PR way back then and it must have muddied my fork's master.

src/data/ACTIONS/layers/501.ts
```diff
@@ -0,0 +1,12 @@
+import {Layer} from 'data/layer'
+import STATUSES from 'data/STATUSES'
```
ackwell: doesn't look like it's being used

src/data/ACTIONS/root/AST.ts
```diff
@@ -161,9 +161,11 @@ export const AST = ensureActions({
 		icon: 'https://xivapi.com/i/003000/003140.png',
 		cooldown: 90,
 		statusesApplied: [
-			'COLLECTIVE_UNCONSCIOUS_SHIELD',
+			'COLLECTIVE_UNCONSCIOUS_DIURNAL_MITIGATION',
```
ackwell: This looks the same as the data in the `5.1` layer - intentional?

src/parser/jobs/ast/modules/ArcanaTracking/ArcanaTracking.tsx
```diff
@@ -85,6 +89,8 @@ export default class ArcanaTracking extends Module {
 		sealState: CLEAN_SEAL_STATE,
 		sleeveState: SleeveType.NOTHING,
 	}]
+
+	private lastDrawnBuff: BuffEvent | undefined = undefined
```
ackwell: nit; below is equivalent
```suggestion
	private lastDrawnBuff?: BuffEvent
```

src/data/ACTIONS/root/AST.ts
suShirow: Actually, when I try to add fewer to the Root layer, it tells me that the `5.1` layer can't have anything but the original statuses applied.
```
TS2322: Type '{ statusesApplied: ("COLLECTIVE_UNCONSCIOUS_DIURNAL_MITIGATION" | "COLLECTIVE_UNCONSCIOUS" | "WHEEL_OF_FORTUNE_DIURNAL" | "COLLECTIVE_UNCONSCIOUS_NOCTURNAL" | "WHEEL_OF_FORTUNE_NOCTURNAL")[]; }' is not assignable to type 'Partial<{ id: number; name: string; icon: string; cooldown: number; statuldown: number; st
atusesApplied: ("COLLECTIVE_UNCONSCIOUS_DIURNAL_MITIGATION" | "COLLECTIVE_UNCONSCIOUS" | "WHEEL_OF_FORTUNE_DIURNAL")[]; } & Action>'.
```

suShirow: Also, the name change was to differentiate the Diurnal shield with the Nocturnal shield - mitigation is more apt since it's not actually a shield in both cases

suShirow:
> Ah, I totally forgot to mention. To use the layered data, you need to use the `Data` module, rather than importing `ACTIONS` and `STATUSES` directly.
> 
> I've done the port for `core` in #737, there's a few examples of different ways of doing it there. Ping me if there's any Qs.

This makes sense, since we need patch info to figure out which layer to use. But does this mean that I can't do QoL stuff like 
```
export enum DrawnType {
	NOTHING = 0,
	BALANCE = STATUSES.BALANCE_DRAWN.id,
	BOLE = STATUSES.BOLE_DRAWN.id,
	ARROW = STATUSES.ARROW_DRAWN.id,
	SPEAR = STATUSES.SPEAR_DRAWN.id,
	EWER = STATUSES.EWER_DRAWN.id,
	SPIRE = STATUSES.SPIRE_DRAWN.id,
	LORD_OF_CROWNS = STATUSES.LORD_OF_CROWNS_DRAWN.id,
	LADY_OF_CROWNS = STATUSES.LADY_OF_CROWNS_DRAWN.id,
}
```
and also what I have in `ArcanaGroups.ts`?

src/data/ACTIONS/root/AST.ts
ackwell: ergh, goddamnit. that's an issue in the typings. i'll have a look into it tonight.

ackwell:
`ArcanaGroups.ts` should be relatively easy to fix, something like:

```ts
// instead of
export const SOME_COLLECTION = [
	ACTIONS.ONE.id,
	ACTIONS.TWO.id,
]
// you can use the root key instead, as the set of keys is guaranteed to be stable:
export const SOME_NEW_COLLECTION: Array<keyof ActionRoot> = [
	'ONE',
	'TWO',
]
```

You can then use `this.data.actions[somethingFromThatArray].id` to get IDs, where necessary.

The enum might be a bit harder to handle, but should be possible using similar methodology.

Issue #687, opened 2019-10-26
Core status uptime breaks on boss invuln


qaphla:
Currently, core status uptime for personal buffs divides buff uptime by fight length minus downtime. This causes problems for buffs that can be maintained even when the boss is invulnerable on untargetable, in the worst cases, making uptime over 100%.

Most likely, when calculating self-buff uptime, there should be a way to subtract out time that the buff was up but the boss was invulnerable/untargetable/etc.

qaphla:
Closed by #739

PR #739, opened 2019-12-01
Refactor entities.getStatusUptime() into separate module

AzariahAeroborn:
This is in support of #670 and the effort to switch to targetability events for invulnerability.  This should break the dependency from entities to invulnerability, and resolve the circular dependency created from switching to targetability events.  

I only lifted out the getStatusUptime method (and duplicated some of the buff tracking) because the existing Entities module is populating entity.buffs for each processed entity, which some modules depend on via entity.hasStatus (Bard Barrage tracking, for one).

I've reviewed a few jobs and confirmed that calculated DoT uptime is exactly the same for most of them.  White Mage DoT tracking on Titan is coming up with (very slightly higher) different numbers, though on fights like Leviathan it's exactly the same, so there seems to be some interaction with Titan Jails that this refactor is changing; not sure if it's something severe enough to be alarmed about or not.

### DIFF
--- a/src/parser/core/modules/Entities.js
+++ b/src/parser/core/modules/Entities.js
@@ -136,8 +136,8 @@ export default class Entities extends Module {
 		// Update the buff's last refreshed time
 		const oldStacks = buff.stacks
 		buff.lastRefreshed = event.timestamp
-		buff.stacks = event.stack || 1
-		buff.stackHistory.push({stacks: event.stack || 1, timestamp: event.timestamp})
+		buff.stacks = 1
+		buff.stackHistory.push({stacks: 1, timestamp: event.timestamp})
 
 		this.triggerChangeBuffStack(buff, event.timestamp, oldStacks, buff.stacks)
 	}
--- a/src/parser/core/modules/Entities.js
+++ b/src/parser/core/modules/Entities.js
@@ -9,19 +9,21 @@ export default class Entities extends Module {
 		super(...args)
 
 		// Buffs
-		this.addHook('applybuff', this.applyBuff)
-		this.addHook('applydebuff', event => this.applyBuff(event, true))
-		this.addHook('applybuffstack', this.updateBuffStack)
-		this.addHook('applydebuffstack', event => this.updateBuffStack(event, true))
-		this.addHook('removebuffstack', this.updateBuffStack)
-		this.addHook('removedebuffstack', event => this.updateBuffStack(event, true))
-		this.addHook('removebuff', this.removeBuff)
-		this.addHook('removedebuff', event => this.removeBuff(event, true))
+		this.addEventHook('applybuff', this.applyBuff)
+		this.addEventHook('applydebuff', event => this.applyBuff(event, true))
+		this.addEventHook('applybuffstack', this.updateBuffStack)
+		this.addEventHook('applydebuffstack', event => this.updateBuffStack(event, true))
+		this.addEventHook('removebuffstack', this.updateBuffStack)
+		this.addEventHook('removedebuffstack', event => this.updateBuffStack(event, true))
+		this.addEventHook('removebuff', this.removeBuff)
+		this.addEventHook('removedebuff', event => this.removeBuff(event, true))
+		this.addEventHook('refreshbuff', this.refreshBuff)
+		this.addEventHook('refreshdebuff', event => this.refreshBuff(event, true))
 
 		// Resources - hooked to init to make sure normaliser runs to determine damage event before hooking events
-		this.addHook('init', () => {
-			this.addHook(this.fflogsEvents.damageEventName, this.updateResources)
-			this.addHook(this.fflogsEvents.healEventName, this.updateResources)
+		this.addEventHook('init', () => {
+			this.addEventHook(this.fflogsEvents.damageEventName, this.updateResources)
+			this.addEventHook(this.fflogsEvents.healEventName, this.updateResources)
 		})
 	}
 
@@ -57,6 +59,7 @@ export default class Entities extends Module {
 		const buff = {
 			...event,
 			start: event.timestamp,
+			lastRefreshed: event.timestamp,
 			end: null,
 			stackHistory: [{stacks: 1, timestamp: event.timestamp}],
 			isDebuff,
@@ -106,14 +109,7 @@ export default class Entities extends Module {
 
 		// If there's no existing buff, fake one from the start of the fight
 		if (!buff) {
-			const startTime = this.parser.fight.start_time
-			buff = {
-				...event,
-				start: startTime,
-				end: null,
-				stackHistory: [{stacks: 1, timestamp: startTime}],
-				isDebuff,
-			}
+			buff = this.synthesizeBuff(event, isDebuff)
 			entity.buffs.push(buff)
 		}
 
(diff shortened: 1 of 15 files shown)


AzariahAeroborn:
On further review, changed EntityStatuses to re-use the entity.buffs dictionary that is already getting created by Entities.js and its child classes.  

While I was cleaning this up, I also added some debug output and was able to trace and fix the behavior change that was changing the WHM DoT uptime calculations vs Titan.  Now showing the same results as live for the logs I've checked.

AzariahAeroborn:
Added a commit to remove the throw when checking for invuln events effecting a buff that had no end value due to target becoming invalid without generating a remove event.  

This fixes #687

src/parser/core/modules/EntityStatuses.ts
```diff
@@ -0,0 +1,142 @@
...
+	// API
+	// -----
+	// TODO: This implementation which I've shamelessly stolen seems to only track overall
+	//       uptime, ignoring potential gains from multidotting, etc. Should that be in here,
+	//       or is that a somewhere-else sort of thing...
+	getStatusUptime(statusId: number, onTargets: TODO, sourceId = this.parser.player.id) {
+		// Search for status maps on valid targets
+		const statusEvents: BuffTrackingEvent[] = Object.values(onTargets)
+			.reduce((statusEvents: BuffTrackingEvent[], target: TODO) => {
+				const appliedStatuses = target.buffs?.filter((statusEvent: BuffTrackingEvent) => {
```
ackwell: that `?.buffs` has a potential to return `undefined`, which JS'll happily concat onto the array as a new item. You'll probably need to return the accumulator early or something similar.

Not sure why TS isn't complaining about it; best guess is that the `any` from the `TODO` is squashing the `| undefined` that `?.` should generate.

src/parser/core/modules/EntityStatuses.ts
```diff
@@ -0,0 +1,142 @@
...
+					return statusEvent.ability.guid === statusId && (statusEvent.sourceID === null || statusEvent.sourceID === sourceId)
+				})
+				return statusEvents.concat(appliedStatuses)
+			}, [])
+
+		let active = 0
+		let start: number | null = null
+		return statusEvents
+			.reduce((statusRanges: StatusRangeEndpoint[], statusEvent) => {
+				return statusRanges.concat(this.getStatusRangesForEvent(statusEvent))
```
ackwell: can inline the function block to avoid the `return`

src/parser/core/modules/EntityStatuses.ts
```diff
@@ -0,0 +1,142 @@
...
+			return statusRanges
+		}, [])
+	}
+
+	splitStatusEventForInvulns(statusEvent: BuffTrackingEvent) {
+		const eventToAdjust = {...statusEvent}
+		const adjustedEvents = [eventToAdjust]
+		if (!statusEvent.end) {
+			this.debug('Unfinished status event detected.  Applying ability duration after last refresh event.')
+			const statusInfo = getDataBy(STATUSES, 'id', statusEvent.ability.guid)
```
ackwell: Esp. given this is core, `this.data.getStatus(statusEvent.ability.guid)` - you'll need to add data as a dep (should be safe, `data` is and should remain completely depless).

src/parser/core/modules/EntityStatuses.ts
```diff
@@ -0,0 +1,142 @@
...
+
+		invulns.forEach((invuln: TODO) => {
+			this.debug(`Target was detected as invulnerable during duration of status.  Invulnerable from ${this.parser.formatTimestamp(invuln.start, 1)} to ${this.parser.formatTimestamp(invuln.end, 1)}`)
+			if (invuln.start < statusEvent.start && invuln.end >= statusEvent.start) {
+				this.debug('Invuln clipped the start of the range - changing beginning of event')
+				eventToAdjust.start = invuln.end
+			} else {
+				this.debug('Invuln clipped the end of the range or split the range - changing end of event')
+				eventToAdjust.end = invuln.start
+				eventToAdjust.stackHistory.splice(-1, 1, {stacks: 0, timestamp: invuln.start, invuln: true})
```
ackwell: This will be modifying the original event's history, as well - `{...obj}` is only a shallow copy. You can do a second shallow copy here with a boring `eventToAdjust.stackHistory = [...eventToAdjust.stackHistory]`, or go nuclear with `const eventToAdjust = _.cloneDeep(statusEvent)` at the top.

src/parser/core/modules/EntityStatuses.ts
```diff
@@ -0,0 +1,142 @@
...
+						start: invuln.end,
+						lastRefreshed: invuln.end,
+						end: statusEvent.end,
+						stackHistory: newStackHistory,
+					})
+				}
+			}
+		})
+
+		adjustedEvents.forEach(event => this.debug(`Active time after adjusting for invulns - start at: ${this.parser.formatTimestamp(event.start)} - end at: ${this.parser.formatTimestamp(event.end!)}`))
```
ackwell: mmh - this is fine, but i should probably add an overload signature to `debug` for a function arg so that potentially expensive argument computations can be done in debug only.

src/parser/core/modules/Entities.js
```diff
@@ -220,6 +119,40 @@ export default class Entities extends Module {
...
+		// If there's no existing buff, fake one from the start of the fight
+		if (!buff) {
+			buff = this.synthesizeBuff(event, isDebuff)
+			entity.buffs.push(buff)
+		}
+
+		// Update the buff's last refreshed time
+		const oldStacks = buff.stacks
+		buff.lastRefreshed = event.timestamp
+		buff.stacks = event.stack || 1
```
ackwell: Hm - the event type defs have `stack` as always-defined - does that need to be updated to be optional?

If it is optional, this should probably use `??` to avoid possible kludge of `0` stack values.

src/parser/jobs/mnk/modules/TwinSnakes.tsx
```diff
@@ -14,6 +14,7 @@ import Enemies from 'parser/core/modules/Enemies'
 import Invulnerability from 'parser/core/modules/Invulnerability'
 import Suggestions, {SEVERITY, Suggestion, TieredSuggestion} from 'parser/core/modules/Suggestions'
 
+import {EntityStatuses} from '../../../core/modules/EntityStatuses'
```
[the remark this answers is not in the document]
ayyaruq: I did what now?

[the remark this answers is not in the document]
AzariahAeroborn: Whoops, sorry, IntelliJ must have auto-inserted that for me.  I usually catch those.

src/parser/core/modules/EntityStatuses.ts
AzariahAeroborn: I didn't inline the function block because I was trying to separate across lines (since it's a really really long one-line otherwise) and then that causes TSLint to complain about no trailing comma (presumably due to the concat)

src/parser/core/modules/Entities.js
AzariahAeroborn: It looks like the current behavior in Entities.js is that an `apply(de)buff` event applies 1 buff stack, and a `apply(de)buffstack` event is applying event.stack || 1 stacks (in the updateBuffStack method).  So I made `refresh(de)buff` behave like the updateBuffStack method, since I wasn't sure which way made more sense.  

Looking at the output from FFLogs, it looks like `refreshbuff` events are shaped more like `applybuff` events, and the stack property doesn't exist on them.  Will change this to always be 1.

PR #741, opened 2019-12-02
DRG: CDDT Module for Jumps

ebshimizu:
Use core `CooldownDowntime` to track High Jump, Spineshatter Dive, and Dragonfire Dive uses.
Default first use offset set to 25s (10 GCDs at 2.5s, the typical full opener for DRG).

image


src/parser/jobs/drg/modules/OGCDDowntime.ts
```diff
@@ -0,0 +1,13 @@
+import ACTIONS from 'data/ACTIONS'
+import {CooldownDowntime} from 'parser/core/modules/CooldownDowntime'
+
+const DEFAULT_FIRST_USE_OFFSET = 25000
+
+export default class OGCDDowntime extends CooldownDowntime {
+	defaultFirstUseOffset = DEFAULT_FIRST_USE_OFFSET
```
AzariahAeroborn: From a UX perspective, this should either track all relevant OGCDs or override the `checklistName` and `checklistDescription` properties to indicate that the module is only tracking Jumps and not all OGCDs.

Personally, I would recommend adding the other OGCDs to this as well... looking through the job guide, it seems the relevant ones are probably:
- Life Surge
- Lance Charge
- Battle Litany
- Dragon Sight

ebshimizu: Ah yeah I kinda forgot about Life Surge lol (I just kinda automatically use it before full thrust these days). The buffs I was hesitant to add to this because I wasn't sure if the module would recommend a number of usages that would require clipping too much into boss invuln windows. I did a check of a few logs (not comprehensive) and it seems to be reasonable, for this tier at least.

PR #743, opened 2019-12-07
Normalise FFLogs events to a standard interface

AzariahAeroborn:
Overall, the goal of this effort is to eliminate the FFLogsEventNormaliser shim and to consolidate the functionality of AOE and the custom deduplication logic in BRD's Barrage module, so that we have stable event types that deduplicate multihit / multitarget actions into a single damage or heal action, and whose interface remains consistent between logs from before or after the `calculatedX` event changes on the FFLogs API.

I've confirmed basic functionality against DRK's MultiHitSkills module, and am looking for feedback on the event shape and whether I should go ahead and update any consumers of the FFLogsEventNormaliser shim or `aoedamage` events to use this new interface, or if we continue supporting both in the short term and encourage job maintainers to migrate their modules separately.


src/parser/core/modules/NormalisedEvents.ts
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	@dependency private hitType!: HitType
+
+	_normalisedDamageEvents = new Map<number, NormalisedDamageEvent>()
+	_normalisedHealEvents = new Map<number, NormalisedHealEvent>()
+
+	normalise(events: Event[]): Event[] {
+		events.forEach((event) => {
+			if (DAMAGE_EVENT_TYPES.includes(event.type)) {
```
ackwell: Purely because of how many times this will be hit, it'd be cool to get the two const arrays defined as `Set`s, so you can use `.has`

src/parser/core/modules/NormalisedEvents.ts
```diff
@@ -0,0 +1,178 @@
...
+				this.normaliseDamageEvent(event as DamageEvent)
+			}
+
+			if (HEAL_EVENT_TYPES.includes(event.type)) {
+				this.normaliseHealEvent(event as HealEvent)
+			}
+		})
+
+		this._normalisedDamageEvents.forEach(this.summarize, this)
```
ackwell: I use [JT]S every bloody day, and TIL that `forEach` takes a `thisArg`

src/parser/core/modules/NormalisedEvents.ts
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	private findRelatedDamageEvent(event: DamageEvent) {
+		if (event.packetID) {
+			return this._normalisedDamageEvents.get(event.packetID)
+		}
+
+		// No packetID set - match based on timestamps
+		const possibleRelatedEvents = Array.from(this._normalisedDamageEvents.values())
+			.filter((normalisedEvent) => {
```
ackwell: nits:

arrows w/ no type and one arg can forgo the parens (`.filter(normalisedEvent => ...)`).

---

you saw this one coming; can afford to remove the `{ return }` from this. It can still be split over multiple lines, just saves the fn body itself.

src/parser/core/modules/NormalisedEvents.ts
```diff
@@ -0,0 +1,178 @@
...
+	'damage',
+	'calculateddamage',
+]
+
+const HEAL_EVENT_TYPES: Array<Event['type']> = [
+	'heal',
+	'calculatedheal',
+]
+
+export interface NormalisedDamageEvent extends Omit<DamageEvent, 'type'> {
```
ackwell: General _thoughts_ on the attrs themselves:

* Hm... I'm aware that "ghost" is a popular term for the issue we're modelling, but it's also a term that's not very self-explanatory, and often has to be explained to people when it gets brought up. I'm not opposed to referencing it as ghosting in xiva, but there might be value in putting some thought into it to be sure it's what we want to codify. If it is, I'll probably req. a comment explaining what it means.
* Events don't _need_ to be flat structures. I'm not coming up with anything clever right now, but I wonder if there's a neater representation with a nested structure? Maybe just something to consider.

src/parser/core/modules/AdditionalEvents.js
```diff
@@ -43,6 +45,7 @@ export default class AdditionalEvents extends Module {
 	static dependencies = [
 		'enemies',
 		'data',
+		'normalisedEvents', // eslint-disable-line @xivanalysis/no-unused-dependencies
```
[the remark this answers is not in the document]
AzariahAeroborn: My two reasons for adding the dependency were to:
1) Ensure the normaliser for normalisedEvents ran, without requiring implementers to specifically add the dependency (under the mental image that this would be a hard replacement for aoedamage and the getFFLogsEvents compatibility shim)
2) Ensure the fabricated events got sorted in with the other events, without having to deal with a bunch of individual splicing when dumping the events out (doubly so since splicing in the middle of Array.forEach requires some funky handling to avoid infinitely looping on the added event).

ackwell:
Ah sorry, this was marked as ready to review. I totally skipped over it.

@AzariahAeroborn this still good to go? I'll do another pass over it on the weekend or thereabouts.

AzariahAeroborn:
@ackwell Yes, this is ready to go.

PR #745, opened 2019-12-13
DRG: Adjust first use offset for CDDT

ebshimizu:
Tightening up the first use offset based on latest use times in both DRG openers (low/high skill speed).

### DIFF
--- a/src/parser/jobs/drg/modules/OGCDDowntime.ts
+++ b/src/parser/jobs/drg/modules/OGCDDowntime.ts
@@ -1,17 +1,45 @@
 import ACTIONS from 'data/ACTIONS'
 import {CooldownDowntime} from 'parser/core/modules/CooldownDowntime'
 
-const DEFAULT_FIRST_USE_OFFSET = 25000
+// at high skill speeds, Battle Litany is first, so the order is a bit fluid,
+// however all are used before the third GCD
+const BUFF_FIRST_USE_OFFSET = 5000
+
+// the high sks opener delays jumps to better line up with later windows,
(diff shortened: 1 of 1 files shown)


src/parser/jobs/drg/modules/OGCDDowntime.ts
```diff
@@ -1,15 +1,39 @@
...
+
+// the high sks opener delays jumps to better line up with later windows,
+// at least that's what I'm told. These timings should account for both openers,
+// assuming the fast sks threshold of 2.35s
+const JUMP_FIRST_USE_OFFSET = 14100		// before 7th gcd
+const SSD_FIRST_USE_OFFSET = 18800		// before 9th gcd
+const DFD_FIRST_USE_OFFSET = 21500		// before 10th gcd
+
+// always before Full Thrust, the 8th GCD
+const LIFE_SURGE_FIRST_USE_OFFSET = 17500
```
Toastdeib: Might be worth building at least a little wiggle room into these numbers; in my experience, there can often be a bit of variance between when everyone's first attack actually lands, so you can't really guarantee that the first GCD of a specific party member will be at exactly t=0.

ebshimizu: I'll take a look! They're currently set to be the next GCD after the oGCD should be used. I'll take a peek at some more logs and check to see if that's enough of a buffer, if not I'll bump them up a little bit. I'll give a ping after I've investigated a bit more.

ebshimizu:
Looks like most of the top logs have a max delay of ~2s before first gcd. Added that in to the timings and they appear to be working well.

PR #751, opened 2019-12-24
Add korean patch date for 5.01

Ra-Workspace:


Ra-Workspace:
KR already updated v5.08 balance

PR #755, opened 2020-01-03
Updates for RDM CD Downtime

Cephid:
This is just a small PR to adjust the values for the RDM Cooldown Downtime due to update of that module.  Also adjusted the first grace of CDs to the 3-8 standard opener.


src/parser/jobs/rdm/modules/GeneralCDDowntime.ts
```diff
@@ -12,19 +12,19 @@ export default class GeneralCDDowntime extends CooldownDowntime {
 		// hope that it's accurate enough
 		{
```
AzariahAeroborn: Should this comment be changed or renewed in light of the code change?

ayyaruq:
Looks fine to me but how common is that standard opener, is it optimal, or actually standard?

src/parser/jobs/rdm/modules/GeneralCDDowntime.ts
Cephid: We're not actually sure honestly, we tried a multitude of tests but weren't entirely happy with any result.  This particular configuration yielded the most correct data across different situations.  We elected in the TC channel to just go with it and see how it goes with more real world data.

Cephid:
It's the pinned opener, the TC channel also confirmed it for me when I was adjusting the old numbers.

PR #759, opened 2020-01-13
DRG: Enhanced Life of the Dragon Window Analysis

ebshimizu:
Number of enhancements for the Life of the Dragon module, based on things in the current balance checklist. Also cleaned up the logic for determining lotd delay timings.

Looking for comments regarding suggestion severity for the various things you could do wrong in a life window.

### Enhancements
1, Added a suggestion for keeping GK casts within -1 of Jump casts (checklist section 2), -2 = minor, -3 = medium.
2. Added suggestions for missing casts within LotD windows (unless the window is truncated by the end of the fight). Triggers on less than 3 Nastronds in a window, and missing a Stardiver cast in a window. Tiered suggestion, 1 miss = medium, 2 = major. (edited, was: 1 minor, 2 medium, 3 major)
3, Added suggestion for not using Stardiver while buffed (checklist section 6). Minor suggestion always.
4, Re-formatted LotD table to highlight windows where one of the above issues appeared. Missed casts are marked as errors, missed buffs are listed as warnings. A note also now appears if a window was supposed to be delayed, but cannot be due to boss downtime sections.

### Images n Stuff
Reformatted window listing: 
image

Expanded window with notices (long image): https://www.dropbox.com/s/qi1shvawfmbqv1f/xiva-lotd-pr-preview.PNG?dl=0


src/parser/jobs/drg/modules/BloodOfTheDragon.js
```diff
@@ -224,10 +236,69 @@ export default class BloodOfTheDragon extends Module {
...
+			if (window.start < end) {
+				return window.start
+			}
+		}
+
+		return null
+	}
+
+	_analyzeLifeWindows() {
+		for (const window of this._lifeWindows.history) {
```
ackwell: nit; aite this one scared me a tad for a second there, 'till i read the loop arg

it's _fine_ per se, but it's shadowing the global `window` object, so it'll fail pretty catastrophically if this line gets misspelled.

src/parser/jobs/drg/modules/BloodOfTheDragon.js
```diff
@@ -274,40 +410,65 @@ export default class BloodOfTheDragon extends Module {
...
-		const delayBuffs = Object.keys(buffsInDelayWindow).filter(id => buffsInDelayWindow[id]).map((id, idx) => {
+		const delayBuffs = Object.keys(window.buffsInDelayWindow).filter(id => window.buffsInDelayWindow[id]).map((id, idx) => {
 			const action = getDataBy(ACTIONS, 'id', parseInt(id))
 			return <Message.Item key={idx}><Trans id="drg.blood.delay-buff"><ActionLink {...action} /> in {this.parser.formatDuration(window.timeToNextBuff[id])}</Trans></Message.Item>
 		})
 
 		return <Fragment>
-			{canBeDelayed && couldBeDelayed && (
+			{window.isLast && (
+				<>
```
ackwell: nit; only need fragments (`<>...</>`) if you're wrapping multiple top-level JSX elements (the act as an invisible "root" node, as you always need one). when you do need them, they have the same structural effect as parens.

should only ever need parenthesis _or_ fragments, not both - in this case, parens should be fine

ebshimizu:
Fixed up some of the overloads on `window` and got the fragments sorted. Good to know on both counts. It didn't occur to me that top level `window` was still in scope at this part of the project but it makes sense.

PR #762, opened 2020-01-25
DRG: Allow partial buff windows

ebshimizu:
Partial windows are now tracked and marked as such.

Also reformatted the rotation tables to use the RotationTable component, and added a column to indicate when a window starts on a non-optimal GCD.
image


src/parser/jobs/drg/modules/Buffs.js
```diff
@@ -251,70 +252,69 @@ export default class Buffs extends Module {
...
-					</Fragment>,
-				},
-				content: {
-					key: 'content-' + window.start,
-					content: <Rotation events={window.casts}/>,
-				},
+	_formatWindowTable(statusId) {
+		return this._buffWindows[statusId].history.map(buffWindow => {
+			const targetsData = {}
+			const notesMap = {}
```
ackwell: teeeechnically not a `Map`, but weh

PR #766, opened 2020-02-09
Replace FFLogsEventNormaliser compatibility shim with NormalisedEvents

AzariahAeroborn:
For this pass, replacing all remaining dependencies on the `calculateddamage` shim except AOE (since AOE will be removed in the next pass and things using it will be migrated to NormalisedEvents).

I confirmed via some E2S logs that the detections for NIN, MNK, and BRD that previously used the shim are getting the same number of detections with NormalisedEvents.
http://localhost:3000/analyse/kRyJafMNcKnh2XGm/10/51/
http://localhost:3000/analyse/wfKv7H3qmjcdn68B/5/7/
http://localhost:3000/analyse/cn29xvh3WmAKbTaH/23/1201/
http://localhost:3000/analyse/9ZYy6k4RmLVwc2XT/7/1/


src/parser/core/modules/Entities.js
```diff
@@ -21,10 +17,8 @@ export default class Entities extends Module {
 		this.addEventHook('refreshdebuff', event => this.refreshBuff(event, true))
 
 		// Resources - hooked to init to make sure normaliser runs to determine damage event before hooking events
```
ackwell: comment is outdated

src/parser/jobs/mnk/modules/Steppies.tsx
```diff
@@ -47,21 +43,20 @@ export default class Steppies extends Module {
...
+		this.addEventHook('normaliseddamage', {by: 'player', abilityId: ACTIONS.BOOTSHINE.id}, this.onDamage)
+		this.addEventHook('complete', this.onComplete)
 	}
 
-	private onDamage(event: DamageEvent): void {
-		const boot = new Boot(event.criticalHit, this.combatants.selected.hasStatus(STATUSES.LEADEN_FIST.id), event.timestamp)
-		this.steppies.push(boot)
+	private onDamage(event: NormalisedDamageEvent): void {
+		if (event.hits > 0) {
+			const boot = new Boot(event.confirmedEvents[0].criticalHit, this.combatants.selected.hasStatus(STATUSES.LEADEN_FIST.id), event.timestamp)
```
ackwell: thought; drilling down to `confirmedEvents[0]` for a single-target skill is a bit of ick - might be nice to improve in future iterations

src/parser/jobs/brd/modules/Sidewinder.js
```diff
@@ -276,51 +262,4 @@ export default class Sidewinder extends Module {
 		}
 		return rawDamage
 	}
-
-	_addInSingleTargetShadowbites() {
```
ackwell: I'm gonna trust you on this lot :meowsweats:

AzariahAeroborn: This function was able to be obsoleted by determining if a shadowbite was ST based on the normaliseddamage hits property, so this module doesn’t have to do the event deduplication on its own any more.

src/parser/jobs/mnk/modules/Steppies.tsx
AzariahAeroborn: Yeah, I wasn’t thrilled with this either. I guess the question is whether this approach is more or less bad than trying to come up with a rule for how to mark a multi hit as isCritical/isDH. I’m not really comfortable with a tri-Boolean (eg returning null on multi hits), and it might be a little nonsensical to return isCritical = true if 1 out of 7 AOE hits crits but the other 6 don’t.

ackwell: Yeah, totally gonna agree on that. Something to think about.

PR #769, opened 2020-02-13
Adjusted allowedDowntime and firstUseOffset for SMN cooldowns.

KelosPelmand:
Most important part here is getting rid of the 4s fuzz time leftover from the previous implementation.  First use offsets also added to improve accuracy.

### DIFF
--- a/src/parser/jobs/smn/modules/GeneralCDDowntime.ts
+++ b/src/parser/jobs/smn/modules/GeneralCDDowntime.ts
@@ -25,7 +25,7 @@ export default class GeneralCDDowntime extends CooldownDowntime {
 			ACTIONS.ASSAULT_I_EARTHEN_ARMOR,
 			ACTIONS.ASSAULT_I_CRIMSON_CYCLONE,
 		],
-		firstUseOffset: 1000,
+		firstUseOffset: 3500,
 		// isAffectedBySpeed: true,
 	}, {
 		cooldowns: [
@@ -34,6 +34,7 @@ export default class GeneralCDDowntime extends CooldownDowntime {
 			ACTIONS.ASSAULT_II_FLAMING_CRUSH,
 		],
 		firstUseOffset: 3500,
+		// isAffectedBySpeed: true,
 	}, {
 		cooldowns: [
 			ACTIONS.ENKINDLE_AERIAL_BLAST,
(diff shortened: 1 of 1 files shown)


src/parser/jobs/smn/modules/GeneralCDDowntime.ts
```diff
@@ -1,38 +1,50 @@
...
+	}, {
+		cooldowns: [
 			ACTIONS.ASSAULT_I_AERIAL_SLASH,
 			ACTIONS.ASSAULT_I_EARTHEN_ARMOR,
 			ACTIONS.ASSAULT_I_CRIMSON_CYCLONE,
-		]},
-		{cooldowns: [
+		],
+		firstUseOffset: 1000,
+		// isAffectedBySpeed: true,
```
AzariahAeroborn: Leftover from an experiment?  Can this be removed?

src/parser/jobs/smn/modules/GeneralCDDowntime.ts
```diff
@@ -1,38 +1,50 @@
...
+	}, {
+		cooldowns: [
 			ACTIONS.ASSAULT_II_SLIIPSTREAM,
 			ACTIONS.ASSAULT_II_MOUNTAIN_BUSTER,
 			ACTIONS.ASSAULT_II_FLAMING_CRUSH,
-		]},
-		{cooldowns: [
+		],
+		firstUseOffset: 3500,
```
AzariahAeroborn: Is it intentional to enforce the order of using Assault 1 then Assault 2 in neighboring weave slots, or would it also be acceptable to use them in the other order, such that both should have an allowable offset of 3.5s?

src/parser/jobs/smn/modules/GeneralCDDowntime.ts
KelosPelmand: It's something that's defined in a comment on CooldownGroup as something we might implement in the future if we get confident in our GCD length estimates.  It would apply to the other assault as well if we leave it on here.

src/parser/jobs/smn/modules/GeneralCDDowntime.ts
KelosPelmand: I listed them in the order they are presented in the opener, but, no, there's no real reason you have to do one before the other.  I'll set them both to 3.5.

PR #770, opened 2020-02-13
Fix Korean translation (grammar, naturalness)

Fleta:
I recently returned on FFXIV and I found xivanalysis using FFlogs. While using the service, I found some misspellings and unnaturalness of the sentence, and then corrected it to promote it.

Thank you for making a good service, and I hope my contribution helps.

My main job is Monk, so I only worked with content that has the `mnk` tag. I will try to contribute to the translation of other jobs.


AzariahAeroborn:
@Fleta Thanks for your interest!  We do our translations through Crowdin, at https://crowdin.com/project/xivanalysis - if you can add them to the dictionaries there, we'll get them picked up in our next translation merge pass.

PR #772, opened 2020-02-14
Deprecate aoeX and migrate consumers to normalisedX

AzariahAeroborn:
(damage and applybuff were the only two being consumed)

This isn't ready for merge yet, as I haven't been able to test it, since FFLogs is melted this evening, but I wanted to go ahead and post the code changes for review/comment.

Jobs/Modules to test:
- [x] GNB Ammo tracking
- [x] DNC allowable combo
- [x] DNC incorrect dances
- [x] DNC esprit generation
- [x] DNC feather generation
- [x] DNC technical step windows
- [x] DRK darkside
- [x] DRK resource generation
- [x] MNK AOE minimum hit checking
- [x] NIN Doton targets hit
- [x] NIN Ninki single target hellfrog
- [x] RDM Gauge
- [x] SAM AOE minimum hit checking
- [x] SAM combo validation (Meikyo override)
- [x] SMN DWT check deathflare targets hit
- [x] SMN pet AOE checker (Garuda vs Ifrit recommendation)
- [x] SMN AOE minimum hit checking (extends core)
- [x] All class combo tracking

### DIFF
--- a/src/parser/core/modules/Combos.tsx
+++ b/src/parser/core/modules/Combos.tsx
@@ -5,8 +5,8 @@ import {Plural, Trans} from '@lingui/react'
 import {RotationTable} from 'components/ui/RotationTable'
 import _ from 'lodash'
 import Module, {dependency} from 'parser/core/Module'
-import {AoeEvent} from 'parser/core/modules/AoE'
 import DISPLAY_ORDER from 'parser/core/modules/DISPLAY_ORDER'
+import {NormalisedDamageEvent} from 'parser/core/modules/NormalisedEvents'
 import Suggestions, {SEVERITY, TieredSuggestion} from 'parser/core/modules/Suggestions'
 import Timeline from 'parser/core/modules/Timeline'
 import React from 'react'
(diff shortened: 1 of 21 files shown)


AzariahAeroborn:
Observations while testing:
This change seems to fix an issue with DNC combo detection on old logs: https://xivanalysis.com/analyse/ap49mA3V6Gdq2rzM/18/85/

I haven't found a log that would specifically let me test whether the DNC's combo break forgiveness for timeouts while dancing is working under the refactor.

AzariahAeroborn:
This PR will fix an issue with NIN doton tracking - currently on live, Doton casts for logs with `calculateddamage` events don't show any successful hits (and thus will flag for an unoptimized Doton with not enough ticks - see https://xivanalysis.com/analyse/6RLVj1NazBvpqJTf/2/5/), because the aoedamage module only looks for `calculateddamage` events if that is the shim'd version, and Doton (and other ground target DOT AOEs) only generates `damage` events.

AzariahAeroborn:
I wasn't able to find a SMN test log that had Ifrit hitting 3+ targets (such that the Garuda recommendation would be made).  Originally, that recommendation only required Ifrit hitting 2+ targets, so the test logs from initial implementation don't trigger that recommendation any more.

src/parser/core/modules/Combos.tsx
```diff
@@ -20,14 +20,17 @@ const ISSUE_TYPENAMES = {
 	failedcombo: <Trans id="core.combos.issuetypenames.failed">Missed or Invulnerable</Trans>,
 }
 
-export interface ComboEvent extends AoeEvent {
-	type: 'combo'
+export class ComboEvent extends NormalisedDamageEvent {
+	constructor(event: NormalisedDamageEvent) {
+		super(event)
+		this.type = 'combo'
```
ackwell: Hm, being able to do this assignment is a little unfortunate - the base class' `type: string` means filters and hooks will be a little less strict on types.

That said, it's partially an unfortunate side-effect of using classes. There's no _nice_ way of overriding the `type` field's type as a class.

Something to think about for later.

AzariahAeroborn: In part, that’s why the normalisedEvents module is doing both an interface and a class for the normalisedX classes, which seems to make typescript a bit happier about the event type and keeps it from just being an arbitrary string to TS. But for some reason Combos wouldn’t take that same pattern, and I just kind of gave up on it. Right now there’s only one module that hooks combo events (DRK resource generation), so the damage is fairly contained.

PR #773, opened 2020-02-14
Improved display of SMN dot table

KelosPelmand:
For all fights, but especially those with multiple targets (E1S, E4S, TEA).

### DIFF
--- a/src/parser/jobs/smn/modules/DoTs.js
+++ b/src/parser/jobs/smn/modules/DoTs.js
@@ -41,7 +41,8 @@ export default class DoTs extends Module {
 		'suggestions',
 	]
 
-	_lastCast = undefined
+	_lastBioCast = undefined
+	_lastMiasmaCast = undefined
 	_lastApplication = {}
 	_clip = {
 		[STATUSES.BIO_III.id]: 0,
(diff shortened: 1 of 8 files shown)


src/parser/jobs/smn/modules/DoTs.js
```diff
@@ -153,51 +173,61 @@ export default class DoTs extends Module {
...
 							</Table.Header>
 							<Table.Body>
-								{this._application[STATUSES.MIASMA_III.id].map(
+								{target[STATUSES.MIASMA_III.id].map(
 									(event) => {
 										totalMiasmaClip += event.clip
+										// Depending on latency of status application, the Bio III instant cast may have
+										// resolved before the previous Miasma III went off.  Assume that this the normal
+										// Miasma III -> Bio III cast pattern happened and adjust the id.
+										if (event.source === ACTIONS.BIO_III.id) { event.source = ACTIONS.MIASMA_III.id }
```
AzariahAeroborn: Would it be better to do this check in the _pushApplication routine, rather than this output-related method having side effects?

Could the "assume they went Miasma -> Bio" assumption be mitigating by tracking lastMiasmaCast and lastBioCast separately, and just having TriD and Bane update both of those values?

KelosPelmand: This was mostly a "simplest thing that worked" situation.  These are now split in the new commits.

PR #776, opened 2020-02-18
[SAM] 5.2 Update. Meditate Fix, OGCD tracker updated, AoeChecker converted to core instead of my own code.

Rjb99:
see title.
short and sweet PR. now I just hope meditate won't break in 5.2


Rjb99:
Alright, so updated on this PR. the shoha rework has been rolled in due to my mess up with branch checkouts from today's early issue with graphs.

src/parser/jobs/sam/modules/Shoha.tsx
```diff
@@ -79,41 +82,42 @@ export default class Shoha extends Module {
...
+		const ticks = Math.min(Math.floor(diff / MEDITATE_TICK_FREQUENCY), MAX_MEDITATE_TICKS)
+
+		const generatedStacks = ticks
 
-			// just a safety net
-			if (this.stacks > MAX_STACKS) {
-				this.stacks = MAX_STACKS
-			}
+		this.stacks += generatedStacks
+		this.totalGeneratedStacks += generatedStacks
```
ayyaruq: Are you counting total generated stacks including overcapping? If not, might wanna move this to only happen when `this.stacks <= MAX_STACKS`

src/parser/jobs/sam/modules/OGCDDowntime.ts
```diff
@@ -4,6 +4,8 @@ import {CooldownDowntime} from 'parser/core/modules/CooldownDowntime'
 // GCD timing: 2.5 seconds for first two GCDs (to get Shifu up), 2.18 seconds for all others.
 // meikyo is used after the 7th gcd
 const FIRSTUSEOFFSET_MEIKYO = 15900
+// Senei/Guren is used after 6th gcd
+const FIRSTUSEOFFSET_50KENKI = 13720
```
ayyaruq: How consistent is this, along with the first Meikyo offset?  Lag happens and there may somehow be players with a 2.5s GCD for whatever reason, plus regional differences on openers. It might be fine but I'd probably take a look at some ~90th percentile SAM logs across NA/EU/JP and see what players that play well but aren't necessarily perfect are doing, particularly in JP where they usually don't follow English guides.

PR #778, opened 2020-02-21
fix: Sort normalisedX events into the event array to maintain expected ordering

AzariahAeroborn:
Per discovery today that the gauges break in fun and exciting ways if they don't get events in timestamp order, and that there isn't anything running after all the normalisers to re-sort the events array.

I refactored this out of AdditionalEvents and both AdditionalEvents and NormalisedEvents are now referencing the same sort logic.  The BLM Gauge normaliser is still doing its own sorting of sorts (tracking insertAfter for events it needs to generate and then splicing in at the appropriate point at the end), but on quick scan I'm not seeing other normalisers that are doing sort-like things with their synthetic events (aside from AOE, which is going to be removed after the NormalisedEvents migration is finished).

### DIFF
--- a/src/parser/core/EventSorting.ts
+++ b/src/parser/core/EventSorting.ts
@@ -27,7 +27,7 @@ const EVENT_TYPE_ORDER: {[key: string]: number} = {
 	normalisedapplybuff: 3.5,
 }
 
-export function SortNormalisedEvents(events: Event[]) {
+export function SortEvents(events: Event[]) {
 	return stable.inplace(events, (a, b) => {
 		if (a.timestamp === b.timestamp) {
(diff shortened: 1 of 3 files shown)


AzariahAeroborn:
Testing: 
AST - n/a
BRD - n/a
BLM - n/a
DNC - esprit fixed, feathers fixed
DRK - MP OK on live, blood fixed
DRG - n/a
GNB - cartridges fixed
MCH - heat and battery fixed
MNK - GL OK on live
NIN - ninki fixed
PLD - n/a
RDM - gauge fixed
SAM - meditation OK on live, Kenki fixed
SCH - Faerie gauge OK on live
SMN - n/a
WAR - gauge fixed
WHM - n/a

PR #783, opened 2020-02-22
Event order module

ackwell:
gotta run for a bit, putting this up now
should be 1:1 with old functionality but with less re-runs of the sort
promoting sorting to a proper module with a proper dep list. kinda didn't want to do it, but it's the sanest way forward tbqh

have left bigass message telling people to not add to it :^)

also lots of breathing room in the mapping so we don't need to keep getting smaller decimals


ackwell:
Don't merge this yet. I'm not entirely convinced this is the best solution.

ackwell:
closing, this solution doesn't handle some pretty gnarly edge cases.

someone remind me to fix the casing on the current impl :meowknife:

PR #784, opened 2020-02-22
DRK typescript part 1, mark supported for 5.2

AzariahAeroborn:
This converts everything DRK but the gauges (ResourceSimulator) to TS, and moves multi-hit skill tracking onto the core module.  I'm separating out the gauges because they're going to be a meatier rework on their own (and I'm still wrestling with a few alligators).  

Also marking as 5.2 support, no job changes in 5.2


src/parser/jobs/drk/modules/Darkside.tsx
```diff
@@ -13,28 +17,25 @@ const DARKSIDE_EXTENSION = {
...
-		super(...args)
+	protected init() {
 		this.addEventHook('normaliseddamage', {by: 'player', abilityId: Object.keys(DARKSIDE_EXTENSION).map(Number)}, this._updateDarkside)
 		this.addEventHook('death', {to: 'player'}, this._onDeath)
 		this.addEventHook('raise', {to: 'player'}, this._onRaise)
 		this.addEventHook('complete', this._onComplete)
 	}
 
-	_updateDarkside(event) {
+	_updateDarkside(event: Event | NormalisedDamageEvent) {
```
ayyaruq: Would this not always be a `NormalisedDamageEvent`? Quick look at the type def has `DamageEvent | NormalisedDamageEvent` anyway but idk, hooks are black magic.

AzariahAeroborn: I’m overloading _updateDarksideEvent both as the receiver for the event hook and as a function called from onDeath. If it gets called from onDeath, it gets a death event passed in. 

I should probably refactor a bit so both of the event receiver functions call a third function instead :facepalm:

AzariahAeroborn:
I think Quietus is a bit easier to recognize than the other AoE options, so I picked it for the suggestion icon. The data table will show the specific skills misused and how many, regardless of which icon is used for the suggestion.

src/parser/jobs/drk/modules/Darkside.tsx
ayyaruq: Makes sense, bit iffy to read but checks out so yolo.

PR #785, opened 2020-02-22
Add crit and DH count to normalisedevent interface

AzariahAeroborn:
Per comment on the original normalisedX migration PR and convo on Discord.


ayyaruq:
Looks like this was gonna conflict either way, I'll just rebase #789 then.

PR #788, opened 2020-02-24
Add korean patch date for 5.1

Ra-Workspace:


ayyaruq:
I think this needs the KR dates for 5.0/01/05/08 as well even if they're all the same? Maybe a question for ack.

ackwell:
nah. non-global dates can be sparse. it'll fall back to the closest patch in the past from time of log.

PR #789, opened 2020-02-25
MNK 5.2 support and smol feex for steppies

ayyaruq:


src/parser/jobs/mnk/modules/Weaving.ts
```diff
@@ -1,15 +1,15 @@
...
 export default class Weaving extends CoreWeaving {
-	isBadWeave(weave: TODO /*, maxWeaves*/) {
-		if (weave.leadingGcdEvent &&
-			weave.leadingGcdEvent.ability &&
-			weave.leadingGcdEvent.ability.guid === ACTIONS.SIX_SIDED_STAR.id) {
-				return weave.weaves.length > SSS_MAX_WEAVES
-			}
+	isBadWeave(weave: WeaveInfo /*, maxWeaves*/) {
+		const gcd = weave.leadingGcdEvent as CastEvent
+		if (gcd?.ability && gcd?.ability?.guid === ACTIONS.SIX_SIDED_STAR.id) {
```
AzariahAeroborn: Stylistically, it would be better to add an isCastEvent type discriminator (similar to the isDamageEvent discriminator in fflogs.ts) and import that to keep TS happy, rather than force-casting the event to a CastEvent -- so then the next check could be something like 
```
if (isCastEvent(gcd) && gcd?.ability?.guid === ACTIONS.SIX_SIDED_STAR.id) 
```
In either case, you can collapse the ?. checks into a single check here, it will bail to null and not equal to SSS the first property it finds that is missing.

src/parser/jobs/mnk/modules/TwinSnakes.tsx
```diff
@@ -74,23 +77,23 @@ export default class TwinSnakes extends Module {
...
 			return
 		}
 
 		switch (action.id) {
 		// Ignore TS itself, plus Form Shift and maybe 6SS?
 		// We use gcdsSinceTS because we don't want to double count FPF
 		case (ACTIONS.TWIN_SNAKES.id):
-			if (this.twinSnake && !this.twinSnake.end && this.gcdsSinceTS < TWIN_SNAKES_CYCLE_LENGTH) {
+			if (!this.twinSnake?.end && this.gcdsSinceTS < TWIN_SNAKES_CYCLE_LENGTH) {
 				this.earlySnakes++
```
AzariahAeroborn: This may not be a straight refactor here... https://xivanalysis.com/analyse/BM9hJdHwpxfFN6qg/11/6/ is showing an extra early refresh with this change in place, which looks to be a detection on the very first Twin Snake use.

Before the change, the if is going to return FALSE if this.twinSnake is null/undefined (from the first element), but now it will return TRUE if this.twinSnake is null/undefined and this.gcdsSinceTS is < CYCLE_LENGTH, since this.twinSnake?.end -> null if twinSnake is null/undefined, and !null -> true from a boolean logic perspective.

src/parser/jobs/mnk/modules/Forms.tsx
```diff
@@ -2,13 +2,13 @@ import {Plural, Trans} from '@lingui/react'
...
 
 import {ActionLink, StatusLink} from 'components/ui/DbLink'
-import {getDataBy} from 'data'
 import ACTIONS from 'data/ACTIONS'
 import STATUSES from 'data/STATUSES'
 import {BuffEvent, CastEvent} from 'fflogs'
 
 import Module, {dependency} from 'parser/core/Module'
 import Combatants from 'parser/core/modules/Combatants'
+import {Data} from 'parser/core/modules/data'
```
AzariahAeroborn: Getting a compile time warning about the lower case `data` here instead of `Data` everywhere else.  Probably worth being consistent to prevent possible case-sensitive-filesystem excitement in the future.

src/parser/jobs/mnk/modules/Forms.tsx
```diff
@@ -85,12 +86,12 @@ export default class Forms extends Module {
...
 			case STATUSES.COEURL_FORM.id:
-				if (OPO_OPO_SKILLS.includes(action)) { this.resetForms++ }
+				if (OPO_OPO_SKILLS.includes(action.id)) { this.resetForms++ }
 				break
 
 			default:
 				// No form used
-				if (OPO_OPO_SKILLS.includes(action)) {
+				if (OPO_OPO_SKILLS.includes(action.id)) {
 					this.formless++
```
AzariahAeroborn: Does this need a check for whether they're under perfect balance?  Showing 19 formless skill usages for http://localhost:3000/analyse/BM9hJdHwpxfFN6qg/11/6/ (compared to no suggestion being made on live), and looking at the ones that are detecting as formless, 16 of them occurred while under the Perfect Balance status.

Also, the recommendation specifically says "Avoid Using Dragon Kick outside of Opo-Opo Form", but this is going to ding if any of the three Opo-Opo skills are used formless, is that intended?  If so, should the recommendation copy be updated?

src/parser/jobs/mnk/modules/Steppies.tsx
```diff
@@ -100,15 +107,9 @@ export default class Steppies extends Module {
 		}))
 	}
 
-	getUnbuffedCount(boots: Boot[]): number {
-		return boots.reduce((total, current) => current.weak ? total : total+1, 0)
-	}
+	getUnbuffedCount = (boots: Boot[]): number => boots.filter(boot => boot.weak).length
```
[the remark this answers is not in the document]
ayyaruq: This seems to have just been when I create, I'm flagging weak as "has Leaden Fist" instead of not having it. Filtering on `boot.weak` is correct, just the value on init was switched.

src/parser/jobs/mnk/modules/TwinSnakes.tsx
ayyaruq: Hm I'm just gonna revert this one then.

src/parser/jobs/mnk/modules/Forms.tsx
ayyaruq: I think this was just broken before. It does need a PB check, but I'm wondering if that will make hide bad dropped forms due to bad PB timing since the check would need to be before the form switch. I'm not sure that really matters tho besides in the opener, but we do kinda care what GCD was used immediately before a PB I guess. Maybe a job for the PB module tho, and thinking of doing similar in RoF and adding BH at some point. Prob need to talk to some expert monks.

ayyaruq:
Gonna leave the weaving stuff for now, `as CastEvent` is gross but I think a larger refactor of the core weaving module is gonna bloat a MNK update too much. Better to handle in a separate PR.

src/parser/jobs/mnk/modules/Steppies.tsx
```diff
@@ -53,8 +55,13 @@ export default class Steppies extends Module {
 	}
 
 	private onDamage(event: NormalisedDamageEvent): void {
-		if (event.hitCount > 0) {
-			const boot = new Boot(event.criticalHits > 0, this.combatants.selected.hasStatus(STATUSES.LEADEN_FIST.id), event.timestamp)
+		if (event.hits > 0) {
```
AzariahAeroborn: TS error - this needs to be event.hitCount due to the signature/naming changes for NormalisedDamageEvent from #794

PR #792, opened 2020-02-27
Proctimeline

Cephid:
Added Verstone, Verfire procs to the Timeline.  Extremely minor cleanup


src/parser/jobs/rdm/modules/Procs.js
```diff
@@ -5,8 +5,40 @@ import {ActionLink, StatusLink} from 'components/ui/DbLink'
 import ACTIONS from 'data/ACTIONS'
 import STATUSES from 'data/STATUSES'
 import Module from 'parser/core/Module'
-import {TieredSuggestion} from 'parser/core/modules/Suggestions'
-import {PROCS, SEVERITY_MISSED_PROCS, SEVERITY_OVERWRITTEN_PROCS, SEVERITY_INVULN_PROCS} from 'parser/jobs/rdm/modules/ProcsEnum'
```
AzariahAeroborn: With bringing this into the module directly, is there anything else using the ProcsEnum, or can that be deleted?

src/parser/jobs/rdm/modules/Procs.js
```diff
@@ -36,29 +61,55 @@ export default class Procs extends Module {
...
-		}, this._onRemove)
-		this.addHook('refreshbuff', {
-			to: 'player',
-			abilityId: PROCS,
-		}, this._onRefresh)
+		this.addHook('applybuff', {by: 'player', abilityId: PROCS}, this._onGain)
+		this.addHook('removebuff', {by: 'player', abilityId: PROCS}, this._onRemove)
+		this.addHook('refreshbuff', {by: 'player', abilityId: PROCS}, this._onRefresh)
 		this.addHook('complete', this._onComplete)
+		//this.addHook('death', { to: 'player' }, this._onDeath)
```
AzariahAeroborn: Is this something planned, or left over from experimenting that should be removed?

src/parser/jobs/rdm/modules/Procs.js
Cephid: It appears not, I had comments about Dualcast in there so I had ignored it.  I'll push another commit with it deleted here shortly.

src/parser/jobs/rdm/modules/Procs.js
Cephid: It was from initial experiments, but with how I finished out onComplete I have what I need for onDeath for the timeline at least.  So I've implemented it for now; it'll be in this upcoming commit.

PR #795, opened 2020-02-28
Update support for WHM to 5.2

nonowazu:
Since no analysis-based breaking changes were made, this should be fine.

### DIFF
--- a/src/parser/jobs/whm/index.js
+++ b/src/parser/jobs/whm/index.js
@@ -35,7 +35,7 @@ export default new Meta({
 	</>,
 	supportedPatches: {
 		from: '5.0',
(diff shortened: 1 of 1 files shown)


ayyaruq:
I'm fine for merging this but is there any change to minimum targets for AoE with the adjustments to damage falloff? Just double checking @nonowazu, since you weren't on Discord that I could find.

nonowazu:
@ayyaruq  - I'm in discord under the name Nono in the healer channel.

Given that the current WHM analysis code doesn't track AoEs, I don't think there's any adjustments in the code to make.

ayyaruq:
Neato, a-mergin' we will go.

PR #796, opened 2020-03-06
Fix minor error of Korean locale file

P17seongbin:
Line 740. 
『최대 가능 횟수 {0} 회 중 {1} 회가 사용되었습니다』 means 『{1} out of {0} possible』 in Korean.


ayyaruq:
Hi @P17seongbin, thanks for your contribution! Your translation fix looks good but the PR is failing to lint. We use CrowdIn for our translations, you can find it [here](https://crowdin.com/project/xivanalysis/ko#). It's simpler to have it in one place and generate PRs for us than to do it all in GitHub. The key in the file you edited is called the context in CrowdIn so you could be able to easily find it there 😃

PR #799, opened 2020-03-07
Add Interrupts to core modules

nonowazu:
This is my first attempt at converting something to TS, and because this is core I really would like the maximum amount of feedback that I can get for this.

* `Interrupts.tsx` core module
* Convert SCH Interrupts to use core module


src/parser/core/modules/Interrupts.tsx
```diff
@@ -0,0 +1,143 @@
...
+const TIMELINE_UPPER_MOD: number = 30000
+
+export abstract class Interrupts extends Module {
+	static handle: string = 'interrupts'
+	static title: MessageDescriptor = t('core.interrupts.title')`Interrupted Casts`
+
+	@dependency private timeline!: Timeline
+	@dependency private globalCooldown!: GlobalCooldown
+	@dependency private suggestions!: Suggestions
+	@dependency private data!: Data
```
ayyaruq: Mostly a style thing but may wanna order these by alphabet.

src/parser/core/modules/Interrupts.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+	private currentCast?: CastEvent | undefined
```
ayyaruq: Optional variables are undefined automatically until you assign something to them.

src/parser/core/modules/Interrupts.tsx
```diff
@@ -0,0 +1,143 @@
...
+	private droppedCasts?: CastEvent[] = []
```
ayyaruq: Since this is being assigned, you can straight up `droppedCasts` without the optional indicator.

src/parser/core/modules/Interrupts.tsx
```diff
@@ -0,0 +1,143 @@
...
+		// that's an interrupted cast
+		if (this.currentCast && guid !== this.currentCast.ability.guid) {
+			this.pushDropCasts(event)
+		}
+		this.currentCast = undefined
+	}
+
+	private pushDropCasts(currentEvent: CastEvent) {
+		this.missedTimeMS += Math.min(this.globalCooldown.getEstimate(), currentEvent.timestamp - this.currentCast!.timestamp)
+		this.droppedCasts!.push(currentEvent)
```
ayyaruq: Since `droppedCasts` is initialised as an empty array, it'll always exist. When you remove the `?` in it's declaration, this ignore null check isn't required. Is using the GCD here a good idea though? I think using the current cast timestamp should be fine, although I'm not sure how safe ignoring it being null is.

src/parser/core/modules/Interrupts.tsx
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	private onComplete() {
+		this.suggestions.add(new TieredSuggestion({
+			icon: this.icon,
+			tiers: this.severity,
+			value: this.droppedCasts!.length,
```
ayyaruq: This can be straight up `this.droppedCasts.length` once you feex the optionalness never being used.

src/parser/core/modules/Interrupts.tsx
```diff
@@ -0,0 +1,143 @@
...
+			content: this.suggestionContent,
+			why: this.suggestionWhy(this.droppedCasts!, this.missedTimeMS),
```
ayyaruq: Same here.

src/parser/core/modules/Interrupts.tsx
```diff
@@ -0,0 +1,143 @@
...
+					<strong><Trans id="core.interrupts.table.time">Time</Trans></strong>
+				</Table.HeaderCell>
+				<Table.HeaderCell>
+					<strong><Trans id="core.interrupts.table.cast">Cast</Trans></strong>
+				</Table.HeaderCell>
+			</Table.Row>
+		</Table.Header>
+		<Table.Body>
+			{
+				this.droppedCasts!.map((cast) =>
```
ayyaruq: And here.

src/parser/core/modules/Interrupts.tsx
nonowazu: I think the reason why I used the estimated GCD here was because there was times where someone would interrupt a cast and then not cast for a while – for example, a BLM might not be able to do anything during E3S transition, which might mean you have a 30s 'gap' in your interrupted casts, and it seemed unfair to penalize them for more than the gcd estimate.

I'm very open to suggestions here, though.

ayyaruq: That explanation seems reasonable to me, the ABC module should handle if they're straight up not casting vs unable to cast.

src/parser/core/modules/Interrupts.tsx
```diff
@@ -0,0 +1,146 @@
...
+	 */
+	protected severity: SeverityTiers = {
+		2: SEVERITY.MEDIUM,
+		5: SEVERITY.MAJOR,
+	}
+
+	/**
+	 * Implementing modules MAY override the default suggestion text
+	 */
+	protected suggestionContent: JSX.Element | string = <Trans id="core.interrupts.suggestion.content">
```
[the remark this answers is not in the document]
nonowazu: Sure thing – more fixes incoming.

src/parser/core/modules/Interrupts.tsx
```diff
@@ -0,0 +1,146 @@
...
+			this.pushDropCasts(event)
+		}
+		this.currentCast = undefined
+	}
+
+	private pushDropCasts(event: CastEvent) {
+		// we shouldn't hit this, since there has to be a valid event to compare to, but check just to be safe
+		if (this.currentCast) {
+			this.missedTimeMS += Math.min(this.globalCooldown.getEstimate(), event.timestamp - this.currentCast.timestamp)
+			this.droppedCasts.push(event)
```
AzariahAeroborn: This is different behavior (pushing the incoming new event onto the droppedCasts array, rather than the outgoing currentCast event) than what is currently live for SCH:
```
		this._droppedCasts.push({
			action: getDataBy(ACTIONS, 'id', this._currentCast.ability.guid),
			timestamp: this._currentCast.timestamp,
		})
```

Which is resulting in the table displaying the "new cast" and time, rather than the interrupted cast.

PR #802, opened 2020-03-09
Modify Weaving in core for message flexibility

nonowazu:
* Add `_suggestionIcon`, `_suggestionContent`, and `_severity` and set
them to the existing defaults so nothing changes for anyone using it
right now
* Modify SCH usage of Weaving to use this new method so I'm not being a
terrible SCH and overriding private methods while still keeping
SCH-specific help visible
* Prepare 9,000 feex bux to bribe Ack for inclusion


src/parser/core/modules/Weaving.js
```diff
@@ -35,6 +35,13 @@ export default class Weaving extends Module {
 
 	static title = t('core.weaving.title')`Weaving Issues`
 
+	// WVR Focused synth lmao
```
ackwell: keeping the comment :meowjustright:

src/parser/core/modules/Weaving.js
```diff
@@ -35,6 +35,13 @@ export default class Weaving extends Module {
...
+	_suggestionIcon = 'https://xivapi.com/i/001000/001785.png'
```
ackwell: only private properties should be prefixed with `_`, given these are "protected", can omit the prefix

nonowazu: They've been protected - secret service isn't cheap though.

ackwell:
Sorry about the RTT on this.

@ayyaruq cmon fam there was even a review which was explicitly addressed ya coulda hit the big fat merge button

PR #805, opened 2020-03-09
[GNB] Quick Simple things

Rjb99:
Just simple things from Lhea's existing notes.

-Added feedback for leftover ammo
-Adjusted the offsets for OGCDs
-Made it so that Both zones are treated the same in regards to the modules that use them
(This is a largely useless feature, but things like UWU and UCoB have a use for it I suppose, so I did it)

Updated the changelog and index to reflect the above and the fact I'm now maintaining GNB


src/parser/jobs/gnb/modules/NoMercy.tsx
```diff
@@ -119,7 +119,10 @@ export default class NoMercy extends Module {
 			switch (actionId) {
 				// Blasting Zone is just a traited version of Danger Zone
 				// Impossible to see both in the same log outside of hacking
+				// Ryan Note: Because of this, I'm going to roll the logic of danger zone under blasting zone so lvl 70's still get their zone counted for.
 				case ACTIONS.BLASTING_ZONE.id:
+					lastNoMercy.numBlastingZones++
+					break
```
ackwell:
```suggestion
```

This addition doesn't actually do anything - in JS, case statements fall through unless you break them, so

```js
// this switch
switch (colour) {
	case RED:
	case BLUE:
		paint(colour)
		break
}
// is doing the same thing as this one
switch (colour) {
	case RED:
		paint(colour)
		break
	case BLUE:
		paint(colour)
		break
}
```

src/parser/jobs/gnb/modules/Ammo.tsx
```diff
@@ -170,6 +175,18 @@ export default class Ammo extends Module {
 			value: this.erroneousCircles,
 		}))
 
+		this.suggestions.add(new TieredSuggestion({
+			icon: ACTIONS.BLOODFEST.icon,
+			content: <Trans id="gnb.ammo.leftover-ammo.content">
+				Avoid having leftover ammo at the end of a fight, consider using the ammo earlier if possible.
+			</Trans>,
+			why: <Trans id="gnb.ammo.leftover-ammo.why">
+				you had <Plural value={this.leftoverAmmo} one="# cartridge" other="# cartridges"/> leftover at the end of the fight.
```
ackwell:
```suggestion
				You had <Plural value={this.leftoverAmmo} one="# cartridge" other="# cartridges"/> remaining at the end of the fight.
```

src/parser/jobs/gnb/modules/Ammo.tsx
```diff
@@ -170,6 +175,18 @@ export default class Ammo extends Module {
 			value: this.erroneousCircles,
 		}))
 
+		this.suggestions.add(new TieredSuggestion({
+			icon: ACTIONS.BLOODFEST.icon,
+			content: <Trans id="gnb.ammo.leftover-ammo.content">
+				Avoid having leftover ammo at the end of a fight, consider using the ammo earlier if possible.
```
ackwell: Might be worth adding a brief _why_. Why should I avoid having ammo left over?

src/parser/jobs/gnb/modules/NoMercy.tsx
Rjb99: TIL. I assumed since the notes left said it wasn't a thing that that code did nothing. Reverted.

src/parser/jobs/gnb/modules/Ammo.tsx
Rjb99: Expanded on this, Burst Strike > Any GCD in the standard combo. Ammo combo is still better than it though.

src/parser/jobs/gnb/index.tsx
```diff
@@ -28,6 +28,7 @@ export default new Meta({
 	contributors: [
 		{user: CONTRIBUTORS.LHEA, role: ROLES.DEVELOPER},
 		{user: CONTRIBUTORS.QAPHLA, role: ROLES.DEVELOPER},
+		{user: CONTRIBUTORS.RYAN, role: ROLES.MAINTAINER},
```
ayyaruq, line 31: :blobsweatsbutfast:

src/parser/jobs/gnb/modules/OGCDDowntime.tsx
```diff
@@ -11,6 +11,10 @@ const FIRST_USE_OFFSET_NO_MERCY = 10000
  */
 const FIRST_USE_OFFSET_BLOODFEST = 22000
 
+const FIRST_USE_OFFSET_BOWBLAST = 17500 // The BowBlast combo is appiled at a maximum of 7 GCDs in current viable openers but can be applied as soon as 5 GCDs
```
ayyaruq, line 14: Ultranit, to reduce further maintenance if this ever changes, would you not want to separate Bowshock and Pewpew Zone? Also did you validate on high damage JP GNBs that they're not doing some weird opener? Obviously don't need to look at low end weirdness, but might be good for high end. NBD if you did tho.

src/parser/jobs/gnb/modules/Ammo.tsx
```diff
@@ -170,6 +175,18 @@ export default class Ammo extends Module {
 			value: this.erroneousCircles,
 		}))
 
+		this.suggestions.add(new TieredSuggestion({
```
ayyaruq, line 178: Future TODO, might be worth attempting to figure out why they had extra ammo. One is unavoidable sometimes and a minor on it is a bit eh. I don't think the other tanks (or most DPS) ping people for leftover gauge etc so it stands out a bit.

Rjb99: I'll have to think on this. Currently the idea behind this is that Ammo GCDs > Skill GCDs, so they should only have leftover ammo if: 
a) They didn't use ammo and just used a weaponskill
b) had 2 ammo before entering a ammo combo and couldn't get to use the second ammo
c) they popped bloodfest like literally before the thing died in vain

The reason one ammo leftover is set to minor is it's on a range from 300-100 potency lost, where as 2 ammo is 800-600 based on what weaponskill they used instead of it. (possibly more if they tossed away an ammo combo, but we yell elsewhere for that mistake)

src/parser/jobs/gnb/index.tsx
Rjb99: I know.

src/parser/jobs/gnb/modules/OGCDDowntime.tsx
Rjb99: Updated the files to the 4 current openers, and separated the offsets to allow for easier changes should a opener fall out of favor and the skills are used sooner.

(Also pewpew zone is here to stay now)

Rjb99: GNB_openers

PR #807, opened 2020-03-17
UI Updates

ceg9498:
In the process of ripping out Semantic & providing comparative custom components.

So far, Header, Table, & Accordion components have been made & implemented in all known locations. These components have been designed to mimic Semantic:
- Visuals: have been matched to Semantic so they'll fit in with the current site
- Functionality: for functionality currently being used

I've also adjusted the CSS to use CSS variables for upcoming dark theme - theming can/will be implemented once all components can support it.


ceg9498:
Closing this since it's bigger than intended. Will submit components separately.

PR #808, opened 2020-03-19
Update theme module css.

ceg9498:
Note: most/all the components I've written rely on the variables here.


ackwell:
Sorry for being slooow, and big thanks for doing all of this!

---

I'm a _little_ hesitant about using `var()` cross-file. I'd initially intended to keep the variable handling in `theme`, and let consumers pull in the css-module `@value`s blindly. Keeps the code cleanly modulated, even if the global variable tree does get quite polluted at runtime. (Can adjust theme at runtime with a class on a high-sitting element that re-defines the theme vars).

That said, I'm not going to go be some draconian overlord about original plans. Do you have any strong feelings / do you forsee any incoming issues with using module values? Happy to go with raw vars if it's going to ease long-term impl. I... realise that's a lot of code churn now it's been split over multiple prs :notlikemeow:.

---

Just OOI, do you expect a theoretical dark mode to invert the text/background light/dark colour picks?

ceg9498:
No worries, take your time!

---

I was mostly using var() because that's what I knew how to use, and was having trouble finding info on / figuring out how to use css-module `@value`s dynamically, but with a bit more digging today I think it finally clicked in my head.

I don't really have strong feelings about it either way. I'm generally happy to learn about new things and such, so we can do the css-module method.

Here's a sample of what I came up with today:
```css
.light-mode {
  --text: rgba(0, 0, 0, 0.87);
  --background: themeWhite;
  --backdrop: themeLight;
  color: var(--text);
  background-color: var(--backdrop);
}

.dark-mode {
  --text: white;
  --background: #2d2d2d;
  --backdrop: #3d3d3d;
  color: var(--text);
  background-color: var(--backdrop);
}

@value themeText: var(--text);
@value themeBackground: var(--background);
@value themeBackdrop: var(--backdrop);
```
`.light-mode` or `.dark-mode` would sit on a root element (locally I'm using the root fragment in `App.js` as a div, but I've no strong preference on this either). Then, as it is in master, components like Segment would pull in the variables needed such as `themeBackground` to apply the color. At least, that's my understanding of how it would work. If this is correct I'll push the changes.

---

For color swapping, I wasn't originally going to just invert the grey colors, though I did consider it a couple of days ago. For the message box I was going to invert the colors as a lazy solution, but some of them (yellow) look yucky. The color values will most definitely need more work in general -- for example, the role colors for `CombatantList` live in JS-Land, so as it is I can't easily/cleanly adjust them based on the theme, and since the background is semi-transparent the text becomes completely illegible in dark mode. I was intending on finalizing color choices once custom components are complete, just before Dark Mode gets enabled.

ackwell:
aaaaaaaagh fuck i missed the replllly. too busy with the bloody timeline, my bad

yeah, that's spot on. that way, the component that "owns" the CSS variables has _full_ control over them - and everything else imports them over `@value`, which is easily traced within the codebase.

i'm not particularly fussed on _where_ the light/dark theme class ends up. the app component is a bit overloaded as-is, but it's kinda appropriate none-the-less.

finalising colors later sgtm. colors being hard-coded in js isn't something that needs to stay as-is - esp. with TS it's relatively simple to type+encode them as a string union which corresponds to classes/values pulled from css.

ceg9498:
Awesome, good to know. And I'm not worried, I had to wait for timeline to be finished before I could finalize the theming work anyways.

ackwell:
oh shit i nearly merged this before realising some of the existing vars would be lost

can either put one theme (probably light, as it's the og) on `:root` and have the dark theme "override" it, or can hold off merging until everything is good to go - whichever suites

ceg9498:
From what I remember, it's only `themeInverse` that's being removed, and I don't believe it was being used anywhere.

ackwell:
It's also getting `themeText` (which has minimal effect, as it's on the `body` anyway), but also `themeBackdrop`, which is the light grey in the content area background.

src/theme.module.css
```diff
@@ -42,12 +61,72 @@
...
+	--warning-text: warningLight;
+	--error-background: errorDark;
+	--error-text: errorLight;
+
+	color: var(--text);
+	background: var(--backdrop);
+}
+
+
+@value themeText: var(--text);
```
ceg9498, line 116: `themeText` and `themeBackdrop` are showing up as new here. I changed the CSS var to be a bit less wordy.

PR #809, opened 2020-03-19
Button component

ceg9498:
Includes a specialized button for Timestamps.


src/akkd/Button/Button.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+interface Props {
+	icon?: any,
+	content?: any,
+	nowrap?: boolean,
+	onClick: ()=>void,
+	pill?: boolean,
+}
+
+export class Button extends React.PureComponent<Props> {
```
ackwell, line 14: nit; this would be a bit cleaner as a `React.memo`'d function component

src/akkd/Button/Button.tsx
```diff
@@ -0,0 +1,38 @@
...
+			icon,
+			content,
+			nowrap,
+			onClick,
+			pill,
+			children,
+		} = this.props
+
+		return (
+			<button className={classNames(
```
ackwell, line 26: nit; generally preference splitting attrs onto next line, i.e.

```tsx
return (
	<button
		className={classNames(
			// ...
		)}
		onClick={onClick}
	>
		...
	</button>
)
```

ceg9498:
Should all be fixed now, but not entirely sure if I handled children in the best way.

PR #810, opened 2020-03-19
Add Header component

ceg9498:


src/akkd/Header/Header.module.css
```diff
@@ -0,0 +1,40 @@
+@value gutter, md from 'theme.module.css';
+
+.header {
+	color: var(--theme-text);
+	position: relative;
+	margin: calc(2rem - .14em) 0 1rem;
```
ackwell, line 6: nit; stuff like the `.14em` would work well as a `@value`

---

purely OOI, where's the .14 coming from? existing semantic stuff?

ceg9498: Yeah, that's exactly where it's coming from (but rounded, because before it was something like .14328582958960368)
I can absolutely change it to something more static and/or rounded, but wasn't sure what kind of value would work best.

ackwell: Huh, so it is. Wonder how they're getting that value themselves - 10/10 bet it's some pre-processor calc.

ackwell:
I kinda let all this slip, hard. Sorry. Between motivation and the analyser migration, it just didn't happen.

I'd like to revisit the UI work at some point in the future (the current UI is starting to really show its age), but for now, I'll be closing these to clear the air.

Once EW is stable, I want to put together a concerted effort to work on the UI from the ground.

PR #811, opened 2020-03-19
Message component

ceg9498:
Update message component to include a box style for ability to replace message boxes on analysis pages.


src/akkd/Message/Message.tsx
```diff
@@ -9,39 +9,45 @@ interface MessageTypes {
...
 	'error',
 	'warning',
 	'info',
 	'success',
+	'default',
 ]
 
 interface Props extends MessageTypes {
+	as?: React.ReactType,
+	className?: string,
```
ackwell, line 25: I'm a little adverse to passing classNames in, though I recognise `Container` does it. is it something avoidable or does it bring significant benefit?

ceg9498: IIRC, I'd passed classNames in to a few components because classNames were currently being passed through somewhere, but AFAIK there's no actual benefit.

PR #812, opened 2020-03-19
Accordion component

ceg9498:
Adding an accordion component. It is capable of taking either panels as an object, or using subcomponents (with tweaks) as is implemented in the ChangeLog.


src/akkd/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,61 @@
+import React from 'react'
+import styles from './Accordion.module.css'
+import {AccordionContent} from './AccordionContent'
+import {AccordionPanel} from './AccordionPanel'
+import {AccordionTitle} from './AccordionTitle'
+
+interface Panel {
+	key: any,
```
ackwell, line 8: Try to avoid using `any`, it's usually avoidable. `key` can _probably_ be `React.Key`, and `content` is probably `React.ReactNode`.

src/akkd/Accordion/AccordionPanel.tsx
```diff
@@ -0,0 +1,36 @@
+import classNames from 'classnames'
+import React from 'react'
+import styles from './Accordion.module.css'
+import {AccordionContent} from './AccordionContent'
+import {AccordionTitle} from './AccordionTitle'
+
+interface Props {
+	panel: any,
```
ackwell, line 8: if you `export` the Panel interface from the other file, should be able to import it for this type

src/akkd/Accordion/AccordionTitle.tsx
```diff
@@ -0,0 +1,26 @@
+import classNames from 'classnames'
+import React from 'react'
+import styles from './Accordion.module.css'
+
+interface Props {
+	index: any,
```
ackwell, line 6: Given index is a straight pass-through, never used, this could naively be types as `unknown`.

To go even more accurate, though, you can genericise the interface with `index: T`, and `onClick?: (event: React.MouseEvent, props: {index: T}) => void` - that way you get a tight typing, but the consumer can use whatever it wants as an index

src/akkd/Accordion/AccordionTitle.tsx
```diff
@@ -0,0 +1,26 @@
...
+export class AccordionTitle extends React.PureComponent<Props> {
+	render() {
+		const {index, className, onClick = () => {}, children} = this.props
+		return (
+			<summary
+				className={classNames(
+					styles.accordionTitle,
+					className,
+				)}
+				onClick={(e)=>onClick(e, {index})}
```
ackwell, line 20: nit; avoid inline function calls in jsx
for class components, can break it out as a private arrow function in the class body, fn components can use a `React.useCallback`

PR #814, opened 2020-03-20
Segment component

ceg9498:
Prepare for Dark Mode.


ceg9498:
This was all assuming using CSS vars, so I'll just close it 😅

PR #820, opened 2020-04-03
Smn Devotion Module

AshleySherron:
This implements a simple information-only module for tracking actions used under the Devotion buff


src/parser/jobs/smn/modules/Devotion.js
```diff
@@ -0,0 +1,144 @@
...
+	_onDevotionRemoved(event) {
+		this._closeWindow(event.timestamp)
+	}
+
+	_onPetDevotionCast(event) {
+		this._openWindow(event.timestamp)
+	}
+
+	_countDevotionTargets(event) {
+		const playersHit = event.confirmedEvents.map(hit => hit.targetID).filter(id => this.parser.fightFriendlies.findIndex(f => f.id === id) >= 0)
```
ayyaruq, line 57: This could prob be broken into a few lines just for readability. Can keep the method chaining, just throw each chain on a new line, like this.
```suggestion
		const playersHit = event.confirmedEvents
			.map(hit => hit.targetID)
			.filter(id => this.parser.fightFriendlies.findIndex(f => f.id === id) >= 0)
```

src/parser/jobs/smn/modules/Devotion.js
```diff
@@ -0,0 +1,144 @@
...
+		playersHit.forEach(id => {
+			if (!this._currentWindow.playersHit.includes(id)) {
```
ayyaruq, line 59: May want to verify `_currentWindow` isn't null.

src/parser/jobs/smn/modules/Devotion.js
```diff
@@ -0,0 +1,144 @@
...
+				this._currentWindow.playersHit.push(id)
+			}
+		})
+	}
+
+	_onCast(event) {
+		const action = getDataBy(ACTIONS, 'id', event.ability.guid)
```
ayyaruq, line 66: You can use `getAction(event.ability.guid)` here I think, bit clearer. No idea if/how it works in JS tho.

src/parser/jobs/smn/modules/Devotion.js
AshleySherron: These changes have all been implemented

KelosPelmand:
Plays SMN. Can verify functionality is good.

ayyaruq:
Awesome, thanks for confirmation @KelosPelmand, and @AshleySherron for the work!

PR #821, opened 2020-04-04
Added pet actions to timeline.

KelosPelmand:
Added as a core module that can be used by any class that wants to show any pet actions on timeline.

It is up to the job if it wants to distinguish between pet "autos" and pet "commands"  if no distinction is made, all actions will show up on the same line.  Jobs will also need to make sure that all pet actions have an icon specified.


src/parser/core/modules/PetTimeline.tsx
```diff
@@ -0,0 +1,87 @@
...
+		} else {
+			this.autoCasts.push(event)
+		}
+	}
+
+	private onComplete() {
+		if (this.canPetBeCommanded()) {
+			this.timeline.addGroup(new Group({
+				id: this.petContainerGroupId,
+				content: 'Pet',
```
ayyaruq, line 45: Could the row name maybe be configurable? `Pet` works for SMN and SCH but sounds a bit weird for MCH or DRK for example.

PR #822, opened 2020-04-04
Added module to show tincture usage.

KelosPelmand:
Display-only, no suggestions or checks.  The buffAction is needed for BuffWindow, but isnt actually used since there are no checks.  The medicated status appears to be the same across all jobs.

The change to the BuffWindow hook is needed to account for buff mirroring to pets.


src/parser/core/modules/Tincture.ts
```diff
@@ -0,0 +1,12 @@
+import {t} from '@lingui/macro'
+import ACTIONS from 'data/ACTIONS'
+import STATUSES from 'data/STATUSES'
+import {BuffWindowModule} from 'parser/core/modules/BuffWindow'
+
+export class Tincture extends BuffWindowModule {
+	static handle = 'tincture'
+	static title = t('core.tincture.title')`Tinctures`
```
ayyaruq, line 8: This could maybe be Medicine or Medicated. Newer players might get confused with pots vs infusions vs tinctures, although we do only care about level cap so tinctures is prob fine. JP calls them all "medicine" anyway, I imagine DE is prob similar.

src/parser/core/modules/Tincture.ts
```diff
@@ -0,0 +1,12 @@
...
+
+	buffAction = ACTIONS.INFUSION_INT
```
ayyaruq, line 10: Why not STR? :mnknife:

KelosPelmand: If you want STR, do it yourself :P  Also, that part never shows up anywhere anyway and is just needed because buffAction is abstract.

src/parser/core/modules/Tincture.ts
KelosPelmand: I'm sure a lot of people call them potions (even though actual Potions don't give you Medicated), but Tinctures is actually what they are at this point.  I'm not tied to the name myself if someone feels strongly.

src/parser/core/modules/Tincture.ts
ayyaruq: Aight, if you can rebase I'll re-approve and merge then.

PR #826, opened 2020-04-05
Add Esteem actions to DRK timeline

ayyaruq:
Along with the actions, also adds the ability to override the action grouping.

Looks like dees:
image


src/parser/core/modules/PetTimeline.tsx
```diff
@@ -18,6 +18,12 @@ export default class PetTimeline extends Module {
 	 * appear on the timeline.
 	 */
 	protected timelineGroupName = 'Pet'
+	/**
+	 * Implementing modules MAY indicate that a pet has a specific summon action.
+	 * If canPetBeCommanded returns true, this field is not used.
+	 * If timelineSummonAction is set to an action ID, actions will be grouped under it.
+	 */
+	protected timelineSummonAction = 0
```
ackwell, line 26:
```suggestion
	protected timelineSummonAction?: number
```

Then check for `this.timelineSummonAction != null`.

0 is technically a valid ID

PR #828, opened 2020-04-06
Timeline neue

ackwell:
This is a fully-functional re-implementation of the timeline UI, with 100% less `vis`.

I'm pulling in some `d3` to handle scaling and such. While `d3`, even as individual modules, is a reasonable code chunk to swallow, I'm expecting the size cost to be reduced at a future point when charts can be moved into `d3` as well.

The new module's codebase is primarily split into 3 sections:
- `config`, which houses the abstract interfaces that the module acts on, as well as some concrete implementations of them that can be consumed by other modules
- `components`, containing the UI that is rendered by the module based on the interfaces from the preceeding, and
- The module itself, which acts as the glue between the two domains, and an external interface for the module system.

Only took 
* ~30 days
* 3 attempts at label handling
* 2 different component paradigm attempts, and
* 1 melty

**TODO List** (in vague but unenforced order)
- [x] Add final label styling
- [x] Add final item styling
- [x] Finalise scale functionality
  - [x] Touch event pinch zoom
  - [x] Touchpad wheel
  - [x] Touchpad pinch
- [x] Integrate with any implementation points I missed
- [x] Remove old timeline callsites
- [x] `timelineNeue -> timeline`
- [x] ***delet vis***
- [x] compare bundle sizes and weep (joy or sorrow? tune in at some point in the future to find out!) (actually came out pretty well, nice job me)
- [x] write a changelog


ackwell:
what could possibly go wrong

PR #829, opened 2020-04-06
Add Swiftcast tracking to core

nonowazu:
* Remembered to run `yarn extract` first this time
* Modified `BuffWindow` to export `BuffWindowExpectedGCDs` interface so I can be a good person and annotate correctly.
* Modify SCH and WHM implementations of Swiftcast to use new core module
* Add Swiftcast tracking to SMN


nonowazu:
I checked with Nemekh, and the default seemed to work for him for context.
As for the single-cell table, it is a tad boring, but the groups seem to want that.

PR #830, opened 2020-04-07
Add pet tracking module to SCH

nonowazu:
* Modify SCH actions to have the correct URL for Seraphic Veil
* Add SchPets class that extends PetTimeline
* Cry because my example parse shows me only using a single consolation during the seraph window


src/parser/jobs/sch/modules/Pets.js
```diff
@@ -0,0 +1,25 @@
...
+	ACTIONS.WHISPERING_DAWN.id,
+	ACTIONS.FEY_ILLUMINATION.id,
+	ACTIONS.CONSOLATION.id,
+	ACTIONS.ANGELS_WHISPER.id,
+	ACTIONS.FEY_UNION.id,
+	ACTIONS.SERAPHIC_ILLUMINATION.id,
+]
+
+export default class SchPets extends PetTimeline {
+	static handle = 'schfaerietimeline'
```
ayyaruq, line 14: NGL I thought this was German.

PR #832, opened 2020-04-09
Bump WAR to 5.25 (latest)

holl0wstar:
Title, more or less.

Warrior should be all good to go and can just get bumped up.


AzariahAeroborn:
No job changes at all for 5.25, so no patch to bump jobs to.

PR #833, opened 2020-04-09
Bugfix negative expected GCDs in Swiftcast module

nonowazu:
* Provide our own `reduceExpectedGCDsEndOfFight` method that totally doesn't just rip off the original at all
* Redefine `expectedGCDs` in init (lemme know if you want me to move this to the constructor instead) so that inheriting classes can provide their own suggestion content and severity tiers.

### DIFF
--- a/src/parser/core/modules/Swiftcast.tsx
+++ b/src/parser/core/modules/Swiftcast.tsx
@@ -63,7 +63,13 @@ export abstract class SwiftcastModule extends BuffWindowModule {
 	// ding them only if they had enough time during the window to use a spell with
 	// swiftcast
 	protected reduceExpectedGCDsEndOfFight(buffWindow: BuffWindowState): number {
-		return Math.min(1, super.reduceExpectedGCDsEndOfFight(buffWindow))
+		if ( this.buffStatus.duration ) {
+			// Check to see if this window is rushing due to end of fight - reduce expected GCDs accordingly
+			const fightTimeRemaining = this.parser.fight.end_time - buffWindow.start
+			const gcdEstimate = this.globalCooldown.getEstimate()
+			return ( fightTimeRemaining > gcdEstimate ) ? 0 : 1
+		}
+		return 0
 	}
 
 	protected considerAction(action: Action) {
--- a/src/parser/core/modules/Swiftcast.tsx
+++ b/src/parser/core/modules/Swiftcast.tsx
@@ -5,7 +5,7 @@ import React from 'react'
 import {ActionLink} from 'components/ui/DbLink'
 import ACTIONS, {Action} from 'data/ACTIONS'
 import STATUSES, {Status} from 'data/STATUSES'
-import {BuffWindowExpectedGCDs, BuffWindowModule} from 'parser/core/modules/BuffWindow'
+import {BuffWindowExpectedGCDs, BuffWindowModule, BuffWindowState} from 'parser/core/modules/BuffWindow'
 import {SEVERITY} from 'parser/core/modules/Suggestions'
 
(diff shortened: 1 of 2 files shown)


src/parser/core/modules/Swiftcast.tsx
```diff
@@ -47,6 +47,25 @@ export abstract class SwiftcastModule extends BuffWindowModule {
...
+			severityTiers: this.severityTiers,
+		}
+	}
+
+	// Provide our own logic for the end of the fight – even though the window is
+	// ~4 GCDs 'wide', we can only use one action with it anyway; this change should
+	// ding them only if they had enough time during the window to use a spell with
+	// swiftcast
+	protected reduceExpectedGCDsEndOfFight(buffWindow: BuffWindowState): number {
+		return Math.min(1, super.reduceExpectedGCDsEndOfFight(buffWindow))
```
AzariahAeroborn, line 66: Contra your comment, what this implementation will do is if they use SC with less than 10 seconds before the end of the fight, it will reduce the expected GCDs to 0 (because super is going to return `Math.ceil((windowDurationMillis - fightTimeRemaining)/this.globalCooldown.getEstimate)`, or 10000 - 9500 / 2500 rounded up to 1 if they use it 9.5s before the end of the fight with a 2.5 GCD.

If you want it to treat using with 2.5s+ left in fight as expect 1, and otherwise expect 0, you'd need to implement something like:
```suggestion
		if ( this.buffStatus.duration ) {
			// Check to see if this window is rushing due to end of fight - reduce expected GCDs accordingly
			const fightTimeRemaining = this.parser.fight.end_time - buffWindow.start
			const gcdEstimate = this.globalCooldown.getEstimate()

			return ( fightTimeRemaining > gcdEstimate ) ? 0 : 1
		}

		return 0
```

src/parser/core/modules/Swiftcast.tsx
```diff
@@ -47,6 +47,25 @@ export abstract class SwiftcastModule extends BuffWindowModule {
 		severityTiers: this.severityTiers,
 	}
 
+	protected init() {
+		super.init()
+		// Inheriting the class doesn't update expectedGCDs's parameters when they
+		// override, so let's (re)define it here... feels mega jank tho
+		this.expectedGCDs = {
```
AzariahAeroborn, line 54: Ah, yes, the joys of JS abstract classes.  Since I chose to have the expectedGCDs property be an object with the three properties on it for compactness, overriding individual object properties with "this.SomeOtherProperty" doesn't work unless you do it in init due to order of operations on class construction.

I'd say this raises a question of whether it's better to split up expectedGCDs (and probably the other three tracking objects) into separate properties rather than a consolidated object.  Alternatively, if you just have Swiftcast consumers override the full expectedGCD object (even at the cost of some code duplication), it won't require the init jank either.

src/parser/core/modules/Swiftcast.tsx
nonowazu: That's a good point, and after checking back in the code I do understand it a bit better. I'll push another commit with the changed code.

src/parser/core/modules/Swiftcast.tsx
nonowazu: I guess I was thinking about how a user might implement this – if I were just trying to override to change only the warning message, I'd only want to set the one property instead of a whole object.

nonowazu: Ah, forgot to update here as well. Per discord, I think we should just leave it in here so I can make the experience more straightforward to the users of the class – unless someone wants to block on me overriding `init`

PR #840, opened 2020-04-12
Added a check in the Ninjutsu module to forgive prepull STDs

Toastdeib:
The meme opener for NIN on fights where the boss gets pulled far enough into middle includes a prepull Doton before the Hide reset into Suiton. The module was dinging on it when there's no harm in getting free damage, so this fixes that.

### DIFF
--- a/src/parser/jobs/nin/modules/Ninjutsu.js
+++ b/src/parser/jobs/nin/modules/Ninjutsu.js
@@ -30,7 +30,7 @@ export default class Ninjutsu extends Module {
 		this.addEventHook('cast', {by: 'player', abilityId: [ACTIONS.HYOTON.id, ACTIONS.HYOTON_TCJ.id]}, () => { this._hyotonCount++ })
 		this.addEventHook('cast', {by: 'player', abilityId: ACTIONS.RABBIT_MEDIUM.id}, () => { this._rabbitCount++ })
 		this.addEventHook('cast', {by: 'player', abilityId: [ACTIONS.DOTON.id, ACTIONS.DOTON_TCJ.id]}, this._onDotonCast)
-		this.addEventHook('normaliseddamage', {by: 'player', abilityId: STATUSES.DOTON.id}, this._onDotonDamage)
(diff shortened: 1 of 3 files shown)


src/parser/jobs/nin/modules/Ninjutsu.js
```diff
@@ -35,19 +36,21 @@ export default class Ninjutsu extends Module {
...
+			prepull: fabricated && (event.timestamp - PREPULL_DOTON_BUFFER) <= this.parser.fight.start_time,
+			// Flag Dotons with no cast event and a first damage tick at or before 3s as prepull so we don't ding on them
 		}
 	}
 
 	_onDotonDamage(event) {
 		// If there are no casts at all, use the damage event to fabricate one
 		if (!this._dotonCasts.current) {
-			this._onDotonCast()
+			this._onDotonCast(event, true)
```
AzariahAeroborn, line 53: On the one hand, I feel like fixing the data files so that PrecastStatus synthesizes the Doton cast would be a better approach (since Doton applies a status to the NIN that falls off when the AOE ends and so PCS can synth the cast).  On the other hand, I suppose this is the least cost change for prepull Doton tracking, since we're already synthing that first Doton cast anyway...

src/parser/jobs/nin/modules/Ninjutsu.js
```diff
@@ -35,19 +35,20 @@ export default class Ninjutsu extends Module {
...
 			ticks: [],
+			prepull: event.timestamp < this.parser.fight.start_time,
 		}
 	}
 
 	_onDotonDamage(event) {
 		// If there are no casts at all, use the damage event to fabricate one
 		if (!this._dotonCasts.current) {
-			this._onDotonCast()
+			this._onDotonCast(event)
```
AzariahAeroborn, line 51: Is this code getting hit at all after the change so that PrecastStatus will pick up the pre-pull Doton?  I would expect this if block should be obsolete now.

Toastdeib: Oh true, didn't even think of that. Will feex.

PR #845, opened 2020-04-14
Unify the utilities function formatDuration with parser.formatDuration

AzariahAeroborn:
This also changes the behavior of formatDuration for pre-pull timestamps to show <0s rather than -01:-01

### DIFF
--- a/src/utilities/strings.ts
+++ b/src/utilities/strings.ts
@@ -1,16 +1,20 @@
 /**
  * Renders a time given into the format `mm:ss`
- * @param duration {number} Seconds
- * @param secondPrecision {number} Optional value for number of decimal places to format seconds values to
- * @param isMilliseconds {boolean} Optional value for whether the input duration is in Milliseconds.  Defaults to false (input value is in seconds)
- * @param hideMinutesIfZero {boolean} Optional value for whether the return string should hide minutes (and only return ss.##s for values less than 1 minute).  Defaults to false (always show minutes)
+ * @param duration {number} Milliseconds
+ * @param options {object} Interface for optional option values
+ * @param options.secondPrecision {number} Optional value for number of decimal places to format seconds values to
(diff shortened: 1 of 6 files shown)


src/utilities/strings.ts
```diff
@@ -1,17 +1,39 @@
 /**
  * Renders a time given into the format `mm:ss`
  * @param duration {number} Seconds
+ * @param secondPrecision {number} Optional value for number of decimal places to format seconds values to
+ * @param isMilliseconds {boolean} Optional value for whether the input duration is in Milliseconds.  Defaults to false (input value is in seconds)
+ * @param hideMinutesIfZero {boolean} Optional value for whether the return string should hide minutes (and only return ss.##s for values less than 1 minute).  Defaults to false (always show minutes)
  * @return {string} Formatted duration
  */
-export function formatDuration(duration: number): string {
+export function formatDuration(duration: number, secondPrecision?: number, isMilliseconds: boolean = false, hideMinutesIfZero: boolean = false): string {
```
ackwell, line 9: arguments are getting a bit bloaty, should probably be moved to an options object, i.e.

```ts
function formatDuration(duration: number, options?: {
	secondPrecision?: number,
	isMilliseconds?: boolean
	hideMinutesIfZero?: boolean
} = {})
```

not super on top of `isMilliseconds`, but dunno

AzariahAeroborn: I suppose it comes down to whether the lesser evil is an isMilliseconds option/argument, or consumers having to remember to divide by 1000 (or enforcing consumers to always send in milliseconds, I suppose... it does looks like the four existing calls to Utilities.formatDuration are doing ms -> s conversions by hand before sending it down 🤔).  Ok, you convinced me, will change this to expect duration in milliseconds.

ackwell: standardising across the board on ms sgtm.

PR #858, opened 2020-04-23
[GNB] Expanded Visual Feedback on Gnashing Fang

Rjb99:
This PR adds a visual table for all Gnashing Fang combos that do not contain all 6 actions, along with a  timestamp on when the combo happened. The table's purpose is to just show which parts of the combo were missed and when. No feedback is given besides that and the normal checklist rule feedback.

The PR also includes some housekeeping in GNB statuses for another module I'm working on.


src/parser/jobs/gnb/modules/AmmoCombo.tsx
```diff
@@ -0,0 +1,262 @@
...
+
+		// cases to ignore, No Mercy any other modules will handle the latter 2.
+
+		if (actionId === ACTIONS.ATTACK.id || actionId === ACTIONS.BURST_STRIKE.id || actionId === ACTIONS.SONIC_BREAK.id ) {
+			return
+		}
+
+		if (actionId === ACTIONS.GNASHING_FANG.id) {
+
+			this.onEndGnashingCombo(event) // Make damn sure the previous window has ended.
```
AzariahAeroborn, line 96: This is going to ALWAYS overwrite the end time for the previous Gnashing Fang combo, is that intended?  It looks like nothing's really happening with the window end time currently; is that planned for future handling or should the end time tracking just be removed altogether as unused?

src/parser/jobs/gnb/modules/AmmoCombo.tsx
```diff
@@ -0,0 +1,262 @@
...
+			this.onEndGnashingCombo(event)
+		}
+
+		// If the action is a gnashingCombo one, log it
+
+		const lastGnashingCombo = this.lastGnashingCombo
+		if (lastGnashingCombo != null && lastGnashingCombo.end == null) {
+
+				switch (actionId) {
+					case ACTIONS.GNASHING_FANG.id:
```
AzariahAeroborn, line 115: Couldn't this be heavily simplified by just 
```
if (COMBO_ACTIONS.contains(actionId))
    lastGnashingCombo.rotation.push(event)
```
and then using lastGnashingCombo.rotation.length as comparison for the isProper check?

Especially if you go with just dropping tracking the window end time altogether, could streamline the logic pretty heavily.  You could also in that case make isProper and the various numAction properties be computed class getters (see NormalisedEvents in core) and not have to track any of that.

src/data/STATUSES/root/GNB.ts
```diff
@@ -53,7 +53,7 @@ export const GNB = ensureStatuses({
 		id: 1840,
 		name: 'Heart of Stone',
 		icon: 'https://xivapi.com/i/013000/013610.png',
-		duration: 18,
+		duration: 7,
```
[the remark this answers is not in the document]
Rjb99: they were never correct in the first place. so I fixed the values.

the housekeeping is for a mitigation usage module I want to do after I finish up everything else I've got in the works atm on GNB/SAM

src/parser/jobs/gnb/modules/AmmoCombo.tsx
Rjb99: unintended override there.

end time is currently planned to be used in a custom use checker, but I'll discuss this more in discord with you, if I can make it work on the standard one, I'd prefer to use cores, but due to to GNB's unique cases I'm hesitating to use it.

src/parser/jobs/gnb/modules/AmmoCombo.tsx
Rjb99: It could. I'll get on that.

PR #863, opened 2020-04-27
Thunder

rexatlanti:
Added a Thunder module to track DoT clipping. Thanks for the SMN module for the inspiration.

In the process I factored out things from the rotation watchdog and fixed a bunch of stuff


src/parser/jobs/blm/modules/Thunder.js
```diff
@@ -0,0 +1,241 @@
...
+
+		// Suggestions to not spam T3 too much
+		const sumClip = this._clip[STATUSES.THUNDER_III.id]
+		const maxExpectedClip = (this.thunder3Casts - 1) * MAX_ALLOWED_T3_CLIPPING
+		if (sumClip > maxExpectedClip) {
+			this.suggestions.add(new Suggestion({
+				icon: ACTIONS.THUNDER_III.icon,
+				content: <Trans id="blm.thunder.suggestions.excess-thunder.content">
+					Casting <ActionLink {...ACTIONS.THUNDER_III} /> too frequently can cause you to lose DPS by casting fewer <ActionLink {...ACTIONS.FIRE_IV} />. Try not to cast <ActionLink showIcon={false} {...ACTIONS.THUNDER_III} /> unless your <StatusLink {...STATUSES.THUNDER_III} /> DoT or <StatusLink {...STATUSES.THUNDERCLOUD} /> proc are about to wear off.
+					Check the <a href="javascript:void(0);" onClick={() => this.parser.scrollTo(this.constructor.handle)}><NormalisedMessage message={this.constructor.title}/></a> module for more information.
```
AzariahAeroborn, line 145: React is emitting deprecation errors for href="javascript:void(0)" links: `Warning: A future version of React will block javascript: URLs as a security precaution. Use event handlers instead if you can. If you need to generate unsafe HTML try using dangerouslySetInnerHTML instead. React was passed "javascript:void(0);"`

The better way to do this would be:
```suggestion
					Check the <a href="#" onClick={e => { e.preventDefault(); this.parser.scrollTo(this.constructor.handle) }}><NormalisedMessage message={this.constructor.title}/></a> module for more information.
```

rexatlanti: alrighty, I dunno enough about links so I'll stick with your suggestion

PR #864, opened 2020-04-29
AST 5.1-5.2 support, partial layered data migration

suShirow:
This PR will mark AST as 5.2 supported. 

- Fixes some bugs in arcana logs caused by changes 5.1
- Supports 5.1 ability ID changes from SE's balance changes (see src/data/ACTIONS/layers/patch5.1.ts)
- Modifies ArcanaTracking to use layered data. Looks like a lot, but actually just shifted the data constants to initialize in constructor to get data at runtime.
- Removes invul/downtime warning on AST page

Some revert commits in there; intended to port some other files to TS in order to fully support layered data, but realized that this would become a very large PR and we'd never land it. 

There will be a lot of overlap with https://github.com/xivanalysis/xivanalysis/pull/736 which had some comments and reviews already


src/parser/jobs/ast/modules/ArcanaTracking/ArcanaSuggestions.tsx
```diff
@@ -200,7 +212,7 @@ export default class ArcanaSuggestions extends Module {
 
 	// Helper for output()
 	RenderSpreadState(artifact: CardLog) {
-		const drawnArcana = getDataBy(STATUSES, 'id', artifact.drawState)
+		const drawnArcana = artifact.drawState ? getDataBy(this.data.statuses, 'id', artifact.drawState) : undefined
```
ayyaruq, line 215: You can use `this.data.getStatus(artifact.drawState)` here, since it's gonna be either the status or `undefined` due to the `getStatus` sig anyway.

ayyaruq:
Not got a huge amount of time rn for proper review, but replacing ActionLink/StatusLink with `this.data` is prob overkill, makes them harder to read. If they're the only thing you're using the ACTIONS and STATUSES imports for tho maybe it is better then.

You removed the charges but didn't add them back when fixing the baseline, do they no longer exist or was that an oversight?

src/parser/jobs/ast/modules/ArcanaTracking/ArcanaSuggestions.tsx
suShirow: Good catch

suShirow:
> Not got a huge amount of time rn for proper review, but replacing ActionLink/StatusLink with `this.data` is prob overkill, makes them harder to read. If they're the only thing you're using the ACTIONS and STATUSES imports for tho maybe it is better then.
> 
> You removed the charges but didn't add them back when fixing the baseline, do they no longer exist or was that an oversight?

I forgot about action charges, if that's what you meant. Added them in. 
About ActionLink and StatusLink, not sure if that's how it's mean to be used. But for those files I've completely removed ACTION and STATUSES... cause... why do we need them if we have this.data

I agree that it made it more verbose and when I saw it, I felt kinda eeww, but I decided it's better since it's consistent with the other data uses

ackwell:
in the theoretical future when I've had time to do the last bit myself, you won't be able to import directly from actions/statuses, you will need to go through the data module.

moving db links on it is a good move

src/parser/jobs/ast/modules/ArcanaGroups.ts
```diff
@@ -1,55 +1,55 @@
-import ACTIONS from 'data/ACTIONS'
-import STATUSES from 'data/STATUSES'
+import {ActionRoot} from 'data/ACTIONS/root'
+import {StatusRoot} from 'data/STATUSES/root'
 // Grouped cards for future convenience - whatever processing desired
 
 /* Grouped actions */
 
-export const PLAY = [
```
AzariahAeroborn, line 7: This change breaks the grouping of card actions on the timeline, since Cooldowns.tsx is importing PLAY for the list of actions to merge under the "Play" heading, and Core Cooldowns expects a list of Action ID numbers, not action name strings.

suShirow: Oh...
I'll feeex

src/parser/jobs/ast/modules/ArcanaTracking/ArcanaSuggestions.tsx
```diff
@@ -200,7 +212,7 @@ export default class ArcanaSuggestions extends Module {
 
 	// Helper for output()
 	RenderSpreadState(artifact: CardLog) {
-		const drawnArcana = getDataBy(STATUSES, 'id', artifact.drawState)
+		const drawnArcana = artifact.drawState ? this.data.getStatus(artifact.drawState) : undefined
```
AzariahAeroborn, line 215: Apologies if this was in before and I missed it, but I also noticed that the Arcana Logs are no longer showing the placeholder in Spread State for when there is no active card, instead showing a red square with an X:
image

suShirow: This PR actually fixes this. The problem was, we used to use "0" as inactive state, but it turns out that "0" represented the "X" icon.
So now we're using undefined if there's no active card, which means it should render the placeholder.

AzariahAeroborn: Ah, whoops, I got my windows crossed up 😊

PR #869, opened 2020-05-02
Initial reorganization of unit test structure

AzariahAeroborn:
Per discussion with @ackwell on Discord this morning, converting test directory structure to be in the `__tests__` subdirectory of each appropriate code directory for better organization.

While I was working on this, observed that several of the test files were having problems with mocking, tried upgrading Jest to latest and that fixed the issue.

Also observed that ESLint was apparently ignoring .tests.js files automatically; changing naming convention caused them to no longer get ignored so then they were failing the git hook.  Updated eslint ignore file accordingly.


.eslintignore
```diff
@@ -5,3 +5,4 @@ locale
 preset-react-app-custom
 build
 **/*.ts
+**/__tests__/**
```
ackwell, line 8: Once linting is Fixed™️, would be good to get tests covered by linting as well. But that's a not-now thing.

package.json
```diff
@@ -32,7 +32,9 @@
     "not op_mini all"
   ],
   "lingui": {
-    "srcPathDirs": ["src"],
+    "srcPathDirs": [
```
ackwell, line 35: fuck i hate the way it autoformats it, like, come the fuck on

src/data/__tests__/getDataBy.ts
```diff
@@ -1,4 +1,4 @@
-import {getDataBy} from './getDataBy'
+import {getDataBy} from 'data/getDataBy'
```
ackwell, line 1: For the sake of consistency, I wanna keep test imports of the thing being imported as `../foo`
```suggestion
import {getDataBy} from '../getDataBy'
```

src/parser/__tests__/Root.tsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 import ReactDOM from 'react-dom'
 
-import Root from './Root'
+import Root from '../../Root'
```
ackwell, line 4: this looks like it ended up in the wrong test folder

AzariahAeroborn: :notlikeblob:

PR #870, opened 2020-05-02
fix: PrecastAction should not consider autos as first casts

AzariahAeroborn:
Per a report on the FFLogs discord, if an auto-attack cast action occurs before the first damage event for an actual action in the fight, PrecastAction bails early on the auto-attack cast action and does not synthesize a cast event for the pre-pull action that generated the first non-auto damage event.  An effected log is: https://xivanalysis.com/analyse/xBY8QdKNazhVTRvk/3/3/

I went ahead and converted PCA to TypeScript and added unit tests for it along with this change.  The tests won't run successfully without the Jest upgrades in #869.


src/parser/core/modules/PrecastAction.ts
```diff
@@ -0,0 +1,38 @@
...
+	static debug = false
+
+	normalise(events: Event[]): Event[] {
+		const startTime = this.parser.fight.start_time
+
+		for (const event of events) {
+			this.debug(`Timestamp: ${event.timestamp} - Event Type: ${String(event.type)} - Action: ${isAbilityEvent(event) ? event.ability.name : ''}`)
+
+			// Only care about events by the player and ignore autoattacks
+			if (!this.parser.byPlayer(event) || !isAbilityEvent(event) || event.ability.name === 'attack')
```
ackwell, line 17: This'll fail for ranged phys (i think?) - their auto is `shot`.

If you dep on `data` (should be more safe to do so), and `data.getAction`, should be able to check `action.autoAttack` which is `boolean | undefined`

src/parser/core/modules/__tests__/PrecastAction.ts
```diff
@@ -0,0 +1,122 @@
+import PrecastAction from 'parser/core/modules/PrecastAction'
```
ackwell, line 1: nit; per other pr
```suggestion
import PrecastAction from '../PrecastAction'
```

That said, keep the imports below as-is - they're not the thing being tested

src/parser/core/modules/__tests__/PrecastAction.ts
```diff
@@ -0,0 +1,122 @@
...
+import Parser from 'parser/core/Parser'
+import {AbilityType, CastEvent, DamageEvent, Event, HitType} from 'fflogs'
+
+jest.mock('parser/core/Parser')
+const MockedParser = Parser as jest.Mock<Parser>
+
+// tslint:disable:no-magic-numbers
+
+const fightStartTime = 0
+const mockCastEvent = (timestamp: number): CastEvent => {
```
ackwell, line 11: nit; can shorthand these as 
```ts
const mockCastEvent = (timestamp: number): CastEvent => ({
	key: value
})
```

PR #871, opened 2020-05-02
[BugFix]: Buff event sorting

AzariahAeroborn:
Since time immemorial, AdditionalEvents has sorted removebuff events and refreshbuff events before applybuff events at the same timestamp.  If the first Excogitation cast of the fight instantly pops, the FFLogs API emits an applybuff and removebuff event at the same timestamp with the applybuff event first.  After AdditionalEvents runs and resorts the events array, we mangle this order and then PrecastStatus "sees" the removebuff before an eventbuff and determines that it needs to synthesize a pre-pull Excogitation.

Fixing sort order to prevent this case.


AzariahAeroborn:
After further conversation on Discord, changing this approach to **preserve** the order of buff/debuff events as presented by FFLogs, rather than enforcing a sort order.

From a technical perspective, as the various normalisers are currently running, this will result in events from the original log (filtered to the player being evaluated) occurring first, then events from Additional Events queries, in case of a timestamp tie.

PR #872, opened 2020-05-03
Feature overheal core

nonowazu:
Okay, there's a lot to unpack here:

* I got a request to add overheal tracking to SCH, to which I did what
any normal dev would do – immediately steal from AST/WHM. Of course,
after solicting some feedback, it seems like the checklist percentage
'goal' was confusing, so I wrote a suggestion instead. That lacked
information, so after the 345th cup of coffee I decided to rig up a pie
chart.
* At this point, cold, dark insanity had seeped in, and I was committed
to this, so I ported what I had to typescript and then to core. After
testing it out with all the different modes enabled (yes, this has modes
like a transformer), SCH seems to want them all enabled.

So, onto the class and other changes:

* It has a TrackedOverheal class which just takes in HealEvents and
tallies up the heal/overheal amounts, and can be used to filter out
HealEvents into noteworthy categories (for example, WHM found value in
separating heals into three categories of direct heals, hots, and
assize), which can then be displayed in any number of ways. If a heal
doesn't belong to a category, it gets dumped into the 'direct' heal bin.
* It has options for spitting out pie charts, suggestions, and/or a
checklist item. I *think* this has feature parity with the existing WHM
and AST implementations.
* It maintains the `InvertedRequirement` class from AST/WHM overheal
tracking (and I somewhat expect a rejection, which is why I allow
overrides to the description for an enterprising dev to provide
programmatic descriptions to get around this)
* I give some SuggestedColors (that Levi at the Balance) came up with
based on other colors found in xiva; I'm not sure if this is the best
way to do it or not, but the SCH impelmentation uses them directly like
that but I'll change to something else if there's a block or a better
suggestion
* Charts do not like 4 rows, so after talking with Ack, he suggested I
override the CSS – this has an additional effect of fixing SMN pie
charts with 4 pet rows
* I actually ran `yarn extract` up front this time

This thing is quite meaty, so I fully expect to iterate on this a few times.


src/parser/core/modules/Overheal.tsx
```diff
@@ -225,6 +223,7 @@ export class CoreOverheal extends Module {
 		const guid = event.ability.guid
 		for (const trackedHeal of this.trackedOverheals) {
 			if (trackedHeal.idIsTracked(guid)) {
+				this.debug(`Heal from ${event.ability.name} at ${event.timestamp} matched into category ${trackedHeal.name.props.defaults}`)
```
AzariahAeroborn, line 226: Can you go ahead and duplicate this debug hook for "matched into Direct" as well down below, to make it easier for future maintainers to see what's going on?

Also, one thing that showing the matched into Direct reveals is that the "Regeneration" events for passive regeneration are also matching into that bucket.  I think that should probably be excluded, as it will make overheal from direct heal actions appear a bit lower than it really was, since the player doesn't really control their own self-Regeneration.

nonowazu: the duplicate is honestly a good idea. I pushed more changes in and added Ids just for additional tracking purposes.

PR #877, opened 2020-05-03
FormatDuration cleanup & fixes

AzariahAeroborn:
- Pass the secondPrecison option from parser.formatDuration to the function, so that the option is actually respected (whoopsie daisy) - restores original behavior
- Add a parameter for showNegative to control whether to format and return the actual negative amount or to return `< 0s` for negative numbers.  Along with this, fix the rounding for minutes so that it rounds in the correct direction on negative numbers, and fix showing -00:-15 if a longer duration.
- Since all of our default option values are falsey, actually make all of the options optional.
- Fix RotationTable so it requests 0 decimals precision for seconds (which it was getting previously with the way the string function behaved), so early windows show as 00:01 instead of 00:01.25, in line with the rest of the table


AzariahAeroborn:
I'll work on adding tests for the string utility functions later, I wanted to get this PR out there since I broke AST's Play table and found a few other uses of formatDuration who had unexpected / undesired behavior changes with the unification.  My bad :notlikeblob:

PR #880, opened 2020-05-04
Suppress interrupt table output if there's no interrupts

nonowazu:
What it says on the tin, except there's now a method to provide alternative output as well. Also, corrected an issue which technically would lead to chaos if someone tried to use ts to provide suggestion why content.

### DIFF
--- a/src/parser/core/modules/Interrupts.tsx
+++ b/src/parser/core/modules/Interrupts.tsx
@@ -59,8 +59,16 @@ export abstract class Interrupts extends Module {
 	 * @param missedCasts
 	 * @param missedTime
 	 */
-	private suggestionWhy(missedCasts: CastEvent[], missedTime: number): JSX.Element {
+	protected suggestionWhy(missedCasts: CastEvent[], missedTime: number): JSX.Element {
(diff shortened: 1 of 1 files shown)


src/parser/core/modules/Interrupts.tsx
```diff
@@ -59,10 +59,18 @@ export abstract class Interrupts extends Module {
 	 * @param missedCasts
 	 * @param missedTime
 	 */
-	private suggestionWhy(missedCasts: CastEvent[], missedTime: number): JSX.Element {
+	protected suggestionWhy(missedCasts: CastEvent[], missedTime: number): JSX.Element {
```
AzariahAeroborn, line 62: Is making this overridable intended?  If so, can you add a doc comment indicating what consumers should provide?

nonowazu: Yes, it's absolutely intended - it would make sense if someone wanted to change the why text to be more about missed cast time vs the number of missed casts (even though the default is to provide both). I could see samurai not caring about time at all (since the new Iaijutsu has such a short casting time), but only that you interrupted.

I'll absolutely add more docs to this, though.

PR #881, opened 2020-05-04
Fix SMN going turbomode during DWT and swiftcast

nonowazu:
* Added helper method to Swiftcast that's much like it's parent class that allows users to filter out swiftcasts.
* Modified the SMN-specific implementation of swiftcast to filter out casts during DWT. To do this, I'm pretty sure I just read a book on necromancy (or figured out some of the pain SMN devs have to go through) to figure out how DWT worked.
* Modified DWT to export DWT_CAST_TIME_MOD so it only has to be edited in one place instead of every file (do we have a 'job constants' file somewhere?)
* Was a good dev and disabled debug mode on EgiCommands.tsx when I saw that someone had left it enabled

### DIFF
--- a/src/parser/core/modules/Swiftcast.tsx
+++ b/src/parser/core/modules/Swiftcast.tsx
@@ -47,6 +47,17 @@ export abstract class SwiftcastModule extends BuffWindowModule {
 		severityTiers: this.severityTiers,
 	}
 
+	/**
+	 * Implementing modules MAY override this if they have special cases not covered
+	 * by the standard 'considerAction' method – for example, SMN with instant ruins during
+	 * DWT.
+	 * @param action - the action to consider
+	 * @returns true to allow the spell; false to ignore the spell
+	 */
+	protected considerSwiftAction(action: Action): boolean {
+		return true
+	}
+
 	protected init() {
 		super.init()
 		// Inheriting the class doesn't update expectedGCDs's parameters when they
@@ -73,11 +84,11 @@ export abstract class SwiftcastModule extends BuffWindowModule {
 	}
 
 	protected considerAction(action: Action) {
+		this.debug('Evaluating action during window:', action)
 		// ignore actions that don't have a castTime
-		if (!action.castTime) {
+		if (!action.castTime ) {
 			return false
 		}
-
-		return true
+		return this.considerSwiftAction(action)
 	}
 }
--- a/src/parser/core/modules/Swiftcast.tsx
+++ b/src/parser/core/modules/Swiftcast.tsx
@@ -86,7 +86,7 @@ export abstract class SwiftcastModule extends BuffWindowModule {
 	protected considerAction(action: Action) {
 		this.debug('Evaluating action during window:', action)
 		// ignore actions that don't have a castTime
-		if (!action.castTime ) {
+		if (!action.castTime) {
 			return false
 		}
 		return this.considerSwiftAction(action)
--- a/src/parser/jobs/smn/modules/Swiftcast.js
+++ b/src/parser/jobs/smn/modules/Swiftcast.js
@@ -12,7 +12,7 @@ const MISSED_SEVERITIES = {
 export default class Swiftcast extends SwiftcastModule {
 	static handle = 'swiftcast'
 
-	@dependency 'dwt'
+	@dependency dwt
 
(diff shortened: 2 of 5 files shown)


src/parser/jobs/smn/modules/Swiftcast.js
```diff
@@ -8,5 +10,48 @@ const MISSED_SEVERITIES = {
...
+		this.addEventHook('cast', {to: 'player'}, this._onCast)
+		this.addEventHook('normaliseddamage', {by: 'player', abilityId: this.data.actions.DEATHFLARE.id}, this._endDWT)
+		this.addEventHook('death', {to: 'player'}, this._endDWT)
+	}
+
+	_onCast(event) {
+		const guid = event.ability.guid
+
+		if (guid === this.data.actions.DREADWYRM_TRANCE.id) {
+			this.debug(`dwt now active ${this.parser.formatTimestamp(event.timestamp)}`)
```
ayyaruq, line 31: Why are all the other `DWT` in caps?

src/parser/jobs/smn/modules/Swiftcast.js
```diff
@@ -8,5 +10,48 @@ const MISSED_SEVERITIES = {
...
+
+	considerSwiftAction(event) {
+		this.debug(`DWT Active (${this._dwtActive}): ${event.name}`)
+		if (this._dwtActive) {
+			if (event.castTime + DWT_CAST_TIME_MOD > 0) {
+				return true
+			}
+			return false
+		}
+		return true
```
ayyaruq, line 55: Could this just not be one conditional, kinda iffy to read, something like this:

```
if (this._dwtActive && event.castTime + DWT_CAST_TIME_MOD <= 0) {
  return false
}

return true
```

src/parser/core/modules/Swiftcast.tsx
```diff
@@ -73,11 +84,11 @@ export abstract class SwiftcastModule extends BuffWindowModule {
 	}
 
 	protected considerAction(action: Action) {
+		this.debug('Evaluating action during window:', action)
 		// ignore actions that don't have a castTime
-		if (!action.castTime) {
+		if (!action.castTime ) {
```
ayyaruq, line 89: bruh

src/parser/jobs/smn/modules/Swiftcast.js
nonowazu: Pretty sure it was like midnight when I *started* work on this. Might have been crazy (my PR messages tend to devolve into the above when tired/crazy). Also still crazy, so I'll change.

src/parser/jobs/smn/modules/Swiftcast.js
nonowazu: That was the one that was hanging me up at midnight-thirty (if you read the caster channel I think my brain BSoD'd at that time) – will change.

src/parser/core/modules/Swiftcast.tsx
nonowazu: 👨💤➡️👨‍💻=🔥
(I will fix)

src/parser/jobs/smn/modules/Swiftcast.js
```diff
@@ -8,5 +9,45 @@ const MISSED_SEVERITIES = {
...
+	static handle = 'swiftcast'
+
 	severityTiers = MISSED_SEVERITIES
+	_dwtActive = false
+	_dwtActiveTimestamp = 0
+
+	constructor(...args) {
+		super(...args)
+
+		this.addEventHook('cast', {to: 'player'}, this._onCast)
```
KelosPelmand, line 21: The DWT module does have an activeAt method that looks like it would get you the information you are tracking here.  Likewise, you should be able to check getCurrentPet in the pet module for if Phoenix is active to determine if Firebird Trance is running (since we only support level 80).

If that method doesn't turn out to work, I would suggest checking for Deathflare casts rather than damage.  DWT hooks damage for reasons that don't apply to this module.  In the DWT module, it is checking to see if multiple enemies were hit, which was more relevant when trance gave a damage buff.

nonowazu:
Changing to draft because something seems broken when I run this from a fresh repo (or my regular repo)

nonowazu:
Yes, less weird now.

PR #888, opened 2020-05-08
Fix the GCD estimation

rexatlanti:
So _apparently_ serverticks are the cause as to why we can get faster/slower GCD estimates, in a 0.05s range around the real GCD.

Initial idea is to locate peaks in the GCD histogram and then rebin the whole thing since it's most likely to have the client side GCD value as the GCD and not GCD +- 0.05.


AzariahAeroborn:
This has been superseded by the work for the SpeedStatAdapter and updating the ABC calculations.  If we determine that we still need to tune the estimates in light of the ABC calculation updates, we will need to port this work to the new data structure.

PR #890, opened 2020-05-09
Add drill/air anchor drift module + CDDT for Machinist

yumiyafangirl:
CDDT

Tokyo Drift


src/parser/jobs/mch/modules/Drift.tsx
```diff
@@ -0,0 +1,95 @@
...
+export default class Drift extends Module {
+	static handle = 'drift'
+	static title = t('mch.drift.title')`GCD Drift`
+
+	@dependency private suggestions!: Suggestions
+	@dependency private downtime!: Downtime
+
+	private castHistory: CastEvent[] = []
+
+	private cumDrift = {
```
nonowazu, line 39: `drifts`? `totalDrift`? c'mon, you're killing me here.
Other than this (I can't stop looking at this), I think it's good – gonna pull down a few machinist parses and verify just to make sure.

src/parser/jobs/mch/modules/Drift.tsx
```diff
@@ -0,0 +1,95 @@
+import {t} from '@lingui/macro'
+import {Plural, Trans} from '@lingui/react'
+import _ from 'lodash'
+import React from 'react'
+
+import ActionLink from 'components/ui/DbLink'
```
nonowazu, line 6: Need to change this to:

```ts
import {ActionLink} from 'components/us/DbLink'
```

src/parser/jobs/mch/modules/Drift.tsx
```diff
@@ -0,0 +1,95 @@
...
+import ACTIONS from 'data/ACTIONS'
+import {CastEvent} from 'fflogs'
+import Module, {dependency} from 'parser/core/Module'
+import Downtime from 'parser/core/modules/Downtime'
+import Suggestions, {SEVERITY, TieredSuggestion} from 'parser/core/modules/Suggestions'
+
+// Escalate severity depending on how long drill/AA were drifted for
+const DRIFT_SEVERITY_TIERS = {
+	0.5: SEVERITY.MINOR,
```
nonowazu, line 15: Thinking more on this – since the math works out to the *number* of missing drills/aa, then this will never be a minor (since you call `Math.ceil` to round up). It's either going to be 0 or 1, which pushes it right to a medium.

If that's the intent (medium at 1 miss) then that's fine, but you might want to remove this line or adjust otherwise.

src/parser/jobs/mch/modules/Drift.tsx
ayyaruq: I for one welcome and appreciate cumdrift. With the Dirty Dancing module on DNC, we just need BRD to come? to the party. I am 12.

src/parser/jobs/mch/modules/Drift.tsx
yumiyafangirl: It's actually filtering on 'driftedDrills' (the unrounded value), so this shouldn't be an issue.

nonowazu: I see – carry on.

src/parser/jobs/mch/modules/Drift.tsx
```diff
@@ -0,0 +1,129 @@
...
+				},
+			}
+		})
+
+		return <Fragment>
+			<Message>
+				<Trans id="mch.drift.accordion.message">
+					<ActionLink {...ACTIONS.DRILL}/> and <ActionLink {...ACTIONS.AIR_ANCHOR}/> are your strongest GCDs and ideally they should always be kept on cooldown,
+					unless you need to insert a filler GCD to adjust for skill speed. Avoid casting <ActionLink {...ACTIONS.HYPERCHARGE}/> if
+					Drill or Air Anchor will come off cooldown within 8-10 seconds.
```
Toastdeib, line 118: Tiny wording nitpick, but it might be worth changing "8-10 seconds" to "8 seconds or less" to make it more clear that you don't want Drill/AA coming off CD during a Hypercharge window.

src/parser/jobs/mch/modules/GeneralCDDowntime.tsx
```diff
@@ -0,0 +1,47 @@
...
+		firstUseOffset: 3000,
+	}, {
+		cooldowns: [ACTIONS.REASSEMBLE],
+		allowedAverageDowntime: 5000,
+		firstUseOffset: 0,
+	}, {
+		cooldowns: [ACTIONS.AIR_ANCHOR],
+		allowedAverageDowntime: DOWNTIME_ALLOWED_GCD,
+		firstUseOffset: 9000,
+		// isAffectedBySpeed: true,
```
Toastdeib, line 24: I'm not actually familiar with this flag; is there a reason it's here but commented out? From the name it seems like it should be uncommented (and if we don't actually need it, it would probably be cleaner to just delete it entirely). Same thing for the Drill/Bioblaster blob below.

src/parser/jobs/mch/modules/GeneralCDDowntime.tsx
```diff
@@ -0,0 +1,47 @@
...
+	}, {
+		cooldowns: [ACTIONS.RICOCHET],
+		firstUseOffset: 3000,
+		resetBy: {actions: [ACTIONS.HEAT_BLAST], refundAmount: 15000},
+	}]
+
+	checklistName = <Trans id="mch.cooldownDowntime.name">Use your cooldowns</Trans>
+	checklistDescription = <Trans id="mch.cooldownDowntime.suggestion">
+		Always make sure to use your abilities when they are available, but do not clip or delay your GCD to use them.
+	</Trans>
```
Toastdeib, line 46: Any particular reason for the overrides here? Personal preference would be to use the default ones defined in the core module (which are nearly the same, just with "OGCDs" where you have "cooldowns") to be more consistent with other jobs.

KelosPelmand: Like SMN, these aren't all oGCDs (Air Anchor/Drill).  We actually had a user point this out for SMN, which lead to the change.  

If the goal is to have MCH use the actual same one as SMN, without making the translators translate it twice, I think you can say

`<Trans` id="smn.cooldownDowntime.name" />`

(empty tag with no value, likewise for description) and the same English and translations should be used.

abodak: @KelosPelmand since there's at least 2 jobs now with GCD cooldowns using this module, maybe it would be worthwhile to generalize the default text?

yumiyafangirl: @KelosPelmand since there are now at least 2 jobs using this module with cooldowns on the GCD, maybe it's best to update the default text? Or maybe a flag for oGCDs vs cooldowns if we really want to keep that separation?

src/parser/jobs/mch/modules/GeneralCDDowntime.tsx
yumiyafangirl: Following convention from the SMN module here, this flag would indeed be useful but it seems like it's been disabled since GCD estimates within xiva aren't very accurate (yet).

PR #894, opened 2020-05-10
wip: Convert GlobalCooldown to TypeScript

AzariahAeroborn:
Still a work in progress, haven't actually converted speedmod classes (or implemented the interface they'll need to use) yet, but up and working for classes that don't have speed mods.  

I have observed that the changed approach for getUptime() is resulting in different ABC uptimes -- I think this should be more accurate, but I haven't dug through yet to see what's causing the different results.


ayyaruq:
Is this maybe dead what with the analyser migration?

AzariahAeroborn:
Yes, this is being abandoned in favor of migrating to an AdapterStep - I'm WIP on that and hope to have something for review this weekend or early next week.

PR #899, opened 2020-05-12
DNC - Adjustments for dirty dances

AkairyuGestalter:
Added leniency for finishing a dance with fewer than the expected steps if:
- The fight would have ended before the dance could be completed in full
- The dance would have ended during a downtime/invuln window if completed in full

Added suggestions to avoid finishing a dance with zero steps used, since that voids the buff portion of the dance. Zero-step dances that fit into the leniency checks described above will not be dinged.


src/parser/jobs/dnc/modules/DirtyDancing.tsx
```diff
@@ -289,6 +304,38 @@ export default class DirtyDancing extends Module {
 				<Plural value={driftedTechnicals} one="# Technical Step was" other="# Technical Steps were"/> lost due to drift.
 			</Trans>,
 		}))
+
+		const zeroStandards = this.danceHistory.filter(dance => dance.dirty && dance.initiatingStep.ability.guid === ACTIONS.STANDARD_STEP.id &&
```
AzariahAeroborn, line 308: This is double-counting the 0 step dance within the generic "missing steps" suggestion as well.  Is there a better way to do the two suggestions so you don't wind up with the same dance generating both suggestions?  Or is it intended to generate a double suggestion?

src/parser/jobs/dnc/modules/DirtyDancing.tsx
```diff
@@ -71,17 +75,28 @@ class Dance {
...
 		return expectedFinish
 	}
 
-	constructor(start: number) {
-		this.start = start
+	public get expectedEndTime(): number {
+		const actionData = getDataBy(ACTIONS, 'id', this.initiatingStep.ability.guid) as TODO
+		return this.start + actionData.gcdRecast * 1000
+			+ EXPECTED_DANCE_MOVE_COUNT[this.expectedFinishId] * DANCE_MOVE_COOLDOWN_MILLIS
+			+ 1000 // Additional 1s leniency to account for network latency
```
AzariahAeroborn, line 91: Nit: this could (should?) be changed to a constant, maybe RUSHING_LATENCY_ALLOWANCE or something?

AkairyuGestalter: Easy enough, done

src/parser/jobs/dnc/modules/DirtyDancing.tsx
AkairyuGestalter: I think at some point I considered this, and then forgot to actually do something about it. Since these are a more important sub-category of the missing steps suggestion, I think it's fair not to double-count them. Fixed.

PR #900, opened 2020-05-13
BRD - add first-use offsets to OGCDDowntime

yumiyafangirl:


ayyaruq:
Looks fine, however I don't know BRD rotations well enough to comment on numbers. Raging Strikes specifically, that's a 1 second window which is a bit too exact I feel. Might be worth loosening that up a bit to handle stuff like ping, and people being unable to time exactly one second if not using an ACT trigger.

ayyaruq:
I was thinking more a larger number than a smaller one. For intentional downtime, you could use `allowedAverageDowntime` I guess, although it doesn't really fit situational usages. Maybe worth talking to @KelosPelmand or @AzariahAeroborn about how it could be more flexible.

yumiyafangirl:
What makes the most sense, just leaving it as default? Or maybe -500?

ayyaruq:
I meant like, -1500 to -3000, to account for people not having perfect timing on pre-pull. 3000 might be a bit much just because it'll lose a GCD at the end though.

yumiyafangirl:
Correct me if I'm wrong but wouldn't -3000 be more restrictive than -1000 since it would predict your 2nd RS to be at 77s vs 79s?

src/parser/jobs/brd/modules/OGCDDowntime.ts
```diff
@@ -6,9 +6,16 @@ import {CooldownDowntime} from 'parser/core/modules/CooldownDowntime'
 
 export default class OGCDDowntime extends CooldownDowntime {
 	trackedCds = [
-		{cooldowns: [ACTIONS.BARRAGE]},
-		{cooldowns: [ACTIONS.RAGING_STRIKES]},
-		{cooldowns: [ACTIONS.SIDEWINDER]},
+		{
+			cooldowns: [ACTIONS.BARRAGE],
+			firstUseOffset: 12000,
+		}, {
```
nonowazu, line 12: Minor stylistic nit more than anything (so I won't block on it), but might be worth adjusting to have each opening `{` on their own line (which I see a lot of precedent for throughout the job modules I've been checking.

For example:

* [drg OGCDDowntime.js](https://github.com/xivanalysis/xivanalysis/blob/master/src/parser/jobs/drg/modules/OGCDDowntime.ts)
* [mnk Cooldowns.js](https://github.com/xivanalysis/xivanalysis/blob/master/src/parser/jobs/mnk/modules/Cooldowns.ts)

Other than that, perhaps set `defaultFirstUseOffset` to `12000` and just adjust the raging strikes cooldown?

yumiyafangirl: Gonna avoid overriding `defaultFirstUseOffset` in case we ever want to add other cooldowns (e.g. EA) in the future.

nonowazu: Yeah, makes sense; I think it's good to go pending any other changes you'd like to make (I see Kelos making some suggestions for additional tracks, so I want to make sure before I approve)

PR #902, opened 2020-05-14
Multi-dancer adjustments

AkairyuGestalter:
- Added a message to the top of the Technical Windows table if multiple dancers are generating windows, to let players know why the display might be weird, and to use good judgement about maximizing damage.
- Updated Esprit generation calculation to only consider Technical Finish buffs sourced to the player, fixes overestimated generation due to other dancer's buffs.


AkairyuGestalter:
Github nub, not sure how to actually respond back to @AzariahAeroborn's comment but:

- Fixed the problem where the message wasn't displaying, by making the identification of this case more robust.
- It's an intended part of the behavior that you see a full Tech window from buff initiation to final drop-off. However, I could've done a better job of calling out that the window wasn't one you started, to help inform the player's reading of the data. That is done now.

AzariahAeroborn:
I'm not a huge fan of the extra space being consumed by the "You opened?" column, it's really scrunching up the Rotation column even on 1080p desktop (and the rotation column just breaks down on mobile from all the other note columns, but that pre-exists this PR so not blocking here).  I'm also concerned that the UI isn't displaying the time of when you used your Tech (if you're the second dancer to tech).

For the dancer who opened both overlapped tech windows in that log, he's still not seeing the "This log contains Technical Step windows from other dancers", and he's not seeing the 04:25 window from the other dancer where he never used Technical Step at all - is that intended?

If you think this is the best approach at this point, and want to go ahead with this and get user feedback, let me know.  I know this whole case is kind of messy to deal with, so just trying to think of what looks weird to me and see if that's worth trying to change before we go to production; if you're happy with this as-is I won't block on my above comments.

AkairyuGestalter:
Updated the table header wording and dropped the column to clean up the UI. Also made sure that the source of the window knows if another DNC extended it, in addition to the extender DNC know that they weren't the source, so that either case triggers the header blurb. Confirmed that the final window didn't hit the other DNC since he was dead at the time, so since all of this is based on the buff, it should be working as intended (even if the intent is kind of gross in the multi-DNC situations...)

PR #905, opened 2020-05-16
PrecastStatus - synth cast if first event for an action is applybuff

AzariahAeroborn:
Currently, if an applybuff/removebuff occurs at the start of the pull but the matching cast event isn't picked up, PrecastStatus won't synth the cast (because it only synths if there's a missing applybuff to synth). 

This is an endeavor to fix that, and convert PrecastStatus to tested TS.  ~~Currently setting to draft because I haven't gone through to confirm I haven't broken other things and haven't actually, you know, written the tests.~~

~~In cursory glancing, I'm pretty sure this has messed up RDM proc statuses, so don't merge this yet...~~ Fixed as of later pushes


yumiyafangirl:
LGTM

Confirmed on a handful of bard parses where the issue was known to occur

src/parser/core/modules/PrecastStatus.ts
```diff
@@ -0,0 +1,153 @@
...
+export class PrecastStatus extends Module {
+	static handle = 'precastStatus'
+	static debug = true
+
+	@dependency private data!: Data
+
+	trackedStatuses = new Map<number, number[]>()
+	trackedActions: number[] = []
+	buffEventsToSynth: AbilityEvent[] = []
+	castEventsToSynth: CastEvent[] = []
```
ackwell, line 20: these seem like internal details of how the module works - should they be private?

src/parser/core/modules/Data.ts
```diff
@@ -25,6 +27,11 @@ export class Data extends Module {
 		return getDataBy(this.statuses, 'id', id)
 	}
 
+	getActionAppliedByStatus(status: Status) {
```
ackwell, line 30: this seems like a very specialised use-case to be putting in the low-level data module...

is it feasible to keep in a more contained module scope? (also the fn name seems the wrong way around, almost - statuses don't apply actions)

src/__mocks__/Data.ts
```diff
@@ -0,0 +1,44 @@
...
+	private actionData: MockAction[] = []
+	private statusData: Status[] = []
+
+	getAction(id: Action['id']) {
+		return this.actionData.find(a => a.id === id)
+	}
+
+	getStatus(id: Status['id']) {
+		// tslint:disable-next-line:no-magic-numbers
+		return this.statusData.find(a => a.id + 1000000 === id)
```
ackwell, line 18:
```ts
import {STATUS_ID_OFFSET} from 'data/STATUSES'
```
should do it i think

src/parser/core/modules/__tests__/PrecastStatus.ts
```diff
@@ -0,0 +1,477 @@
+// region CodeUnderTest
```
ackwell, line 1: I'm not a vast fan of regions, but I can see their use. If you're going to add them, can you use `// #region` and `// #endregion` so other editors pick them up properly?

src/parser/core/modules/__tests__/PrecastStatus.ts
```diff
@@ -0,0 +1,477 @@
...
+}
+data.mockStatus(mockStatusStackData)
+// endregion
+
+// endregion
+
+// endregion
+
+
+describe('The PrecastStatus module', () => {
```
ackwell, line 83:
```suggestion
describe('PrecastStatus', () => {
```

src/parser/core/modules/__tests__/PrecastStatus.ts
```diff
@@ -0,0 +1,477 @@
...
+		jest.spyOn(data, 'getAction')
+		jest.spyOn(data, 'getActionAppliedByStatus')
+		jest.spyOn(data, 'getStatus')
+	})
+
+	afterEach(() => {
+		jest.clearAllMocks()
+	})
+
+	it('has a normalise method', () => {
```
ackwell, line 99: a lot of this seems to be testing the minutae of the implementation details of _how_ PCS works, not _what_ it does.

Assuming an identical end-result, a total rewrite of implementation details should require minimal to no changes to the tests.

src/__mocks__/Events.ts
```diff
@@ -0,0 +1,57 @@
...
+export const mockApplyBuffEvent = (timestamp: number, ability: Ability, targetID: number): BuffEvent => ({
+	type: 'applybuff',
+	timestamp,
+	ability,
+	sourceIsFriendly: true,
+	targetIsFriendly: true,
+	targetID,
+})
+
+export const mockApplyBuffStackEvent = (timestamp: number, ability: Ability, targetID: number, stack: number): BuffStackEvent => ({
```
ackwell, line 20: at these sort of arg counts, it's usually better to move towards an object param. In cases of providing defaults &c, I usually do something like 

```ts
const createFoo = (overrides: Partial<Foo>): Foo => ({
	// write out
	// default values
	// here
	...overrides,
})
```

src/parser/core/modules/PrecastStatus.ts
```diff
@@ -0,0 +1,153 @@
...
+			if (event.type === 'cast') {
+				const action = this.data.getAction(event.ability.guid)
+				if (action && !this.trackedActions.includes(action.id)) {
+					this.markActionAsTracked(action.id)
+					continue
+				}
+			}
+
+			if (isTrackedBuffEvent(event)) {
+				if (event.targetID) {
```
ackwell, line 34: these two can be combined into a single conditional

src/parser/core/modules/PrecastStatus.ts
```diff
@@ -0,0 +1,153 @@
...
+					const trackedStatusesForTarget = this.trackedStatuses.get(event.targetID)
+					if (trackedStatusesForTarget != null && trackedStatusesForTarget.includes(event.ability.guid)) {
+						// Status is already tracked for this target and no synth needs to take place
+						continue
+					}
+
+					const statusInfo = this.data.getStatus(event.ability.guid) as Status
```
ackwell, line 41: shouldn't need the cast here i don't think - the ugly af union `getStatus` returns should be a superset of `Status`, so anything that accepts `Status` should accept the union.

the union is intentional, fwiw

src/parser/core/modules/PrecastStatus.ts
```diff
@@ -0,0 +1,153 @@
...
+					if (!statusInfo) {
+						// No valid status data for this event, skip to next event
+						continue
+					}
+
+					this.debug(`Checking ${event.type} of ${event.ability.name} at ${this.parser.formatTimestamp(event.timestamp, 1)} with stacks ${event.stack ?? 0}`)
+
+					if (statusInfo.stacksApplied && statusInfo.stacksApplied > 0) {
```
ackwell, line 49: Don't do bool ops on `number | undefined`, `0` is falsey. (tbqh i've moved towards always using `==null` for undef checks, saves on headaches down the road from falsey scalar values)

src/parser/core/modules/PrecastStatus.ts
```diff
@@ -0,0 +1,153 @@
...
+						// This action applies stacks, expected first cast will have an applybuff and an applybuffstack with max stacks at the same timestamp
+						// Ignore the applybuff event and check the applybuffstack event for validity
+						if (event.type === 'applybuff') { continue }
+						if (event.type === 'applybuffstack')
+						{
```
ackwell, line 54: akknit; curlies on same line. same goes for the else below.

surprised lint didn't pick this up, somethings seriously fucky with ts linting atm i think

src/parser/core/modules/PrecastStatus.ts
```diff
@@ -0,0 +1,153 @@
...
+
+	@dependency private data!: Data
+
+	trackedStatuses = new Map<number, number[]>()
+	trackedActions: number[] = []
+	buffEventsToSynth: AbilityEvent[] = []
+	castEventsToSynth: CastEvent[] = []
+	private startTime = this.parser.fight.start_time
+
+	normalise(events: Event[]) {
```
ackwell, line 23: this function is a prime target for splitting up a bit, it's a bit chonky and hard to get through

src/parser/core/modules/PrecastStatus.ts
```diff
@@ -0,0 +1,153 @@
...
+		// Determine if this buff comes from a known action, fab a cast event
+		const actionInfo = this.data.getActionAppliedByStatus(statusInfo)
+
+		if (!actionInfo) {
+			this.debug('No known action found, no cast event to synthesize')
+		} else {
+			if (this.trackedActions.includes(actionInfo.id)) {
+				this.debug(`Cast event already seen for ${actionInfo.name}, no synth necessary`)
+			} else {
+				this.debug(`Fabricating cast event for action ${actionInfo.name} by ${event.sourceID}`)
```
ackwell, line 122: this feels like a lot of extra mental overhead to read for the sake of some logging - while i'm on board with logging debug info, it shouldn't be at the cost of the code being readable.

if you've got any ideas on how to improve logging so this is less of an issue in-place i'm all ears

src/parser/core/modules/Data.ts
AzariahAeroborn: Whoops, I must have been drunk when I named that.  

I added the helper method for testability.  I spent about 30 minutes hacking at trying to mock statusesApplied the way we currently do it and had only heartbreak and sorrow, so I tried the (mockable) helper method instead.

src/parser/core/modules/__tests__/PrecastStatus.ts
AzariahAeroborn: Sorry, TDD habit from my work code base, we don't usually clean up those sort of things along the development way.  Will feex.

src/parser/core/modules/__tests__/PrecastStatus.ts
```diff
@@ -0,0 +1,477 @@
+// region CodeUnderTest
+import {PrecastStatus} from '../PrecastStatus'
+// endregion
+// tslint:disable:no-magic-numbers
+// region Mocks
+import {MockedParser} from '__mocks__/Parser'
```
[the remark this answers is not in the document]
AzariahAeroborn: So, maybe I'm doing something wrong, but with the MockedParser set up this way, I have a parameterless constructor I can call and not have to deal with mocking out / setting up all the rest of the dependencies.  TS gets thoroughly unhappy if I just call jest.mock('parser/core/Parser'), unless I wind up redoing the same `const MockedParser = Parser as jest.Mock<Parser>` as I'm doing within the mocks class

src/parser/core/modules/PrecastStatus.ts
AzariahAeroborn: I feel like I ran into some sort of TS problem without the cast, but it's not happening now so YOLO

src/parser/core/modules/PrecastStatus.ts
AzariahAeroborn: I could at least improve the if/else nesting by switching to an "if/log/early return" pattern instead for checking the pre-conditions.  Personally, I really want this pre-condition debug logging to stay in place one way or another -- the hardest thing to trace in normalisers is figuring out why they are or are not choosing to normalise an event, so I consider this logging fairly high priority, even at the expense of "noop" if blocks.

src/__mocks__/Events.ts
AzariahAeroborn: I'm not sure I agree that it's worth the overhead to do a params object here - none of the parameters are or should be optional, and I would expect most IDEs/Intellisense to handle hinting what the parameters are anyway.  I could maybe see making the mock<T>Event stuff more generic, but I kind of suck with breaking down generics in my mental model and I'm not 100% sure it's really worth the candle to hoist a big generic structure for functions designed to provide syntactic sugar for creating mock events.

src/__mocks__/Parser.ts
```diff
@@ -0,0 +1,4 @@
+jest.mock('parser/core/Parser')
```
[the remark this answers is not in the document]
AzariahAeroborn: Same as the previous comment - this approach removes the need for mocking out the Meta/Report/Fight/Actor dependencies of the parser, which I don't really want to have to hoist since the rest of the mocking approach is not relying on those behaviors of the Parser.

AzariahAeroborn:
Closing this for now; enough other changes have happened to fix PCS issues that this work will need substantial updates.  Will re-evaluate and re-submit after touching things up.

PR #907, opened 2020-05-22
Warrior CDDT and timeline order update

ayyaruq:
What it says on the tin

Also resolves some naming inconsistencies between code/UI and the game.


src/parser/jobs/war/modules/OGCDDowntime.ts
```diff
@@ -0,0 +1,36 @@
...
+			// IR can sit for up to 4 GCDs, enough to reapply Eye and avoid overcapping gauge
+			allowedAverageDowntime: 10000,
+			// IR can be delayed in the double IC opener, as late as just before the 6th GCD
+			firstUseOffset: 15000,
+		},
+
+		{
+			cooldowns: [ACTIONS.UPHEAVAL],
+			// With delayed IR, and weaving many defence CDs, final weavable CD in IR (ie before 11th GCD)
+			firstUseOffset: 24300,
```
qaphla, line 33: I remain a bit skeptical of this ever being a good idea, just because of how likely it is to lose a use. Maybe something to consult on?

src/parser/jobs/war/modules/Cooldowns.ts
```diff
@@ -1,29 +1,30 @@
...
 		// Personal Mitigation
-		ACTIONS.HOLMGANG.id,
 		ACTIONS.VENGEANCE.id,
 		ACTIONS.RAMPART.id,
-		ACTIONS.THRILL_OF_BATTLE.id,
+		ACTIONS.NASCENT_FLASH.id,
 		ACTIONS.RAW_INTUITION.id,
+		ACTIONS.THRILL_OF_BATTLE.id,
 		ACTIONS.EQUILIBRIUM.id,
+		ACTIONS.HOLMGANG.id,
```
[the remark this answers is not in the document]
ayyaruq: I usually try to group stuff by how they're used and their frequency more than alphabet. Prob preference, but I find it makes stuff like CD stacking and openers easier to see.

src/parser/jobs/war/modules/OGCDDowntime.ts
ayyaruq: This is the first use offset, but that's fair. Over the course of a fight, this will allow for maaaaybe 1 use lost. If user does it earlier, it's not an issue, but depending on their kill time they may see an overcap; ie 21/20.

src/parser/jobs/war/modules/Cooldowns.ts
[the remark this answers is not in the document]
qaphla: 👍  I was mostly inferring from the IR/Infuriate switch. I don't think the order is terribly important in general, though (as long as everything's there).

src/parser/jobs/war/modules/OGCDDowntime.ts
qaphla: I'm not sure if there's an overall philosophy to follow here --- it seems like this will complain essentially never if something even remotely reasonable is being done, but at the cost of overcapping some portion of the time (hard to tell what portion without better data).

Hard to tell what the right timing is to balance not showing an overcap too much with not complaining for something reasonable.

ayyaruq: Checking the numbers a bit, it's enough to lose 1 use over most fights, and potentially 2 in E8S and other fights of similar length, if the average downtime is hit commonly (ie 32.5s rather than on CD). If it's hit on CD, even starting at 24s, it won't lose a cast. That's based on total fight duration tho so obviously diff story with different kill times. I think this is ok, but maybe good to get feedback from other warriors. Maybe 3rd FC in IR can be a cutoff, so 20s. I wouldn't want to go below 17s though.

src/parser/jobs/war/modules/Cooldowns.ts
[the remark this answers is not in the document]
ayyaruq: Infuriate is used more often and not paired with IR so I reordered those two to make visualising what's in IR easier.

src/parser/jobs/war/modules/OGCDDowntime.ts
ayyaruq: As per discussion in Discord, adjusted to 18500ms.

PR #911, opened 2020-05-25
removed the "this is not even remotely done" warning :c

rexatlanti:


src/parser/jobs/blm/index.js
```diff
@@ -16,12 +15,6 @@ export default new Meta({
 
 	Description: () => <>
 		<TransMarkdown source={description}/>
-		<Message warning icon>
-			<Icon name="warning sign"/>
-			<Message.Content>
-				<Trans id="blm.about.description.warning">This isn&apos;t even remotely done.</Trans>
-			</Message.Content>
-		</Message>
```
nonowazu, line 24: Since you removed this, can you run `yarn extract` again?

rexatlanti: oh yea, sorry

PR #912, opened 2020-05-25
Add a CooldownDowntime module to NIN

Toastdeib:
What it says on the tin. Also, to address one question before it gets asked; I deliberately excluded Shadow Fang from the CDDT module (at least for now) because its cooldown is affected by Huton. Since we aren't properly accounting for that in calculations, it acts as though it's 70s no matter what and throws off the math.

### DIFF
--- a/src/data/ACTIONS/root/NIN.ts
+++ b/src/data/ACTIONS/root/NIN.ts
@@ -261,7 +261,7 @@ export const NIN = ensureActions({
 		name: 'Ten Chi Jin',
 		icon: 'https://xivapi.com/i/002000/002922.png',
 		onGcd: false,
-		cooldown: 100,
+		cooldown: 120,
 		statusesApplied: ['TEN_CHI_JIN'],
 	},
 
--- a/src/parser/jobs/nin/index.js
+++ b/src/parser/jobs/nin/index.js
@@ -21,6 +21,11 @@ export default new Meta({
 		{user: CONTRIBUTORS.TOASTDEIB, role: ROLES.MAINTAINER},
 	],
 	changelog: [{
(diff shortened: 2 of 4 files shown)


AzariahAeroborn:
Would it make sense to either change the cooldown of Shadow Fang in the data files to match its Huton effected CD, or add an additional property to indicate the speed buffed cooldown and update CDDT to override the base cooldown if a speed-effected cooldown is in place?

src/data/ACTIONS/root/NIN.ts
```diff
@@ -261,7 +261,7 @@ export const NIN = ensureActions({
 		name: 'Ten Chi Jin',
 		icon: 'https://xivapi.com/i/002000/002922.png',
 		onGcd: false,
-		cooldown: 100,
+		cooldown: 120,
```
ayyaruq, line 264: This (and a bunch of other NIN adjustments) were technically in 5.1, so maybe use layers here? The complete mudra rework in 5.1 prob complicates stuff tho.

Toastdeib: I didn't bother because NIN isn't supported by xiva pre-5.1 anyway. The reworked changed a *lot* of stuff.

Toastdeib:
@AzariahAeroborn The problem with changing it in the data files is that it varies based on both Huton *and* the player's SkS. With an optimal 0 SkS build, it'll be a flat 59.5s, but it could go lower than that as well. That said, I *could* set it to 59.5 in the data files (or just round it to 60), since in practice you're using it every 60s anyway to keep it under Trick, even if you have SkS for some ungodly reason and you can use it sooner/misalign it deliberately.

PR #916, opened 2020-05-27
Remove warning messages

nonowazu:
* Remove warning for SCH
* Remove warning for WHM for both invuln status and module incompleteness
* Additionally for WHM, amend the lily suggestions per balance mentor


src/parser/jobs/whm/index.js
```diff
@@ -13,25 +11,14 @@ This analyser aims to identify some of the low-hanging fruit that could be used
 Note that since healing is a very fight-dependent process, it might be that some suggestions do not apply to you, but they may they help to inform you about things you have missed.
```
AzariahAeroborn, line 11: I realize this is a pre-existing condition, but they might they benefit from cleaning this up while we're here 🙃

src/parser/jobs/whm/index.js
```diff
@@ -13,25 +11,14 @@ This analyser aims to identify some of the low-hanging fruit that could be used
...
 
 If you would like to learn more about WHM, check the guides over at [The Balance](https://thebalanceffxiv.com/), and have a chat in the #whm_questions channel.
+
+Currently this module can track DoT uptime, note when your healing lily gauge is full, detect clipping, detect interrupted casts, and report about missed oGCD casts.
```
AzariahAeroborn, line 15: Clipping is a term that gets overloaded in the community (for overweaving / clipping GCD as well as for re-applying a DOT too soon) -- is there a more clear phrasing we can use here?

nonowazu: Not sure, to be honest. I let the WHM mentor drive the wording here.

Rjb99: Perhaps detect incorrect weaving?

PR #919, opened 2020-05-30
Remove warning from WAR

ayyaruq:
Also feex unused imports on MNK index


src/parser/jobs/mnk/index.tsx
```diff
@@ -1,6 +1,5 @@
 import {Trans} from '@lingui/macro'
 import React from 'react'
-import {Icon, Message} from 'semantic-ui-react'
```
suShirow, line 3: Sneaky! But gud acc

ayyaruq: This import isn't used anywhere, I'm surprised it doesn't trip the linter being left in there tbh.

PR #920, opened 2020-06-02
Update PLD.ts

YuuSmz:
SENTINEL cooldown change 180 to 120.


ayyaruq:
This should probably have been done in layers, but Sentinel isn't really used anywhere except timeline anyway. We don't seem to have patch dates for CN even though they're already on 5.1 afaik, so it may show incorrectly on some Chinese logs.

PR #921, opened 2020-06-02
[BRD] Clean up snapshot module

yumiyafangirl:
- Blacklisted all statuses known not to affect bard DoTs
- Sorted Medicated and Raging Strikes to the front of the statuses
- Moved DoT statuses to a separate column (to more easily see where DoTs were dropped, for example)

Before: 
image

After:
image


src/parser/jobs/brd/modules/SnapshotBlacklist.ts
```diff
@@ -0,0 +1,60 @@
...
+	STATUSES.SHIELD_SAMBA.id,
+	STATUSES.STRAIGHT_SHOT_READY.id,
+	STATUSES.TACTICIAN.id,
+	STATUSES.THE_WARDENS_PAEAN.id,
+	STATUSES.TROUBADOUR.id,
+	STATUSES.WEAKNESS.id,
+	STATUSES.WELL_FED.id,
+	STATUSES.WHEEL_OF_FORTUNE_DIURNAL.id,
+	STATUSES.WHEEL_OF_FORTUNE_NOCTURNAL.id,
+	STATUSES.WHISPERING_DAWN.id,
```
suShirow, line 59: Do you think it would have made more sense to make a whitelist instead?
We do kind of have a list of raidbuffs at `/parser/core/modules/RaidBuffs.js` so I would have thought it's possible to write a getter to pull these out.
This list will be unwieldy to maintain x_x

src/parser/jobs/brd/modules/Snapshots.js
```diff
@@ -47,13 +48,46 @@ export default class Snapshots extends Module {
...
+				STATUSES.WINDBITE.id,
+			]
+
+			const personalStatuses = [
+				STATUSES.MEDICATED.id,
+				STATUSES.RAGING_STRIKES.id,
+			]
+
+			// Sort personal buffs to the front of the status list (why do I have to cast this to Array???? JS explain??)
+			Array(snapshot.statuses).some((status, index) => {
```
suShirow, line 65: `event.snapshot` comes from `brd/AdditionalStats`, it's not a standard CastEvent prop.
From the looks of it these statuses are actually not arrays, they are key/value pairs (objects) where the key is the ID of a status.
For this you could try [ Object.entries() ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries) and it might be a little safer than casting it.

PR #922, opened 2020-06-04
Dragoon: Ability drift calculation module for High Jump and Geirskogul.

Rietty:
Added a module to Dragoon calculate High Jump and Geirskogul drift, since these are the most critical of the Dragoon skills and drifting them can cost you uses of Life of the Dragon windows, as well as move them out of your burst/buff windows. Code is borrowed from the MCH Drill/Air Anchor Drift module, table is formatted after SMN DoT table.


src/parser/jobs/drg/modules/Positionals.js
```diff
@@ -177,7 +177,7 @@ export default class Positionals extends Module {
 			return <Table.Row key={combo.time}>
 				<Table.Cell>{this.createTimelineButton(combo.time)}</Table.Cell>
 				<Table.Cell><ActionLink {...action} /></Table.Cell>
-				<Table.Cell textAlign="center">{this._checkIcon(combo.trueNorthCharges > 0, '')} (<Plural id="drg.positionals.tn-charges" value={combo.trueNorthCharges} one="# charge" other="# charges" />)</Table.Cell>
+				<Table.Cell>{this._checkIcon(combo.trueNorthCharges > 0, '')} (<Plural id="drg.positionals.tn-charges" value={combo.trueNorthCharges} one="# charge" other="# charges" />)</Table.Cell>
```
ebshimizu, line 180: oh no my center text alignment D:
Unless there's a design reason to change this, I'd suggest leaving it centered.

Rietty: It looks really awkward as it is right now, so I did that to fix the alignment!

src/parser/jobs/drg/modules/Drift.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+	private onDriftableCast(event: CastEvent) {
+		let actionId: number
+		actionId = event.ability.guid
+
+		const window = this.currentWindows[actionId]
+		window.end = event.timestamp
+		const downtime = this.downtime.getDowntime(window.start, window.end)
+		const cd = COOLDOWN_MS[actionId]
+		window.drift = Math.max(0, window.end - window.start - cd - downtime)
```
AzariahAeroborn, line 78: I'm a little concerned that this is going to just blanket forgive downtime that occurs during the cooldown, even if the skill never came off of cooldown during downtime (e.g. skill used at 1:00, downtime from 1:10-1:20, skill used again at 1:40 would wind up calculating this as 0s of drift, even though the skill became available at 1:30 and could have been used then).

Similarly, I'm not sure I understand what line 81 is doing - from the comment, I would have expected the compare to be downtime > cooldown.

src/parser/jobs/drg/modules/Drift.tsx
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	private onDriftableCast(event: CastEvent) {
+		let actionId: number
+		actionId = event.ability.guid
+
+		const window = this.currentWindows[actionId]
+		window.end = event.timestamp
+		const downtime = this.downtime.getDowntime(window.start, window.end)
+		const cd = COOLDOWN_MS[actionId]
```
AzariahAeroborn, line 77: nit: I know cd is a pretty common abbreviation, but could this be better as cooldown or skillCooldown instead, to be more obvious to a future maintainer?

src/parser/jobs/drg/modules/Drift.tsx
Rietty: I'll admit I just kinda copy pasted the math from the MCH module, so I'll have to review it and maybe check it again. Do you have any suggestion for the first thing you pointed out? How to resolve/avoid that issue?

src/parser/jobs/drg/modules/Drift.tsx
Rietty: Fixed this in e021fe5.

src/parser/jobs/drg/modules/Drift.tsx
yumiyafangirl: Line 81 checks that the drift > buffer (drift is significant enough to report) and downtime < cd (downtime isn't longer than the CD itself, in which case when the downtime ends we'll likely have lots of "reopening" to do and we don't want to count this as drifting).

yumiyafangirl: As for line 76, I agree the calculation might be too lenient. If you have an idea for a more accurate implementation I would love to port it to the MCH module (and/or move it to core).

Rietty: In regards to:

> I haven't run this in local yet, so I'm basing this comment on screenshots posted in Discord, but I wonder if the presentation would be better if the data table had every usage (whether it drifted or not) and then showed 0s drift on the "on time" uses, more similar to what the SMN dot tables are displaying. It was a little jarring to me to see the screenshots with 3+ minute gaps between High Jump entries but only a few seconds of drift on the second use. 

I followed the suggestion and changed the table a bit, so it will have all the abilities showing, with a highlight per substantial drift. 
image In commit d86ae19.

Rietty:
Personally I think I'm done with any extra changes and what not regarding this module. A bit of the math regarding cooldowns and downtime is a bit iffy, so if anyone has a better option/idea on how to handle it, it'd be nice to have that input. But as it currently is, the module is functionally useful and usable for all intents and purposes.

src/parser/jobs/drg/modules/Drift.tsx
```diff
@@ -0,0 +1,172 @@
...
+	[ACTIONS.HIGH_JUMP.id]: ACTIONS.HIGH_JUMP.cooldown * 1000,
+	[ACTIONS.GEIRSKOGUL.id]: ACTIONS.GEIRSKOGUL.cooldown * 1000,
+}
+
+class DriftWindow {
+	actionId: number
+	start: number
+	end: number = 0
+	drift: number = 0
+	abilityRotation: CastEvent[] = []
```
yumiyafangirl, line 32: `abilityRotation` and associated methods don't need to be here for DRG's implementation.

Rietty: Er- What do you mean? They're being used later on in the module? If you remove them and the methods, can't really push anything to window?

Toastdeib: Unless I'm misreading the code, seems like each window only ever has a single ability in it; as soon as you use `addAbility()` and push a HJ or GSK, it closes the window and creates a new one for the next cast. I'm assuming that was yumi's point - that you could replace the array with a single `CastEvent` and just assign it rather than futzing around with adding it and then having to slice in `getLastActionId()`.

Rietty: Oh, I see now. I'll see if I can edit that and fix it up.

AzariahAeroborn: Or, from that perspective, since all this module cares about is the start & end times of the DriftWindow, you could just assign those based on the appropriate event.timestamp values and not have any ability / abilityRotation (or addAbility method) at all.  Whichever approach makes more sense.

Rietty: Addressed in 301a373

PR #924, opened 2020-06-05
Report sources - part 1

ackwell:
This PR is the first part of re-implementing the "report flow" used to drill down to the analysis page, and is the first step of unlinking our implementation details from the direct shape of fflogs reports.

In brief, it:
- Adds the initial skeleton of the new "report source" system, which we'll be using to remove our current reliance on FFLogs' report shape.
- Wires the new report source system through a new "find"/"analyse" route framework.
- In service of the above, rewrites the entire breadcrumb system to better serve how we'll be using it.
- Sets up a number of "legacy" adapters to utilise the existing report and analysis code within the new system while porting efforts are underway.


ackwell:
Aite I think this is good to go. Will merge tomorrow or so.

If anyone can get some testing in that would be awesome. Existing paths through the site have a number of changes under the hood, but should act identically. Additionally, replacing `/analyse` or `/find` in the URL with `/fflogs` should drop you into the new report flow system - it's ugly as fuck, but should be functionally equivalent.

ackwell:
you know who it is

PR #927, opened 2020-06-08
DRG: Add Battle Litany Module

ebshimizu:
Added a module to track uses of Battle Litany for DRG.

**Implementation notes**
- Only tracks the windows started by the character being analyzed
- If a window clips into another DRG's Litany, an additional message will show on the analysis and an extra column will show up on the table indicating if a window overlapped/clipped another Litany.
- Tracks number of players affected as a target out of 8 (assumes full party for now)
- Does not output suggestions right now (could output a minor for not affecting all party members every cast?)

**Previews**
Clean 
image

Overlapping
image


src/parser/jobs/drg/modules/BattleLitany.tsx
```diff
@@ -0,0 +1,232 @@
...
+
+	private onCast(event: CastEvent) {
+		const lastWindow: BLWindow | undefined = _.last(this.history)
+
+		// If we don't have a window, bail
+		if (!lastWindow) {
+			return
+		}
+
+		const action = getDataBy(ACTIONS, 'id', event.ability.guid) as TODO
```
ayyaruq, line 138: This should be using the data dependency for `getAction(event.ability.guid)` which returns an `Action` type.

PR #933, opened 2020-06-12
Fix warning duration for whm spells when they never use them

nonowazu:
* Make 2 lines of change after over an hour of work, proving that people in charge need to reevaluate my skills as a dev
* Modify said file to modify the time held duration based on casts – if 0 casts, then it's the fight duration; otherwise the regular logic should hold


ayyaruq:
Looks good to me, maybe some Typescript magic to validate `held` exists could help, I'm not 100% sure tho.

nonowazu:
It looks like the interface `CooldownTrack` has `held` set as a not-nullable field, and the method that creates it always initializes it with a 0 as the default.

This took me a bit longer than I liked, but I'm not 100% familiar with this module.

src/parser/jobs/whm/modules/oGCDs.tsx
```diff
@@ -139,9 +145,10 @@ export default class OGCDs extends Module {
 		}))
 
 		OTHER_COOLDOWNS_TRACKED.forEach(action => {
+			const cooldownUsage = this.getCooldownUsage(action.id)
 			const maxUses = Math.ceil(this.parser.fightDuration / (action.cooldown * 1000))
-			const uses = this.cooldownUsage[action.id].uses
-			const held = uses > 0 ? this.cooldownUsage[action.id].held : pullDuration
+			const uses = cooldownUsage.uses
+			const held = uses > 0 ? cooldownUsage.held : pullDuration
```
ayyaruq, line 151: I'm still not sure about this, as I wrote in Discord, it doesn't get closed out for end of fight. I think if you took each window and did a final pass at the start of `onComplete()`, for anything that came off CD earlier than end of fight minus the CD, and set held based on that, you'd catch any missed final windows, and the held time would be pullDuration if they never used it anyway.

PR #934, opened 2020-06-12
Report flow: ActorList UI

ackwell:
Style up the new report flow actor list. We gettin close to flippin the switch bois

Before
image

After
image


ackwell:
Let's break some stuff

PR #936, opened 2020-06-14
MCH: add detonator to actions

yumiyafangirl:
Detonator now shows up on the timeline and as an incorrect weave (when applicable). Doesn't appear to affect any other MCH modules. Here's a log where detonator is weaved incorrectly (2:48 weave window): https://www.fflogs.com/reports/fHgM438FwAXWt9KJ

### DIFF
--- a/src/parser/jobs/mch/modules/Cooldowns.js
+++ b/src/parser/jobs/mch/modules/Cooldowns.js
@@ -4,6 +4,7 @@ import ACTIONS from 'data/ACTIONS'
 export default class Cooldowns extends CoreCooldowns {
 	static cooldownOrder = [
 		ACTIONS.WILDFIRE.id,
+		ACTIONS.DETONATOR.id,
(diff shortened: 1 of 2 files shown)


src/parser/jobs/mch/modules/Cooldowns.js
```diff
@@ -11,6 +11,7 @@ export default class Cooldowns extends CoreCooldowns {
 		ACTIONS.REASSEMBLE.id,
 		ACTIONS.AUTOMATON_QUEEN.id,
 		ACTIONS.QUEEN_OVERDRIVE.id,
+		ACTIONS.DETONATOR.id,
```
Toastdeib, line 14: I was going to suggest sorting this directly below Wildfire, since the two are tied together. I'm don't feel too strongly either way though, since it should show up on the timeline so infrequently, so ping me once you've either dismissed this comment or made the change and I'll go ahead with the merge.

PR #938, opened 2020-06-15
Knoweth thy Fist

ayyaruq:
When players pull in RoW (or any other Fist, or none really) and never change over the fight, we're unable to detect precisely what Fist they're in. This PR adds a suggestion for when we don't know, and updates the stats block for Fists and Table for RoF with the relevant information.

image
image

### DIFF
--- a/locale/de/messages.json
+++ b/locale/de/messages.json
@@ -560,6 +560,7 @@
   "mnk.rof.suggestions.tackle.content": "",
   "mnk.rof.suggestions.tackle.why": "",
   "mnk.rof.table.header.gcds": "GCDs",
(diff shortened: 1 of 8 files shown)


ayyaruq:
Pro tip for reviewing this - set GitHub diff to ignore whitespace, the diff looks huge and unwieldy if you don't.
image

Rjb99:
FWIW looks good to me. 

Though regarding your fist pie chart, You don't translate the other fists either, so it might be related to trying to pass in a trans Id to the PieChartStatistic.tsx that's causing it to freak out. The jobs that use a PieChart that translate like SCH, use actionLinks to translate it seems.

PR #939, opened 2020-06-17
[Bug Fix] [Core] BuffWindow's reduceExpectedGCDsEndOfFight Function

Rjb99:
BuffWindow's reduceExpectedGCDsEndOfFight function could make Samurai's expectedGCD go negative. Instead of making an override for it in SAM's BuffWindow file, I've fixed Core to only reduce if the new calculated expectedGCD is less than the current one. to avoid getting 0 or -X expected GCDs for other jobs that may have a similar problem now or in the future.

If this is an issue, I can just override SAM's BuffWindow and leave core alone.

### DIFF
--- a/src/parser/core/modules/BuffWindow.tsx
+++ b/src/parser/core/modules/BuffWindow.tsx
@@ -219,7 +219,13 @@ export abstract class BuffWindowModule extends Module {
 
 			if (windowDurationMillis >= fightTimeRemaining) {
 				const gcdEstimate = this.globalCooldown.getEstimate()
-				return Math.ceil((windowDurationMillis - fightTimeRemaining) / gcdEstimate)
+				const newExpectedGCDs = Math.ceil((windowDurationMillis - fightTimeRemaining) / gcdEstimate)
+
+				if ( this.expectedGCDs ) {
+					if ( newExpectedGCDs <= this.expectedGCDs.expectedPerWindow ) {
+						return newExpectedGCDs
+					}
+				}
 			}
 		}
 


src/parser/core/modules/BuffWindow.tsx
```diff
@@ -219,7 +219,13 @@ export abstract class BuffWindowModule extends Module {
 
 			if (windowDurationMillis >= fightTimeRemaining) {
 				const gcdEstimate = this.globalCooldown.getEstimate()
-				return Math.ceil((windowDurationMillis - fightTimeRemaining) / gcdEstimate)
+				const newExpectedGCDs = Math.ceil((windowDurationMillis - fightTimeRemaining) / gcdEstimate)
+
+				if ( this.expectedGCDs ) {
+					if ( newExpectedGCDs <= this.expectedGCDs.expectedPerWindow ) {
+						return newExpectedGCDs
```
AzariahAeroborn, line 226: This is going to result in not allowing any reduction in expected GCDs for a situation like "used Meikyo with 1.5 seconds left in the fight", since the duration vs gcdEstimate is coming up with 4 expected GCDs to forgive.

Also, naming this variable `newExpectedGCDs` is wrong - what's being calculated is the number of GCDs to remove, not what the expected amount per the window is (getBuffWindowExpectedGCDs subtracts the result of this function from the baseline expectation per window, in addition to adding the values the class overrides provide, if any).

I think capping the overall result of getBuffWindowExpectedGCDs to a minimum of 0 or 1 may make some sense, but it's probably better for classes with "non-standard" buffs like Acceleration or Meikyo (whose duration is much longer than the number of casts required to consume all the stacks and early-end the buff) to override reduceExpectedGCDsEndOfFight directly with logic that properly accounts for their implicitly shorter "actual" duration.

Rjb99: Point Taken, I'll close this PR and just redo the logic on a override using just the fight duration left / estimatedGCD

PR #946, opened 2020-06-25
AoE Usages cleanup

yumiyafangirl:
- Updated wording from "abilities" to "actions" per the conversation here: https://discordapp.com/channels/441414116914233364/470050726014615563/709067420035186688

- Added a column "Minimum targets" that exposes our threshold to presume correctness of AoE usages

- Adjusted the wording/capitalization of the columns to bring it in line with other modules

- Added an implementation for bard:

image


ayyaruq:
I kinda feel like the interface should be `AoEAction` because the E is a proper part of the name. Re making table titles lower case... IMO having properly capitalised words is correct, and the other modules should fix their stuff, rather than making this one equally incorrect.

KelosPelmand:
All our other tables except the Barrage table use Title Casing for their headers, so it would be better for consistency to have the titles be "AoE Action Used" and such.

Wording nit: If the title of the column is "Minimum targets", I think the values should be just "2" rather than "2+".  To use "2+", the title would be "Required targets" or such.

ayyaruq:
Maybe even Optimal Targets. I'm really not sure where the "other modules" bit re titles is coming in from tho, BRD does make sense given other contents of this PR.

yumiyafangirl:
It would be nice to standardize this one way or the other. The core suggestions use sentence case ("Always be casting", "Use your cooldowns"), but I do see now that a few suggestions and tables are using full capitalization. I suppose it's a regional thing but where I'm from table headers and subheaders only capitalize the first word (Chicago Manual of Style & APA off the top of my head).

I think "Minimum Targets" is clearest so I'll just remove the +.

ayyaruq:
Core suggestions and sentences are not titles. As far as standards go, we were standardised from the beginning and use the Queen's English (ie British/Australian English).

yumiyafangirl:
British English is a dialect, not a style, and I don't think we're fully standardized there either - I've definitely seen American spellings in a handful of places.

ayyaruq:
Those should prob be cleaned up over time, and we should maybe document our preferred styles. I've never seen more than a single style for capitalisation through university in Australia or the UK so I always thought it was a single standard, I could be wrong though. This PR isn't really for how we label stuff so for now, let's revert that particular change, address the rest of the PR, and make a GitHub issue to solidify our style?

src/parser/core/modules/AoEUsages.tsx
```diff
@@ -8,16 +8,16 @@ import Suggestions, {SEVERITY, TieredSuggestion} from 'parser/core/modules/Sugge
...
 import {Table} from 'semantic-ui-react'
 
-export interface AoeAbility {
+export interface AoEAction {
 	/**
-	 * The AoE ability to check usages of.
+	 * The AoE action to check usages of.
 	 */
-	aoeAbility: Action
+	AoEAction: Action
```
ayyaruq, line 15: Hypernit, we mostly use snake-case for these kinda things, if you look at other core modules they're all like `aoeAction`. My initial comment about the E was for the interface title where we case appropriately. Not a big deal, just a consistency thing.

PR #948, opened 2020-06-26
[Bug Fix][Sam] Meikyo Window End of Fight rushing

Rjb99:
Sam has a bug currently that the default end of fight rushing can make the expected GCDs go negative. this PR fixes that problem.

There is however currently a minor problem in the way I reduce the GCDs. If a samurai uses Meikyo in the beginning of the weave window, the time stamp for the buff can sometimes cause the system to think they can get an extra GCD when in reality they can not. I'm unsure of how I can fix that at the moment and would love some ideas.

### DIFF
--- a/src/parser/jobs/sam/modules/Meikyo.tsx
+++ b/src/parser/jobs/sam/modules/Meikyo.tsx
@@ -74,8 +74,8 @@ reduceExpectedGCDsEndOfFight(buffWindow: BuffWindowState): number  {
 			const fightTimeRemaining = this.parser.pull.duration - (buffWindow.start - this.parser.eventTimeOffset)
 
 			if (windowDurationMillis >= fightTimeRemaining) {
-				// const gcdEstimate = this.globalCooldown.getEstimate()
-				const possibleGCDs = Math.ceil(fightTimeRemaining / SAM_BASE_GCD_SPEED_BUFFED)
+				// This is using floor instead of ceiling to grant some forgiveness to first weave slot casts at the cost of 2nd weaves might be too forgiven
+				const possibleGCDs = Math.floor(fightTimeRemaining / SAM_BASE_GCD_SPEED_BUFFED)
 
 				if (possibleGCDs < SEN_GCDS) {
 					const reduceGCDsBy = (SEN_GCDS - possibleGCDs)
--- a/src/parser/jobs/sam/modules/Meikyo.tsx
+++ b/src/parser/jobs/sam/modules/Meikyo.tsx
@@ -14,8 +14,8 @@ const ONLY_SHOW = new Set([ACTIONS.HAKAZE.id, ACTIONS.JINPU.id, ACTIONS.ENPI.id,
 const SEN_GCDS = 3
 
 // A set const for SAM speed with 0 speed and shifu up, not sure I like this idea tbh but Aza requested it.
-// TODO: Check when adding Iais to table that this set speed doesn't cause missed skills to be not counted against
-const SAM_BASE_GCD_SPEED_BUFFED = 2.18
+// GCD = 2.18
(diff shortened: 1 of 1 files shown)


src/parser/jobs/sam/modules/Meikyo.tsx
```diff
@@ -58,4 +64,27 @@ considerAction(action: Action) {
...
+			if (windowDurationMillis >= fightTimeRemaining) {
+				// This is using floor instead of ceiling to grant some forgiveness to first weave slot casts at the cost of 2nd weaves might be too forgiven
+				const possibleGCDs = Math.floor(fightTimeRemaining / SAM_BASE_GCD_SPEED_BUFFED)
+
+				if (possibleGCDs < SEN_GCDS) {
+					const reduceGCDsBy = (SEN_GCDS - possibleGCDs)
+					return reduceGCDsBy
+				}
+			}
+		}
```
ayyaruq, line 85: Kinda over complicated imo. Why not initialise like `let reduceGCDsBy = 0`, always return it, and modify it, or just clamp or `Math.max` the value.

```
let reduceGCDsBy = 0

if (windowDurationMillis >= fightTimeRemaining) {
  const possibleGCDs = Math.floor(fightTimeRemaining / SAM_BASE_GCD_SPEED_BUFFED)

  if (possibleGCDs < SEN_GCDS) {
    reduceGCDsBy -= (SEN_GCDS - possibleGCDs)
  }
}

return reduceGCDsBy
```

or even like, 
```
if (windowDurationMillis >= fightTimeRemaining) {
  const possibleGCDs = Math.floor(fightTimeRemaining / SAM_BASE_GCD_SPEED_BUFFED)

  return Math.max((SEN_GCDS - possibleGCDs), 0)
}

return 0
```

src/parser/jobs/sam/modules/Meikyo.tsx
```diff
@@ -6,10 +6,16 @@ import ACTIONS, {Action} from 'data/ACTIONS'
...
+import {BuffWindowState} from 'parser/core/modules/BuffWindow'
 
 // Set for stuff to ignore TODO: revisit this and get it to show iaijutsu properly
 // const IGNORE_THIS = new Set([ACTIONS.MIDARE_SETSUGEKKA.id, ACTIONS.TENKA_GOKEN.id, ACTIONS.HIGANBANA.id, ACTIONS.KAESHI_SETSUGEKKA.id, ACTIONS.KAESHI_GOKEN.id, ACTIONS.KAESHI_HIGANBANA])
 const ONLY_SHOW = new Set([ACTIONS.HAKAZE.id, ACTIONS.JINPU.id, ACTIONS.ENPI.id, ACTIONS.SHIFU.id, ACTIONS.FUGA.id, ACTIONS.GEKKO.id, ACTIONS.MANGETSU.id, ACTIONS.KASHA.id, ACTIONS.OKA.id, ACTIONS.YUKIKAZE.id])
+const SEN_GCDS = 3
+
+// A set const for SAM speed with 0 speed and shifu up, not sure I like this idea tbh but Aza requested it.
+// GCD = 2.18
+const SAM_BASE_GCD_SPEED_BUFFED = 2180
```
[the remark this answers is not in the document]
Rjb99: to be quite honest, no one is going to want run this slow. 2.18 isn't really a supported SAM GCD.
2.15, 2.07, 2.0 are the 3 speeds that a user will normally run with a +/- 0.01 variance. 

This to be honest might end up more forgiving for faster samurai in terms of end of fight hits but will cover more consistently then using GCDEstimates in some cases with users that have really bad GCD uptime

src/parser/jobs/sam/modules/Meikyo.tsx
ayyaruq: I think the latter was cleaner fwiw but yolo

PR #953, opened 2020-06-29
More adjustments to DRG Modules.

Rietty:
Changed display order of modules to make more sense, copied the samurai tincture buff overwrite to account for some core stuff regarding STR pot/infusions & added a small header to BL to be more consistent with DS/LC.


Rietty:
> Some broken imports to resolve, and I'd make the table headers less inflected like it's from Australia.
Fixed all of this by making it consistently singular in 0e0bb08.

PR #956, opened 2020-07-02
Moving some UI components to typescript

MikeMatrix:
Did some moving to typescript stuffies.

👀 I was bored ok?


src/utilities/random.ts
```diff
@@ -1,9 +1,11 @@
+type RandomRollFnc = () => 4
+
 const RANDOM_NUMBER = 4
```
ayyaruq, line 3: I hate you.

src/components/ui/ContributorLabel.tsx
```diff
@@ -0,0 +1,59 @@
...
+		]).isRequired,
+		detail: PropTypes.node,
+	}
+
+	normalizeContributor = (contributor: string | Contributor): Contributor => {
+		if (typeof contributor === 'string') {
+			return {
+				name: contributor,
+				jobs: [],
+			}
```
ayyaruq, line 29: Is this not gonna need a stubbed avatar as well? I can't think of any reason contributor would be a string tho.

src/components/ui/ContributorLabel.tsx
```diff
@@ -0,0 +1,59 @@
...
+		}
+	}
+
+	render() {
+		const {contributor, detail} = this.props
+
+		const {avatar, name, jobs} = this.normalizeContributor(contributor)
+
+		const style: CSSProperties = {}
+		if (jobs && jobs.length) {
```
ayyaruq, line 41: Would `jobs?.length` not work here?

src/components/ui/ContributorLabel.tsx
```diff
@@ -0,0 +1,59 @@
...
+			// I'm assuming the first job they list is their 'main'
+			const job = jobs[0]
+			style.backgroundColor = job.colour
```
ayyaruq, line 44: It'd be real nice if we could somehow get context on what job the user is currently looking at, and opt for that if the specific contributor has that job in their list, but eh.

src/components/Home/Home.tsx
```diff
@@ -36,7 +36,6 @@ class Home extends Component {
 				<img
 					src={process.env.PUBLIC_URL + '/logo.png'}
 					alt="logo"
-					className={styles.logoImage}
```
ayyaruq, line 39: Did this style even exist for this component? I know it's on App but not Home?

src/utilities/random.ts
MikeMatrix: Hey I could've made it worse by making a random number type

src/components/ui/ContributorLabel.tsx
MikeMatrix: No because `avatar` is optional and being checked in the render later if it even exists. As to why contributor would be a string... I have no idea, I was just refactoring what was there.

src/components/ui/ContributorLabel.tsx
MikeMatrix: Yes, that should be `jobs?.length`.
I'll fix it up in a sec. Must have just variable replaced it when I was doing that part.

src/components/ui/ContributorLabel.tsx
MikeMatrix: Maybe I can look at that down the line, but the component is pretty non-context related right now. Maybe I can add a "preferred job" option to it later.

src/components/Home/Home.tsx
MikeMatrix: Nope. I was wondering that as well, because typescript was ree'ing at me as soon as I switched that stuff over. It's not in the styles at all.

PR #957, opened 2020-07-03
BRD - more TS-ification

yumiyafangirl:
The first of several planned rewrites to bard modules, with the end goal of removing all dependencies on additionalstats.js ("yumicode") which is outdated and difficult to maintain. The output remains unchanged.

I'm still learning how to wrestle with TS so I'm certain parts of this PR can/need to be improved upon. 

image


src/parser/jobs/brd/modules/Snapshots.tsx
```diff
@@ -0,0 +1,197 @@
...
+	private snapshots: Snapshot[] = []
+	private currentSnapshot?: Snapshot
+
+	private targets = new Map<string, AbilityEvent[]>()
+
+	protected init() {
+		this.addEventHook('cast', {by: 'player', abilityId: SNAPSHOTTERS}, this._onSnapshot)
+		this.addEventHook('normaliseddamage', {by: 'player', abilityId: DOT_STATUSES}, this._onDotTick)
+		this.addEventHook(['applydebuff', 'refreshdebuff'], this._onApply)
+		this.addEventHook('removedebuff', this._onRemove)
```
ayyaruq, line 65: Do you not want to filter these to more relevant events, or just getting everythang?

src/parser/jobs/brd/modules/Snapshots.tsx
```diff
@@ -0,0 +1,197 @@
...
+			const index = targetStatuses.findIndex(status =>
+				status.ability.guid === event.ability.guid)
+
+			if (index > -1) {
+				targetStatuses.splice(index, 1)
+			}
+		}
+	}
+
+	private _getStatuses(event: CastEvent) {
```
ayyaruq, line 100: Prob more of a nit since TS is good at stuff but with a more complex return type like this, may be good to explicitly declare it.

src/parser/jobs/brd/modules/Snapshots.tsx
```diff
@@ -0,0 +1,197 @@
...
+						}
+					})
+				}
+			</Table.Cell>
+
+			const snapshotBuffCell = <Table.Cell>
+				{
+					snap.statusIDs.map(id => {
+						// Avoid showing statuses we do not currently know of and statuses known not to affect bard DoTs
+						const status = getDataBy(STATUSES, 'id', id)
```
ayyaruq, line 161: Same deal here, plus could just filter when adding stuff to the snapshot in the first place.

src/parser/jobs/brd/modules/Snapshots.tsx
```diff
@@ -0,0 +1,197 @@
...
+				}
+			</Table.Cell>
+
+			const event = snap.snapEvent
+			return <Table.Row key={event.timestamp}>
+				<Table.Cell>
+					{this.util.createTimelineButton(event.timestamp)}
+				</Table.Cell>
+				<Table.Cell>
+					<ActionLink {...getDataBy(ACTIONS, 'id', event.ability.guid)}/>
```
ayyaruq, line 175: Here too, `this.data.getAction`.

src/parser/jobs/brd/modules/Snapshots.tsx
```diff
@@ -0,0 +1,197 @@
...
+
+	private _onApply(event: AbilityEvent) {
+		const targetKey = `${event.targetID}-${event.targetInstance}`
+
+		if (!this.targets.has(targetKey)) {
+			this.targets.set(targetKey, [])
+		}
+
+		const target = this.targets.get(targetKey)
+		if (target) {
```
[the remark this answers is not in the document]
yumiyafangirl: You would think so, but TS disagrees :(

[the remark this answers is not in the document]
ayyaruq: Prob because `Map.get()` can be undefined I guess.

yumiyafangirl:
I added a bunch of stuff so this PR is outdated, I'll make a new one.

PR #959, opened 2020-07-03
Add in the old IDs for Chi and Jin to fix buggy timelines

Toastdeib:
What it says on the tin. I also refactored `TEN_KASSATSU` to `TEN_NEW` since my assumption about when 18805 gets used over 2259 doesn't apply to Chi and Jin and I'd rather keep their names consistent.

### DIFF
--- a/src/parser/jobs/nin/modules/TrickAttackUsage.js
+++ b/src/parser/jobs/nin/modules/TrickAttackUsage.js
@@ -32,7 +32,9 @@ export default class TrickAttackUsage extends Module {
 	_onCast(event) {
 		const action = getDataBy(ACTIONS, 'id', event.ability.guid)
 		if (event.timestamp >= this.parser.fight.start_time && action && action.onGcd && !(
-			action.id === ACTIONS.TEN.id || action.id === ACTIONS.TEN_KASSATSU.id || action.id === ACTIONS.CHI.id || action.id === ACTIONS.JIN.id)) {
+			action.id === ACTIONS.TEN.id || action.id === ACTIONS.TEN_NEW.id ||
+			action.id === ACTIONS.CHI.id || action.id === ACTIONS.CHI_NEW.id ||
+			action.id === ACTIONS.JIN.id || action.id === ACTIONS.JIN_NEW.id)) {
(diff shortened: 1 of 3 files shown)


src/parser/jobs/nin/modules/TrickAttackUsage.js
```diff
@@ -32,7 +32,9 @@ export default class TrickAttackUsage extends Module {
 	_onCast(event) {
 		const action = getDataBy(ACTIONS, 'id', event.ability.guid)
 		if (event.timestamp >= this.parser.fight.start_time && action && action.onGcd && !(
-			action.id === ACTIONS.TEN.id || action.id === ACTIONS.TEN_KASSATSU.id || action.id === ACTIONS.CHI.id || action.id === ACTIONS.JIN.id)) {
+			action.id === ACTIONS.TEN.id || action.id === ACTIONS.TEN_NEW.id ||
+			action.id === ACTIONS.CHI.id || action.id === ACTIONS.CHI_NEW.id ||
```
AzariahAeroborn, line 36: This is getting pretty unwieldy. Can this be converted to a MUDRAS constant array instead?

src/parser/jobs/nin/modules/TrickAttackWindow.js
```diff
@@ -90,7 +90,9 @@ export default class TrickAttackWindow extends BuffWindowModule {
 
 	considerAction(action) {
 		// Ten, Chi, and Jin should be ignored for purposes of GCD counts
-		return !(action.id === ACTIONS.TEN.id || action.id === ACTIONS.TEN_KASSATSU.id || action.id === ACTIONS.CHI.id || action.id === ACTIONS.JIN.id)
+		return !(action.id === ACTIONS.TEN.id || action.id === ACTIONS.TEN_NEW.id ||
+			action.id === ACTIONS.CHI.id || action.id === ACTIONS.CHI_NEW.id ||
```
AzariahAeroborn, line 94: :ditto:

src/parser/jobs/nin/modules/TrickAttackUsage.js
Toastdeib: It can, just not sure where the best place to put it would be since I wouldn't want to define the array twice. I'll think on it when it isn't 2 AM.

PR #961, opened 2020-07-03
Retiring now outdated dragoon rotation watchdog module.

Rietty:
This module can now be retired due to changes to the core Dragoon rotation from an 11 GCD to a 10 GCD system. As well as the existence of the new combo and positional modules in XIVA. Both of these combined provide the same information as this module did.


src/parser/jobs/drg/modules/notes.md
```diff
@@ -26,4 +26,4 @@
 - [x] Chaos Thrust uptime and clipping (Debuffs.js)
 
 ### Rotation
-- [ ] Adhering to the standard 11-GCD rotation
+- [-] Adhering to the standard 11-GCD rotation (This has been removed.)
```
ayyaruq, line 29: You can cross this out with a strikethrough in another PR, not gonna block merging for it but just FYI. No idea what's up with the `/` and `-`, GitHub markdown just supports `-` as far as I know.

PR #963, opened 2020-07-07
BRD - Snapshots Rewrite

yumiyafangirl:
Rewrote the Snapshots module in TS and culled some undesirable AdditionalStats.js code (the end goal is to refactor/remove all of it).  Also split the output table by target, and generally cleaned up the way statuses are handled - it's more accurate now.

Old:
image

New:
image

New suggestion:
image


src/parser/jobs/brd/modules/Snapshots.tsx
```diff
@@ -0,0 +1,260 @@
...
+		return statuses
+	}
+
+	private onComplete() {
+		const badRefreshes = Array.from(this.targets.values(), target => target.snapshots)
+			.reduce((acc, snap) => acc.concat(snap))
+			.filter(snap => snap.badRefresh)
+			.length
+
+		const moduleLink = <a href="javascript:void(0);" onClick={() => this.parser.scrollTo(Snapshots.handle)}><NormalisedMessage message={Snapshots.title}/></a>
```
AzariahAeroborn, line 161: React is complaining:
> Warning: A future version of React will block javascript: URLs as a security precaution. Use event handlers instead if you can. If you need to generate unsafe HTML try using dangerouslySetInnerHTML instead. React was passed "javascript:void(0);".

I know we use this elsewhere in the codebase, but we really need to move away from doing it, so we should avoid it here.  Using `a style={{cursor: 'pointer'}}` (or better yet, a custom component that handles the onClick and the pointer styling itself and can be reused in other places) keeps the "this is a link" pointer without using the JS URL that's going to break on us in the future.

src/parser/jobs/brd/modules/Snapshots.tsx
```diff
@@ -0,0 +1,260 @@
...
+
+		this.suggestions.add(new TieredSuggestion({
+			icon: ACTIONS.IRON_JAWS.icon,
+			content: <Trans id="brd.snapshots.suggestions.content">
+				Avoid using <ActionLink {...ACTIONS.IRON_JAWS} /> when <ActionLink {...ACTIONS.CAUSTIC_BITE} /> and <ActionLink {...ACTIONS.STORMBITE} /> are not present on the target. These Iron Jaws are highlighted in the {moduleLink} module below.
```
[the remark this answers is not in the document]
yumiyafangirl: Works for me, I think we can safely consider an IJ on just Stormbite a "mistake" even though there's no potency loss on that specific GCD. I'll rewrite the suggestion including the comments from acc/toast about the 'why' text.

src/parser/jobs/brd/modules/Snapshots.tsx
```diff
@@ -0,0 +1,260 @@
...
+			target.removeStatus(status)
+		}
+	}
+
+	private onSnapshot(event: CastEvent) {
+		const target = this.getTarget(event)
+		target.addSnapshot(event, this.playerStatuses)
+	}
+
+	private getTarget(fields: {targetID?: number, targetInstance?: number}): Target {
```
ayyaruq, line 133: Do these fields not exist on BuffEvent? I can't recall what came out of our Discord chat but I feel like the `onApply`/`onRemove` handlers should be using BuffEvent rather than AbilityEvent, and then this entire function could be replaced with a getter in the Target class, along with what Az suggested.

src/parser/jobs/brd/modules/Snapshots.tsx
```diff
@@ -0,0 +1,260 @@
...
+		this.statuses.add(status)
+	}
+
+	public removeStatus(status: Status) {
+		this.statuses.delete(status)
+	}
+
+	public addSnapshot(snap: CastEvent, playerStatuses: Status[]) {
+		// If IJ didn't refresh a caustic bite, it was a misuse
+		// (IJ on storm bite is technically potency neutral, though odd)
```
ayyaruq, line 76: Do we care about potency neutrality, or encouraging players to best practice? I don't think it matters much, but worth a thought.

src/parser/jobs/brd/modules/Snapshots.tsx
```diff
@@ -0,0 +1,260 @@
...
+
+const BAD_REFRESH_SEVERITY = {
+	1: SEVERITY.MINOR,
+	2: SEVERITY.MEDIUM,
+	3: SEVERITY.MAJOR,
+}
+
+interface Snapshot {
+	// The event that applied the DoT(s).
+	snapEvent: CastEvent
```
ayyaruq, line 45: This could prob be called `snapshotter` or something, to be more aligned with table output and easier to read.

src/parser/jobs/brd/modules/Snapshots.tsx
```diff
@@ -0,0 +1,260 @@
...
+import Util from './Util'
+import {Data} from 'parser/core/modules/Data'
+import Suggestions, {SEVERITY, TieredSuggestion} from 'parser/core/modules/Suggestions'
+
+import styles from './Snapshots.module.css'
+
+const SNAPSHOTTERS = [
+	ACTIONS.IRON_JAWS.id,
+	ACTIONS.CAUSTIC_BITE.id,
+	ACTIONS.STORMBITE.id,
```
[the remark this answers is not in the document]
yumiyafangirl: But they're statuses, not actions 👀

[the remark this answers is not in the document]
ayyaruq: I can't read 💯

src/parser/jobs/brd/modules/Snapshots.tsx
yumiyafangirl: I changed all AbilityEvent types to BuffEvent, I'm not sure what you mean about a getter in the Target class though. Do you mean a static method in the Target class that takes the map as an arg?

ayyaruq: Basically yeah, but in retrospect it's a bad idea.

PR #964, opened 2020-07-08
Fixed double Devotion issue

KelosPelmand:
With two Summoners, the window would show skills cast during the other Summoner's Devotion.


AzariahAeroborn:
There seems to be a negative side effect, in that this is prematurely closing the devotion window if someone else dies with devotion on: https://www.fflogs.com/reports/dbHTn9MLfmkJgxVv#fight=9&type=auras&view=events&ability=1001213 (the 9:10 window here).  Should the hook passing to _onDevotionRemoved just be for { to: player } instead of every friendly in the fight?

PR #969, opened 2020-07-14
DRG: Refactor and Re-Order Buffs

ebshimizu:
- Re-factored Lance Charge and Dragon Sight into separate modules. These modules derive from a drg-specific buff window parent class (`DrgBuffWindow.tsx`)
- `Buffs.js` -> `Buffs.tsx`. There were a few functions related to Life Surge and Disembowel that were left in there, while the LC/DS code was moved elsewhere. This module has suggestions and a checklist item, but no output
- Modules were re-ordered so that buffs are the last thing on the page before the timeline.
- And a small change while I'm here, Positionals silences its output if all positionals were hit correctly.
- These changes don't really add new stuff, so no changelog entry was added.


src/parser/jobs/drg/modules/DrgBuffWindow.tsx
```diff
@@ -0,0 +1,210 @@
...
+		return this.casts
+			.map(e => this.data.getAction(e.ability.guid))
+			.filter(a => a && a.onGcd)
+			.length
+	}
+}
+
+// the main reason that this doesn't extend core buff windows is that DRG
+// needs a few extra hooks to determine if the buff had an optimal placement,
+// which requires us to know which GCDs were used immediately before the buff
```
AzariahAeroborn, line 47: Given the reasoning of "don't use right after one of the 3d hit skills" and "include 8 GCDs per window", does that mean the expectation is that each buff window should include a total of 2 3d hit skills (1 of each in single target, I would assume, but if multi-dotting is a thing then maybe make the good skills check a little more forgiving to take 2 Chaos Thrust instead of one of each)?  Would that be a better presentation to end users rather than "don't use it right after the big hits" and/or allow for simplifying this logic and directly extending core BuffWindow?

src/parser/jobs/drg/modules/DrgBuffWindow.tsx
```diff
@@ -0,0 +1,210 @@
...
+		} else {
+			return <Fragment>
+				<Message warning>
+					<Trans id="drg.buffs.unused"><ActionLink {...this.buffAction} /> was not used during this fight.</Trans>
+				</Message>
+			</Fragment>
+		}
+	}
+
+	output() {
```
AzariahAeroborn, line 207: Same concern as BattleLitany.tsx - is there a reason for having output() just be a wrapper on another function?

src/parser/jobs/drg/modules/BattleLitany.tsx
```diff
@@ -164,7 +164,7 @@ export default class BattleLitany extends Module {
 
 	// just output, no suggestions for now.
 	// open to maybe putting a suggestion not to clip into other DRG windows? hitting everyone with litany?
-	output() {
+	renderTable() {
```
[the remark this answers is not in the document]
ebshimizu: whoops yeah that (and the other one noted) are changes left over from experiments with combining output from all three buff modules. No reason to have a separate function anymore.

src/parser/jobs/drg/modules/DrgBuffWindow.tsx
ebshimizu: I believe the current thinking wants each window to include CT, FT, and FC/WT, so I think Lance Charge could be swapped pretty easily to core. Dragon Sight needs to be able to track two statuses and add an additional hook for checking solo DS use (but that check could easily go into a different module), so the core BuffWindow would need a slight adjustment for that to work.

We should get a comment from @Rietty or other DRG theorycrafters about which suggestion format would be better (I'm leaning towards switching to core myself).

AzariahAeroborn: Per our discussion on Discord, I’m going to go ahead and merge this and we can move forward with investigating alternatives for the buff windows separately.

PR #970, opened 2020-07-15
SAM - some minor changes

yumiyafangirl:
- Styled the higanbana output so enemy names are capitalized.

image

- "Meditation Stack Waste by Source" is now hidden unless there was waste, and rows with 0 waste are no longer shown.

image

- Also cleaned up some unused imports / deps.


Rjb99:
Looks good to me.

PR #978, opened 2020-08-05
add link to CrowdIn in README.md

AzariahAeroborn:
Start with a simple change, and then we can figure out if we want to add something to the i18n docs page or otherwise.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,11 +27,11 @@ If you would like to contribute new or corrected translation strings, please joi
 
 We support the languages that FFXIV officially supports:
 
-- Chinese Simplified
-- French
-- German
-- Japanese
-- Korean
+- Japanese (日本語)
+- German (Deutsch)
+- French (Français)
+- Korean (한국어)
+- Chinese Simplified (简体中文)
 
 ## Structure of the parser
 
--- a/README.md
+++ b/README.md
@@ -27,11 +27,11 @@ If you would like to contribute new or corrected translation strings, please joi
 
 We support the languages that FFXIV officially supports:
 
-- Chinese Simplified :cn:
-- French :fr:
-- German :de:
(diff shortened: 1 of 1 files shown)


KelosPelmand (dismissed):
Looks good to me.  Most I might suggest would be some sort of "we only do officially supported languages", but the people that statement would target probably come straight to discord, so they wouldn't read the readme file.

AzariahAeroborn:
> I'm with Kelos, would be good to have a note about supported language and also have a list of what those languages are, prob with the English name, localised name, locale name, and maybe a flag because it's cool.

I added the English name.  Not certain of what the correct localised names would be, and they're not on our Crowdin project either.  I experimented with a couple options for flags but I don't think they're really worth the effort to try to style inline in markdown and emoji-flags don't display as flags when someone's viewing via Windows.

README.md
```diff
@@ -12,13 +12,27 @@ If you'd like to contribute, check past [our developer documentation](https://do
...
+We use [Crowdin](https://crowdin.com/project/xivanalysis) to manage and maintain our translations.
+If you would like to contribute new or corrected translation strings, please join our [CrowdIn project](https://crowdin.com/project/xivanalysis).
+
+We support the languages that FFXIV officially supports:
+
+- Chinese Simplified
+- French
+- German
+- Japanese
+- Korean
```
ayyaruq, line 34:
```suggestion
- Japanese (日本語)
- German (Deutsch)
- French (Français)
- Korean (한국어)
- Chinese Simplified (简体中文)
```

ayyaruq (changes requested):
TIL emoji flags don't work in Windows, nice. Here's the localised strings for you.

ayyaruq (approved):
Noice (ナイス)

PR #981, opened 2020-08-08
MCH - Tinctures

yumiyafangirl:
Adds a separate event hook for pet casts in BuffWindows. This hook is ignored unless an optional `petActions: Action[]` exists and is non-empty. These actions will be considered in the rotation section of the table, and can also be added to trackedActions if desired (here we're tracking Pile Bunker since it's by far the biggest flexible damage cd MCH has). 

image

### DIFF
--- a/src/parser/jobs/mch/modules/Tincture.tsx
+++ b/src/parser/jobs/mch/modules/Tincture.tsx
@@ -1,0 +1,66 @@
+import {Trans} from '@lingui/react'
+import {ActionLink} from 'components/ui/DbLink'
+import ACTIONS, {Action} from 'data/ACTIONS'
+import Downtime from 'parser/core/modules/Downtime'
+import {dependency} from 'parser/core/Module'
+import {SEVERITY} from 'parser/core/modules/Suggestions'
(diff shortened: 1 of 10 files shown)


src/parser/jobs/mch/modules/Tincture.tsx
```diff
@@ -0,0 +1,66 @@
...
+				action: ACTIONS.PILE_BUNKER,
+				expectedPerWindow: 1,
+			},
+			{
+				action: ACTIONS.DRILL,
+				expectedPerWindow: 2,
+			},
+		],
+		suggestionContent: <Trans id="mch.tincture.suggestions.trackedActions.content">
+			Try to cover as much damage as possible with your Infusions of Dexterity.
```
Toastdeib, line 37: Wording nit, this should be "Tinctures" not "Infusions". Infusions are soooooo last expansion.

suShirow:
```suggestion
			Try to cover as much damage as possible with your Tinctures of Dexterity.
```

yumiyafangirl:
Should be good to re-review

PR #983, opened 2020-08-08
BRD - Apex Arrow suggestions

yumiyafangirl:
Nothing super fancy. It catches some but not all of the possible Apex misuses, including ghosts and uses that were very obviously too low gauge. This action has a lot of nuance in when it should be used and it's impossible for us to accurately predict gauge, so this module will by no means catch ALL bad uses, but it should catch the most egregious ones at least.

image


suShirow:
Did anything here change due to 5.3?

yumiyafangirl:
> Did anything here change due to 5.3?

Nope, apex is unchanged.

src/parser/jobs/brd/modules/ApexArrow.tsx
```diff
@@ -0,0 +1,108 @@
...
+		this.suggestions.add(new TieredSuggestion({
+			icon: ACTIONS.APEX_ARROW.icon,
+			content: <Trans id="brd.apex.suggestions.bad.content">
+				Avoid using <ActionLink {...ACTIONS.APEX_ARROW} /> without a sufficient amount of Soul Voice.
+			</Trans>,
+			why: <Trans id="brd.apex.suggestions.bad.why">
+				Apex Arrow dealt very low damage <Plural value={this.badApexCasts} one="# time" other="# times"/>.
+			</Trans>,
+			tiers: {
+				1: SEVERITY.MINOR,
```
suShirow, line 101: Isn't just 1 misuse a massive waste of at least medium? How much less damage would it do? Almost as bad as if it were just ghosted. I don't know bard very well though, ignore if this has been thought through already ^^

src/parser/jobs/brd/modules/ApexArrow.tsx
```diff
@@ -0,0 +1,108 @@
...
+	static debug = false
+
+	@dependency private stats!: AdditionalStats
+	@dependency private suggestions!: Suggestions
+
+	private ghostedApexCasts: number = 0
+	private badApexCasts: number = 0
+
+	protected init() {
+		this.addEventHook('normaliseddamage', {by: 'player', abilityId: ACTIONS.APEX_ARROW.id}, this.onApex)
```
suShirow, line 32: I recommend calling 
`	@dependency private data!: Data`
and then using
`this.data.actions.APEX_ARROW.id` for all uses of ACTIONS.xyz in order to start supporting layered data. Especially since this is a new module, better to keep up!

yumiyafangirl: Thanks, I didn't even know this was necessary in order to make use of the layers.

yumiyafangirl:
I added a little more to the debug output in onApex() so we can better track down issues with the potency calc if they ever arise. Here's what the output looks like in a dungeon log where I know apex was always used at 90+ gauge: 

image

Max potency is 600, so it's looking reasonably accurate to me. I also confirmed on a handful of known bad apex casts. This module is still too lenient, but I think it's better to err on the safe side and only make a suggestion when the potency was **very** low (lower than a filler GCD). Accordingly, I agree that it makes sense to start the severity at medium for 1 bad use, and I've added this into the PR.

PR #989, opened 2020-08-10
Bump GNB to 5.3

csteward24:
First PR, let me know if I did anything wrong


src/parser/jobs/gnb/changelog.tsx
```diff
@@ -52,6 +52,14 @@ export const changelog = [
 		</>,
 		contributors: [CONTRIBUTORS.RYAN],
 	},
+	{
+		date: new Date('2020-08-11'),
+		Changes: () => <>
+			Brutal Shell duration has been extended in patch 5.3.
```
nonowazu, line 58: Really a nit more than anything – do you think it might be better to have this set as:

```tsx
Changes: () => <>Brutal Shell duration changes and added support for patch 5.3</>,
```

Both in that it's one line and should be readable even without the whitespace and to expand the comment to note that you extended the duration via layer *and* that support for 5.3 is live as of this PR.

csteward24: Sounds good to me. When I wrote this originally I forgot it was public. Will commit the changes.

PR #999, opened 2020-08-11
AST 5.3 bump

suShirow:
Seems like all I do these days it just deprecate code out of the AST section ;_; 

PLEASE SQUEENIX GIVE ALL MY CARDS BACK

- [x] **arcanalogs**: removal of sleevedraw from current patch logic
- [x] **arcanalogs**:  addition of SD doing the same thing that Draw does
- [x] **lightspeed**, draw: change to MP in messaging
- [x] **horoscope**: something about them not expiring? Needs testing
- [x] **layered data**: updates to 503 AST MP values
- [x] **Neutral sect**: check for changes to status buff IDs (SHAKES FIST)


suShirow:
Draft mode until I can test updates in-game.

src/data/ACTIONS/layers/patch5.3.ts
```diff
@@ -13,5 +13,13 @@ export const patch530: Layer<ActionRoot> = {
 
 		// MNK 5.3 cooldown changes
 		PERFECT_BALANCE: {cooldown: 90},
+
+		// AST 5.3 changes
```
ayyaruq, line 17: Summary like, `AST 5.3 MP cost changes` or something makes this comment more meaningful.

src/data/STATUSES/layers/patch5.3.ts
```diff
@@ -6,5 +6,11 @@ export const patch530: Layer<StatusRoot> = {
 	data: {
 		// 5.3 GNB Changes - Brutal Shell updated duration
 		BRUTAL_SHELL: {duration: 30},
+
+		// 5.3 AST Changes - Neutral Sect (Noct)
+		DIURNAL_BALANCE: {
+			name: 'Diurnal Balance (Benefic)',
+		},
+
```
ayyaruq, line 14: Note to self, this is gonna conflict with WAR changes, whichever is more ready for merge can go in and Ay will feex the other.

src/parser/jobs/ast/index.tsx
```diff
@@ -36,6 +36,13 @@ export default new Meta({
 		{user: CONTRIBUTORS.SUSHIROU, role: ROLES.MAINTAINER},
 	],
 	changelog: [
+		{
+			date: new Date('2020-07-13'),
```
ayyaruq, line 40: I think you're a month off. Sentences also have periods but I only care about grammar for @nonowazu I guess.

src/parser/jobs/ast/modules/ArcanaTracking/ArcanaTracking.tsx
```diff
@@ -119,6 +119,11 @@ export default class ArcanaTracking extends Module {
 	}
 
 	normalise(events: Event[]) {
+		// The whole point of this module was to support "pre-5.3" Three-card Sleeve Draw
+		if (this.parser.patch.before('5.3')) {
+			return events
+		}
+
```
ayyaruq, line 126: If this only was needed before 5.3, should the conditional not be `!this.parser.patch.before('5.3')`? This is ignoring it for before 5.3 instead of including and after.

src/parser/jobs/ast/modules/ArcanaTracking/ArcanaTracking.tsx
```diff
@@ -347,7 +352,9 @@ export default class ArcanaTracking extends Module {
 			cardStateItem.drawState = undefined
 		}
 
-		if (actionId === this.data.actions.SLEEVE_DRAW.id) {
+		if (actionId === this.data.actions.SLEEVE_DRAW.id
+			&& this.parser.patch.before('5.3')) {
```
ayyaruq, line 356: This is short enough to be one line, it's even shorter than the one above.

src/parser/jobs/ast/modules/Draw.tsx
```diff
@@ -15,7 +15,7 @@ import ArcanaTracking from './ArcanaTracking/ArcanaTracking'
 // Track them using Draw when they still have a minor arcana (oopsie) or a card in the spread
 
 const CARD_DURATION = 15000
-const SLEEVE_DRAW_PLAYS_GIVEN = 3
+const SLEEVE_DRAW_PLAYS_GIVEN = 1
```
ayyaruq, line 18: Is this patch conditional? You can take a look at MNK's Forms module and use a ternary or something to switch it out as appropriate.

src/parser/jobs/ast/modules/Draw.tsx
```diff
@@ -79,6 +79,12 @@ export default class Draw extends Module {
 	}
 
 	private onDraw(event: CastEvent) {
+
+		// ignore precasted draws
+		if (event.timestamp < this.parser.fight.start_time) {
+			return
+		}
```
ayyaruq, line 86: Does ignoring it break pre-5.3, or is this a bug fix?

src/parser/jobs/ast/modules/Draw.tsx
```diff
@@ -162,10 +168,10 @@ export default class Draw extends Module {
 		// Prepull consideration: + 1 play
 
 		// Begin Theoretical Max Plays calc
-		const fightDuration = this.parser.fight.end_time - this.parser.fight.start_time
-		const maxSleeveUses = Math.floor((fightDuration - (CARD_DURATION*2)) / (this.data.actions.SLEEVE_DRAW.cooldown * 1000)) + 1
+		const fightDuration = this.parser.pull.duration
+		const maxSleeveUses = Math.floor(Math.max(0, (fightDuration - (CARD_DURATION*2))) / (this.data.actions.SLEEVE_DRAW.cooldown * 1000)) + 1
 		const playsFromSleeveDraw = maxSleeveUses * SLEEVE_DRAW_PLAYS_GIVEN
-		const playsFromDraw = Math.floor((fightDuration - CARD_DURATION) / (this.data.actions.DRAW.cooldown * 1000)) + 1
+		const playsFromDraw = Math.floor(Math.max(0, (fightDuration - CARD_DURATION)) / (this.data.actions.DRAW.cooldown * 1000)) + 1
```
ayyaruq, line 174: I assume this is for clarity or bugfix?

src/parser/jobs/ast/modules/Horoscope.tsx
```diff
@@ -1,24 +1,26 @@
...
 import Suggestions, {SEVERITY, TieredSuggestion} from 'parser/core/modules/Suggestions'
 import React from 'react'
 
 // Things to track:
-// Did they forget to activate horoscope?
 // Did they cast a Helios/Aspected Helios without Horoscope despite having it up?
 // Maybe track how they used it?
 
-const SEVERETIES = {
+const SEVERITIES = {
```
ayyaruq, line 15: 🤩

src/parser/jobs/ast/modules/Horoscope.tsx
```diff
@@ -50,23 +61,48 @@ export default class Horoscope extends Module {
...
+					Plan to have <ActionLink {...this.data.actions.HOROSCOPE} /> up before you need to cast <ActionLink {...this.data.actions.HELIOS} /> or <ActionLink {...this.data.actions.ASPECTED_HELIOS} /> .
+				<StatusLink {...this.data.statuses.HOROSCOPE_HELIOS} /> may help to cover more damage later without needing to cast more AOE heals.
+			</Trans>,
+				tiers: SEVERITIES.WASTED_AOE_HEAL_TIERS,
+				value: this.nonHoroscopeHeals,
+				why: <Trans id="ast.synastry.suggestion.why">
+					<Plural value={this.nonHoroscopeHeals} one="# single-target GCD heal was cast" other="# single-target GCD heals were cast" /> without synastry even though it was available.
+			</Trans>,
+			}))
+		}
```
ayyaruq, line 106: Should the i18n keys here match up? This doesn't appear to be Synastry.

src/parser/jobs/ast/modules/Draw.tsx
suShirow: This is a fix from back when we started spoofing precast stuff. Didn't catch this. It's a bug fix. Forgot to make the distinction clear sry

src/parser/jobs/ast/modules/Draw.tsx
suShirow: You're right, this will break prepatch. I'll check out punchypunch

src/parser/jobs/ast/index.tsx
suShirow: Oh gawd my time machine keeps acting up

src/parser/jobs/ast/modules/Draw.tsx
suShirow: Both

src/parser/jobs/ast/modules/Horoscope.tsx
suShirow: _(:3 」∠)_

src/parser/jobs/ast/modules/Horoscope.tsx
suShirow: sloppyyyyy

src/parser/jobs/ast/modules/Lightspeed.js
```diff
@@ -125,17 +125,22 @@ export default class LIGHTSPEED extends Module {
...
 			return {
 				key: lightspeed.start,
 				title: {
 					content: <>
 						{this.parser.formatTimestamp(lightspeed.start)}
 						&nbsp;-&nbsp;<Trans id="ast.lightspeed.rotation.gcd"><Plural value={numGcds} one="# GCD" other="# GCDs"/></Trans>
-						&nbsp;-&nbsp;{mpSavings} <Trans id="ast.lightspeed.rotation.mp-saved">MP saved</Trans>
+						{mpSavingsDisplay}
 					</>,
 				},
```
ayyaruq, line 145: Disregarding how much I hate that snippet, this might be cleaner to just set and run the patch conditional inside the content, like `{this.parser.patch.before('5.3') && mpSavingsDisplay}` I think will do it. Prob need to check this in browser but it'll let you save the const and avoid a useless reassignment.

src/parser/jobs/ast/modules/Horoscope.tsx
```diff
@@ -50,23 +61,48 @@ export default class Horoscope extends Module {
...
+			}))
 
+		} else {
+			/*
+				SUGGESTION: AOE heal without horoscope
+			*/
+			this.suggestions.add(new TieredSuggestion({
+				icon: this.data.actions.HOROSCOPE.icon,
+				content: <Trans id="ast.horoscope.suggestion.usage.content">
+					Plan to have <ActionLink {...this.data.actions.HOROSCOPE} /> up before you need to cast <ActionLink {...this.data.actions.HELIOS} /> or <ActionLink {...this.data.actions.ASPECTED_HELIOS} /> .
```
ayyaruq, line 97: Maybe "Try to plan", bit less forceful.

src/parser/jobs/ast/index.tsx
```diff
@@ -30,12 +30,19 @@ export default new Meta({
...
+		to: '5.3',
 	},
 	contributors: [
 		{user: CONTRIBUTORS.SUSHIROU, role: ROLES.MAINTAINER},
 	],
 	changelog: [
+		{
+			date: new Date('2020-08-13'),
+			Changes: () => <>
+				Updated AST for 5.3, adjustments to calculations for counting theoretical maximum card draws/plays
```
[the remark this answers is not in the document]
suShirow: y u du dis 2 me ;_;

src/parser/jobs/ast/modules/Draw.tsx
```diff
@@ -59,7 +60,13 @@ export default class Draw extends Module {
 	private PLAY: number[] = []
 	private ARCANA_STATUSES: number[] = []
 
+	private SLEEVE_DRAW_PLAYS_GIVEN: number = SLEEVE_DRAW_PLAYS_GIVEN_530
```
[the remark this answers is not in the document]
suShirow: I actually can't... init() isn't the constructor. So tslint prods me about it

src/parser/jobs/ast/modules/Lightspeed.js
suShirow: I haet this snippet too :(
your point is well received thx

src/parser/jobs/ast/modules/Draw.tsx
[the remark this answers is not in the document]
ayyaruq: You only use this in the `_onComplete()` method (idk why underscore but yolo), so it doesn't need to be defined in the init to begin with I guess. You don't need to type or assign it tbh, like in MNK I just do 
```
    const FORM_TIMEOUT_MILLIS = this.parser.patch.before('5.05')
      ? FORM_TIMEOUT_MILLIS_200
      : FORM_TIMEOUT_MILLIS_505

File: src/parser/jobs/ast/modules/Draw.tsx [hunk 15]
```diff
@@ -175,7 +188,7 @@ export default class Draw extends Module {
 		const pullState = this.arcanaTracking.getPullState()
 		this.prepullPrepped = !!pullState.drawState
 
-		const totalCardsObtained = (this.prepullPrepped ? 1 : 0) + this.draws + (this.sleeveUses * SLEEVE_DRAW_PLAYS_GIVEN)
+		const totalCardsObtained = (this.prepullPrepped ? 1 : 0) + this.draws + (this.sleeveUses * this.SLEEVE_DRAW_PLAYS_GIVEN)
```
ayyaruq, line 191: If you use my const idea in this specific function, you don't need this change.

src/parser/jobs/ast/modules/Draw.tsx
```diff
@@ -193,7 +206,7 @@ export default class Draw extends Module {
 			<ul>
 				<li><Trans id="ast.draw.checklist.description.prepull">Prepared before pull:</Trans>&nbsp;{this.prepullPrepped ? 1 : 0}/1</li>
 				<li><Trans id="ast.draw.checklist.description.draws">Obtained from <ActionLink {...this.data.actions.DRAW} />:</Trans>&nbsp;{this.draws}/{playsFromDraw}</li>
-				<li><Trans id="ast.draw.checklist.description.sleeve-draws">Obtained from <ActionLink {...this.data.actions.SLEEVE_DRAW} />:</Trans>&nbsp;{this.sleeveUses * SLEEVE_DRAW_PLAYS_GIVEN}/{playsFromSleeveDraw}</li>
+				<li><Trans id="ast.draw.checklist.description.sleeve-draws">Obtained from <ActionLink {...this.data.actions.SLEEVE_DRAW} />:</Trans>&nbsp;{this.sleeveUses * this.SLEEVE_DRAW_PLAYS_GIVEN}/{playsFromSleeveDraw}</li>
```
ayyaruq, line 209: Same here.

src/parser/jobs/ast/modules/Lightspeed.js
```diff
@@ -125,17 +125,17 @@ export default class LIGHTSPEED extends Module {
 				.map(cast => getDataBy(ACTIONS, 'id', cast.ability.guid))
 				.filter(action => action && action.onGcd)
 			const numGcds = gcdActions.length
-			const mpSavings = gcdActions
-				.reduce((totalSavings, action) => action.mpCost / 2 + totalSavings, 0)
-			// TODO: Use mpCostFactor instead of mpCost to be level agnostic
+
+			// pre-5.3 mp savings calcuation
+			const  mpSavingsDisplay = <>&nbsp;-&nbsp; {gcdActions.reduce((totalSavings, action) => action.mpCost / 2 + totalSavings, 0)} <Trans id="ast.lightspeed.rotation.mp-saved">MP saved</Trans></>
```
ayyaruq, line 130: Double spaace.

src/parser/jobs/ast/modules/SleeveDraw.tsx
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	private onCast(event: CastEvent) {
+		this.uses++
+		if (this.lastUse === 0) { this.lastUse = this.parser.fight.start_time }
+
+		const held = event.timestamp - this.lastUse - (ACTIONS.CELESTIAL_OPPOSITION.cooldown * 1000)
+		if (held > 0) {
+			this.totalHeld += held
+		}
```
ayyaruq, line 45: This is redundant, unless held can go negative. `this.totalHeld += held` where held is 0 will just add zero. Neater to clamp the value (if needed) and just yolo add it imo, but more a question of preference I guess.

src/parser/jobs/ast/modules/SleeveDraw.tsx
```diff
@@ -0,0 +1,88 @@
...
+		// update the last use
+		this.lastUse = event.timestamp
+	}
+
+	private onSect(event: BuffEvent) {
+		this.activeSect = getDataBy(STATUSES, 'id', event.ability.guid)
```
ayyaruq, line 51: `getDataBy`, in 2020?

ayyaruq (changes requested):
Needs a lingui extract. Bit of nitpicking on the sleeve draw plays given still, since you don't actually need to reassign it, seems weird to declare and do that weird assignment, especially when it's only used in one function. I'm also a bit ??? why there's both Draw and SleeveDraw now, was SleeveDraw part of 5.3 support or added on?

src/parser/jobs/ast/modules/Horoscope.tsx
```diff
@@ -50,23 +61,48 @@ export default class Horoscope extends Module {
...
+				<StatusLink {...this.data.statuses.HOROSCOPE_HELIOS} /> may help to cover more damage later without needing to cast more AOE heals.
+			</Trans>,
+				tiers: SEVERITIES.WASTED_AOE_HEAL_TIERS,
+				value: this.nonHoroscopeHeals,
+				why: <Trans id="ast.horoscope.suggestion.usage.why">
+					<Plural value={this.nonHoroscopeHeals} one="# single-target GCD heal was cast" other="# single-target GCD heals were cast" /> without synastry even though it was available.
```
CasualSuperman, line 103: This says synastry, should it be horoscope?

src/parser/jobs/ast/modules/Horoscope.tsx
CasualSuperman, line 97: While I'm here, just something that bothers me personally (and probably nobody else) - there's a space between the `ASPECTED_HELIOS` link and the period.

CasualSuperman (commented):
@suShirow, I looked over this as requested. Hopefully I explained myself enough that you can tell me why I'm wrong.

src/parser/jobs/ast/modules/Draw.tsx
```diff
@@ -162,10 +174,10 @@ export default class Draw extends Module {
 		// Prepull consideration: + 1 play
 
 		// Begin Theoretical Max Plays calc
-		const fightDuration = this.parser.fight.end_time - this.parser.fight.start_time
-		const maxSleeveUses = Math.floor((fightDuration - (CARD_DURATION*2)) / (this.data.actions.SLEEVE_DRAW.cooldown * 1000)) + 1
+		const fightDuration = this.parser.pull.duration
+		const maxSleeveUses = Math.floor(Math.max(0, (fightDuration - (CARD_DURATION*2))) / (this.data.actions.SLEEVE_DRAW.cooldown * 1000)) + 1
```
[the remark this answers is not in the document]
suShirow: Well that was a very well written comment. I've struggled on this for awhile and it's been stuck at "good enough" until now. And with the new sleeve changes it's probably the best time to rewrite the whole algorithm to something that makes more sense.

1. Extra time after pressing SD is not much needed anymore since it only gives you one card, which can be swiftly delivered without having to weave a few times. I agree with you here, we don't need to account for that anymore.
2. Your point about this being a count of the max number of times the button is pressed, rather than a count of whether it's considered meaningful holds weight. It's important to make the distinction as to what this helps count. I believe in pulling data out first and making it visible, rather than skipping steps so people can see clearly the facts. Will make this change.
3. I'll look over your formula when I'm not bleary eyed and look into adapting it. 

Thank you so much for taking the time to look through all that spaghetti

src/parser/jobs/ast/modules/Draw.tsx
```diff
@@ -162,10 +174,10 @@ export default class Draw extends Module {
...
 		const playsFromSleeveDraw = maxSleeveUses * SLEEVE_DRAW_PLAYS_GIVEN
-		const playsFromDraw = Math.floor((fightDuration - CARD_DURATION) / (this.data.actions.DRAW.cooldown * 1000)) + 1
+		const playsFromDraw = Math.floor(Math.max(0, (fightDuration - CARD_DURATION)) / (this.data.actions.DRAW.cooldown * 1000)) + 1
 		const theoreticalMaxPlays = playsFromDraw + playsFromSleeveDraw + 1
```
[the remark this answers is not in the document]
suShirow: That's something new. In 5.2 we didn't need to prepull draw, and the general consensus was that it's not important to hold your group back 30 seconds just to have it up. So we tried not to penalize on it. But in 5.3, it's a HUGE dps loss not to have div up when it's supposed to. So yea let's do that. I'll add that over the weekend

src/parser/jobs/ast/modules/Horoscope.tsx
suShirow: That makes you and everyone else here. No bulli ;_;

src/parser/jobs/ast/modules/Draw.tsx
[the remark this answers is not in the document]
CasualSuperman: Thank you! I was worried there was going to be too much rambling, since I was writing it at about 4 in the morning. :sweat_smile: 

1. The function I wrote scales for multiple cards pre-5.3, and in 5.3 looks for one GCD for Sleeve Draw and another for Play, so that's probably close enough? I recognize these aren't actually GCD skills and could maybe be played faster, but between trying to ensure optimal card plays and weaving Draw > Play, I think one GCD per card works as a baseline? Feel free to disagree, I would like to hear your thoughts - I could very easily be over/underestimating how long it takes to identify and play a card, I'm usually too busy trying to actually play the card to to track how long it usually takes me.
    
    If we don't want to require a GCD for Play at all starting in 5.3, we could adapt the function to return a list of possible activations and the number of cards there was time to play during that activation. If we did that:
    - Before 5.3 we would just have to sum the card counts from the results
    - Starting in 5.3, we could pass in `0` for `cardsToPlay`, making the function only check for enough time to activate the ability. Then we could use the length of the list of activations as the number of possible plays, since each activation would equal 1 Play.
2. Thanks, I'm glad my confusion helped. :smile: Can you clarify for me if you're saying the number you want to start reporting is the "physical button press" maximum, or the "enough time to be effective" maximum? It sounds like the physical maximum, but just double-checking.
3. I tested the function locally as I was writing it and it seemed to work for 5.3 parse, so hopefully there won't be much adaptation required.

src/parser/jobs/ast/modules/Draw.tsx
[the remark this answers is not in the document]
CasualSuperman: What do you think about removing prepull draws from the card play checklist if the fight is from pre-5.3? Since you said it wasn't considered important in 5.2, it was probably relatively uncommon, so including it would artificially lower the checklist score/percentage in historical fights compared to 5.3.
(I swear I'm not trying to make a bunch of extra work for you - I'd be willing to make that update after you finish this one if you think it's a good idea)

ayyaruq (approved):
Fixed the lingui crap myself, I think there's enough going on here to mark support. Changes to pre-cast logic and code quality can be done in another PR so we can expedite this and push back a bit on all these giant PRs.

PR #1000, opened 2020-08-11
Update WAR support for 5.3

ayyaruq:
Pending confirming changes to fflogs events on Storm's Eye, I just wanted PR 1000.

**Update**:
Basic support, this is mostly to avoid penalising people for old Eye behaviour and keep things working properly for KR/CN. A separate PR will be made to handle rules around overcapping and dropping Eye once theory peeps are in agreement.


ackwell:
Closing pr just to meme #1000

ayyaruq:
ngl I fucked up commit history so... good luck!

suShirow:
Seeing PR 1000 closed made my heart sink (;´Д｀)

src/parser/jobs/war/modules/StormsEye.tsx
```diff
@@ -0,0 +1,86 @@
...
+		this.lastRefresh = event.timestamp
+		this.totalEyes++
+
+		if (this.totalEyes < 2) {
+			return
+		}
+
+		if (this.parser.patch.before('5.3')) {
+			const remaining = event.timestamp - this.lastRefresh
+			if (remaining < STATUSES.STORMS_EYE.duration - STORMS_EYE_BUFFER) {
```
suShirow, line 46: Your new layered data updates for 5.3 aren't going to work without using `this.data.actions.STORMS_EYE.duration` I think!

suShirow: Okay technically they will work since these will be grabbing the StatusRoot values, which is the original 30s, but STILL

ayyaruq: I'm pretty sure it uses the final merged list, since I see this working fine in MNK which does a similar thing. Prob should update it to have nicer typing tho, just gotta fill array objects like what you do in Horoscope or whatever.

suShirow: If it works then I'll just give it a pass and we can go back and conform everything later

ayyaruq:
I think not using the data module for now is fine, it flattens anyway so it works properly, have verified. Definitely a task for another PR tho. Should be good to go.

suShirow (approved):

PR #1004, opened 2020-08-16
BRD - BuffWindow for Raging Strikes

yumiyafangirl:
This took some work. The logic for IJs under RS is a bit more precise than BuffWindow normally allows for, so this PR uses the functionality added recently to define a custom comparator for IJ. It also makes some small adjustments to how tracked actions get counted:

- If a custom comparator is defined, and it doesn't return negative for a given window, don't count it
- If this was a "rushed" window, also don't count it (we might not have had time to use all the actions we would in a full buff window)

Onto the bard stuff. 9 GCDs are expected in RS unless we didn't have muse (opener, reopener, song cycle adjustments). 1 barrage is always expected, and in **single target only** one IJ is expected. If we detected more than one enemy, we simply don't highlight the IJ cell of the table as the logic for calculating correctness here would be much too complicated. 

image
image
image

- At 4:09, we had 3 targets so the IJ cell is not flagged as right or wrong
- Only 2 tracked actions were suggested as missing. This ignores the 2 missed in the final RS window (rushed), and ignores the IJ at 4:09 even if it wasn't 3/3


src/parser/jobs/brd/modules/RagingStrikes.tsx
```diff
@@ -0,0 +1,117 @@
...
+			.length === 1
+
+		return isMuseUp ? 1 : 0
+	}
+
+	private getEventTargetKey = (event: CastEvent): string => `${event.targetID}-${event.targetInstance}`
+
+	protected getBaselineExpectedTrackedAction(buffWindow: BuffWindowState, action: BuffWindowTrackedAction): number {
+		// If the action was Iron Jaws, the upper limit = the number of enemies we cast something on during this RS window
+		if (action.action === ACTIONS.IRON_JAWS) {
```
suShirow, line 91: What do you think about flattening this function by escaping early.
```tsx
if (action.action !== ACTIONS.IRON_JAWS) {
return action.expectedPerWindow || 0
}
// iron jaws logic starts
```
That way like 90% of the function can be an indent less. Since this is the bulk of the function logic anyway.

suShirow (dismissed):
Mostly about the data references. I'll look over the BuffWindow calcs again tomorrow after some sleep and coffee.

src/parser/core/modules/BuffWindow.tsx
```diff
@@ -302,6 +302,25 @@ export abstract class BuffWindowModule extends Module {
 		return undefined
 	}
 
+	private countMissedTrackedActions(buffWindow: BuffWindowState, action: BuffWindowTrackedAction): number {
+		const expected = this.getBuffWindowExpectedTrackedActions(buffWindow, action)
+		const actual = buffWindow.getActionCountByIds([action.action.id])
+		const comparator = this.changeComparisonClassLogic(buffWindow, action)
```
AzariahAeroborn, line 308: This is a little hard to follow what's going on here.  I'm not sure I have a better alternative at hand that could still share / reuse the rotation table's highlighting logic, but it took me 15 solid minutes to grok this.  Might be worth some further investigation / discussion of whether the RotationTableOutcome interface can be simplified somehow in the future?

src/parser/core/modules/BuffWindow.tsx
```diff
@@ -302,6 +302,25 @@ export abstract class BuffWindowModule extends Module {
...
+
+		// If a custom comparator is defined for this action, and it didn't return negative, don't count this window
+		if ( comparator && !comparator(actual, expected).negative ) { return 0 }
+
+		// If this buff window was rushed, again don't count it
+		if ( this.buffStatus.duration ) {
```
AzariahAeroborn, line 314: Should this be accounted for in getBuffWindowExpectedTrackedActions instead, similar to what we do for expected GCDs?  Probably either through a reduceExpectedTrackedActionEndOfFight overrideable function, since it may vary from class to class when/whether they want to "forgive" expected actions during a rushing end of fight window.

AzariahAeroborn (changes requested):

src/parser/jobs/brd/modules/RagingStrikes.tsx
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	private onApplyMuse = (event: BuffEvent) => this.museTimestamps.push(event.timestamp)
+
+	protected considerAction = (action: Action) => !this.SUPPORT_ACTIONS.includes(action.id)
+
+	protected changeExpectedGCDsClassLogic(buffWindow: BuffWindowState): number {
+		// Expect one extra GCD if we had muse up for this RS
+		const museDuration = STATUSES.ARMYS_MUSE.duration * 1000
+		const isMuseUp = this.museTimestamps
```
AzariahAeroborn, line 89: This seems too aggressive.  I'm a little dubious of adding 1 GCD for Muse even if the Muse buff was applied within the first 10 seconds of Raging Strikes, since if it's only a 1-3 stack muse, they're gaining less than 0.1s per GCD and only going to gain for ~4 GCDs depending on where in the weave window they switched songs and whether Muse follows the Blood Weapon school of timing or the more conventional "10.75 seconds" timing.  

But this is also catching a case like the 06:57 window here: http://localhost:3000/fflogs/dFAy6pwP4L9vRbmM/42/2020 -- Muse happened with the song switch at 06:51 and only caught 2 GCDs at the beginning of the Raging Strikes window.

src/parser/core/modules/BuffWindow.tsx
```diff
@@ -288,7 +288,7 @@ export abstract class BuffWindowModule extends Module {
 	}
 
 	private getBuffWindowExpectedTrackedActions(buffWindow: BuffWindowState, action: BuffWindowTrackedAction): number {
-		return this.getBaselineExpectedTrackedAction(buffWindow, action) + this.changeExpectedTrackedActionClassLogic(buffWindow, action)
+		return this.getBaselineExpectedTrackedAction(buffWindow, action) + this.changeExpectedTrackedActionClassLogic(buffWindow, action) - this.reduceExpectedGCDsEndOfFight(buffWindow)
```
AzariahAeroborn, line 291: This should be a separate function rather than reusing the GCD reducer.  By default this would automatically reduce any expected single use per window action to 0 if the window was rushed at all, and this doesn't provide a way for implementing classes to determine how to handle rushing per-action.

AzariahAeroborn (changes requested):

src/parser/jobs/brd/modules/RagingStrikes.tsx
yumiyafangirl: At infinitestory's suggestion I changed the approach here to require 3 GCDs under muse in order to expect another GCD under RS. There's going to be a lot of 9/8s as a result, but I guess it's better to err on the side of 9/8 instead of 8/9.

AzariahAeroborn (approved):

PR #1009, opened 2020-08-23
MCH - Tincture improvements

yumiyafangirl:
Uses layered action data now. Also added a buffer for the tincture buff application on fight start / downtime to deal with the case where a tincture is used prepull but the buff doesn't apply until postpull. Ideally I'd like to be able to track reassembled GCDs themselves instead of reassemble casts, but that would involve tearing up a lot of BuffWindow code so this'll do for now.

### DIFF
--- a/src/parser/core/modules/BuffWindow.tsx
+++ b/src/parser/core/modules/BuffWindow.tsx
@@ -123,10 +123,10 @@ export abstract class BuffWindowModule extends Module {
 	 */
 	protected rotationTableNotesColumnHeader?: JSX.Element
 
-	@dependency private data!: Data
+	@dependency protected data!: Data
+	@dependency protected globalCooldown!: GlobalCooldown
 	@dependency private suggestions!: Suggestions
 	@dependency private timeline!: Timeline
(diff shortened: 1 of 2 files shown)


yumiyafangirl:
Don't merge yet, need to incorporate the precast status bugfix.

yumiyafangirl:
After playing around with PrecastStatus, it looks like the fixes in this PR are still necessary since the pot applybuff event happens *after* the synthed reassemble cast. Leaving it as is. @AzariahAeroborn

AzariahAeroborn:
> After playing around with PrecastStatus, it looks like the fixes in this PR are still necessary since the pot applybuff event happens _after_ the synthed reassemble cast. Leaving it as is. @AzariahAeroborn

Oh, yeah, that makes sense... PCS is synthesizing the reassemble cast at -0.002 and synthesizing the tincture status at -0.001... hmmm.  This seems like a sane adjustment for that limitation, not sure it'd really be worth the effort to alter BW to allow overriding the "actual" count to deal with situations like this.

AzariahAeroborn (approved):

PR #1019, opened 2020-09-15
BRD - if apex hits multiple targets, choose the best hit for potency calcs

yumiyafangirl:
Fixes a tricky edge case with apex arrow where the 1st confirmedEvent does 'very low damage' (usually due to immunity) but subsequent hits don't.


ayyaruq (approved):
Holy fuck tho. All this math for crit when the straight up crit modifier is in the event packet if Kihra could be bothered exposing it on the API.

ayyaruq:
I'll point out the best hit isn't necessary the best for potency calcs. What you'd prob rather do is map all hits, discard bad RNG outliers within 5% variance, and use the median. But stuff doing that in JS.

PR #1021, opened 2020-09-29
Dualcast severity update for missed casts.

Cephid:
Per Meru updated the severity rating for missed Dual Casts shifting values down one and dropping minor as missing is never considered minor.

### DIFF
--- a/src/parser/jobs/rdm/index.js
+++ b/src/parser/jobs/rdm/index.js
@@ -32,7 +32,7 @@ export default new Meta({
 	changelog: [
 		{
 			date: new Date('2020-09-29'),
-			Changes: () => <>Changed Severity ratings for Missed Dualcasts per Meru 1 missed is now medium, 2 missed is now major.</>,
+			Changes: () => <>Changed Severity ratings for Missed Dualcasts 1 missed is now medium, 2 missed is now major.</>,
 			contributors: [CONTRIBUTORS.LEYLIA],
 		},
 		{
--- a/src/parser/jobs/rdm/index.js
+++ b/src/parser/jobs/rdm/index.js
@@ -31,6 +31,11 @@ export default new Meta({
 
 	changelog: [
 		{
+			date: new Date('2020-09-29'),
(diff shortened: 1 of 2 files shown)


src/parser/jobs/rdm/modules/Dualcast.js
```diff
@@ -36,9 +36,8 @@ export default class DualCast extends Module {
 	//The last timestamp for a change in CastType
 	_castTypeLastChanged = null
 	_severityMissedDualcast = {
-		1: SEVERITY.MINOR,
-		2: SEVERITY.MEDIUM,
-		3: SEVERITY.MAJOR,
+		1: SEVERITY.MEDIUM,
+		2: SEVERITY.MAJOR,
```
ayyaruq, line 40: Any reason major is also changing here?

src/parser/jobs/rdm/index.js
```diff
@@ -30,6 +30,11 @@ export default new Meta({
 	],
 
 	changelog: [
+		{
+			date: new Date('2020-09-29'),
+			Changes: () => <>Changed Severity ratings for Missed Dualcasts per Meru 1 missed is now medium, 2 missed is now major.</>,
```
ayyaruq, line 35: I don't think users care who suggested what, the changelog just says *what* changed. IMO anyway.

ayyaruq (commented):
Seems sane but I'm not seeing why major was also lowered.

src/parser/jobs/rdm/index.js
Cephid: Fair enough I'll remove that piece, it was on my mind when I typed it which is why it appeared there heh.

src/parser/jobs/rdm/modules/Dualcast.js
Cephid: The reasoning was that short of an invuln trigger preventing you from casting there's no reason to waste a dualcast, whether you use it on the proper spell or anything else it should never just fall off.  Thus if you let two drop that's a major problem you should look at.

Cephid: It should be noted that the miss logic does handle for invuln.

ayyaruq (approved):
Neato

PR #1022, opened 2020-09-30
 AST 5.3 bump

### LINKED ISSUE #1014: bug: AST 'no sect' suggestion appears while sect-related skills are logged
Log info: KOR server, patch 5.20
Original Log: https://www.fflogs.com/reports/P7gcfGQZka8dF1h9#fight=last&type=summary
Example: https://xivanalysis.com/fflogs/P7gcfGQZka8dF1h9/10/3578

In this report, the suggestion module suggests 'use Diurnal Sect or Nocturnal Sect'. but below the suggestion module, the used sect has appeared and sect-related skills are logged.

(I'm not sure issue report is allowed and the format of this report is valid, if this is invalid, please close this issue)

CasualSuperman:
Updates #999 with fixes for the remaining issues. I can't update that PR since I'm not the original creator.
Fixes #1014 - the module was being tricked by the PrecastStatus module. I hope there's a better way to detect this than what I've got (see Sect.tsx)


src/parser/jobs/ast/modules/Draw.tsx
```diff
@@ -59,7 +80,9 @@ export default class Draw extends Module {
 	private PLAY: number[] = []
 	private ARCANA_STATUSES: number[] = []
 
+
 	protected init() {
+
 		PLAY.forEach(actionKey => {
```
kuwoyuki, line 86: nit and maybe not this pr scope but this can just be a `map()`
```js
this.PLAY = PLAY.map(actionKey => this.data.actions[actionKey].id)
```

CasualSuperman: Makes sense, I'll make that change.

CasualSuperman:
I found an example of a [pre-pull sleeve draw](https://www.fflogs.com/reports/P7gcfGQZka8dF1h9#fight=10&source=3578&type=auras&translate=true&ability=1001926), and it's throwing off the card counts. Technically it's already broken on prod because we get the same number (15/12), but unless I figure it out before this lands, I'll have a follow-up PR for that.

ayyaruq:
#999 has been merged, closing this one. Let's make a separate PR to handle pre-pull sleeve draws, and try to keep code style/structure changes a bit more isolated from features so they're easier to review and PRs don't end up huge.

PR #1023, opened 2020-10-01
AST - Fix starting sect not recognized

### LINKED ISSUE #1014: bug: AST 'no sect' suggestion appears while sect-related skills are logged

CasualSuperman:
Fixes #1014
The Sect module was treating the action fabricated by the PrecastStatus module as an actual stance activation. There might a better way to detect this than what I've got, but I didn't see any extra properties added by PrecastStatus that would allow distinguishing between fabricated and real events.


src/parser/jobs/ast/index.tsx
```diff
@@ -34,8 +34,16 @@ export default new Meta({
 	},
 	contributors: [
 		{user: CONTRIBUTORS.SUSHIROU, role: ROLES.MAINTAINER},
+		{user: CONTRIBUTORS.CASUALSUPERMAN, role: ROLES.DEVELOPER},
 	],
 	changelog: [
+		{
+			date: new Date('2020-09-30'),
+			Changes: () => <>
+				Fixed bug where starting stance was always reported forgotten.
```
ayyaruq, line 43: Can we reword this so that it doesn't sound like you had a stroke mid-sentence?

src/parser/jobs/ast/modules/Sect/Sect.tsx
```diff
@@ -142,8 +142,9 @@ export default class Sect extends Module {
 	}
 
 	private onCast(event: CastEvent) {
-		// If they casted a sect at anytime it means they pulled without one on
-		this.pullWithoutSect = true
+		// If they used a sect after the fight started, it means they pulled without one on; otherwise, it was likely added by the PrecastStatus module.
+		this.pullWithoutSect = event.timestamp >= this.parser.eventTimeOffset
```
ayyaruq, line 146: Any reason to use `eventTimeOffset` over `start_time`?

ayyaruq (changes requested):
Question on your timestamp aside, the changelog message could really be clearer as to what it's doing. I'm not sure how Precast could cause the issue you're talking about in the comment though unless they remove it after combat, because Precast relies ons seeing the status fall off.

src/parser/jobs/ast/modules/Sect/Sect.tsx
CasualSuperman: Nope

> Question on your timestamp aside, the changelog message could really be clearer as to what it's doing. I'm not sure how Precast could cause the issue you're talking about in the comment though unless they remove it after combat, because Precast relies ons seeing the status fall off.

I thought the aspected heals had a `statusesApplied` section, but looking at it they actually don't. It's being done by this module's own `normalize`.

src/parser/jobs/ast/index.tsx
CasualSuperman: ... sure

ayyaruq (approved):

PR #1031, opened 2020-10-05
AST - Fix Earthly Star counts

CasualSuperman:
We weren't ever incrementing uses, which was the main issue.
Also copy the logic from Draw for calculating lost time if it stopped being used at some point during the fight.
Fixes issue reported on discord.

### DIFF
--- a/src/parser/jobs/ast/index.tsx
+++ b/src/parser/jobs/ast/index.tsx
@@ -40,6 +40,13 @@ export default new Meta({
 		{
 			date: new Date('2020-10-05'),
 			Changes: () => <>
+				Fixed <ActionLink {...ACTIONS.EARTHLY_STAR} /> drift calculation.
+			</>,
+			contributors: [CONTRIBUTORS.CASUALSUPERMAN],
+		},
+		{
+			date: new Date('2020-10-05'),
+			Changes: () => <>
 				Fixed bug where sect was always reported forgotten, even when it wasn't.
 			</>,
 			contributors: [CONTRIBUTORS.CASUALSUPERMAN],
--- a/src/parser/jobs/ast/modules/EarthlyStar.tsx
+++ b/src/parser/jobs/ast/modules/EarthlyStar.tsx
@@ -29,7 +29,7 @@ export default class EarthlyStar extends Module {
 	static handle = 'earthlystar'
 	static title = t('ast.earthly-star.title')`Earthly Star`
 
-	static debug = true
+	static debug = false
 
 	@dependency private data!: Data
 	@dependency private suggestions!: Suggestions
--- a/src/parser/jobs/ast/modules/EarthlyStar.tsx
+++ b/src/parser/jobs/ast/modules/EarthlyStar.tsx
@@ -29,10 +29,13 @@ export default class EarthlyStar extends Module {
 	static handle = 'earthlystar'
 	static title = t('ast.earthly-star.title')`Earthly Star`
 
+	static debug = true
+
 	@dependency private data!: Data
 	@dependency private suggestions!: Suggestions
 	@dependency private precastStatus!: PrecastStatus
 
+	private prepull = false
 	private uses = 0
 	private lastUse = 0
 	private totalHeld = 0
@@ -50,9 +53,11 @@ export default class EarthlyStar extends Module {
 	}
 
 	private onPlace(event: CastEvent) {
+		this.uses++
+
 		// this was prepull
 		if (event.timestamp < this.parser.fight.start_time) {
-			this.debug('prepull')
+			this.prepull = true
 		}
 
 		// TODO: Instead determine how far back they used it prepull by checking explosion time.
@@ -60,10 +65,21 @@ export default class EarthlyStar extends Module {
 			this.lastUse = this.parser.fight.start_time
 		}
 
-		const held = event.timestamp - this.lastUse - (this.data.actions.EARTHLY_STAR.cooldown * 1000)
-		if (held > 0) {
-			this.totalHeld += held
+		let drift = 0
+
+		if (this.uses === 1 && !this.prepull) {
+			// The first use, take holding as from the start of the fight
+			drift = event.timestamp - this.parser.fight.start_time
+		} else {
+			// Take holding as from the time it comes off cooldown
+			drift = event.timestamp - this.lastUse - (this.data.actions.EARTHLY_STAR.cooldown * 1000)
 		}
+
+		// Keep track of total drift time not using star
+		if (drift > 0) {
+			this.totalHeld += drift
+		}
+
 		// update the last use
 		this.lastUse = event.timestamp
 	}
@@ -94,6 +110,11 @@ export default class EarthlyStar extends Module {
 				tiers: SEVERETIES.UNCOOKED,
 				value: earlyBurstCount,
 			}))
+		}
+
+		// If they stopped using Star at any point in the fight, this'll calculate the drift "accurately"
+		if (this.parser.fight.end_time - this.lastUse > (this.data.actions.EARTHLY_STAR.cooldown * 1000)) {
+			this.totalHeld += (this.parser.fight.end_time - (this.lastUse + (this.data.actions.EARTHLY_STAR.cooldown * 1000)))
 		}
 
 		/*
--- a/src/parser/jobs/ast/modules/EarthlyStar.tsx
+++ b/src/parser/jobs/ast/modules/EarthlyStar.tsx
@@ -29,8 +29,6 @@ export default class EarthlyStar extends Module {
 	static handle = 'earthlystar'
 	static title = t('ast.earthly-star.title')`Earthly Star`
 
-	static debug = false
-
 	@dependency private data!: Data
 	@dependency private suggestions!: Suggestions
 	@dependency private precastStatus!: PrecastStatus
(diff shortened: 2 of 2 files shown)


src/parser/jobs/ast/modules/Sect/Sect.tsx
```diff
@@ -142,8 +142,12 @@ export default class Sect extends Module {
 	}
 
 	private onCast(event: CastEvent) {
-		// If they casted a sect at anytime it means they pulled without one on
-		this.pullWithoutSect = true
+		if (event.timestamp < this.parser.eventTimeOffset) {
+			// This was before the pull and likely added by the PrecastStatus module.
```
nonowazu, line 146: blank if section? Is there supposed to be logic here?

nonowazu (changes requested):

src/parser/jobs/ast/index.tsx
```diff
@@ -34,8 +34,23 @@ export default new Meta({
...
+			</>,
+			contributors: [CONTRIBUTORS.CASUALSUPERMAN],
+		},
+		{
+			date: new Date('2020-09-30'),
+			Changes: () => <>
+				Fixed bug where starting stance was always reported forgotten.
+			</>,
+			contributors: [CONTRIBUTORS.CASUALSUPERMAN],
+		},
```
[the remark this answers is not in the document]
CasualSuperman: It's meant to be separate - I made this (and most of the other ones) as a draft and was going to publish it after #1023 got merged so I wasn't adding myself as a contributor in multiple PRs, getting conflicts later, etc.

src/parser/jobs/ast/modules/EarthlyStar.tsx
```diff
@@ -29,10 +29,13 @@ export default class EarthlyStar extends Module {
 	static handle = 'earthlystar'
 	static title = t('ast.earthly-star.title')`Earthly Star`
 
+	static debug = true
```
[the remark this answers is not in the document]
CasualSuperman: Whoops.

src/parser/jobs/ast/modules/Sect/Sect.tsx
CasualSuperman: It was there just to document why the check was necessary.

src/parser/jobs/ast/index.tsx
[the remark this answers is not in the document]
nonowazu: That's perfectly fine, but it's confusing from a 'hey, which PR made this change' perspective when there's two changes in the history list under a job associated only with a singular PR.

src/parser/jobs/ast/modules/Sect/Sect.tsx
nonowazu: I still find the blank logic confusing - code shouldn't be a substitute for comments, especially when it leads to noop-style logic to have 'self documenting code'. Instead of using a blank `if` block, document as part of a comment and invert the logic to be:

```js
if (event.timestamp >= this.parser.eventtimeOffset) {
```

src/parser/jobs/ast/index.tsx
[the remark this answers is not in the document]
AzariahAeroborn: Once you merge one, if @CasualSuperman back-merges master into the other branch, it will automatically remove the "duplicated" changeset from the other, in either order.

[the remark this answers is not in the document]
CasualSuperman: Agreed, I wasn't sure what the best route was (didn't even know drafts were publicly listed tbh)

src/parser/jobs/ast/modules/Sect/Sect.tsx
CasualSuperman: Fixed this in #1023, since that's the PR those changes came from.

src/parser/jobs/ast/index.tsx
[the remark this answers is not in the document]
CasualSuperman: Would it help if I abandoned the dependent PRs, and reopened them once the prior changes have been landed?

src/parser/jobs/ast/modules/Sect/Sect.tsx
nonowazu: I'll let this go, but in the future document when PRs have ordered dependencies, and please try not to base your PR off another PR branch - small atomic commits are much easier to reason about.

CasualSuperman: I did say in the PR description that it was branched off of #1023..

nonowazu: I understand this was branched off of #1023, but I didn't know about the ordered dependency until we discussed it., and the suggestion was to base your PRs off of master moving forward (unless there's a compelling reason not to).

CasualSuperman: Shouldn't be an issue since this was based off of that to avoid duplicating/conflicting adding myself as a contributor, and I didn't add it to that one so I could keep the individual PR small. This was a bit of a one-off situation for me.

nonowazu (dismissed):
I'm going to approve this because it appears that my requested changes were implemented, but Ay is going to give this another set of eyes on this because I usually only check for code sanity, not full logic flow.

CasualSuperman:
I don't know who Ay is short for, can you tag them to see when they can give this a look? Someone joined the discord to ask about this bug.

karashiiro:
@ayyaruq

src/parser/jobs/ast/modules/EarthlyStar.tsx
```diff
@@ -50,20 +53,33 @@ export default class EarthlyStar extends Module {
...
+			drift = event.timestamp - this.parser.fight.start_time
+		} else {
+			// Take holding as from the time it comes off cooldown
+			drift = event.timestamp - this.lastUse - (this.data.actions.EARTHLY_STAR.cooldown * 1000)
+		}
+
+		// Keep track of total drift time not using star
+		if (drift > 0) {
+			this.totalHeld += drift
 		}
```
ayyaruq, line 81: This is generally messy and could be done more simply with a clamp. In the interest of getting it through, ignoring, but at some point drift and also held times should probably be rethought. Could just initialise lastUse to be fight start time.

src/parser/jobs/ast/modules/EarthlyStar.tsx
```diff
@@ -96,6 +112,11 @@ export default class EarthlyStar extends Module {
 			}))
 		}
 
+		// If they stopped using Star at any point in the fight, this'll calculate the drift "accurately"
+		if (this.parser.fight.end_time - this.lastUse > (this.data.actions.EARTHLY_STAR.cooldown * 1000)) {
+			this.totalHeld += (this.parser.fight.end_time - (this.lastUse + (this.data.actions.EARTHLY_STAR.cooldown * 1000)))
+		}
```
ayyaruq, line 118: Is it not accurate already? The held bump could prob be a method of it's own so you end up with less funky math like this. Again, won't block on this, but the whole module could use a cleanup in this regard.

ayyaruq (approved):
I'm taking some issue with code structure and dead code but it works so whatever. Could prob rethink the way this module does stuff as a whole at some point.

src/parser/jobs/ast/modules/EarthlyStar.tsx
CasualSuperman: Easy enough to change

src/parser/jobs/ast/modules/EarthlyStar.tsx
CasualSuperman: I think some form of logic for this is required.
Say they use Earthly Star once at the beginning of the fight, and never again. `onPlace` won't be called again, so nothing will increase the `totalHeld` variable, so at the end of the fight we'll need to see how many more they could have cast.

src/parser/jobs/ast/modules/EarthlyStar.tsx
```diff
@@ -29,10 +29,13 @@ export default class EarthlyStar extends Module {
 	static handle = 'earthlystar'
 	static title = t('ast.earthly-star.title')`Earthly Star`
 
+	static debug = false
```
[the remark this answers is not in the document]
CasualSuperman: Nope, left it in from while I was figuring out what was going wrong. I'll remove it.

src/parser/jobs/ast/modules/EarthlyStar.tsx
ayyaruq: So?

ayyaruq (approved):

ayyaruq:
Merged so we can get this done without further superfluous edits.

src/parser/jobs/ast/modules/EarthlyStar.tsx
CasualSuperman: Maybe I'm missing your point. It sounded like you thought this code wasn't required based on "Is it not accurate already?"

ayyaruq: I get what it does, my point was about fixing the initial drift conditional and collapsing the 3 diff calculations into a single method that can be called in onPlace and onComplete to avoid duplicating complex-looking math and making it easier to maintain.

PR #1037, opened 2020-11-15
Fix Provoke cooldown duration

hofstee:
Provoke is a 30s CD, not 40s.


hofstee:
image

ayyaruq (approved):
Hi, thanks for the PR. Nice catch!

ayyaruq:
I'm approving but gonna hold off on a merge for a bit so we can resolve the linter issue with localisation data in CI. This isn't your fault, just looks like something has broken on the previous commit and we can't merge or deploy until the pipeline is resolved.

ayyaruq (approved):

PR #1040, opened 2020-11-18
Update AST.ts with correct Collective Unconscious cooldown

danielylee01:
Changed Collective Unconscious's cooldown from 90s to 60s (as it currently is in-game)

### DIFF
--- a/src/data/ACTIONS/layers/patch5.1.ts
+++ b/src/data/ACTIONS/layers/patch5.1.ts
@@ -24,6 +24,7 @@ export const patch510: Layer<ActionRoot> = {
 				'COLLECTIVE_UNCONSCIOUS_NOCTURNAL',
 				'WHEEL_OF_FORTUNE_NOCTURNAL',
 			],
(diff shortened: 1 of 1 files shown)


ayyaruq (changes requested):
As this differs per patch, please use layers, you can find the layer override for 5.1 at https://github.com/xivanalysis/xivanalysis/blob/master/src/data/ACTIONS/layers/patch5.1.ts

Since there's already a block for Collective Unconsciousness, you just need to add a line for the cooldown.

danielylee01:
Thanks for the response! Hopefully I did this correctly - let me know if not

ayyaruq (approved):
Thanks for the fix, I made a small edit for the CI fault on linting. Lines should start with tab indentation instead of spaces or the linter will fail it.

danielylee01:
Awesome! Thanks so much for your hard work on the tool (I use it a lot!) and I'm glad to have been able to help in some way :D

PR #1041, opened 2020-11-24
BRD - PrecastStatus Bug Fix

yumiyafangirl:
Currently there's a bug where almost every bard log will have a synthed Heavy Shot cast at the beginning of the timeline because the status Straight Shot Ready is applied by many actions, but only Heavy Shot has SSR as an applied status in the data files. Adding SSR to all the relevant actions doesn't solve the problem because of the way PrecastStatus looks for corresponding statuses, and anyway there is no current need for SSR being tracked as an applied status, so this bandaid fix seems appropriate for now.

cc @AzariahAeroborn


ayyaruq (approved):
Def a bandaid fix. I remember doing something for SSR and Barrage a while back but not why so this may be a leftover from that. Approved but waiting for Az before merging.

PR #1043, opened 2020-11-28
Add patch 5.3 actions layer to the index

ayyaruq:
For some reason, patch 5.3 actions layer was never added to the layers index so these changes weren't actually being used. Ay feex.

Looks like 5.2 only had potency changes, so while we have that data, nothing seems to actually use it for the affected jobs and so there's no layer for it.


nonowazu (approved):
LGTM. Frankly, I'm surprised this was missed for so long.

PR #1044, opened 2020-11-28
MNK patch 5.4 support

ayyaruq:
Currently WIP. These changes will not be backwards compatible due to the scope, so we'll be bumping MNK support to only be for patch 5.4+. Unfortunately, this will break compatibility for the Korean and Chinese servers until they catch up in 4-5 months.

**There is some pending work in the RotationTable component that I think is useful to have for this PR, but isn't included as it needs more discussion. Would be good to get that in around the same time, but it doesn't block merging.**

Also, I **_strongly_** advise reading the diff with whitespace changes disabled, it'll make some conditional wrapper changes way easier to read.

## Job Module Changes
- [x] Remove Gauge and update logic that depends on it
- [x] Remove RoE module
- [x] Update Speedmod
- [x] Update RoF rules for TK and FS
- [x] Update RoF math for 6SS
- [x] Rewrite PB entirely for stack counting and suggestions for dropping before completion
- [x] Update CDDT offsets (pending feedback on opener changes)

## Data Layer Changes
- [x] Add Formless Fist status
- [x] Update actions
- [x] Update statuses

Depends on #1042


src/data/STATUSES/layers/patch5.4.ts
```diff
@@ -0,0 +1,16 @@
...
+	patch: '5.4',
+	data: {
+		// MNK 5.4 status updates
+		PERFECT_BALANCE: {duration: 18},
+		FORMLESS_FIST: {
+			id: 0,
+			name: 'Formless Fist',
+			icon: 'https://xivapi.com/i/000000/000000.png',
+			duration: 15,
+		},
```
ayyaruq, line 14: This needs to be filled in once patch is available for download so we can get the ID and correct icon URL.

src/parser/jobs/mnk/modules/OGCDDowntime.ts
```diff
@@ -1,17 +1,38 @@
 import ACTIONS from 'data/ACTIONS'
 import {CooldownDowntime} from 'parser/core/modules/CooldownDowntime'
 
-// Approximate downtime window to account for aligning derped setups and GCD pinning:
-// 2 GCDs under slow GL3 (2.07s GCD), 700ms for second half of GCD activation
+// Approximate downtime window to account for aligning GCDs and unreliable FBC procs:
+// 2 GCDs at a slow 2.0s GCD, 600ms for animation lock, and another 250ms to deal with
+// high end of "sane" ping or a late FBC proc clipping the GCD.
 const DEFAULT_ALLOWED_DOWNTIME = 4850
```
AzariahAeroborn, line 7: Given your comments, you need to reduce this number to 2425 -- if the allowable downtime window is 2GCDs plus the other acoutrements, the average value that CDDT should allow is half that (to account for times that skills get used right on time and thus have 0 downtime).

ayyaruq: Oh good find, I completely forgot this is an average.

AzariahAeroborn (approved):

PR #1057, opened 2020-12-08
Pld 5.4 support

qaphla:
No actual change was needed. This just version bumps (and fixes the sorting of the changelog).


ayyaruq (approved):
lol nice reordering, thanks bb

PR #1058, opened 2020-12-08
Updated SMN to 5.4 and removed clipping suggestion

KelosPelmand:
No changes due to the patch.

The clipping suggestion has actually been outdated since the change to 55s trances, since some clipping will be required.  I am now getting around to removing it.


ayyaruq (approved):
onice

PR #1060, opened 2020-12-10
Bend time(line) to my will, recognize TAU for e9s

nonowazu:
Not too hard. I totally still have sanity left.

* Create directory / file scaffolding for e9 under `parser/bosses`
* Wander into this nicely marked ~~maze~~ path to quickly find the boss ID
* Update the tau for boss to firsttap until overkill
* Reload the page dozens of times wondering why it isn't working
* Realize I need to import all this somewhere.
* Descend into the third ~~layer of the labyrinth~~ file of AVAILABLE_MODULES and find the right section to add
* Find the right section in `data/bosses` and guess at the ID
* Get the ID right
* Perform PR

### DIFF
--- a/src/data/BOSSES.ts
+++ b/src/data/BOSSES.ts
@@ -57,6 +57,7 @@ const BOSSES = ensureBosses({
 	E6: {logId: 70},
 	E7: {logId: 71},
 	E8: {logId: 72},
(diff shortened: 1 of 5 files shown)


ackwell (approved):
thankeee

PR #1061, opened 2020-12-10
(Bad) first pass at tau update documentation

nonowazu:
I expect Ack to have input / commentary on this, but it's probably good to start pushing for more things to be documented so that nobody gets lost in the labyrinth again.

### DIFF
--- a/docs/tau-updates.md
+++ b/docs/tau-updates.md
@@ -84,9 +84,15 @@ const AVAILABLE_MODULES: AvailableModules = {
 
 In the `Invulnerability.ts` file for your fight, there are 3 options per actor ID:
 
-* start (takes a string)
-* end (takes a string)
-* exclude (takes a boolean)
+
+| Param | Option | Description |
+| ----- | ---------: | ----------- |
+| start | 'tau'      | Use targetability update from fflogs (the default) |
+|       | 'firstTap' | Use the first registered hit on the enemy as a synthetic targetability update |
+| end   | 'tau'   | use targetability update from fflogs (the default) |
+|       | 'overkill' | Use the last hit registered on a boss (that causes overkill) as a synthetic targetability update |
+| mirrorToInvuln | boolean | "mirrors" untargetable moments with invulnerable moments. Set to 'false' for enemies that can be untargetable but still take damage (dot tickets, other effects, etc). Defaults to true. |
+| exclude | boolean | exclude the target from the fight in terms of targeting (default is false) |
 
 For example, let's say there are 3 actors (4567, 4568, and 4569) with different properties:
 
--- a/docs/tau-updates.md
+++ b/docs/tau-updates.md
@@ -1,0 +1,122 @@
+# Targetability Updates (TAU)
+
+In cases where fflogs sends insufficient information to determine the correct information and state of a boss (invulnerability window, ephemeral enemies, etc), there is a mechanism to override this on a per-fight basis under `src/parser/bosses`.
+
+# Setup
+
+## Scaffolding
+
+At current, you'll need to provide the following scaffolding / directory structure:
+
+```
+src/parser/bosses/
+	${IDENTIFIER}/
+		index.ts
+		modules/
+			Invulnerability.ts
+			index.ts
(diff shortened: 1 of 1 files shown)


AzariahAeroborn:
Thanks much for the effort on this, very well written.

nonowazu:
Needs a slightly better overall set of docs around the actorconfig themselves. For example, I need to list out the options for start/end/exclude and what they mean, more like in the [Invulnerability](https://github.com/xivanalysis/xivanalysis/blob/master/src/parser/core/modules/Invulnerability.tsx#L21) module itself.

docs/tau-updates.md
```diff
@@ -0,0 +1,122 @@
...
+```
+src/parser/bosses/
+	${IDENTIFIER}/
+		index.ts
+		modules/
+			Invulnerability.ts
+			index.ts
+```
+
+Where `${IDENTIFIER}` is a meaningful identifier for a boss. It's important to note that in the case of savage fights, the boss id is the same between normal and savage, so there's no need to make normal and savage scaffolding for a fights.
```
ayyaruq, line 20: I don't think this is right, unless Kihra fixed shit, E9 and E9S seem to have different IDs.

ayyaruq: Seems fflogs IDs rather than game, disregard.

ayyaruq (dismissed):
LGTM, pending ack taking a look.

docs/tau-updates.md
```diff
@@ -0,0 +1,128 @@
+# Targetability Updates (TAU)
+
+In cases where fflogs sends insufficient information to determine the correct information and state of a boss (invulnerability window, ephemeral enemies, etc), there is a mechanism to override this on a per-fight basis under `src/parser/bosses`.
```
ackwell, line 3: Hmmmm... semi-accurate. More precise; the mechanism is part of the `invuln` module, `src/parser/bosses` is just the part of the meta system that loads the invuln override. Same as overriding core modules for job logic.

docs/tau-updates.md
```diff
@@ -0,0 +1,128 @@
...
+
+# Setup
+
+## Scaffolding
+
+At current, you'll need to provide the following scaffolding / directory structure:
+
+```
```
ackwell, line 11: just thinking;

while this makes sense to keep in-band for now; this part is essentially docs on scaffolding a meta group, rather than invuln specifically. should we ever have two boss-specific override modules, would likely move scaffolding (if still required) into a separate doc.

docs/tau-updates.md
```diff
@@ -0,0 +1,128 @@
...
+export class Invulnerability extends CoreInvulnerability {
+	protected actorConfig: ActorsConfig = {
+		// TAU data here
+	}
+}
+```
+
+## Other file(s) to update
+
+1. In `src/data/BOSSES.ts`, add a short identifier and the logID to the file. For example:
```
ackwell, line 59: we should probably include docs on how to get the logid from fflogs

docs/tau-updates.md
```diff
@@ -0,0 +1,128 @@
...
+		4568: {start: 'firstTap'},
+		// Target 3 - can't target them
+		4569: {exclude: true},
+	}
+}
+```
+
+# Testing
+
+(Temporarily) modify `parser/core/modules/invulnerability.tsx` and set `static debug` to `true`. With your updates in place (or while tweaking your updates), the timeline should have an additional section indicating target(s) in various states. If the target doesn't show up and it needs to be there, add the actor ID and configure as appropriate. If the actor is already there but it's lingering or not showing up at the right time, also tweak these values.
```
ackwell, line 128: might be worth mentioning the debug mode earlier in the doc - people will usually want to check that first, then set up overrides

---

as with log id, probably should outline how to get actor ids - the debug mode does output ones it's aware of, but as with the discussion in dev-general the other day, isn't always all-inclusive.

ackwell (commented):
thanks for writing this down!

main points to include, i think, would be some details on what the ids actually mean, and how to get them

PR #1063, opened 2020-12-17
DRG: Move Dragon Sight and Lance Charge to core BuffWindow Module

ebshimizu:
Removed (most) class specific logic for Dragon Sight and Lance Charge buff windows.

- Core: Modified to include a `secondaryBuffStatus` parameter because DRG has a separate id for a solo Dragon Sight. Existing modules are unaffected and should function as before with no modifications required.
- Deleted `drgBuffWindow` module and set LC/DS modules to inherit from core `BuffWindow`
- Suggestions re-written for both windows to highlight the specific things that we're looking for in each window. Previously, the suggestion just told people to use DS/LC any time except after a Chaos Thrust or Full Thrust and did not explain why. Now, the suggestion and the table highlights the GCDs that we're expecting to be within the window instead of just saying "bad start."
- Dragon Sight now indicates if a window is short because you died or your tether partner died,

Review Notes
- Please advise on changes to core and if there's a better way to accommodate DRG's split buff ids
- Set expected uses for FC/WT to 1. It's definitely possible to get 2 but I'm not sure it's a hard requirement? 1 FT/CT definitely is.

Table Preview
image

Suggestion Preview
image


ebshimizu:
PR updated to include notes on the cause of truncated Dragon Sight windows. The buff disappears if your partner dies, of if you die. In the event of partner death, the expected GCDs and tracked actions are adjusted to not penalize the DRG (you cannot control your partner). If the DRG dies, no adjustments to expected actions are performed (you can control yourself).

implementation note for this: the notes column is set to appear only if a truncated window shows up, though this is done somewhat awkwardly through relying on a helper being called before the table is output. This could probably be addressed better by a modification to the core module that would only show the notes header if at least one window returns something that's not undefined.

src/parser/core/modules/BuffWindow.tsx
```diff
@@ -172,7 +177,8 @@ export abstract class BuffWindowModule extends Module {
 	}
 
 	private onApplyBuff(event: BuffEvent) {
-		if (!this.buffStatus || event.ability.guid !== this.buffStatus.id) {
+		if ((!this.buffStatus || event.ability.guid !== this.buffStatus.id) &&
```
AzariahAeroborn: Per discussion [on Discord](https://discord.com/channels/441414116914233364/441424599310270464/793185644699582494) - we should refactor this to use the utilities class method `ensureArray` and change buffStatus to accept `Status | Status[]`

I POC'd this with:
```
	private onApplyBuff(event: BuffEvent) {
		if (!this.buffIsTracked(event.ability.guid)) {
			return
		}

		this.startNewBuffWindow(event.timestamp)
	}

	private onRemoveBuff(event: BuffEvent) {
		if (!this.buffIsTracked(event.ability.guid)) {
			return
		}

		if (this.activeBuffWindow) {
			this.activeBuffWindow.end = event.timestamp
		}
	}

	protected buffIsTracked(buffId: number) {
		let buffStatusArray = ensureArray(this.buffStatus)
		return buffStatusArray.some(bs => bs.id == buffId)
	}

	protected trackedBuffDuration() {
		let buffStatusArray = ensureArray(this.buffStatus)
		return buffStatusArray[0].duration
	}
```

and it seemed to WOMM.

AzariahAeroborn (changes requested):

AzariahAeroborn (approved):

PR #1064, opened 2020-12-17
DRG: Fix hard crash on drift error

ebshimizu:
Fixes hard crash on instances where `downtime.isDowntime()` returns true, but `downtime.getDowntimeWindows()` returns nothing. Doesn't fix the root cause, but now checks for the existence of a window and falls back to assuming there is no downtime if the window isn't defined.

### DIFF
--- a/src/parser/jobs/drg/modules/Drift.tsx
+++ b/src/parser/jobs/drg/modules/Drift.tsx
@@ -77,7 +77,11 @@ export default class Drift extends Module {
 
 		if (this.downtime.isDowntime(plannedUseTime)) {
 			const downtimeWindow = this.downtime.getDowntimeWindows(plannedUseTime)[0]
-			expectedUseTime = downtimeWindow.end
+
+			// there's a case where getDowntimeWindows will return null even though the planned use time is marked as downtime
+			// unsure the root cause but it looks like this occurs when invulns aren't properly configed for a fight yet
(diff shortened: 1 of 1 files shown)


src/parser/jobs/drg/modules/Drift.tsx
```diff
@@ -77,7 +77,10 @@ export default class Drift extends Module {
 
 		if (this.downtime.isDowntime(plannedUseTime)) {
 			const downtimeWindow = this.downtime.getDowntimeWindows(plannedUseTime)[0]
-			expectedUseTime = downtimeWindow.end
+
+			// there's a case where getDowntimeWindows will return null even though the planned use time is marked as downtime
+			// adding a fallback to ensure the parse is still viewable.
+			expectedUseTime = (downtimeWindow) ? downtimeWindow.end : plannedUseTime
```
AzariahAeroborn, line 83: This can be simpler, and a little more accurate at only coalescing on null, rather than falsey (not that it should be falsey if it isn't null but 🤷) 
```suggestion
			expectedUseTime = downtimeWindow?.end ?? plannedUseTime
```

ebshimizu: works for me! wasn't aware of this syntax but I am now and I like it.

AzariahAeroborn (approved):

PR #1065, opened 2020-12-17
Break data for scholar because events from fflogs is coming in reverse

nonowazu:
Pretty much every event that uses dissipation within the first 240s of the fight conjures a precast synthesis to 'helpfully' pair up the event. This is because fflogs is sending the status before the associated cast event:

Screen Shot 2020-12-17 at 4 21 52 PM

I can't think of anything in SCH analysis that gets broken by this change, only what it *fixes* in this case. Yes, this is gross.


nonowazu:
Closing in favor of #1066

PR #1067, opened 2020-12-18
DRG: actually fix drift error

ebshimizu:
Turns out we should just use `getDowntimeWindows` correctly and then we won't have errors. Leaving in the safety check just in case. Thanks to Kelos for pointing this out.


nonowazu (approved):
always frustrating when it's the single-line fix, isn't it?

PR #1068, opened 2020-12-23
Wire report flow events into the parser

ackwell:
This PR wires the new report flow-based event system all the way through to the parser, allowing migration of existing modules to onto the new events to begin.

Most of the code changes here should be relatively "behind the scenes" - if you're writing new code or migrating old, the main difference you'll notice is the new `Analyser` class (replaces `Module`), and associated new APIs when working with them.

Worth noting that, due to how the new and legacy events are interleaved, while `Module`s may depend on both `Module`s and `Analyser`s, the reverse is not true - new `Analyser`s must only ever depend on `Analysers` - you'll get an error if you try otherwise. This is to ensure a one-way migration, and prevent mishaps in data consumption.

Once this hits `master`, a few notes:
- I've renamed the environment variable used as config for the fflogs remote URL. This will break your `.env.local` files. This is intentional - feel free to delete those files. You do _not_ want to sit through the new system pulling down events on your own key, trust me.
- There is zero normalisation capability in the new system. This is intentional. I don't want to introduce normalisation just yet - there's a few different approaches, and I want to collect some usage before I make a call.
- Anything else I think up?


ackwell:
I think this is ready. Will merge in a day or few - please feel free to look through this lot and raise concerns if you have any.

PR #1072, opened 2020-12-29
Feex bad switching on FPF

ayyaruq:


src/parser/jobs/mnk/modules/TwinSnakes.tsx
```diff
@@ -82,45 +88,33 @@ export default class TwinSnakes extends Module {
...
 
-		switch (action.id) {
-		// Ignore TS itself, plus Form Shift, Anatman, and Meditation
-		// Should we double count 6SS?
-		// We use gcdsSinceTS because we don't want to double count FPF
-		case (this.data.actions.TWIN_SNAKES.id):
-			if (this.twinSnake && !this.twinSnake.end && this.gcdsSinceTS < TWIN_SNAKES_CYCLE_LENGTH) {
-				this.earlySnakes++
-			}
+		// Ignore FS and Meditation
```
AzariahAeroborn, line 91: Comment says form shift and meditation, but the skill list is Anatman and Meditation.  I'm assuming the skill list is wrong, since you're doing a check if Anatman extended twin snakes further down.

src/parser/jobs/mnk/modules/TwinSnakes.tsx
```diff
@@ -82,45 +88,33 @@ export default class TwinSnakes extends Module {
...
-		// Ignore Antman, but check if it's a bad one
-		case (this.data.actions.ANATMAN.id):
-			if (!this.combatants.selected.hasStatus(this.data.statuses.TWIN_SNAKES.id)) {
-				this.failedAnts++
+			if (action.id === this.data.actions.FOUR_POINT_FURY.id) {
+				this.failedFury++
 			}
 
-			break
+			return
```
AzariahAeroborn, line 103: Do you intend to _not_ count either Four Point Fury or Anatman as GCDs in between twin snakes usages?  My understanding from our conversation on Discord was that FPF was supposed to be counted as a GCD.

AzariahAeroborn (changes requested):

src/parser/jobs/mnk/modules/TwinSnakes.tsx
ayyaruq: Yes, good catch.

AzariahAeroborn (approved):

PR #1073, opened 2020-12-31
Update About's UI a bit

ackwell:
image
image


ayyaruq:
Got a screenshot of before/after?

ackwell:
In OP.

PR #1074, opened 2021-01-02
AST - Add spaces to overheal checklist

yumiyafangirl:
Immersion ruined please fix


ayyaruq (commented):
Looks ok but is there any reason not to use core overheal module?

ackwell (approved):
itt acc tries to push scope creep onto other people

PR #1078, opened 2021-01-06
[SAM] Sam Display Order

Rjb99:
What is this, 2018? SAM has Display Order now.

Current Logic for display order in my head is:

1) Checklists
2) Suggestions
3) Stats that can be changed/ could have obvious errors (Dots, Buff Windows, etc.)
4) Stats like gauge amount and stacks of Shoha, This in my head is the normal "End of page" for SAM focused critique/Information
5) Things such as interrupted Casts that were stated as a suggestion before that are now expanded in detail.

I'm not sure if I should flip the last 2 slots though

### DIFF
--- a/src/parser/jobs/sam/modules/DISPLAY_ORDER.js
+++ b/src/parser/jobs/sam/modules/DISPLAY_ORDER.js
@@ -1,0 +1,9 @@
+export default {
+	HIGANBANA: 1,
+	MEIKYO: 2,
+	SEN: 3,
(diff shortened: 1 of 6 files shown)


ayyaruq (approved):
Order looks fine to me, but I'd suggest at some point moving Higanbana to use core dots and Kenki+Shoha to core gauge. IMO also for UX, the Sen title could be shorter, the "non-standard" part of it prob isn't needed and I'm not sure it's really a window.

PR #1079, opened 2021-01-07
Nuketrino

ackwell:
This PR completely re-tools xiva from the ground up. We're nuking neutrino. And TSC, tslint, and a whole host of other stuff besides. Nuketrino, for short.

There is a _lot_ of lint rules disabled here - most of them because tslint was kinda broken and TS linting slipped. I wanted to keep code changes minimal. I'll be sweeping through in a follow up to enable those.


babel.config.js
```diff
@@ -0,0 +1,42 @@
...
+	}],
+	['@babel/plugin-proposal-decorators', {legacy: true}],
+	['@babel/plugin-proposal-class-properties', {loose: true}],
+	'babel-plugin-macros',
+	'babel-plugin-lodash',
+	'./locale/babel-plugin-transform-react.js',
+	['@babel/plugin-transform-runtime', {
+		corejs: {version: 3},
+		// TODO: Enable for web build only?
+		// useESModules: true,
```
Jessidhia, line 18: just enable this, preset-env will handle the "real" module stuff

src/reportSources/legacyFflogs/eventAdapter.ts
```diff
@@ -98,6 +99,7 @@ class EventAdapter {
 		// I mean if Kihra doesn't know, how am I supposed to?
 		case 'unknown':
 			break
+		/** eslint-enable no-fallthrough */
```
ackwell, line 102: akk note: extra `*`, remove

webpack.config.ts
```diff
@@ -0,0 +1,194 @@
...
+					mode !== 'development'
+						? {loader: MiniCssExtractPlugin.loader, options: {esModule: true}}
+						: {loader: 'style-loader'},
+					{
+						loader: '@teamsupercell/typings-for-css-modules-loader',
+						options: {
+							banner: '// This file is automatically generated. Do not edit.',
+							eol: '\n',
+							disableLocalsExport: true,
+              verifyOnly: mode !== 'development',
```
ackwell, line 139: akk note: indent is fucky

babel.config.js
ackwell: Ah, on fiddling I remember - I'm running the webpack config (`.ts`, so needs transpilation) through babel, which is using this config as well. Enabling that (even with `preset-env`) throws errors, as webpack/node doesn't handle the output. Thanks for the heads up to config in `preset-env` though, I've thrown together a caller condition for `@babel/register` that catches the WP config transpiles and swaps target to `node`.

PR #1081, opened 2021-01-08
Flag WHM support for 5.4

Shirk:
No changes to WHM between 5.3 and 5.4 so it should be safe to mark it supported.


src/parser/jobs/whm/index.js
```diff
@@ -22,14 +22,19 @@ export default new Meta({
...
 	contributors: [
 		{user: CONTRIBUTORS.NIV, role: ROLES.MAINTAINER},
 		{user: CONTRIBUTORS.VULCWEN, role: ROLES.DEVELOPER},
 		{user: CONTRIBUTORS.SUSHIROU, role: ROLES.DEVELOPER},
 	],
 	changelog: [
+		{
+			date: new Date('2021-01-08'),
+			Changes: () => <>Add 5.4 support for WHM</>,
+			contributors: [CONTRIBUTORS.NONO],
```
ayyaruq, line 36: Why did you mark this as Nono's work?

ayyaruq (changes requested):
Thanks for the PR, please add yourself as a contributor and set it appropriately.

ackwell:
(also a side note that merging this basically means you're on line for any support that comes up for whm until the next patch break point - make sure you drop by the discord server!

src/parser/jobs/whm/index.js
Shirk: Copy paste was too strong.. fixed

ayyaruq (approved):
Thanks!

PR #1084, opened 2021-01-08
SMN - Move EF statusApplied from enkindle to FBT

yumiyafangirl:
I believe this is just a longstanding typo from #636 since I can't find any justification for attaching Everlasting Flight to Enkindle Phoenix. Paging @ackwell just in case there's a good reason for having it this way that I missed.

Fixes a bug where PrecastStatus synths a cast of Enkindle Phoenix at the start of the timeline of most SMN logs.


KelosPelmand:
Not ackwell, but I agree that this looks like the correct fix for the reported bug.

KelosPelmand (approved):

ayyaruq (approved):
I've got NFI on the logic but code looks good and Kelos approved so, lets go.

PR #1085, opened 2021-01-10
events: port core/Death

ackwell:
Title.

This PR:
- Adds all the limit breaks to `ACTION` data because I thought I was going to need the healer ones.
- Tweaks some new event stuff 'cus it's still early days
- Ports core/Death to the new system


src/parser/core/modules/Death.tsx
```diff
@@ -0,0 +1,179 @@
...
+		// Queue the raise notification.
+		// Also fabricating a legacy event for backwards compatibility.
+		this.parser.queueEvent({
+			type: 'raise',
+			timestamp: event.timestamp,
+			actor: this.parser.actor.id,
+		})
+		this.parser.fabricateLegacyEvent({
+			type: 'raise',
+			timestamp: event.timestamp,
```
ackwell, line 129: akk note; this is inccorrect. event timestamp will be epoch relative, should use `parser.currentTimestamp` or similar

src/data/ACTIONS/root/LIMIT_BREAK.ts
```diff
@@ -0,0 +1,194 @@
...
+		name: 'Crimson Lotus',
+		icon,
+	},
+
+	// Ranged LB2
+	DESPERADO: {
+		id: 4239,
+		name: 'Desperado',
+		icon,
+	},
```
ayyaruq, line 158: Should this maybe be in the order of LB1, LB2, BRD/MCH/DNC LB3?

src/data/STATUSES/root/SHARED.ts
```diff
@@ -25,6 +26,12 @@ export const SHARED = ensureStatuses({
 		icon: 'https://xivapi.com/i/015000/015011.png',
 	},
 
+	TRANSCENDENT: {
+		id: 418,
+		name: 'Transcendent',
+		icon: 'https://xivapi.com/i/015000/015273.png',
+	},
```
ayyaruq, line 33: IIRC this has a max duration of 5s if it matters.

ayyaruq (commented):
Looks pretty good, and is a solid example of using the new event system. I do wonder tho, would it be worth also hooking the "you have been given a raise" status and track time where it's sitting there but not accepted/declined? Sometimes you wait for stuff to avoid raising at a bad time (ie in the middle of a protean where you're likely to clip someone) but I figure it's not something that's generally held.

src/data/STATUSES/root/SHARED.ts
ackwell: I was hesitant to give it a duration given it drops the moment the player uses something. Duno.

src/data/ACTIONS/root/LIMIT_BREAK.ts
ackwell: Yup. Typo.

ackwell (commented):
> Sometimes you wait for stuff to avoid raising at a bad time (ie in the middle of a protean where you're likely to clip someone) but I figure it's not something that's generally held.

this is what i was concerned about. not sure

src/data/STATUSES/root/SHARED.ts
ayyaruq: I kinda agree but there's other statuses people can lose early (ie Meditate, Anatman, any other channeling shit), plus actually removing stuff I guess although that's rare outside tanking dungeons with a derpy healer and poor regen timing.

PR #1087, opened 2021-01-11
show scholar dots

mairii:
I took the smn code that shows dots on tagets and modified it to show dots and if they were late or early.. theres a lot of problems with this that I can't figure out tho.. pls help 😭


mairii:
I understand this better.. should I put all the code from dotstwo.js into dots.js?

Rjb99:
You should drop by the discord server and talk in either #dev-general or #dev-healer before getting too deep in to this project just to make sure you have all your bases set up right.

But yes, All functions for dots should be contained within the dots.js, 1 file for everything related to a skill is the standard here.

In addition the current lint error on this PR is from line 86:

if (index != 0) { 

lint is expecting !== instead of !=, the difference between the two is that in addition to checking if the value is the same, !== and === will make sure that type of the 2 objects being compared is also the same

mairii:
okies, I will start rewriting

mairii:
@Rjb99 I moved all the code but now every dot is happening twice and I dont know what I did wrong 😭

src/parser/jobs/sch/modules/DoTs.js
```diff
@@ -28,12 +29,146 @@ export default class DoTs extends CoreDoTs {
...
+				return <Trans id="sch.dots.drift.application">reapplication</Trans>
+			}
+			if (drift > lateAllow) {
+				return <Trans id="sch.dots.drift.late">{this.parser.formatDuration(Math.abs(drift))} late</Trans>
+			}
+			if (drift < earlyAllow) {
+				return <Trans id="sch.dots.drift.early">{this.parser.formatDuration(Math.abs(drift))} early</Trans>
+			}
+		}
+		return <Trans id="sch.dots.drift.application">normal application</Trans>
```
yumiyafangirl, line 105: Since you're adding new translations, you need to run `yarn extract` so that other languages are aware of the new strings that need translating.

mairii:
k, I fixed all the issues I think? there are more in discord I know and Ill work on those next

src/parser/jobs/sch/modules/DoTs.js
```diff
@@ -28,12 +29,140 @@ export default class DoTs extends CoreDoTs {
...
+	}
+
+	_pushApplication(targetKey, statusId, event) {
+		const target = this._application[targetKey] = this._application[targetKey] || {[STATUSES.BIOLYSIS.id]: []}
+		const source = this._lastBioCast
+		target[statusId].push({event, source})
+	}
+
+	_onTrackedDotCast(event) {
+		this._lastBioCast = event.ability.guid
```
AzariahAeroborn, line 64: This isn't doing anything meaningful, since Biolysis's ID is the same all the time.  SMN tracks which skill applied the DOT since they have two different ways to apply it, but that doesn't really apply to SCH, so this should be removed, along with the cast hook and anything else using _lastBioCast.

src/parser/jobs/sch/modules/DoTs.js
```diff
@@ -28,12 +29,140 @@ export default class DoTs extends CoreDoTs {
...
+
+		// Make sure we're tracking for this target
+		const applicationKey = `${event.targetID}|${event.targetInstance}`
+		//save the application for later use in the output
+		this._pushApplication(applicationKey, statusId, event)
+	}
+
+	_formatDriftText(timestamp, previous=0) {
+		let drift = 0
+		const lateAllow = 1000
```
AzariahAeroborn, line 78: Our convention is to define constants like this in the header (outside the module).  Typically something like `const LATE_ALLOW` or similar (personally, I'd be a little more verbose in the naming, but I won't block on exact names).

I'm also a little leery of the allowed ranges here - 1 second late reapplication doesn't allow for much wiggle room in terms of the time to finish your previous GCD as the DOT is expiring.  Our usual convention for re-use of oGCDs is to allow for up to 1 GCD worth of "drift" per cast, and I believe other DOT modules "forgive" up to 3 seconds of early re-application as convention.

src/parser/jobs/sch/modules/DoTs.js
```diff
@@ -28,12 +29,140 @@ export default class DoTs extends CoreDoTs {
...
+		const earlyAllow = -2500
+		if (previous !== 0) {
+			const delta = timestamp - previous
+			drift = delta - (STATUSES.BIOLYSIS.duration * 1000)
+			if (drift > (STATUSES.BIOLYSIS.duration * 1000)) {
```
AzariahAeroborn, line 83: As written, this isn't going to give any forgiveness in a case such as:
- Applied Biolysis 20 seconds before the boss went untargetable
- Boss goes untargetable for 20 seconds
- DOT is reapplied immediately after the Boss becomes available (so roughly 10 seconds after it "fell off" while the boss was untargetable).

The "use your cooldowns" module adjusts for downtime by moving the "time of next expected use" to the end of the downtime window - would that make sense here?  So in the above example, the drift calculation would update "previous" to be the end of the downtime window, maybe plus a GCD of forgiveness in case there's something else priority to cast after the downtime?

AzariahAeroborn (changes requested):

src/parser/jobs/sch/modules/DoTs.js
mairii: sorry, but where is the use your cooldowns module? is it cooldowns or cooldowndowntime?

AzariahAeroborn: The module name in core is CooldownDowntime, yes.  Which results in the "Use your cooldowns" checklist item for most classes.

ackwell:
Heyo, sorry, just doing some PR queue cleanup. This PR targets the shadowbringers master, which will no longer be a thing after today.

While I imagine things have changed quite a lot since this PR was made, if you're still interested in sch contributions, please do drop by on discord!

PR #1088, opened 2021-01-13
MCH - flag double weaves after a hypercharge GCD as weaving issues

yumiyafangirl:
Sequel to #1086, these GCDs have a fixed 1.5s recast and we really really don't want to double weave after them.

### DIFF
--- a/src/parser/jobs/mch/modules/Weaving.ts
+++ b/src/parser/jobs/mch/modules/Weaving.ts
@@ -1,17 +1,24 @@
-import ACTIONS from 'data/ACTIONS'
+import {ActionRoot} from 'data/ACTIONS/root'
 import {CastEvent} from 'fflogs'
 import CoreWeaving, {WeaveInfo} from 'parser/core/modules/Weaving'
 
-const HYPERCHARGE_ACTION_IDS = [
-	ACTIONS.AUTO_CROSSBOW.id,
-	ACTIONS.HEAT_BLAST.id,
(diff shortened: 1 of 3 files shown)


src/parser/jobs/mch/modules/Weaving.ts
```diff
@@ -0,0 +1,21 @@
+import ACTIONS from 'data/ACTIONS'
+import {CastEvent} from 'fflogs'
+import CoreWeaving, {WeaveInfo} from 'parser/core/modules/Weaving'
+
+const HYPERCHARGE_ACTION_IDS = [
+	ACTIONS.AUTO_CROSSBOW.id,
+	ACTIONS.HEAT_BLAST.id,
+]
```
ackwell, line 8: This doesn't respect potential patch layering. The more forward-compatible way of defining this stuff is to define it with root keys:

```suggestion
const HYPERCHARGE_ACTIONS: (keyof ActionRoot)[] = [
	'AUTO_CROSSBOW',
	'HEAT_BLAST',
]
```

which you can then map to the IDs at parse-time when patch info is available

```ts
// (you'll probably want to do this once in init or similar)
const theIds = HYPERCHARGE_ACTIONS.map(key => this.data.actions[key].id)
```

There's some good examples of this in AST

yumiyafangirl: 👍 added it for DNC too

ackwell (approved):

PR #1091, opened 2021-01-22
fix(pld.ts): total eclipse now starts a combo for prominence

pogramos:
Total Eclipse wasn't starting the combo correctly, making all usages of TE and Prominence result in an `Uncomboed` Prominence within the analysis


ayyaruq (approved):

ayyaruq:
Thanks for the PR!

PR #1093, opened 2021-01-23
BuffWindow - handle simple statuses as tracked actions

yumiyafangirl:
Adds an optional status field to BuffWindowTrackedActions. When this field is present, usages are counted by RemoveBuff events rather than CastEvents. This allows for actions like Reassemble and Barrage to behave properly as trackedActions (the necessary changes for both are also included in this PR).

Before:
image

After:
image

https://xivanalysis.com/fflogs/zjtxA4GQLZgkC2Ph/10/8


Rjb99:
immediate question that springs to mind without looking to deep at things, what about removebuff events that occur when the buff is simply not consumed by a skill/use?

AzariahAeroborn (approved):

PR #1096, opened 2021-01-30
MCH - Add queen actions to the timeline

yumiyafangirl:
image

### DIFF
--- a/src/parser/core/modules/PetTimeline.tsx
+++ b/src/parser/core/modules/PetTimeline.tsx
@@ -92,7 +92,6 @@ export default class PetTimeline extends Module {
 
 			// TOOD: Type cooldowns so this cast isn't required
 			const parentRow: ContainerRow = this.cooldowns.getActionTimelineRow(summonAction)
-			parentRow.order = -99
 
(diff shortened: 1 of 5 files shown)


src/parser/core/modules/PetTimeline.tsx
```diff
@@ -92,6 +92,7 @@ export default class PetTimeline extends Module {
 
 			// TOOD: Type cooldowns so this cast isn't required
 			const parentRow: ContainerRow = this.cooldowns.getActionTimelineRow(summonAction)
+			parentRow.order = -99
```
ayyaruq, line 95: Looks ok but what's this change for?

yumiyafangirl: Since the pet group gets ordered at the top in the other two cases, I think it makes sense to do the same thing for this case.

yumiyafangirl: Since the pet group gets ordered at the top in the other two cases, I think it makes sense to do the same thing for this case.

ayyaruq (approved):

PR #1098, opened 2021-01-31
PLD - Tinctures

yumiyafangirl:
Adds a tincture module with 13 required GCDs and 1 use of sw/soc per pot.

image

### DIFF
--- a/src/parser/jobs/pld/modules/Tincture.tsx
+++ b/src/parser/jobs/pld/modules/Tincture.tsx
@@ -32,7 +32,7 @@ export default class Tincture extends CoreTincture {
 			},
 		],
 		suggestionContent: <Trans id="pld.tincture.suggestions.trackedactions.content">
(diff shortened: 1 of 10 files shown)


src/parser/jobs/pld/modules/Tincture.tsx
```diff
@@ -0,0 +1,43 @@
...
+				action: this.data.actions.SPIRITS_WITHIN,
+				expectedPerWindow: 1,
+			},
+			{
+				action: this.data.actions.CIRCLE_OF_SCORN,
+				expectedPerWindow: 1,
+			},
+		],
+		suggestionContent: <Trans id="pld.tincture.suggestions.trackedactions.content">
+			One use of <ActionLink {...this.data.actions.SPIRITS_WITHIN}/> and one use of <ActionLink {...this.data.actions.CIRCLE_OF_SCORN}/> should occur during every <ActionLink {...this.data.actions.INFUSION_STR}/> window.
```
qaphla, line 35: Nitpicking: Probably should say *at least* one use of Circle of Scorn --- one will be common but people shouldn't be surprised to see 2 (and shouldn't take the suggestion to mean that having 2 is a bad thing).

yumiyafangirl: Fixed (even though I wrote SoC in the commit for some reason)

AzariahAeroborn (approved):

PR #1101, opened 2021-02-05
Pld copy fix

qaphla:
Fixes an issue with the intro/overview text for PLD where skill links were broken.

### DIFF
--- a/src/parser/jobs/pld/index.tsx
+++ b/src/parser/jobs/pld/index.tsx
@@ -12,7 +12,7 @@ export default new Meta({
 		<Trans id="pld.about.description">
 			<p>As the illegitimate child of a WHM and BLM, you chose the gory path of a shield lobbing, sword swinging Mage that also tries to help everybody out.
 			</p>
-			<p>This analyzer attempts to find just the right things to get you to be a fearsome tank that will show no <ActionLink {...ACTIONS.CLEMENCY}/> to your enemies, while being the true <ActionLink {...ACTIONS.HOLY_SPIRIT}/> of the party.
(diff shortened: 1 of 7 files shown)


src/parser/jobs/pld/index.tsx
```diff
@@ -1,19 +1,20 @@
...
-
 export default new Meta({
 	modules: () => import('./modules' /* webpackChunkName: "jobs-pld" */),
 
 	Description: () => <>
-		<TransMarkdown source={description} key="pld.about.description"/>
+		<Trans id="pld.about.description">
+			<p>As the illegitimate child of a WHM and BLM, you chose the gory path of a shield lobbing, sword swinging Mage that also tries to help everybody out.
+			</p>
+			<p>This analyzer attempts to find just the right things to get you to be a fearsome tank that will show no <ActionLink {...ACTIONS.CLEMENCY}/> to your enemies, while being the true <ActionLink {...ACTIONS.HOLY_SPIRIT}/> of the party.
```
ayyaruq, line 15: inb4 ack kills me, did you mean `analyser`? I feel like the first half of this sentence reads weirdly and could maybe be rethought tho. Won't block on it but hm.

qaphla: I don't like it much either but I'm bad at writing this sort of thing, so all I did was clean up what was there originally (in particular fixing the actionlinks), but also a bit of spelling and making the last sentence less gendered.

Any better writing suggestions welcome.

ayyaruq: Maybe something like this? The wording is mostly fine, just reads a bit odd grammatically.

```suggestion
			<p>This analyser attempts to find just the right things to turn you into a fearsome tank that will show no <ActionLink {...ACTIONS.CLEMENCY}/> to your enemies while being the true <ActionLink {...ACTIONS.HOLY_SPIRIT}/> of the party.
```

AzariahAeroborn (dismissed):

AzariahAeroborn (approved):

PR #1102, opened 2021-02-10
GNB: Cart Timeline Update (Overcap + Extra Info)

VorpalSwords:
Adding two features for the gunbreaker - 
1. Showing source for the ammo timeline.
2. Having "waste" or "overcap" part of the timeline, so players could see when it happend.

I am not sure how good the code is, would love to have some help with that to make it more maintain-able
but it does the work.
I can add comments if it helps.

This is how it looks:
Generator: https://cdn.discordapp.com/attachments/808985736962506752/809188772217880586/unknown.png
Spender: https://cdn.discordapp.com/attachments/808985736962506752/809188810121805884/unknown.png
Overcap Before: https://cdn.discordapp.com/attachments/808985736962506752/809188855710089286/unknown.png
Overcap After: https://cdn.discordapp.com/attachments/808985736962506752/809188883283443712/unknown.png

### DIFF
--- a/src/parser/jobs/gnb/modules/Ammo.tsx
+++ b/src/parser/jobs/gnb/modules/Ammo.tsx
@@ -61,6 +61,11 @@ class AmmoState {
 	y?: number
 	source: string
 	action: number
+
+	constructor() {
+		this.source = ''
+		this.action = 0
+	}
 }
 
 class OvercapState {
@@ -69,6 +74,12 @@ class OvercapState {
 	wasted: number
 	source: string
 	timing: number
+
+	constructor() {
+		this.wasted = 0
+		this.source = ''
+		this.timing = 0
+	}
 }
 
 export default class Ammo extends Module {
@@ -78,6 +89,7 @@ export default class Ammo extends Module {
 
 	private ammo = 0
 	private currentAbility = ''
+	private currentAbilityImpact = 0
 	private ammoHistory: AmmoState[] = []
 	private overcapHistory: OvercapState[] = []
 	private wasteBySource = {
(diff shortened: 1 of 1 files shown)


AzariahAeroborn:
Hey, thanks for the contribution, and much thanks for the sample screenshots, makes it much easier to review.

I'm still doing some digging into whether the charting library can support JSX <Trans> tags for the tooltip output rather than raw strings, so that the tooltips can be internationalized, but I wanted to point out a couple other things.

1. The lint tests aren't passing, because there's a number of code style guide things that need to be fixed with the code.  You should be able to see them by clicking on the "Details" link next to the lint test - if you have problems pulling them up, let me know.
2. I find it a little jarring that the tooltip wording/structure for the overcap tooltips is different than the structure for standard tooltips.  I suppose it makes sense that you'll never overcap by using a spender, but my eye was kind of expecting Generator: instead of the Before/After.  I also kind of wonder if it would simplify the tracking and handling code if both "regular" Generator events and "overcap" Generator events could use the same tooltip, aside from overcap events including some indication of how many cartridges were lost... but I definitely think this is a nice improvement, and if you're happy with this interface over trying to fiddle with those tooltips, it's fine by me.

VorpalSwords:
> Hey, thanks for the contribution, and much thanks for the sample screenshots, makes it much easier to review.
> 
> I'm still doing some digging into whether the charting library can support JSX tags for the tooltip output rather than raw strings, so that the tooltips can be internationalized, but I wanted to point out a couple other things.
> 
> 1. The lint tests aren't passing, because there's a number of code style guide things that need to be fixed with the code.  You should be able to see them by clicking on the "Details" link next to the lint test - if you have problems pulling them up, let me know.
> 2. I find it a little jarring that the tooltip wording/structure for the overcap tooltips is different than the structure for standard tooltips.  I suppose it makes sense that you'll never overcap by using a spender, but my eye was kind of expecting Generator: instead of the Before/After.  I also kind of wonder if it would simplify the tracking and handling code if both "regular" Generator events and "overcap" Generator events could use the same tooltip, aside from overcap events including some indication of how many cartridges were lost... but I definitely think this is a nice improvement, and if you're happy with this interface over trying to fiddle with those tooltips, it's fine by me.

Hey, thanks for the quick review!
As for the charting tooltip being more profesional than what I have done, it might require some more knowledge on the subject which I don't have.
I have fixed the lint problems except for a type problem that I don't know how to handle (the chartjs docs says one thing, but lint does not accept it sadly).
As for your second point, I wonder how "Generator" and "Spender" works with "Overcap"? I have put two coordinates for the overcap (one for before, and one for after) mostly to create a visable line, 
otherwise it would just be a dot up at (t: timestamp, y: 3).
How many ammo was lost - that is a good addition and I should add it.

VorpalSwords:
image
image
image
image

VorpalSwords:
Let me know what do you think!

AzariahAeroborn:
Everything looks good for basic functionality.  I'd like to request / encourage you to work on adding i18n support in a follow-up PR, but let's get this in front of people and start getting some value from it while we work on that.  

Please feel free to ping me in Discord if you'd like assistance with i18n support.

AzariahAeroborn (approved):

PR #1103, opened 2021-02-12
Melee combo severity change

Cephid:
Was asked to make all melee combo breaks Major.

I'm not sure why the other commits are there; as this is a new branch.


ayyaruq (approved):
Extra commits prob coming from your branch not being from latest master, either way they'll be squashed. LGTM, thank.

PR #1107, opened 2021-02-19
DNC - update guide references

yumiyafangirl:
Since Tsuki's FAQ and Ellunavi's writeup are both outdated, it makes sense to update the DNC index to ringabel's more recent guide (with his permission).


AkairyuGestalter:
Minor request: It looks like the assumption about how frequently DNC's generate esprit for themselves is now a 30% chance, as opposed to the 25% chance folks were guessing when I originally wrote this all up. Can you go ahead and update the ESPRIT_RATE_SELF const in Esprit.tsx to match this guide?

Otherwise, I'm happy with this. Reading through that guide was a good reminder of some gameplay aspects that I'd forgotten :)

yumiyafangirl:
> Minor request: It looks like the assumption about how frequently DNC's generate esprit for themselves is now a 30% chance, as opposed to the 25% chance folks were guessing when I originally wrote this all up. Can you go ahead and update the ESPRIT_RATE_SELF const in Esprit.tsx to match this guide?
> 
> Otherwise, I'm happy with this. Reading through that guide was a good reminder of some gameplay aspects that I'd forgotten :)

Good catch! Fixed it.

AkairyuGestalter (approved):

PR #1112, opened 2021-03-16
Low effort April Fools Clippy

AkairyuGestalter:
What it says on the tin (I'm not going to be hurt if this gets rejected 🤣)


yumiyafangirl:
- [X] Necessary change

PR #1113, opened 2021-03-18
Adapter step for precast actions

yumiyafangirl:
Adds a new fflogs adapter step that synthesizes 'action' events for actions used pre-pull. The old version (PCA) needs to stay for now since it's still depended on by legacy modules.

image

### DIFF
--- a/src/reportSources/legacyFflogs/eventAdapter/prepull/action.ts
+++ b/src/reportSources/legacyFflogs/eventAdapter/prepull/action.ts
@@ -1,27 +1,37 @@
 import {Event, Events} from 'event'
 import {FflogsEvent} from 'fflogs'
+import {action} from 'mobx'
+import {isDefined} from 'utilities'
 import {AdapterStep} from '../base'
 
 export class PrepullActionAdapterStep extends AdapterStep {
 	private finished = false
+	private initialEventLocation?: Event[]
 
 	adapt(baseEvent: FflogsEvent, adaptedEvents: Event[]): Event[] {
-		const synthesizedEvents: Event[] = []
-
 		if (this.finished) {
 			return adaptedEvents
 		}
 
+		if (!this.initialEventLocation) {
+			this.initialEventLocation = adaptedEvents
+		}
+
+		const synthesizedEvents: Event[] = []
 		for (const event of adaptedEvents) {
-			if (event.type === 'action') {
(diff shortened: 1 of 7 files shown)


src/reportSources/legacyFflogs/eventAdapter/prepull/action.ts
```diff
@@ -0,0 +1,34 @@
...
+	private finished = false
+
+	adapt(baseEvent: FflogsEvent, adaptedEvents: Event[]): Event[] {
+		const synthesizedEvents: Event[] = []
+
+		if (this.finished) {
+			return adaptedEvents
+		}
+
+		for (const event of adaptedEvents) {
```
ackwell, line 15: hm, this is still effectively iterating over the fflogs events one at a time, linearly. while it's _technically_ unlikely that there's >1 snapshot before the first action, if we assume a case that there _was_, we'd get
```
action
snapshot
action
snapshot
```

i... _might_ be over-thinking this, _but_ i think we should ensure the precast synths are _always_ before the first translated event.

spitball on how to do that;

```ts
class Etc {
	initialEvents?: Event[]
	adapt(baseEvent, adaptedEvents) {
		if (finished) { return ... }
		// grab a reference to the very first adapted events array
		if (initialEvents == null) { initialEvents = adaptedEvents }

		// logic
		
		// mutate the initial events array with the synthed events at the start
		initialEvents.splice(0, 0, ...synthedEvents)
	}
}
```

for that to work you'll need to tweak the adapter to something like

```ts
return events
	.map(baseEvent => theReduceStuffWeAlreadyHave)
	.flat()
```

(i.e. map -> flat, rather than flatmap)

yumiyafangirl:
I added a basic test but I'm not 1000% sure I'm understanding the event flow properly, so you might want to double check that. We end up going from calculateddamage to [action, snapshot, actorUpdate, actorUpdate].

ackwell:
I realised my suggestion above is pretty fragile if any other step does a spread anywhere. As discussed on discord, have added a `postprocess` step to adapters, and moved prepull into it.

Other tweaks
- moved `prepull/action` to `prepullAction` to keep the adapters flat for _now_ (didn't notice earlier oops)
- tweaked the test to make sure it's inserting before all base adaptions.

ackwell (approved):

PR #1114, opened 2021-03-19
Allow one Triplecast weave in JP opener & fix BLM description actionlink

AkairyuGestalter:
1. The JP opener (both versions) require a single clipped Triplecast weave after the first Fire IV. Converting the 'Opener Enochian' check to work for both of these skills, since they have the same opener considerations.
2. The About description for the BLM module had a busted action link in it. Switching from TransMarkdown to Trans to resolve that.


rexatlanti:
Looks good. Clever idea to use the eno logic for that :3

ackwell (approved):

ackwell:
i really need to fix that md renderer

PR #1115, opened 2021-03-25
Modify getDataBy to return all matching entries

yumiyafangirl:
Adds a new function, `getDataArrayBy` with the same signature as gDB except it returns all matches instead of just the last one. Also includes a safeguard against a bug with `precastStatus` that occurs when there are multiple actions in data that apply the same status.

### DIFF
--- a/src/data/getDataBy.ts
+++ b/src/data/getDataBy.ts
@@ -25,7 +25,7 @@ export function getDataArrayBy<
 	data: Data,
 	by: Key,
 	value: FlattenArray<Value[Key]>,
-):  Value[] | undefined {
+): Value[] | undefined {
 	// Sanity check in case someone is using this from JS and misspelled a key
 	const dataKeys = Object.keys(data)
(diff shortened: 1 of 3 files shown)


src/data/getDataBy.ts
```diff
@@ -51,13 +51,32 @@ export function getDataBy<
 
 			// Keys can be arrays (see STATUSES), handle it
 			const newKeys = ensureArray(val[by])
-			newKeys.forEach(key => map.set(key, val))
+			newKeys.forEach(key => {
+				if (map.has(key)) {
+					map.get(key)?.push(val)
+				} else {
+					map.set(key, [val])
+				}
```
ackwell, line 59: while races are _technically_ unlikely because >js runtime, i generally preference the get-once method;

```ts
let values = map.get(key)
if (values == null) {
	values = []
	map.set(key, values)
}
values.push(val)
```

ackwell (approved):

PR #1116, opened 2021-03-26
Adapter step for precast statuses

yumiyafangirl:
#### Depends on #1115 

Ports the logic in `PrecastStatus` to the new event pipeline. I tried to cover all the edge cases but this is a particularly edgy area of the code base so I would appreciate some extra eyes on it (_cough_ @AzariahAeroborn).


src/reportSources/legacyFflogs/eventAdapter/prepullAction.ts
```diff
@@ -24,7 +24,7 @@ export class PrepullActionAdapterStep extends AdapterStep {
 		return {
 			...event,
 			type: 'action',
-			timestamp: this.pull.timestamp,
+			timestamp: this.pull.timestamp - 1,
```
ackwell, line 27: ah, thanks for adding this. i'm _tempted_ to suggest it should be in `base` or a new `consts` given we're sharing this between prepull action and status, but equally i'm not sure that the actual value is important - i guess, if it's important that the `-1` and `-2` are in that particular order, it should probably be shared.

src/reportSources/legacyFflogs/eventAdapter/adapter.ts
```diff
@@ -1,15 +1,18 @@
...
 import {DeduplicateActorUpdateStep} from './deduplicateActorUpdates'
 import {DeduplicateStatusApplicationStep} from './deduplicateStatus'
 import {PrepullActionAdapterStep} from './prepullAction'
+import {PrepullStatusAdapterStep} from './prepullStatus'
 import {TranslateAdapterStep} from './translate'
 
 /** Adapt an array of FFLogs APIv1 events to xiva representation. */
 export function adaptEvents(report: Report, pull: Pull, events: FflogsEvent[]): Event[] {
 	const adapter = new EventAdapter({report, pull})
+	sortEvents(events)
```
ackwell, line 15: selfish request;

```suggestion
	// TODO: Move sort logic into adapter scope once legacy is removed
	sortEvents(events)
```

src/reportSources/legacyFflogs/eventAdapter/prepullStatus.ts
```diff
@@ -0,0 +1,110 @@
+import {getDataBy, getDataArrayBy} from 'data'
+import ACTIONS from 'data/ACTIONS'
+import STATUSES, {Status} from 'data/STATUSES'
+import {Event, Events} from 'event'
+import _ from 'lodash'
+import {Actor} from 'report'
+import {AdapterStep} from './base'
+
+type StatusEvent = Events['statusApply'] | Events['statusRemove']
+type StatusKey = keyof typeof STATUSES
```
ackwell, line 10: should be able to import `StatusKey` directly from `data/STATUSES`

src/reportSources/legacyFflogs/eventAdapter/prepullStatus.ts
```diff
@@ -0,0 +1,110 @@
...
+		return this.precastEvents.length > 0
+			? [...this.precastEvents, ...adaptedEvents]
+			: adaptedEvents
+	}
+
+	private observeAction(actionId: number, sourceId: Actor['id']) {
+		if (!this.observedActions.has(sourceId)) {
+			this.observedActions.set(sourceId, new Set())
+		}
+		this.observedActions.get(sourceId)?.add(actionId)
```
ackwell, line 58: same story as before re:get-once

src/reportSources/legacyFflogs/eventAdapter/prepullStatus.ts
```diff
@@ -0,0 +1,110 @@
...
+	postprocess(adaptedEvents: Event[]): Event[] {
+		for (const event of adaptedEvents) {
+			if (event.type !== 'statusApply' && event.type !== 'statusRemove') {
+				if (event.type === 'action') {
+					this.observeAction(event.action, event.source)
+				}
+				continue
+			}
+
+			const status = getDataBy(STATUSES, 'id', event.status)
```
ackwell, line 28: ngl i'm really pleased the event narrowed the way it did. i know it _should_, but it's nice to see it actually doing it in the wild

src/reportSources/legacyFflogs/eventAdapter/prepullStatus.ts
```diff
@@ -0,0 +1,110 @@
...
+				// No data for this status, skip to next event
+				continue
+			}
+
+			if (event.type === 'statusApply') {
+				this.synthesizeActionIfNew(event, status)
+				this.observeStatus(status.id, event.target)
+
+			} else if (event.type === 'statusRemove') {
+				if (this.observedStatuses.get(event.target)?.has(status.id)) {
```
ackwell, line 39: for the sake of explicitnessnessness

```suggestion
				if (this.observedStatuses.get(event.target)?.has(status.id) ?? false) {
```

src/reportSources/legacyFflogs/eventAdapter/prepullStatus.ts
```diff
@@ -0,0 +1,110 @@
...
+
+	private observeStatus(statusId: number, targetId: Actor['id']) {
+		if (!this.observedStatuses.has(targetId)) {
+			this.observedStatuses.set(targetId, new Set())
+		}
+		this.observedStatuses.get(targetId)?.add(statusId)
+	}
+
+	private synthesizeActionIfNew(event: StatusEvent, status: Status) {
+		const statusKey = _.findKey(STATUSES, status) as StatusKey | undefined
```
ackwell, line 69: hm - i'd _strongly_ like to avoid the old default exports, esp. in the new system. but we're kind of in a corner, given the layered data logic is in a parser module. 😩

will think about it. ideally we'd use layered data here, too - as realistically this logic is running on patch-specific data.

ackwell (commented):
mostly lookin good, nice.
main sticking point is the data imports, though as i mention, it's not something i have a good answer for straight away. i'll have a think about it - feel free to poke if i forget

src/reportSources/legacyFflogs/eventAdapter/prepullAction.ts
yumiyafangirl: Seems like a good idea, maybe something like this in `base`?

```typescript
const PREPULL_OFFSETS = {
	STATUS_ACTIONS: -3,
	STATUS_APPLICATIONS: -2,
	PULL_ACTION: -1,
} as const
```

ackwell: sgtm

ackwell (approved):
i feex 👍

PR #1117, opened 2021-03-26
Add actors module for new event system

ackwell:
Omitting the timeline stuff I posted in dev-gen, it was getting a bit big. Will PR separately when it's done.

This adds a new `Analyser`, `actors`. It's designed to replace the _original_ intent of the old-event `Entities` system. To that end, it tracks all actor update events, and provides simple access to information about the current status of an actor, as well as snapshots of its state in the past.

```ts
// Retrieve an actor
actor = this.actors.get(someActorId)
actor = this.actors.current // the actor being analused

// Retrieve metadata and resources
actor.name
actor.hp.current
actor.position.bearing

// Retrieve info in the past
snapshot = actor.at(someTimestamp)
snapshot.mp.current
snapshot.position.x

// etc
```

Of note, the status logic handling currently present in the old entities system is not implemented here. I'm as yet undecided if it should live here, or in a separate module. Will flesh out more when known.


ackwell:
tfw left default title. will check tests on this tomorrow or something

PR #1118, opened 2021-03-28
Ignore precast events when counting weaves

AkairyuGestalter:
Events synthed as precast shouldn't affect the weave count, since they happened before the first damage event. Kick out any events with a timestamp earlier than the beginning of the fight.


ayyaruq (approved):
LGTM

PR #1120, opened 2021-03-29
Run yarn extract, fix build lint issues in prod

redshadowhero:
Seems like messages are in `locale/en/messages.json` that aren't in the
other ones, so seems like either crowdin is missing the memo or
something else is going on. Regardless, should fix the build issue.

Obviously if this doesn't help, let me know here (@ me), and I'll try to work further.


ackwell (approved):
yeah, crowdin merge last night was missing an extract. dumb system. thanks!

PR #1123, opened 2021-04-04
Medicated row in Timeline

ayyaruq:
Uses the new Analyser module, prob good to double check I didn't pull an acc and fail to read.

Screenshots:
image

With pets:
image


src/parser/core/modules/GlobalCooldown.js
```diff
@@ -54,7 +54,7 @@ export default class GlobalCooldown extends Module {
 
 		this.timelineRow = this.timeline.addRow(new ContainerRow({
 			label: 'GCD',
-			order: -98,
+			order: -97,
```
ackwell, line 57: i should consolidate this stuff so it's less of a pain to reorder

src/parser/core/modules/Medicated.ts
```diff
@@ -0,0 +1,71 @@
...
+		}
+
+		this.pots.push({end: event.timestamp - this.parser.pull.timestamp, ...this.pot})
+
+		this.pot = undefined
+	}
+
+	private onComplete() {
+		const status = this.data.statuses.MEDICATED
+		if (!status) { return }
```
ackwell, line 54: don't think this should be necessary, you're accessing the data directly rather than executing a search

src/parser/core/modules/Medicated.ts
```diff
@@ -0,0 +1,71 @@
...
+			type: 'statusRemove',
+			target: this.parser.actor.id,
+			status: this.data.statuses.MEDICATED.id,
+		}, this.offMedication)
+
+		this.addEventHook('complete', this.onComplete)
+	}
+
+	private onMedication(event: Events['statusApply']) {
+		this.pot = {start: event.timestamp - this.parser.pull.timestamp}
```
ackwell, line 38: I'd avoid storing the data in timeline-specific format. Preference storing event timestamps, and adjusting for timeline display values when actually adding to the timeline. Timeline will eventually be adjusted to handle proper event timestamps.

src/parser/core/modules/Medicated.ts
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	private offMedication(event: Events['statusRemove']) {
+		// Pre-pull should handle if we're missing a start time
+		if (!this.pot) {
+			this.pot = {start: this.parser.pull.timestamp}
+		}
```
ackwell, line 45: Does this ever trip? Pretty sure the prepull adapter will ensure there's a `statusApply`.

Also, for the _current_ pre-zeroed values, this should be 0, though it's the correct value when storing event timestamps.

src/parser/core/modules/Medicated.ts
```diff
@@ -0,0 +1,71 @@
...
+		if (this.pots.length > 0) {
+			const row = new SimpleRow({
+				label: 'Medicated',
+				order: -99,
+			})
+
+			this.pots.forEach(pot => row.addItem(new StatusItem({
+				status: status,
+				start: pot.start,
+				end: pot.end ?? (this.parser.pull.timestamp + this.parser.pull.duration),
```
ackwell, line 65: timeline is currently zeroed, the rightmost X value is just `this.parser.pull.duration`

AzariahAeroborn (dismissed):

AzariahAeroborn (approved):

PR #1127, opened 2021-04-06
Add resource graphs to core gauge + rewrite MCH gauge to use core

yumiyafangirl:
I know this is going to get rewritten eventually, but might as well add an initial implementation of resource graphing to core gauge and move MCH there to make the job a little easier in the future.

(can't move everything to an Analyser immediately, since some other gauge implementations depend on modules that haven't been ported yet)

image


src/parser/core/modules/Gauge/CounterGauge.ts
```diff
@@ -28,13 +32,21 @@ export interface CounterChartOptions {
 	color?: string | Color
 }
 
+export interface ResourceGraphOptions {
+	/** Label to display on the tooltip. */
+	label: string
+	/** Colour to draw the resource graph in. */
+	colour: string | Color
```
yumiyafangirl, line 39: ResourceGraphs uses `colour`, but gauge uses `color` 😰

PR #1129, opened 2021-04-08
Audit & remove look-ahead downtime checking

ackwell:
Title. Tin. Etc.

This PR audits every call to `downtime`, `unableToAct`, and `invuln` that I could find, fixing any logic that relied on their ability to "see the future" due to normalisation.

I've also put in some errors if anything else tries to do it, which'll hopefully be caught by sentry if I missed something.

I'm not going to pretend I know the intricacies of every job. I took care to avoid changing logic too much, but it's impossible to avoid in some cases. I may well have broken your job. Please check the code for your modules!

I'll be adding some comments below shortly.

### DIFF
--- a/src/parser/core/modules/CooldownDowntime.tsx
+++ b/src/parser/core/modules/CooldownDowntime.tsx
@@ -188,14 +188,14 @@ export abstract class CooldownDowntime extends Module {
 		if (gRep.cooldown === undefined) {
 			return 0
 		}
-		const maxCharges = gRep.charges || 1
+		const maxCharges = gRep.charges ?? 1
 
 		// Skill with charges get their allowed downtime from the charge build up time,
 		// so ignore the value on the group object
-		const step = gRep.cooldown * 1000 + ((maxCharges > 1) ? 0 : (group.allowedAverageDowntime || this.defaultAllowedAverageDowntime))
+		const step = gRep.cooldown * 1000 + ((maxCharges > 1) ? 0 : (group.allowedAverageDowntime ?? this.defaultAllowedAverageDowntime))
 
-		const gResets = this.resets.get(group) || []
-		const gUsages = (this.usages.get(group) || [])
+		const gResets = this.resets.get(group) ?? []
+		const gUsages = (this.usages.get(group) ?? [])
 		const dtUsages = gUsages
 			.filter(u => this.downtime.isDowntime(u.timestamp))
 			.map(u => this.downtime.getDowntimeWindows(u.timestamp)[0])
@@ -206,11 +206,12 @@ export abstract class CooldownDowntime extends Module {
 		this.debug(`Checking downtime for group ${gRep.name} with default first use ${group.firstUseOffset} and step ${step} and ${maxCharges} charges`)
 		let charges = maxCharges
 		let count = 0
-		const expectedFirstUseTime = this.parser.eventTimeOffset + (group.firstUseOffset || this.defaultFirstUseOffset)
+		const expectedFirstUseTime = this.parser.eventTimeOffset + (group.firstUseOffset ?? this.defaultFirstUseOffset)
 		const actualFirstUseTime = gUsages[0]
+		const pullEndTimestamp = this.parser.pull.duration + this.parser.eventTimeOffset
 
 		let currentTime = expectedFirstUseTime
-		if ((group.firstUseOffset || 0) < 0 && maxCharges === 1) {
+		if ((group.firstUseOffset ?? 0) < 0 && maxCharges === 1) {
 			// check for pre-fight usages, which cause synthesized usage events
 			// that will have timestamps that don't accurartely indicated when
 			// exactly they were used pre-fight
@@ -232,7 +233,7 @@ export abstract class CooldownDowntime extends Module {
 			currentTime = Math.min(actualFirstUseTime.timestamp, expectedFirstUseTime)
 		}
 
-		while (currentTime - this.parser.eventTimeOffset < this.parser.pull.duration) {
+		while (currentTime < pullEndTimestamp) {
 			// spend accumulated charges
 			count += charges
 			this.debug(`Expected ${charges} usages at ${this.parser.formatTimestamp(currentTime)}. Count: ${count}`)
@@ -262,7 +263,11 @@ export abstract class CooldownDowntime extends Module {
 				gResets.shift()
 			}
 
-			while (charges < maxCharges && this.downtime.isDowntime(currentTime)) {
+			while (
+				currentTime < pullEndTimestamp
+				&& charges < maxCharges
+				&& this.downtime.isDowntime(currentTime)
+			) {
 				this.debug(`Saving charge during downtime at ${this.parser.formatTimestamp(currentTime)}. ${charges} charges stored`)
 
 				const window = this.downtime.getDowntimeWindows(currentTime)[0]
@@ -277,7 +282,10 @@ export abstract class CooldownDowntime extends Module {
 			}
 
 			// full charges were built up during a downtime.  Move to the end of the downtime to spend charges.
-			if (this.downtime.isDowntime(currentTime)) {
+			if (
+				currentTime < pullEndTimestamp
+				&& this.downtime.isDowntime(currentTime)
+			) {
 				const window = this.downtime.getDowntimeWindows(currentTime)[0]
 				this.debug(`Downtime detected at ${this.parser.formatTimestamp(currentTime)} in window from ${this.parser.formatTimestamp(window.start)} to ${this.parser.formatTimestamp(window.end)}`)
 
--- a/src/parser/core/modules/CooldownDowntime.tsx
+++ b/src/parser/core/modules/CooldownDowntime.tsx
@@ -188,7 +188,8 @@ export abstract class CooldownDowntime extends Module {
 		if (gRep.cooldown === undefined) {
 			return 0
 		}
-		const maxCharges = gRep.charges ?? 1
+		// 0 charges is nonsensical at this point, default up to 1.
+		const maxCharges = gRep.charges || 1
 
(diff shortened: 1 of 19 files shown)


src/parser/core/modules/CooldownDowntime.tsx
```diff
@@ -188,14 +188,14 @@ export abstract class CooldownDowntime extends Module {
 		if (gRep.cooldown === undefined) {
 			return 0
 		}
-		const maxCharges = gRep.charges || 1
+		const maxCharges = gRep.charges ?? 1
```
ackwell, line 191: numeric `0` is falsey, swapping to `??` to prevent unlikely edge cases

src/parser/core/modules/CooldownDowntime.tsx
```diff
@@ -262,7 +263,11 @@ export abstract class CooldownDowntime extends Module {
 				gResets.shift()
 			}
 
-			while (charges < maxCharges && this.downtime.isDowntime(currentTime)) {
+			while (
+				currentTime < pullEndTimestamp
+				&& charges < maxCharges
+				&& this.downtime.isDowntime(currentTime)
+			) {
```
ackwell, line 270: added a check to this and the `if` below to prevent the `currentTime` trying to overrun the end of the pull

cc @AzariahAeroborn i think i followed the logic enough this should be safe, but more eyes would be nice.

src/parser/core/modules/DoTs.js
```diff
@@ -75,12 +75,20 @@ export default class DoTs extends Module {
...
+				event.timestamp + this._statusDuration[statusId] + clip,
+				this.parser.eventTimeOffset + this.parser.pull.duration,
+			),
+			({timestamp}) => {
+				clip -= this.invuln.getInvulnerableUptime('all', event.timestamp, timestamp)
+
+				// Capping clip at 0 - less than that is downtime, which is handled by the checklist requirement
+				this._clip[statusId] = (this._clip[statusId] || 0) + Math.max(0, clip)
+			},
+		)
```
ackwell, line 91: core dots as well as a few old modules with similar logic: `this._clip` does not get _read_ outside `onComplete`, so timestamp hooking the invuln check timestamp (maxed at end of fight) allows us to get the same data as a look behind, rather than ahead.

src/parser/core/modules/EntityStatuses.ts
```diff
@@ -162,12 +162,18 @@ export class EntityStatuses extends Module {
...
+				this.debug(`Updating status event for status ${statusInfo.name}.  Adding ${statusInfo.duration} seconds, effective end time set to ${this.parser.formatTimestamp(end, 1)}`)
 			} else {
 				this.debug(`No matching status duration information found for status ${statusEvent.ability.guid}, setting to end of fight so invuln detection can clip the end to when the target went untargetable`)
-				statusEvent.end = this.parser.pull.duration + this.parser.eventTimeOffset
+				end = this.parser.currentTimestamp
+			}
+
+			statusEvent = {
+				...statusEvent,
+				end,
```
ackwell, line 176: While this is primarily called at the end of the fight, tightening usage anyway. I'm limiting to the current timestamp, rather than the end of the fight, which could be in the future.

I'm also shallow cloning the `end` in so it doesn't pollute potential other consumers with this data.

src/parser/core/modules/Invulnerability.tsx
```diff
@@ -549,7 +549,16 @@ export class Invulnerability extends Module {
...
 
-	getInvulns(
+	// Public interface for invuln windows that ensures consumers can't look ahead to the future
+	getInvulns = (...args: Parameters<typeof Invulnerability.prototype.getInvulnsInternal>) =>
+		this.getInvulnsInternal(...args)
+			.filter(window => window.start <= this.parser.currentTimestamp)
+			.map(window => ({
+				...window,
+				end: Math.min(window.end, this.parser.currentTimestamp),
+			}))
```
ackwell, line 559: we need the full data set for normalisation purposes on this current impl, so I've kept the logic as-is, and added this wrapper which exposes only what's technically possible during the main run.

src/parser/jobs/brd/modules/PitchPerfect.js
```diff
@@ -464,7 +468,11 @@ export default class PitchPerfect extends Module {
 				stacksUsedInCurrentWM = 0
 				castsInCurrentWM = []
 			}
-			if (this.downtime.isDowntime(pp.lastTickOnEnemy + DOT_TICK_FREQUENCY + ANIMATION_LOCK)) {
+			const nextPPTick = pp.lastTickOnEnemy + DOT_TICK_FREQUENCY + ANIMATION_LOCK
+			if (
+				nextPPTick > this.parser.eventTimeOffset + this.parser.pull.duration
+				|| this.downtime.isDowntime(nextPPTick)
+			) {
```
ackwell, line 475: figures the end of the fight is about the same as downtime as far as forgiven early PPs goes

src/parser/jobs/dnc/modules/DirtyDancing.tsx
```diff
@@ -197,9 +197,12 @@ export default class DirtyDancing extends Module {
 		dance.end = finisher.timestamp
 
 		// Count dance as dirty if we didn't get the expected finisher, and the fight wouldn't have ended or been in an invuln window before we could have
-		if (finisher.ability.guid !== dance.expectedFinishId && !(this.parser.fight.end_time < dance.expectedEndTime || this.invuln.isInvulnerable('all', dance.expectedEndTime))) {
-			dance.dirty = true
+		if (finisher.ability.guid !== dance.expectedFinishId && dance.expectedEndTime <= this.parser.eventTimeOffset + this.parser.pull.duration) {
+			this.addTimestampHook(dance.expectedEndTime, ({timestamp}) => {
+				dance.dirty = !this.invuln.isInvulnerable('all', timestamp)
+			})
```
ackwell, line 203: dance dirty status isn't read until end of fight.
dance overshooting end of fight is fine, so narrowing that case out and just hooking cases where it wasn't to check invuln

src/parser/jobs/pld/modules/FightOrFlight.tsx
```diff
@@ -173,13 +166,20 @@ export default class FightOrFlight extends Module {
...
-	private onRemoveFightOrFlight() {
-		if (!this.fofState.isRushed) {
+	private onRemoveFightOrFlight(event: BuffEvent) {
+		// If the enemy is untargetable at the end of FoF, the player was rushed - forgive missed hits.
+		// While technically this will never be called beyond the end of the fight, let's be really sure
+		const wasRushed = event.timestamp >= this.parser.eventTimeOffset + this.parser.pull.duration
+			|| this.invuln.isInvulnerable('all', event.timestamp)
+			|| this.invuln.isUntargetable('all', event.timestamp)
+
+		if (!wasRushed) {
```
ackwell, line 176: this module was storing the `goringTooClose` straight in the `errorResult`.
given it considers both pre-downtime _and_ pre-end-of-fight as rush, it's safe to wait for the removal of the buff, when rush is checked, and check then, rather than the predicted duration in advance

src/parser/jobs/smn/modules/Ruin2.js
```diff
@@ -60,8 +59,8 @@ export default class Ruin2 extends Module {
 		// Calc the time in the GCD that the boss can't be targeted - R2ing before an invuln to prevent an R3 cancel is good
 		const invulnTime = this.invuln.getUntargetableUptime(
 			'all',
+			this._lastGcd.timestamp,
 			event.timestamp,
-			event.timestamp + this.gcd.getEstimate(),
```
ackwell, line 64: this was a weird one, enough that I was doubting myself.
previous logic was checking untarget for the current GCD _forward_, to permit R2 on the _previous_ GCD. This fixes it to check untarget time for the GCD that was actually an R2.

src/parser/jobs/dnc/modules/DirtyDancing.tsx
AkairyuGestalter: Yep, this change looks good to me 👍

src/parser/core/modules/CooldownDowntime.tsx
AzariahAeroborn: A skill with 0 charges is non-sensical, from the perspective of this module, so defaulting 0 maxCharges back to 1 was intended.

src/parser/core/modules/CooldownDowntime.tsx
AzariahAeroborn: The later if probably doesn't need it, since the `while` loop this is all running within (start at 236) is already checking that condition, but I don't think it will hurt anything to add it to this inner while and might cover some oddities around end of fight downtime if the cast locks like e12s wind up triggering downtime detection.

src/parser/core/modules/CooldownDowntime.tsx
ackwell: ah, kk. will fix and comment so i don't do that again

src/parser/core/modules/CooldownDowntime.tsx
ackwell: the `while` is incrementing `currentTime` - i'm imagining there's an instance where it'll get out of that while loop with `currentTime > pullEndTime`. if it does, `isDowntime`'ll throw.

src/parser/jobs/smn/modules/Ruin2.js
KelosPelmand: This change looks correct, though I was also doubting it at first as well.

ebshimizu (commented):
DRG looks good

ayyaruq (approved):
MNK, DRK, and WAR don't seem affected, LGTM

PR #1131, opened 2021-04-09
DNC - Convert procs to TypeScript (plus 2 fixes)

AkairyuGestalter:
- Converted Procs.js to .tsx
- Fixed a display problem in the timeline, proc overwrites weren't showing the new status application
- Stopped counting procs lost on death as 'dropped' since we're already dinging them for dying


AkairyuGestalter:
Dropping this PR, since I'm superseding it with a Core procs solution, that also includes the fixes I developed here.

PR #1132, opened 2021-04-10
BLM - Convert Gauge to TypeScript

AkairyuGestalter:
- Change Gauge module to TypeScript
- Create proper BLM Gauge Event interface
- Add BLM Gauge Event to Legacy Events Repository (3/4 deps are on old system, not ready to port to Analyser yet)
- Updated BLM Gauge Event consumers to use standardized event
- Refactored Astral/Umbral and Polyglot timers to use timestamp hooks for improved accuracy

Should be no apparent changes to the end user


AkairyuGestalter:
Do not merge until after #1133, I will need to port the changes in Gauge.js over to the new Gauge.tsx after that moves into master

AkairyuGestalter:
Branch is caught up with the UTA changes introduced on #1133, so it should be good to merge now after review

AkairyuGestalter:
Closing PR since I'll be fully converting BLM gauge to Analyser now that its deps have all been converted in the interim. Will reopen when that's ready

PR #1135, opened 2021-04-13
Update SCH for 5.5 support

redshadowhero:
I mean, it's 100 mp cost. We don't even do mp-based analysis on SCH.

### DIFF
--- a/src/parser/jobs/sch/index.js
+++ b/src/parser/jobs/sch/index.js
@@ -18,7 +18,7 @@ export default new Meta({
 	</>,
 	supportedPatches: {
 		from: '5.0',
-		to: '5.4',
+		to: '5.5',
 	},
 	contributors: [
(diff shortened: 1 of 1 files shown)


redshadowhero:
Depends on #1136; please merge that one first.

ayyaruq (approved):
LGTM, can merge once the 5.5 patch entry is in, thanks for the PR!

src/parser/jobs/sch/index.js
```diff
@@ -27,6 +27,11 @@ export default new Meta({
 		{user: CONTRIBUTORS.NIV, role: ROLES.DEVELOPER},
 	],
 	changelog: [{
+		date: new Date('2020-4-13'),
```
ayyaruq, line 30:
```suggestion
		date: new Date('2021-04-13'),
```

ayyaruq (changes requested):
Let's not live in the past.

ayyaruq (approved):

PR #1136, opened 2021-04-13
Add patch 5.5

ayyaruq:
Layers are empty however DRG and WAR need the actions layer, and MNK needs the statuses one. SCH doesn't appear to use the mpCost so that won't need a layer update I guess.

### DIFF
--- a/src/data/ACTIONS/layers/patch5.5.ts
+++ b/src/data/ACTIONS/layers/patch5.5.ts
@@ -1,0 +1,9 @@
+import {Layer} from 'data/layer'
+import {ActionRoot} from '../root'
+
+// tslint:disable:no-magic-numbers
+
+export const patch550: Layer<ActionRoot> = {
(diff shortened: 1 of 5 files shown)


src/data/ACTIONS/layers/patch5.5.ts
```diff
@@ -0,0 +1,9 @@
+import {Layer} from 'data/layer'
+import {ActionRoot} from '../root'
+
+// tslint:disable:no-magic-numbers
```
ackwell, line 4: don't need this, we don't even have tslint any more. can you nuke it in one of your other prs?

ackwell (approved):

src/data/ACTIONS/layers/patch5.5.ts
ayyaruq: Yeah no worries

PR #1139, opened 2021-04-13
Add mitigation timeline

wxiaoying:
Shows mitigations used in timeline (default collapsed). Also adds some mitigations that were missing.

### DIFF
--- a/src/data/STATUSES/root/PLD.ts
+++ b/src/data/STATUSES/root/PLD.ts
@@ -60,6 +60,12 @@ export const PLD = ensureStatuses({
 		icon: 'https://xivapi.com/i/012000/012512.png',
 		duration: 18,
 	},
+	ARMS_UP: {
+		id: 1176,
+		name: 'Arms Up',
+		icon: 'https://xivapi.com/i/012000/012512.png',
+		duration: 18,
+	},
 	DIVINE_VEIL: {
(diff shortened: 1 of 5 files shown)


ackwell (changes requested):
Heyo! Thanks for the contribution.

It looks like this is using the old `Module` system - while a vast majority of the codebase is still using this system, we're currently undergoing work to port analysis onto the new `Analyser` system.

As a result, _especially_ for `parser/core`, I'm not going to be accepting _new_ modules written on the old system. (Doubly so if it's touching `AdditionalEvents`!)

The only dependency here that you might be missing on said new system is the `isActive` check, but bar that most stuff should port across.

Please feel free to drop past the discord server to chat if you've got any questions on how to get started &c!

wxiaoying:
Okay, that sounds good to me, this is heavily copied from the raid buffs module, so I suppose I'll wait for that to be ported and revisit.

ayyaruq:
RaidBuffs has been migrated to Analyser now, so may want to look at this again now.

ackwell:
Heyo! This has sit dormant for a while sorry - we're doing some cleanup for the EW branch cut.
If you'd like to continue with this, feel free to open a new PR targeting the endwalker branch (with the new analyser system!)

PR #1149, opened 2021-04-15
MNK - assume 100% uptime when fist is unknown

yumiyafangirl:
Fixes this "bug"; since it's already assuming flame fisties for the full duration we might as well call it 100% uptime too.

image


ayyaruq (approved):
Remind me to clean up this module a bit, some of the conditionals are a bit ehhh. Looks good tho, thank.

PR #1126, opened 2021-04-06
DNC Gauge Overhaul

AkairyuGestalter:
The great smashification. Changes:
- Combined Esprit and Feathers modules into a single module
- Moved graph output into the timeline ResourceGraph section
- DRYed up the gauge generation/spend/graph smoothing code
- Converted module to analyser
- Heavy rework of hooks to more accurately simulate Esprit generation
- Assorted cleanup related to above

PR #1153, opened 2021-04-19
feat(pld): adds oath gauge usage module to the paladin analysis

pogramos:
Includes Oath Gauge usage with overcap metrics and suggestion regarding the usage of `sheltron`/ `intervention`
image
image


AkairyuGestalter (dismissed):
Hi there! Always good to see new contributors! Two main notes, the first of which you'll need to address, the second is potentially optional but would make things cleaner:

1. All brand new parser code ought to be using the new Analyser system that we're in the midst of converting to. Since this Oath Gauge file only depends on Suggestions, that should be a relatively easy switch to make. Alternatively, you might look into extending the core Gauge modules instead, so this file will ride along with them when those get switched over.
2. You may want to consider using the new-ish ResourceGraphs component to display the Oath Gauge graph, rather than rolling your own TimelineChart, since it puts it into the consolidated timeline view and makes it easier for users to see what's going on. If you decide to switch to extending the core Gauge, you should(?) get that by default. Otherwise, you could take a look at PR #1126 for an example of how I kept track of DNC's Feather and Esprit gauge history myself, and used that history to add to the resource graphs.

If you have any questions about any of that, feel free to drop by the Discord to chat!

pogramos:
> Hi there! Always good to see new contributors! Two main notes, the first of which you'll need to address, the second is potentially optional but would make things cleaner:
> 
> 1. All brand new parser code ought to be using the new Analyser system that we're in the midst of converting to. Since this Oath Gauge file only depends on Suggestions, that should be a relatively easy switch to make. Alternatively, you might look into extending the core Gauge modules instead, so this file will ride along with them when those get switched over.
> 2. You may want to consider using the new-ish ResourceGraphs component to display the Oath Gauge graph, rather than rolling your own TimelineChart, since it puts it into the consolidated timeline view and makes it easier for users to see what's going on. If you decide to switch to extending the core Gauge, you should(?) get that by default. Otherwise, you could take a look at PR #1126 for an example of how I kept track of DNC's Feather and Esprit gauge history myself, and used that history to add to the resource graphs.
> 
> If you have any questions about any of that, feel free to drop by the Discord to chat!

Hi!
regarding the first point, I've chosen to change `OathGauge` to extend from `CoreGauge` so that it would be (i hope) ready for future changes coming from that module.

and about the second point, I would prefer to postpone that for now. I will study more the component and its usage, and maybe i'll try to refactor it afterwards.

src/parser/jobs/pld/modules/Gauge.tsx
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	// HELPERS
+	private onDamage() {
+		this.oathGauge.modify(GAUGE_GAIN_AMOUNT)
+	}
+
+	private onConsumeOath() {
+		this.oathGauge.modify(-GAUGE_SPEND_AMOUNT)
+	}
```
ayyaruq, line 53: Do these really need to be separate functions? You can do this directly in the hook, with `() => this.oathGauge.modify(GAUGE_GAIN_AMOUNT)`.

src/parser/jobs/pld/modules/Gauge.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+	private onComplete() {
+		this.suggestions.add(new Suggestion({
+			icon: ACTIONS.SHELTRON.icon,
+			content: <Trans id="pld.gauge.waste.suggestion.content">
+					You should periodically use your gauge either on you with a <ActionLink {...ACTIONS.SHELTRON}/> or on your tank partner with an <ActionLink {...ACTIONS.INTERVENTION} /> in case you're off-tanking.
```
ayyaruq, line 59: We generally avoid telling people what to do and use more passive language, like "Try to periodically use" rather than "You should periodically use". Mostly a nit here, but it's one of the few UX things we really focus on how we present.

src/parser/jobs/pld/modules/Gauge.tsx
```diff
@@ -0,0 +1,68 @@
...
+import ACTIONS from 'data/ACTIONS'
+import JOBS from 'data/JOBS'
+import {dependency} from 'parser/core/Module'
+import {CounterGauge, Gauge as CoreGauge} from 'parser/core/modules/Gauge'
+import Suggestions, {SEVERITY, Suggestion} from 'parser/core/modules/Suggestions'
+import React from 'react'
+
+// Gauge numbers
+const GAUGE_GAIN_AMOUNT = 5
+const GAUGE_SPEND_AMOUNT = 50
```
ayyaruq, line 13: Might be cleaner to make this negative rather than inverting it in the gauge adjustment, it's easier to read and clearer to maintain.

src/parser/jobs/pld/modules/Gauge.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+	private oathGauge = this.add(new CounterGauge({
+		chart: {label: 'Oath Gauge', color: JOBS.PALADIN.colour},
+	}))
+
+	private oathConsumers = [
+		ACTIONS.SHELTRON.id,
+		ACTIONS.INTERVENTION.id,
+		ACTIONS.COVER.id,
+	]
```
ayyaruq, line 28: More of a personal thing but I wonder if it wouldn't be simpler to just make this a map with the earn/spend costs built in so you can have a single hook and ref it. The WAR gauge is a good example of this.

ayyaruq (changes requested):
Kinda spitballing for the most part. Request change is specifically for the Suggestion text. Rolling the hooks into a single gauge modifier hook, or doing the adjustments inline, are kinda counter to each other so it's kind of pick one or the other, I'd personally lean more towards the latter since everything has static costs anyway. Since you're not using the data dependency, oathConsumers can prob be made into a const also, or alternatively switch it to the data system. Happy to chat in discord since I'm not really giving you a decision either way and it's better to not leave this floating in the void.

src/parser/jobs/pld/modules/Gauge.tsx
pogramos: I don't like using nameless blocks tbh, that's why I used separate functions for each of them. I could change them if there's a convention that I must follow tho.

src/parser/jobs/pld/modules/Gauge.tsx
pogramos: Not sure if simpler, but cleaner (?)... either way, a good idea. I'm changing that.

src/parser/jobs/pld/modules/Gauge.tsx
pogramos: I'll remove it in favor of this change https://github.com/xivanalysis/xivanalysis/pull/1153#discussion_r621833815

AzariahAeroborn (approved):

PR #1156, opened 2021-04-26
GNB - Tinctures

yumiyafangirl:
As requested by balance tank_tc. Requires at least 1 blasting zone and 1 GF combo per tincture, which should pretty much always be achievable.

image

### DIFF
--- a/src/parser/jobs/gnb/modules/Tincture.tsx
+++ b/src/parser/jobs/gnb/modules/Tincture.tsx
@@ -1,0 +1,30 @@
+import {Trans} from '@lingui/react'
+import {SEVERITY} from 'parser/core/modules/Suggestions'
+import {Tincture as CoreTincture} from 'parser/core/modules/Tincture'
+import React from 'react'
+
+export default class Tincture extends CoreTincture {
+	buffAction = this.data.actions.INFUSION_STR
+
+	trackedActions = {
+		icon: this.data.actions.INFUSION_STR.icon,
+		actions: [
+			{
+				action: this.data.actions.WICKED_TALON,
+				expectedPerWindow: 1,
+			},
+			{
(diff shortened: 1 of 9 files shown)


src/parser/jobs/gnb/modules/Tincture.tsx
```diff
@@ -0,0 +1,30 @@
...
+import React from 'react'
+
+export default class Tincture extends CoreTincture {
+	buffAction = this.data.actions.INFUSION_STR
+
+	trackedActions = {
+		icon: this.data.actions.INFUSION_STR.icon,
+		actions: [
+			{
+				action: this.data.actions.WICKED_TALON,
```
ayyaruq, line 13: So if a user pots between Savage Claw and Wicket Talon, or early pot and it falls off after Wicked Talon without catching Eye Gouge, that's ok? I'm not entirely sure how this presents but might want to include the full continuation combo, maybe also DoTs; basically everything that goes into No Mercy at a minimum.

ayyaruq (commented):
Looks alright but I'm not sure about the rules. I get asking tanks_tc but this may need more discussion in xiva feedback channels rather than private channels in an unrelated server.

src/parser/jobs/gnb/modules/Tincture.tsx
yumiyafangirl: Requires modifications to core, to be addressed in https://github.com/xivanalysis/xivanalysis/issues/1158.

ayyaruq: DoTs are ignored for the same reason No Mercy is, namely the cooldown not lining up with 4:30 pots, the rest addressed in linked issue.

ayyaruq (approved):

PR #1159, opened 2021-04-28
[GNB]Convert NoMercy to core, add burst strike tracking to it, also some guide updates

Rjb99:
The main thing for this PR is that No Mercy(NM)has been converted to BuffWindow from Core.
the following has been added:
Checking if the user used the ideal amount of burst strikes
-Checking if Bloodfest was used and adjusting the ideal amount of burst strikes (currently set to 3)

I'm currently on the fence about Bloodfest NM however. I've set it to just increase by 1 as many of the fights this tier do not lend themselves to the 4/4 bloodfest NM window

https://images-ext-2.discordapp.net/external/gBj8QlvMmdi48tHcKfiaclXM4tZCnTiziE9KRDJnCro/https/i.imgur.com/o8hza9e.png?width=618&height=676

Many of the top GNBs are doing the modified one that only gets 3 instead. I feel that suggesting that there is a 4th possible use where there is not is more harmful than good and think it would be better to just let them hit the 4/3 whenever possible for bloodfest NMs

Also apparently the gamebreaker guide we linked to was deleted, so I changed the link to saltedXIV's guide and added Rin's no mercy window graphic. Also changed the suggestion on where to leave suggestions/feedback from Lhea's DMs to the XIVA discord.

Example output of the table using Rin's 100th precentile parse in o12s 
image


src/parser/jobs/gnb/modules/NoMercy.tsx
```diff
@@ -37,228 +34,87 @@ const EXPECTED_USES = {
...
 	static handle = 'nomercy'
 	static title = t('gnb.nomercy.title')`No Mercy Windows`
 
-	@dependency private suggestions!: Suggestions
-	@dependency private timeline!: Timeline
+	buffAction = ACTIONS.NO_MERCY
+	buffStatus = STATUSES.NO_MERCY
 
-	private noMercyWindows: NoMercyState[] = []
+	rotationTableNotesColumnHeader = <Trans id="gnb.nomercy.notes.header"> Bloodfest Used </Trans>
```
ayyaruq, line 44: nbd but what's with the extra spaces before/after the string here?

src/parser/jobs/gnb/modules/NoMercy.tsx
```diff
@@ -37,228 +34,87 @@ const EXPECTED_USES = {
...
-
-export default class NoMercy extends Module {
+export default class NoMercy extends BuffWindowModule {
 	static handle = 'nomercy'
 	static title = t('gnb.nomercy.title')`No Mercy Windows`
 
-	@dependency private suggestions!: Suggestions
-	@dependency private timeline!: Timeline
+	buffAction = ACTIONS.NO_MERCY
+	buffStatus = STATUSES.NO_MERCY
```
ayyaruq, line 42: Should these use the data system?

src/parser/jobs/gnb/modules/NoMercy.tsx
```diff
@@ -37,228 +34,87 @@ const EXPECTED_USES = {
...
 
-			// Pre-process on the number of certain things we did
-			if (action.onGcd) {
-				lastNoMercy.numGcds++
-			}
+		suggestionContent: <Trans id="gnb.nomercy.suggestions.expected-uses.content">
+			Watch your uses of certain abilities during <ActionLink {...ACTIONS.NO_MERCY} />. Under ideal conditions, you should
+				be using <ActionLink {...ACTIONS.SONIC_BREAK} />, a full <ActionLink {...ACTIONS.GNASHING_FANG} /> combo, and all of
+				your off-GCD skills - <ActionLink {...ACTIONS.BLASTING_ZONE} />, <ActionLink {...ACTIONS.BOW_SHOCK} />, and at least one
+				charge of <ActionLink {...ACTIONS.ROUGH_DIVIDE} /> - under the buff duration.
```
ayyaruq, line 93: Ultra nit, are these hyphens really necessary? First could be a colon and last could just not be there at all.

src/parser/jobs/gnb/modules/NoMercy.tsx
```diff
@@ -37,228 +34,87 @@ const EXPECTED_USES = {
...
-						},
-						rotation: window.rotation,
-					})
-				})
-			}
-			onGoto={this.timeline.show}
-		/>
+	getBuffWindowNotes(buffWindow: BuffWindowState) {
+		return buffWindow.rotation.find(cast => cast.ability.guid === ACTIONS.BLOODFEST.id) ?
+			<Trans id="gnb.nomercy.chart.notes.yes"> Yes </Trans> : <Trans id = "gnb.nomercy.chart.notes.no"> No </Trans>
```
ayyaruq, line 117: Again what's up with the spacing? I wonder if we could consider using a tick/cross or something but then localising would be weird since idk how crowdin would handle unicode emojis (JP needs a circle instead of tick).

src/parser/jobs/gnb/modules/NoMercy.tsx
```diff
@@ -37,228 +34,87 @@ const EXPECTED_USES = {
...
-		}
-	}
+		if (action.action.id === ACTIONS.BURST_STRIKE.id) {
 
-	private onComplete() {
-		// Exclude any window at the end of a fight which could not possibly have gotten all of its GCDs
-		const missedGcds = this.noMercyWindows
-			.filter(window => !window.isRushing)
-			.reduce((sum, window) => sum + Math.max(0, EXPECTED_USES.GCD - window.numGcds), 0)
+			if (buffWindow.rotation.find(cast => cast.ability.guid === ACTIONS.BLOODFEST.id) && (this.getBaselineExpectedTrackedAction(buffWindow, action) === 2)) {
```
ayyaruq, line 103: Could you break this up a bit? It's longer than me blogging for an hour in discord.

src/parser/jobs/gnb/modules/NoMercy.tsx
```diff
@@ -37,228 +34,87 @@ const EXPECTED_USES = {
...
 
-	private onRemoveNoMercy(event: BuffEvent) {
-		const lastNoMercy = this.lastNoMercy
+	changeExpectedTrackedActionClassLogic(buffWindow: BuffWindowState, action: BuffWindowTrackedAction) {
 
-		if (lastNoMercy != null) {
-			lastNoMercy.end = event.timestamp
-		}
-	}
+		if (action.action.id === ACTIONS.BURST_STRIKE.id) {
```
ayyaruq, line 101: Nit, I'd consider changing the name of the param here if possible just so it's a bit less weird to read than `action.action`.

ayyaruq (changes requested):
Couple of nits, mostly around using the old ACTIONS system over the newer Data system, and that expected adjustment being a bit hairy. I'm surprised linter didn't rage at line length.

src/parser/jobs/gnb/modules/NoMercy.tsx
Rjb99: The spaces just made it easier to read, but I can remove them

src/parser/jobs/gnb/modules/NoMercy.tsx
Rjb99: At the moment we've got some jobs using green checks and red x's. and others just using Yes and No. some uniformity might be useful

src/parser/jobs/gnb/modules/NoMercy.tsx
Rjb99: changed to const Id

src/parser/jobs/gnb/modules/NoMercy.tsx
Rjb99: I pressed the enter key

src/parser/jobs/gnb/modules/NoMercy.tsx
Rjb99: the other implementations of BuffWindow don't use data, so I didn't.

Is there a reason to use data over just using actions/statuses directly here?

src/parser/jobs/gnb/modules/NoMercy.tsx
ayyaruq: They're fine if that's the case, just wondered if it was accidental.

src/parser/jobs/gnb/modules/NoMercy.tsx
ayyaruq: You just want the ID so not really. The main benefit of the data system is it more properly manages layers for patch changes, but you literally just want the ID here anyway as far as I can tell.

src/parser/jobs/gnb/modules/NoMercy.tsx
ayyaruq: We prob should really standardise this, there's a few other things that are kinda mix and match (like some jobs calling stuff using buff window as "$thing Windows" and others just "$thing".

src/parser/jobs/gnb/modules/NoMercy.tsx
ayyaruq: I'd still rename the param (and leading upper case names like this trigger me ngl), it's awkward to read either way.

ayyaruq (approved):

PR #1172, opened 2021-05-15
Core: Convert CastTime to Analyser system

AkairyuGestalter:
- Change CastTime to Typescript
- Convert to Analyser
- Provide auto-deprecated fflogs functions
- Update consumers to use fflogs functions


src/parser/core/modules/CastTime.ts
```diff
@@ -0,0 +1,109 @@
...
+		// Get any cast time modifiers active when the event took place
+		const matchingTimes = this.castTimes.filter(ct =>
+			(ct.actions === 'all' || ct.actions.includes(actionId)) &&
+			ct.start <= timestamp &&
+			(ct.end === null || ct.end >= timestamp),
+		)
+
+		// Mimicking old logic w/ the undefined. Don't ask.
+		const action = this.data.getAction(actionId)
+		const defaultCastTime = action ? action.castTime : undefined
```
AzariahAeroborn, line 90: Given that this is now in TS, can use a null coalesce to make this a little cleaner:
```suggestion
		const defaultCastTime = action?.castTime
```

src/parser/core/modules/CastTime.ts
```diff
@@ -0,0 +1,109 @@
...
+
+		// If there were no modifiers, just use the default (or if the default comes back undefined, shouldn't happen but eh)
+		if (!matchingTimes.length || defaultCastTime === undefined) {
```
AzariahAeroborn, line 93: Project style guide is to use `== null` (or != null as appropriate) for null/undefined checks.  Would also dovetail well with the suggested null coalesce cleanup for defaultCastTime above.
```suggestion
		if (!matchingTimes.length || defaultCastTime == null) {
```

AzariahAeroborn (changes requested):

AzariahAeroborn (approved):

PR #1181, opened 2021-05-22
Create a Core Procs analyser and migrate existing modules

AkairyuGestalter:
It's a big one. Changes:
- Consolidate proc status tracking logic from BLM, DNC, RDM, and DRG into a single Analyser
- Provides APIs for inserting job-specific logic into the proc tracking pipeline
- Provides a default set of suggestions for subclasses
- Provides configuration options for the default suggestions
- Migrate the separate proc modules to subclasses of the core analyser

As much as is possible, this change should be mainly cosmetic to end users. RDM has a batch of more specific suggestion text to discretely handle the mana balance nature of their procs, and I think there were a few bugs in the individual modules that got fixed along the way, but this shouldn't by and large change anything for most parses.


src/parser/jobs/rdm/modules/Procs.tsx
```diff
@@ -0,0 +1,192 @@
...
+			icon: invulnFire > invulnStone ? this.data.actions.VERFIRE.icon : this.data.actions.VERSTONE.icon,
+			content: this.getInvulnProcContent(invulnFire, invulnStone),
+			tiers: SEVERITY_INVULN_PROCS,
+			value: invulnFire + invulnStone,
+			why: this.getInvulnProcWhy(invulnFire, invulnStone),
+		}))
+	}
+
+	output() {
+		const allInvulns = this.getUsagesForStatus(this.data.statuses.VERFIRE_READY.id).concat(this.getInvulnsForStatus(this.data.statuses.VERSTONE_READY.id)).sort((a, b) => a.timestamp - b.timestamp)
```
AzariahAeroborn, line 176: I suspect this should be getInvulnsForStatus instead, given the sheer number of invuln entries I'm seeing on random RDM logs I pull up. 🤔 

```suggestion
		const allInvulns = this.getInvulnsForStatus(this.data.statuses.VERFIRE_READY.id).concat(this.getInvulnsForStatus(this.data.statuses.VERSTONE_READY.id)).sort((a, b) => a.timestamp - b.timestamp)
```

src/parser/jobs/drg/modules/Procs.tsx
```diff
@@ -0,0 +1,75 @@
...
+			value: droppedMirage,
+			why: <Trans id="drg.procs.suggestions.mirage-dropped.why">
+				You dropped <Plural value={droppedMirage} one="# Mirage Dive proc" other="# Mirage Dive procs"/>.
+			</Trans>,
+		}))
+
+		this.suggestions.add(new TieredSuggestion({
+			icon: this.data.actions.MIRAGE_DIVE.icon,
+			content: <Trans id="drg.procs.suggestions.mirage-overwritten.content">
+				Avoid casting <ActionLink {...this.data.actions.JUMP}/> and <ActionLink {...this.data.actions.SPINESHATTER_DIVE}/> when you already have a <StatusLink {...this.data.statuses.DIVE_READY}/> procs, as it overwrites them and can delay your Life of the Dragon windows and potentially cost you a lot of DPS.
```
AzariahAeroborn, line 63:
```suggestion
				Avoid casting <ActionLink {...this.data.actions.HIGH_JUMP}/> when you already have a <StatusLink {...this.data.statuses.DIVE_READY}/> procs, as it overwrites them and can delay your Life of the Dragon windows and potentially cost you a lot of DPS.
```

Per discussion in #dev-melee - looks like this never got updated for ShB, might as well finally fix it while we're touching it lol

AzariahAeroborn (changes requested):

src/parser/jobs/drg/modules/Procs.tsx
AkairyuGestalter: It's dead, Jim.

src/parser/jobs/rdm/modules/Procs.tsx
AkairyuGestalter: 🤦

AzariahAeroborn (approved):

PR #1182, opened 2021-05-22
Refactor Debug logging to a module that can be shared with Analysers

AzariahAeroborn:
This was originally done in support of #1180, but I pulled it out for ~~ease of review~~ simplicity of troubleshooting the type errors I had.

I verified against https://xivanalysis.com/fflogs/F8ZfDGXrkbajxhYR/19/1258 that the broken log triggers were still generating the same output with the change to Parser._gatherErrorContext.

### DIFF
--- a/src/parser/core/Loggable.ts
+++ b/src/parser/core/Loggable.ts
@@ -1,0 +1,42 @@
+import {seededColor} from 'utilities/color'
+
+// Helper types for debug
+type LogParameters = Parameters<typeof console.log>
+type DebugCallback = (opts: {
+	/** Log the provided arguments. */
+	log: (...data: LogParameters) => void,
+}) => void
+
+export abstract class Loggable {
+	/**
+	 * Set to `true` to enable debug mode for this loggable class, allowing the execution
+	 * of any calls to the debug method.
+	 */
+	static debug = false
+
+	/** Execute the provided callback if the loggable is in debug mode. */
+	protected debug(callback: DebugCallback): void
+	/** Log the provided arguments if the loggable is in debug mode. */
+	protected debug(...data: LogParameters): void
+	protected debug(...args: [DebugCallback] | LogParameters) {
+		const constructor = this.constructor as typeof Loggable
+		if (!constructor.debug || process.env.NODE_ENV === 'production') {
+			return
+		}
+
+		typeof args[0] === 'function'
+			? args[0]({log: this.debugLog})
+			: this.debugLog(...args)
+	}
+
+	private debugLog = (...data: LogParameters) => {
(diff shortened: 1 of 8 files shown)


src/parser/core/Loggable.ts
```diff
@@ -0,0 +1,42 @@
+import {seededColor} from 'utilities/color'
+
+// Helper types for debug
+type LogParameters = Parameters<typeof console.log>
+type DebugCallback = (opts: {
+	/** Log the provided arguments. */
+	log: (...data: LogParameters) => void,
+}) => void
+
+export abstract class Loggable {
```
ackwell, line 10:
```suggestion
export abstract class Debuggable {
```

I mean `log` is the only debug function atm but we could expand it out a bit in future. Few modules use it for debug semantics already i.e. `this.debug(() => addShitToTimeline())` for invuln's debug output

src/reportSources/legacyFflogs/__tests__/eventAdapter.ts
```diff
@@ -11,7 +11,7 @@ import {ReassignUnknownActorStep} from '../eventAdapter/reassignUnknownActor'
 // TODO: If we need to do this >1 times, work out a cleaner way of doing this.
 jest.mock('../eventAdapter/reassignUnknownActor')
 type ReassignUnknownActorStepParams = ConstructorParameters<typeof ReassignUnknownActorStep>
-const MockReassignUnknownActorStep = ReassignUnknownActorStep as jest.Mock<ReassignUnknownActorStep>
+const MockReassignUnknownActorStep = ReassignUnknownActorStep as unknown as jest.Mock<ReassignUnknownActorStep>
```
ackwell, line 14: i mean tests so who cares, but what was the error?

ackwell (approved):

src/reportSources/legacyFflogs/__tests__/eventAdapter.ts
AzariahAeroborn: It was a can't cast typescript error that specified "cast to unknown before casting to the Mock".  The cast for injectable.constructor in the Parser class was the same error, but I inferred from other similar code and the context of that code that changing to injectable.constructor seemed reasonable, and that didn't break anything else that I found.

ackwell: Yeah the other spot was probably a bug. Oops.

ackwell (approved):
gotta love auto dismiss 😬

ackwell (approved):
gotta.
love.
auto.
dismiss.

PR #1185, opened 2021-05-23
Change semantics of snapshot vs execute for damage/heal actions

AzariahAeroborn:
Update the new system to invert the snapshot -> damage/heal paradigm to be damage/heal -> execute instead, to more accurately match how the game data handles it and simplify the interface for future consumers (e.g. gauges)

### DIFF
--- a/src/event.ts
+++ b/src/event.ts
@@ -90,11 +90,11 @@ interface EventStatusRemove extends FieldsTargeted {
 	status: number
 }
 
-/** The server has snapshot the effects of an action on its target. */
-interface EventSnapshot extends FieldsTargeted {
+/** The server has confirmed the execution of an action on its target. */
+interface EventExecute extends FieldsTargeted {
 	/** XIV Action ID */
 	action: number
-	/** Unique numeric ID that will match the snapshot to each of the resulting effects. */
+	/** Unique numeric ID that will match the execution to each of the actions that caused it. */
 	sequence: number
 }
 
@@ -130,7 +130,7 @@ interface EventDamage extends FieldsTargeted {
 	amount: number
 	/** Amount of total damage that was overkill. */
 	overkill: number
-	/** Unique numeric ID that will match this damage to the snapshot it stems from. If omitted, no snapshot was performed (status ticks, etc). */
+	/** Unique numeric ID that will match this damage to its execution confirmation. If omitted, no confirmation occurred (status ticks, etc). */
 	sequence?: number
 	// TODO: Are these exclusive? Merge?
 	/** Source damage modifier. */
@@ -147,7 +147,7 @@ interface EventHeal extends FieldsTargeted {
 	amount: number
 	/** Amount of total healing that was overheal. */
 	overheal: number
-	/** Unique numeric ID that will match this heal to the snapshot it stems from. If omitted, no snapshot was performed (status ticks, etc). */
+	/** Unique numeric ID that will match this heal to its execution confirmation. If omitted, no confirmation occurred (status ticks, etc). */
 	sequence?: number
 	/** Source healing modifier. */
 	sourceModifier: SourceModifier
@@ -188,7 +188,7 @@ export interface EventTypeRepository {
 	action: EventAction
 	statusApply: EventStatusApply
 	statusRemove: EventStatusRemove
-	snapshot: EventSnapshot
+	execute: EventExecute
 	damage: EventDamage
 	heal: EventHeal
 	actorUpdate: EventActorUpdate
--- a/src/event.ts
+++ b/src/event.ts
@@ -130,7 +130,7 @@ interface EventDamage extends FieldsTargeted {
 	amount: number
 	/** Amount of total damage that was overkill. */
 	overkill: number
-	/** Unique numeric ID that will match this damage to its execution confirmation. If omitted, no confirmation occurred (status ticks, etc). */
+	/** Unique numeric ID that will match this damage to its execution confirmation. If omitted, no confirmation will occur (status ticks, etc). */
 	sequence?: number
 	// TODO: Are these exclusive? Merge?
 	/** Source damage modifier. */
@@ -147,7 +147,7 @@ interface EventHeal extends FieldsTargeted {
 	amount: number
 	/** Amount of total healing that was overheal. */
 	overheal: number
-	/** Unique numeric ID that will match this heal to its execution confirmation. If omitted, no confirmation occurred (status ticks, etc). */
+	/** Unique numeric ID that will match this heal to its execution confirmation. If omitted, no confirmation will occur (status ticks, etc). */
 	sequence?: number
 	/** Source healing modifier. */
 	sourceModifier: SourceModifier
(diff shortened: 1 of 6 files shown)


src/event.ts
```diff
@@ -130,7 +130,7 @@ interface EventDamage extends FieldsTargeted {
 	amount: number
 	/** Amount of total damage that was overkill. */
 	overkill: number
-	/** Unique numeric ID that will match this damage to the snapshot it stems from. If omitted, no snapshot was performed (status ticks, etc). */
+	/** Unique numeric ID that will match this damage to its execution confirmation. If omitted, no confirmation occurred (status ticks, etc). */
```
ackwell, line 133: nit; "no confirmation **will occur**", given it's looking forward rather than back

src/reportSources/legacyFflogs/__tests__/__snapshots__/eventAdapter.ts.snap
```diff
@@ -83,19 +83,25 @@ exports[`Event adapter individual events adapts calculateddamage 1`] = `
 Array [
   Object {
     "action": 3554,
-    "sequence": 16647,
```
ackwell, line 86: whoops

src/reportSources/legacyFflogs/eventAdapter/translate.ts
```diff
@@ -123,16 +122,18 @@ export class TranslateAdapterStep extends AdapterStep {
...
 	}
 
-	private adaptSnapshotEvent(event: DamageEvent | HealEvent): Array<Events['snapshot' | 'actorUpdate']> {
+	private adaptFFLogsEffectEvent(event: DamageEvent | HealEvent): Array<Events['execute' | 'damage' | 'heal' | 'actorUpdate']> {
 		// Calc events should all have a packet ID for sequence purposes. Let sentry catch outliers.
 		const sequence = event.packetID
 		if (sequence == null) {
-			throw new Error('Calculated damage event encountered with no packet ID.')
+			if (event.type === 'damage') { return this.adaptDamageEvent(event) }
+			if (event.type === 'heal') { return this.adaptHealEvent(event) }
```
ackwell, line 130: comment on this redirection to damage/heal is probably worthwhile (i.e. something about damage/heal without a sequence means it's a status which doesn't get executes, etc)

src/reportSources/legacyFflogs/eventAdapter/translate.ts
```diff
@@ -123,16 +122,18 @@ export class TranslateAdapterStep extends AdapterStep {
 		}
 	}
 
-	private adaptSnapshotEvent(event: DamageEvent | HealEvent): Array<Events['snapshot' | 'actorUpdate']> {
+	private adaptFFLogsEffectEvent(event: DamageEvent | HealEvent): Array<Events['execute' | 'damage' | 'heal' | 'actorUpdate']> {
```
ackwell, line 125: `adaptEffectEvent` or `adaptExecuteEvent` methinks. "fflogs" isn't meaningful in a fflogs specific file

src/reportSources/legacyFflogs/eventAdapter/translate.ts
```diff
@@ -123,16 +122,18 @@ export class TranslateAdapterStep extends AdapterStep {
...
 		// Calc events should all have a packet ID for sequence purposes. Let sentry catch outliers.
 		const sequence = event.packetID
 		if (sequence == null) {
```
ackwell, line 128: possibly worth checking that it _is_ coming from a status? it's safe to check the offset const in here.
not sure if there's a special cases that'd make this dangerous or not (ground effects maybe?)

src/reportSources/legacyFflogs/__tests__/__snapshots__/eventAdapter.ts.snap
```diff
@@ -83,19 +83,25 @@ exports[`Event adapter individual events adapts calculateddamage 1`] = `
 Array [
```
ackwell, line 83: doesn't look like there's any `execute`s coming through here - something funky?

src/event.ts
```diff
@@ -90,11 +90,11 @@ interface EventStatusRemove extends FieldsTargeted {
 	status: number
 }
 
-/** The server has snapshot the effects of an action on its target. */
-interface EventSnapshot extends FieldsTargeted {
+/** The server has confirmed the execution of an action on its target. */
+interface EventExecute extends FieldsTargeted {
 	/** XIV Action ID */
 	action: number
```
ackwell, line 96: hm. wonder if we need the action id here any more - it would need to be fabricated on any event source that isn't matching them up for us.

src/reportSources/legacyFflogs/__tests__/__snapshots__/eventAdapter.ts.snap
ackwell: Oh, might be because both of the example events happen to be statuses, haven't checked. Probably worth testing all 3 code paths for that.

src/reportSources/legacyFflogs/__tests__/__snapshots__/eventAdapter.ts.snap
AzariahAeroborn: This was an intended change as a result of my updating how PrePullAction synthed... before, the synth was spreading and was pulling forward fields that weren't in the typedef for an action.  Once I had to change how I got the action ID, I went ahead and just explicitly defined only the "right" fields and ditched the spread, thus effecting that test output.

ackwell: haha, yeah, i mean that as in "whoops, that shouldn't've been there in the first place"

src/reportSources/legacyFflogs/eventAdapter/translate.ts
AzariahAeroborn: Doton and Salted Earth are both being tracked as Status DOT ticks on FFLogs, so that sounds sane & safe to me.  Worst case I break prod again 🤣

src/event.ts
AzariahAeroborn: My thinking is to keep it for now, and if we run into it being undesirable with a future event source we can nuke it then.

src/reportSources/legacyFflogs/eventAdapter/translate.ts
```diff
@@ -122,17 +131,22 @@ export class TranslateAdapterStep extends AdapterStep {
...
 
-	private adaptFFLogsEffectEvent(event: DamageEvent | HealEvent): Array<Events['execute' | 'damage' | 'heal' | 'actorUpdate']> {
+	private adaptEffectEvent(event: DamageEvent | HealEvent): Array<Events['execute' | 'damage' | 'heal' | 'actorUpdate']> {
 		// Calc events should all have a packet ID for sequence purposes. Let sentry catch outliers.
 		const sequence = event.packetID
 		if (sequence == null) {
-			if (event.type === 'damage') { return this.adaptDamageEvent(event) }
-			if (event.type === 'heal') { return this.adaptHealEvent(event) }
-			throw new Error('FFLogs Effect event encountered with no packet ID, did not match to DoT or HoT')
+			const cause = resolveCause(event.ability.guid)
```
ackwell, line 138: ah good idea to use the resolveCause helper

ackwell (approved):

PR #1187, opened 2021-05-24
Updated CastTime API signature

AkairyuGestalter:
- Removed CastTime's singular set function in favor of:
  1. setTimeAdjustment: For setting adjustments that remove fixed amounts from the base cast time (ie. Lightspeed, DWT/FBT)
  2. setPercentageAdjustment: For setting adjustments that are multiplicative to the base cast time (ie. Ley Lines).
  3. setInstantCastAdjustment: Shorthand function for a '0%' setPercentageAdjustment (ie. Swiftcast, Triplecast, Dualcast)
- Updated forAction to handle calculating the modified cast time based on the cumulative time and percentage adjustments in effect at that timestamp
- Provided Module-space equivalent API calls for compatibility
- Attempted to document the snot out of the public APIs
- Updated current consumers of CastTime to use new API calls
- Bonus: Technically this should now accommodate cast time increases (slows) as well as reductions, should that ever need to be a thing


src/parser/core/modules/CastTime.ts
```diff
@@ -31,52 +32,128 @@ export default class CastTime extends Analyser {
...
+	 * setInstantCastAdjustment Module-compatibility function
+	 * @deprecated
+	 */
+	public setInstantCastAdjustmentFflogs(actions: number[] | 'all' = 'all', start: number = this.parser.currentTimestamp, end: number | null = null): number {
+		return this.set(actions, 'percentage', 0, this.parser.fflogsToEpoch(start), end ? this.parser.fflogsToEpoch(end) : null)
+	}
+	/**
+	 * Shorthand function for setting all casts to instant (ie. Swiftcast, Triplecast)
+	 * @actions The actions this adjustment applies to. Either an array of IDs, or the string 'all'. Defaults to 'all'
+	 * @param start
```
AzariahAeroborn, line 72:
```suggestion
	 * @param start The beginning of the adjustment time range. Defaults to the current epoch timestamp
	 * @param end The end of the adjustment time range. May be left null if the end of the range is not yet known
	 * @returns The index number within the cast time adjustments collection, can be used to reset/end this adjustment later
```

GitHub isn't letting me do a multi-line comment right, but I think you had some line drift on this doc comment

AzariahAeroborn (changes requested):

src/parser/core/modules/CastTime.ts
AkairyuGestalter: Copy paste, thou hast failed me

src/parser/core/modules/CastTime.ts
```diff
@@ -90,20 +167,39 @@ export default class CastTime extends Analyser {
...
+				if (ct.type === 'percentage' || ct.adjustment < 0) { return increase }
+				if (ct.adjustment > increase) { return ct.adjustment }
+				return increase
+			}, 0)
+		// Get the total percentage adjustment
+		const percentageAdjustment = matchingTimes.reduce(
+			(adjustment, ct) => {
+				if (ct.type === 'time') { return adjustment }
+				return adjustment * ct.adjustment
+			}, 1)
```
ackwell, line 193: this effectively loops the entire array three times, ignoring a different subset each time.
would be cleaner to use a single `for` loop with some conditional on type and adjustment polarity

src/parser/core/modules/CastTime.ts
```diff
@@ -90,20 +167,39 @@ export default class CastTime extends Analyser {
 		const defaultCastTime = action?.castTime
 
 		// If there were no modifiers, just use the default (or if the default comes back undefined, shouldn't happen but eh)
-		if (!matchingTimes.length || defaultCastTime == null) {
+		if (!matchingTimes.length || !defaultCastTime) {
```
ackwell, line 170: if this is to catch the 0 case as well, i'd be inclined to include that as a separate conditional check.  it's a bit all over the place in old code, but more and more i'm trying to move away from assuming falsiness of `null`/`undefined` (and others as well), as it makes intention clearer when checking two values (i.e. here, it's checking both "is it defined", and "if it's defined, but it's zero")

src/parser/core/modules/CastTime.ts
```diff
@@ -7,7 +7,8 @@ import {Data} from './Data'
 
 export interface CastTimeAdjustment {
 	actions: number[] | 'all',
-	castTime: number,
+	type: 'time' | 'percentage'
+	adjustment: number,
 	start: number,
 	end: number | null
```
ackwell, line 13: realise it's pre-existing, but this should probably be `end?: number`. will tie in well with other suggestions about function sigs

src/parser/core/modules/CastTime.ts
```diff
@@ -31,52 +32,128 @@ export default class CastTime extends Analyser {
...
 	}
 
-	/** @deprecated */
-	public setFflogs(actions: number[] | 'all', castTime: number, start: number = this.parser.currentTimestamp, end: number | null = null): number {
-		return this.set(actions, castTime, this.parser.fflogsToEpoch(start), end ? this.parser.fflogsToEpoch(end) : null)
+	/**
+	 * setTimeAdjustment Module-compatibility function
+	 * @deprecated
+	 */
+	public setTimeAdjustmentFflogs(actions: number[] | 'all', adjustment: number, start: number = this.parser.currentTimestamp, end: number | null = null): number {
```
ackwell, line 47: oof, i wrote these sigs originally didn't i. that null 😬.

uuh, this should probably be `start: number = curTs, end?: number`.

---

`Adjustment` seems a bit redundant on these methods but i'm not sure which way it should be taken, given `setTime` is inaccurate.

src/parser/core/modules/CastTime.ts
```diff
@@ -31,52 +32,128 @@ export default class CastTime extends Analyser {
...
+	 */
 	public forFflogsEvent(event: CastEvent): number | undefined {
 		return this.forAction(event.ability.guid, this.parser.fflogsToEpoch(event.timestamp))
 	}
+	/**
+	 * Returns the effective cast time for the specified event
+	 * @param event The event in question
+	 * @returns The actual cast time, either as the default, or the modified time if any modifiers were in effect. Returns undefined if the action cannot be determined, or has no default cast time
+	 */
 	public forEvent(event: Events['action']): number | undefined {
```
ackwell, line 140: random thought in passing; while i don't really see the point in this method any more, probably worth accepting `prepare` as well if we keep it.

AkairyuGestalter: Weaving's using the (currently fflogs-version) forEvent function to check if the leading GCD event for a weave was instant. Probably not a bad idea to accept prepare as well though. Done.

src/parser/core/modules/CastTime.ts
AkairyuGestalter: I can't think of a better name for the functions either. English major I was not .-.

src/parser/core/modules/CastTime.ts
AkairyuGestalter: I feel like I've written !<variable> elsewhere recently, I'll have to keep this in mind and see if I can find/fix those when the opportunity arises, but this one at least won't be hanging around

src/parser/core/modules/CastTime.ts
AkairyuGestalter: Yeah, that's dumb of me. Fixed.

src/parser/core/modules/CastTime.ts
AzariahAeroborn: I think setXAdjustment makes sense as a naming scheme, since the methods are configuring when a given cast time adjustment (window) is occurring.

src/parser/core/modules/CastTime.ts
```diff
@@ -31,52 +32,128 @@ export default class CastTime extends Analyser {
...
+	 * @deprecated
+	 */
+	public setInstantCastAdjustmentFflogs(actions: number[] | 'all' = 'all', start: number = this.parser.currentTimestamp, end?: number): number {
+		return this.set(actions, 'percentage', 0, this.parser.fflogsToEpoch(start), end ? this.parser.fflogsToEpoch(end) : undefined)
+	}
+	/**
+	 * Shorthand function for setting casts to instant (ie. Swiftcast, Triplecast)
+	 * @param actions The actions this adjustment applies to. Either an array of IDs, or the string 'all'. Defaults to 'all'
+	 * @param start The beginning of the adjustment time range. Defaults to the current epoch timestamp
+	 * @param end The beginning of the adjustment time range. Defaults to the current epoch timestamp
```
AzariahAeroborn, line 73: MMM, sloppy pasta

```suggestion
	 * @param end The end of the adjustment time range. May be left null if the end of the range is not yet known
```

AzariahAeroborn (changes requested):

AzariahAeroborn (approved):

PR #1194, opened 2021-05-28
BRD - Proc Suggestions

yumiyafangirl:
Simple implementation of the shiny new core module.


src/parser/jobs/brd/modules/RefulgentProcs.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+export default class RefulgentProcs extends Procs {
+	trackedProcs = [
+		{
+			procStatus: this.data.statuses.STRAIGHT_SHOT_READY,
+			consumeActions: [this.data.actions.REFULGENT_ARROW],
+		},
+	]
+
+	protected addJobSpecificSuggestions() {
```
AkairyuGestalter, line 15: Since you're not doing any job-specific logic for the suggestions (and since you're only tracking one proc status), you can actually get away with using the default dropped/overwritten proc suggestions and simply overriding the default property values. DNC has a good example of this that you can follow for the icon/content, and you can define your own overrides for the override[Dropped/Overwrote]ProcsWhy functions to take care of the why pieces.

AkairyuGestalter (commented):
Up to you if you wanna hook into the default suggestions, or leave this as-is. They both work, and imo it'd be cleaner to just override the default content/icon properties, but I'm not going to block it if you'd rather go this route.

AkairyuGestalter (approved):

PR #1195, opened 2021-05-30
Support Analyser-space events in RotationTable

AkairyuGestalter:
Analyser-space events that might be shown in a RotationTable will have their action IDs defined in either:
- event.action, or
- event.cause.action

Update the typing in Rotation.tsx's RotationProps and RotationTable.tsx's RotationTable entry, plus handing in Rotation.tsx's render function to accommodate that, while still maintaining support (for now) for the legacy event.ability.guid ID retrieval.


src/components/ui/Rotation.tsx
```diff
@@ -17,7 +18,16 @@ export default class Rotation extends Component<RotationProps> {
...
 			{events.map((event, index) => {
-				const action = getDataBy(ACTIONS, 'id', event.ability.guid) as TODO
+				let actionId = event.action
+				if (!actionId) {
+					if (event.ability) {
+						actionId = event.ability.guid
+					} else if (event.cause && event.cause.type === 'action') {
+						actionId = event.cause.action
+					}
+				}
```
AzariahAeroborn, line 28: I think this is probably gnarly enough to merit extracting into a separate function, and then using early return or an if/elseif pattern to keep it to a single indent level:
```suggestion
function getActionId(event): number | undefined {
    if (event.action != null) { 
        return event.action
    }
    if (event.ability != null) {
        return event.ability.guid
    }
    if (event.cause != null && event.cause.type === 'action') {
        return event.cause.action
    }
    return undefined
}
```

src/components/ui/Rotation.tsx
```diff
@@ -2,13 +2,14 @@ import cn from 'classnames'
...
 import ACTIONS, {ITEM_ID_OFFSET} from 'data/ACTIONS'
+import {Cause} from 'event'
 import {Ability} from 'fflogs'
 import React, {Component} from 'react'
 import overlayStyle from './Procs/ProcOverlay.module.css'
 import styles from './Rotation.module.css'
 
 interface RotationProps {
-	events: Array<{ability: Ability, isProc?: boolean}>
+	events: Array<{ability?: Ability, cause?: Cause, action?: number, isProc?: boolean}>
```
AzariahAeroborn, line 12: While we're touching this anyway... can we extract this object as an interface that can then be shared with RotationTable, so we don't have to keep editing the two in tandem?

AzariahAeroborn (changes requested):

src/components/ui/Rotation.tsx
AkairyuGestalter: Sure, that makes sense. Done.

AzariahAeroborn (approved):

PR #1197, opened 2021-06-01
[Sam] Fix wrong Yukikaze amount

Rjb99:
2 years later and Yukikaze is finally getting that extra 5 kenki from 5.0


ayyaruq (approved):
Still faster than SE fixing bugs tho.

PR #1198, opened 2021-06-01
Add new interrupt event to core structure

ackwell:
Title. Cast interruption/cancellation is a discrete concept in the game (pretty sure it's an ActorControl packet or similar), and is even logged by ACT. Unfortunately it isn't exposed by fflogs, so here we are, fabricating it. Great.

I'm opting to keep both a source and target field on this event, as realistically, interruptions _are_ targeted. Movement and explicit cancellation is self -> self, but players can interrupt some boss abilities, and (at least in the past?) bosses can interrupt us with damage.

### DIFF
--- a/src/parser/jobs/blm/modules/Procs.tsx
+++ b/src/parser/jobs/blm/modules/Procs.tsx
@@ -12,6 +12,7 @@ import React from 'react'
 
 declare module 'event' {
 	interface FieldsTargeted {
+		/** BLM specific logic. Do not use. Please remove ASAP. */
 		overrideAction?: number
 	}
 }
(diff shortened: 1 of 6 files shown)


src/parser/jobs/blm/modules/Procs.tsx
```diff
@@ -12,6 +12,7 @@ import React from 'react'
 
 declare module 'event' {
 	interface FieldsTargeted {
+		/** BLM specific logic. Do not use. Please remove ASAP. */
```
ackwell, line 15: @AkairyuGestalter 
image

left a message in `#dev-caster`, feel free to drop a line, I feel there should be a way to model this without polluting the event structure

PR #1200, opened 2021-06-01
BLM - Execute the overrideAction property

AkairyuGestalter:
Adding the overrideAction property to the legacy CastEvent was a thing. Adding it to the Events['action'] signature was an entirely different thing, and also apparently wasn't actually working anyways.

Now it's not a thing anymore.

- Moved/refactored the 'was this action a proc?' API from BLM Procs to Core Procs
- Updated RotationWatchdog to use new API instead of the overrideAction property for determining which T3s were a T3P
- Retrieved from the dustbin of history the ability to show that an action was a proc in a RotationTable
- ToDo eventually: Make the proc overlay display work for Timeline after GCD is moved to Analyser


src/components/ui/RotationTable.tsx
```diff
@@ -83,7 +83,7 @@ export interface RotationTableEntry {
 	/**
 	 * Rotation to display that occurs during this entry
 	 */
-	rotation: Array<{ability: Ability}>
```
ackwell, line 86: mmm. this'll probably need some thinkin' 'bout for analyser events, but it'll work nice for now

ackwell (approved):

PR #1201, opened 2021-06-02
Core Procs - Use new interrupt event

AkairyuGestalter:
Procs now resets the 'castingSpellId' variable when a cast interrupt is detected. Should help ensure the accuracy of the 'did this cast actually consume a proc?' checks.

Also some coding style cleanup based on recent conversations with ack.


src/parser/jobs/blm/modules/Procs.tsx
```diff
@@ -40,7 +40,7 @@ export default class Procs extends CoreProcs {
 
 	protected jobSpecificCheckConsumeProc(_procGroup: ProcGroup, event: Events['action']): boolean {
 		// If we were already hardcasting this spell, it does not consume the proc
-		return !(this.lastCastingSpellId && this.lastCastingSpellId === event.action)
+		return !(this.castingSpellId && this.castingSpellId === event.action)
```
AzariahAeroborn, line 43:
```suggestion
		return !(this.castingSpellId != null && this.castingSpellId === event.action)
```

AzariahAeroborn (changes requested):

AzariahAeroborn (approved):

PR #1202, opened 2021-06-05
Invalid prop `text` of type `object` supplied to `Dropdown`, expected `string`.

akazhuo:
I'm not using React, but i seen this error, and then i found these answer.
https://github.com/Semantic-Org/Semantic-UI-React/issues/1864#issuecomment-315640247
https://github.com/Semantic-Org/Semantic-UI-React/issues/477#issuecomment-245774707


AzariahAeroborn (approved):
Thanks for the contribution.  Looks like we just need to re-merge master, since you happened to catch a point where some of our merge orderings got the tests all out of kilter, and then this will be good to go.

PR #1204, opened 2021-06-05
Add rudimentary speed adjustments module

ackwell:
This is just a minimal scaffold for a speed adjustments module so we've got an API to build against. Further work at a later date will be required to hook this up with proper tracking and suchforth.

I don't expect this will be the final full API - helper methods and additional parameters will no doubt be needed as usecases grow.

This also wires up the new attributes data into the actors module.

### DIFF
--- a/src/parser/core/modules/Actors/Actor.ts
+++ b/src/parser/core/modules/Actors/Actor.ts
@@ -1,5 +1,5 @@
 import {JobKey} from 'data/JOBS'
-import {Events, Position, Resource} from 'event'
+import {Attribute, AttributeValue, Events, Position, Resource} from 'event'
 import {Actor as ReportActor, Team} from 'report'
 import {Object} from 'ts-toolbelt'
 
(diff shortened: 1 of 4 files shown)


src/parser/core/modules/Actors/Actor.ts
```diff
@@ -59,6 +61,23 @@ class ActorResources {
 		}
 	}
 
+	private buildAttributes() {
+		const getHistoricalAttribute = (attribute: Attribute) =>
+			this.getHistoricalValue(
+				event => event.attributes?.find(eventAttr => eventAttr.attribute === attribute),
+				{attribute, value: 0, estimated: true}
```
AzariahAeroborn, line 68: It'd be better if the default used the BASE_SPEED value for the tier, rather than zero.  The getEstimatedTime function doesn't safety check for "negative" time adjustments, so using a default of zero will produce somewhat nonsensical results.

ackwell: Ah yeah good catch

ackwell: Doneskies. I renamed the exported const to avoid referencing speed specifically, 'cus that value is used for all sub attrs.

Probably should be moved to another file but it can wait.

PR #1205, opened 2021-06-05
BLM Gauge to Analyser

AkairyuGestalter:
Re-PRing this now that BLM's Gauge is fully Analysered, instead of just converted to TypeScript. 

Changes within:
- BLM Gauge is an Analyser now (title, obviously)
- BLM's Gauge events are now queued or fabricated during the parsing pass, rather than created during the normalisation pass
- Added API to retrieve the gauge state at a particular timestamp
- Updated modules that depended on BLM Gauge's normalisation to hook events and retrieve state from new API
- Simulation of Astral Fire/Umbral Ice states, as well as Polyglot gains now use Timestamp hooks for increased accuracy
- BLM gauge state information is now visible in the timeline
- Updated suggestions to remove as many instances of ACTIONS and STATUSES as I could, and to use ActionLink and StatusLink where possible instead of raw text
- Update GlobalCooldown to respect Gauge-set cast time modification data and...
- EXECUTED BLM Speedmod \[T]/


src/parser/core/modules/CastTime.ts
```diff
@@ -56,15 +57,15 @@ export default class CastTime extends Analyser {
 	 * @returns The index number within the cast time adjustments collection, can be used to reset/end this adjustment later
 	 */
 	public setTimeAdjustment(actions: number[] | 'all', adjustment: number, start: number = this.parser.currentEpochTimestamp, end?: number): number {
-		return this.set(actions, 'time', adjustment, start, end)
+		return this.set(actions, 'time', adjustment, false, start, end)
```
ackwell, line 60: this signatures in this file are _very_ quickly approaching, or perhaps have reached, the "this should really be an object parameter" region. not going to block on it - it's probably a seperate PR's worth of changes, but honestly anything over about 3 parameters is in pushing-it territory.

src/parser/core/modules/CastTime.ts
```diff
@@ -155,6 +174,34 @@ export default class CastTime extends Analyser {
...
+	}
+
+	/**
+	 * Returns the adjusted time (either cast or recast) for the specified action at the specified point in time
+	 * @param actionId The action in question
+	 * @param timestamp The timestamp in question
+	 * @param forRecast Do we want the recast for this action, or the cast time?
+	 * @returns The adjusted time, if any adjustments exist at this timestamp, or the default if not. Will return undefined if the base time (recast/cooldown/cast) can't be determined
+	 */
+	private getAdjustedTime(actionId: number, timestamp: number = this.parser.currentEpochTimestamp, forRecast: boolean = false): number | undefined {
```
ackwell, line 204: i was gonna say 'would be nicer to type this as any numeric key on action so we can use it for arbitrary stuff in the future', but, well... `type Foo = {[K in keyof Action]: number extends Action[K] ? K : never}[keyof Action]` gets the "notpretty" badge of meh from me.

I'd like to avoid an unnamed bool arg though - either swapping to a string union or preferably naming via an object

src/parser/core/modules/CastTime.ts
```diff
@@ -164,18 +211,20 @@ export default class CastTime extends Analyser {
...
+		if (!matchingTimes.length || defaultTime == null || defaultTime === 0) {
+			return defaultTime
 		}
-
 		let flatReduction=0
 		let flatIncrease=0
 		let percentageAdjustment=1
 
 		matchingTimes.forEach(ct => {
+			if (!ct.affectsRecast && forRecast) { return }
```
ackwell, line 227: can probably be moved into the original filter so that the short circuit can pick it up

src/parser/core/modules/GlobalCooldown.js
```diff
@@ -156,16 +156,8 @@ export default class GlobalCooldown extends Module {
...
-		let castTime = action.castTime
-
-		// HACK NOTE TODO: Need to properly account for abilities that alter only the cast or recast of attacks.
-		// Thinking of moving this into a module like speedmod, that can be called with a timestamp to grab modified base castTime/cooldown values
-		const HACK_ASTRAL_UMBRAL_SPEED_SCALAR = 0.5
-		if (speedMod <= HACK_ASTRAL_UMBRAL_SPEED_SCALAR) {
-			speedMod /= HACK_ASTRAL_UMBRAL_SPEED_SCALAR
-			castTime *= HACK_ASTRAL_UMBRAL_SPEED_SCALAR
-		}
+		const speedMod = this.speedmod.get(gcdInfo.event.timestamp)
```
ackwell, line 159: probably out of scope for this PR, can be done when upgrading GCD - probably want to move this onto speed adjustments

src/parser/jobs/blm/modules/index.js
```diff
@@ -20,6 +19,5 @@ export default [
 	Procs,
 	RotationWatchdog,
 	Sharpcast,
-	Speedmod,
```
ackwell, line 23: 🥂 🎉

ackwell (commented):
yeah, 'aite. i def. think the time adjustments should be applied by cast time, but it's not feasible to do that until GCD is partially gutted and using the upstream data. now that cd2 is at least in PR, i'll focus on getting something worked out in that space.

not a full review, i've left most of the blm stuff because >late >blm >me

src/parser/core/modules/CastTime.ts
AkairyuGestalter: Yeah, that was on my mind when I added the recast parameters, but this PR was already ballooning, so I opted to leave that for another day. Would've done the object-param changeover as part of the cast time signature PR if I'd realized I was going to need the recast-oriented options at the time but well... I didn't .-.

TODO note to self to deal with that in the not-too-distant future as we collectively get the cd2/GCD stuff sorted out

src/parser/core/modules/CastTime.ts
AkairyuGestalter: Good idea. Hopefully I didn't totally brainfuzz the boolean logic there...

src/parser/core/modules/CastTime.ts
AkairyuGestalter: Changed the 'which time does this adjustment affect?' tracking to be an explicit type, and updated the matchingTimes filter/function signatures to handle it. Tried to make it a transparent change for the callers, aside from Leylines needing to set the new affectsWhich = 'both' param. Did leave myself a todo note to transform the parameter vomit into object parameters in a separate PR, really don't want to balloon this one any farther than it already is 🙃

src/parser/core/modules/CastTime.ts
AkairyuGestalter: Definitely no brainfuzz now that it's not a boolean parameter anymore :)

src/parser/core/modules/GlobalCooldown.js
AkairyuGestalter: Yeah, I was just removing BLM junk and resolving a lint grump about the speedMod value now being a const there. Wasn't aiming to change anything, just clean it up and maintain parity with the previous state. Someone can certainly do more with it/throw it out and redo it in the speed/cd2/gcd/what have you updates that are coming.

AzariahAeroborn (approved):

PR #1206, opened 2021-06-08
BLM - Don't flag rotations with all expected fires as having 'extra T3's

AkairyuGestalter:
Depending on the exact ordering of the MP status updates vs the rotation starting, opener MP can come in as 8000, which is below our 'Max MP for full rotation' threshold. That doesn't mean you'll actually be missing fires however...

Similarly, thunder spam under ice shouldn't _really_ flag as 'extra hardcast t3s' since that metric is about losing Fires because of it, which isn't the case for ice phase thunders.

There's probably some better calculation work we can put into the 'did you hardcast more T3s than you should have?' logic, but for right now, just including a backstop of checking if you got all your fires should resolve the bug.


rexatlanti:
SGTM

AzariahAeroborn (approved):

Issue #600, opened 2019-08-25
Track Thin Air usage


nivthefox:
- [ ] Track where the highest MP spells are cast, and suggest that Thin Air be used there. 
- [ ] This should be a suggestions module, not a warning module, especially if the WHM isn't running OOM.

SaltedyLevity:
This looks accurate!

I don't know if this is helpful because I'm really not knowledgeable on how XIVA operates, but where my head is going: Thin air lasts 12s, but only snapshots at the end of a cast (so if it wears off before the cast is finished, you lose mp), and I've seen GCD estimates in XIVA be a bit inaccurate, so for the sake of looking for mp saving windows, you may want to check mp spent across a 10s window (maybe 9.5s?) to account for what was actually possible to save, rather than across 10.
(You were probably going to do something like this already maybe? Figured I'd spill my thoughts)

Then the wording could be something like "you could have possibly saved at least x amount of mp by using thin air at #:##" or however you intended to state it. So the "at least" covers the fact that you might've saved even more?

(Let me know if this doesn't make sense!)

kkragenbrink:
Makes perfect sense. :) I'll have to see what I can do with it!

IsaacKhor:
The idea sounds good, but there are a few issues when we dive into the details of how one would actually implement it.

We can calculate the usages of thin air that would maximise mana saved given the entire timeline of casts in the fight. This would produce something like 'use thin air at :30, 2:50, 5:20 to save 7200 mp; you used at xx:xx, yy:yy, and zz:zz, saving 6000 mp instead'. This is *sorta* helpful, in the sense that if you could travel back in time and reoptimise your thin air usages that's what you should do, but not very helpful in determining "that's what you should do the next time you run this fight". The target audience are people trying to improve their gameplay, not "I've spreadsheeted every gcd" player.

This is why I'm resistant to recommendations of 'use thin air at xx:xx', since it's so fight specific. I'm thinking a better implementation would be 'you rezzed at 1:30 without thin air even thought it was available', but that still runs into the issue of maybe you had planned on using it later in the fight for a cure3 spam (*cries in terminal*). 

All that is to say I'm not sure what's a good approach here. I think the 'you rezzed without thin air' is a good baseline, but anything more starts running into issues I've outlined above.

ayyaruq:
Closing as no longer relevant in Endwalker.

PR #1209, opened 2021-06-11
WHM: Thin air MP tracking module

IsaacKhor:
Add module to track Thin Air mana usage. Kinda what is suggested in #600, but scanning for a specific spot of time to use thin air doesn't feel quite right as rotations are different per fight and giving specific 'use thin air here' advice doesn't sit quite well with me. 

For now this just tracks each usage of thin air and how much MP they save each. Maybe add something about total MP saved and missed usages?

Edit: also please help test that the MP calculation is correct. In theory this only counts casts completed while thin air is active but that may be buggy because I don't know how the codebase actually works :/

image

### DIFF
--- a/src/data/ACTIONS/root/WHM.ts
+++ b/src/data/ACTIONS/root/WHM.ts
@@ -184,7 +184,7 @@ export const WHM = ensureActions({
 		onGcd: true,
 		speedAttribute: Attribute.SPELL_SPEED,
 		castTime: 2000,
-		mpCost: 700,
+		mpCost: 1000,
 	},
 
 	CURE_III: {
--- a/src/data/ACTIONS/root/WHM.ts
+++ b/src/data/ACTIONS/root/WHM.ts
@@ -41,6 +41,7 @@ export const WHM = ensureActions({
 		onGcd: true,
 		speedAttribute: Attribute.SPELL_SPEED,
 		statusesApplied: ['DIA'],
+		mpCost: 400,
 	},
 
 	GLARE: {
@@ -50,6 +51,7 @@ export const WHM = ensureActions({
 		onGcd: true,
 		speedAttribute: Attribute.SPELL_SPEED,
 		castTime: 2500,
+		mpCost: 400,
 	},
 
 	PLENARY_INDULGENCE: {
@@ -67,6 +69,7 @@ export const WHM = ensureActions({
 		onGcd: true,
 		speedAttribute: Attribute.SPELL_SPEED,
 		castTime: 2500,
+		mpCost: 400,
 	},
 
 	TETRAGRAMMATON: {
@@ -105,6 +108,7 @@ export const WHM = ensureActions({
 		onGcd: true,
 		speedAttribute: Attribute.SPELL_SPEED,
 		castTime: 3000,
+		mpCost: 600,
 	},
 
 	PRESENCE_OF_MIND: {
@@ -122,6 +126,7 @@ export const WHM = ensureActions({
 		onGcd: true,
 		speedAttribute: Attribute.SPELL_SPEED,
 		castTime: 2500,
+		mpCost: 400,
 	},
 
 	DIVINE_BENISON: {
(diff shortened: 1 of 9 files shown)


src/data/ACTIONS/root/WHM.ts
```diff
@@ -176,6 +184,7 @@ export const WHM = ensureActions({
 		onGcd: true,
 		speedAttribute: Attribute.SPELL_SPEED,
 		castTime: 2000,
+		mpCost: 700,
```
AzariahAeroborn, line 187:
```suggestion
		mpCost: 1000,
```
Incorrect MP cost

src/parser/jobs/whm/modules/Thinair.tsx
```diff
@@ -0,0 +1,143 @@
...
+import {getDataBy} from 'data'
+import ACTIONS from 'data/ACTIONS'
+import STATUSES from 'data/STATUSES'
+import {Event, Events} from 'event'
+import {Analyser} from 'parser/core/Analyser'
+import {filter} from 'parser/core/filter'
+import React, {Fragment, ReactNode} from 'react'
+import {Accordion} from 'semantic-ui-react'
+
+type ThinAirRecord = {
```
AzariahAeroborn, line 14: Please use an interface instead to be consistent with the project style.

```suggestion
interface ThinAirRecord {
```

src/parser/jobs/whm/modules/Thinair.tsx
```diff
@@ -0,0 +1,143 @@
...
+	start: number,
+	end: number,
+	casts: number[],
+	mpsaved: number,
+}
+
+export default class Thinair extends Analyser {
```
AzariahAeroborn, line 21: nit: We prefer new modules use named exports instead of a default export.  This would also require changing the import statement in modules/index.js below.

```suggestion
export class Thinair extends Analyser {
```

src/parser/jobs/whm/modules/Thinair.tsx
```diff
@@ -0,0 +1,143 @@
...
+	private onThinAirRemove(ev: Events['statusRemove']) {
+		this.stopAndSave(ev.timestamp)
+	}
+
+	private onCast(ev: Events['action']) {
+		const actionid = ev.action
+
+		if (actionid === ACTIONS.THIN_AIR.id) {
+			this.startThinAir(ev.timestamp)
+		}
```
AzariahAeroborn, line 53: You should only start windows when the status applies - the game won't actually start applying the MP reduction until the status goes on, and if you cast an instant before the status applies, it won't get a reduced MP cost.

src/parser/jobs/whm/modules/Thinair.tsx
```diff
@@ -0,0 +1,143 @@
...
+		// Don't track autos
+		const action = getDataBy(ACTIONS, 'id', actionid)
+		if (!this.active || !action || action.autoAttack) {
+			return
+		}
+
+		if (action.onGcd) {
+			const mpcost = action.mpCost === undefined ? 0 : action.mpCost
+			this.currentRecord.mpsaved += mpcost
+		}
```
AzariahAeroborn, line 64: While all current WHM spells with MP cost happen to be on the GCD, that isn't necessarily a guarantee going forward as other classes have resource consuming oGCDs.  If you change this check to ensure there's an mpCost on the action, you'll simplify the remaining logic accordingly.

```suggestion
		if (action.mpCost != null) {
			this.currentRecord.mpsaved += action.mpCost
		}
```

AzariahAeroborn (changes requested):
A few style guide nitpicks and a few data/logic changes.  My first impression of the explanation message is that it's too mushy ("Don't be afraid to hold it and lose a use over the fight..."), but I'm not sure that I have a better suggestion.

IsaacKhor:
I've been thinking about this a lot, and I still think that a specific recommendation for where to use thin air doesn't really work. As long as you don't run out of mana, mana wasn't an issue, so I think what would be more apt would be trying to detect if you run OOM at any point and if so, say something about use thin air more or meld more piety.

AzariahAeroborn (approved):

PR #1211, opened 2021-06-16
Theme switcher (Debug Only For Now)

AkairyuGestalter:
Tried to take a look at what we had for dark mode, mainly because I was curious, and found it was nigh-impossible without hard-code shenanigans to even look at. PR implements a currently-debug-mode-only Theme selector in the options menus (sidebar and sidebar-less Home), to facilitate future theme development.


ackwell:
Boop. We'll be revisiting this with a different approach on endwalker.

PR #1212, opened 2021-06-17
upgrade Weaving to Analyzer, port dependencies

AzariahAeroborn:
What it says on the tin.  

95% lift and shift, 5% noticing that we had a couple classes that had weaving overrides specifically for dealing with short recast skills (DNC, MCH), so I leveraged CastTime's new recastForEvent function to get the recast of the leading GCD and adjust the max weaves if it was a short recast.  Picked the 2s cutoff semi-arbitrarily, let me know if a different cutoff makes more sense.

Also, PLD was doing some special casing for Requiescat, which I moved to having the RQ module set castTime overrides.  And BLM was BLM.  I moved the Triplecast castTime override out to a separate module for clarity, because I kept overlooking it in the Weaving module.  And the rest of the special casing for certain acceptable clips was ported as-is.


src/parser/core/modules/Weaving.tsx
```diff
@@ -0,0 +1,241 @@
...
+
+	static override title = t('core.weaving.title')`Weaving Issues`
+
+	protected suggestionIcon: string = 'https://xivapi.com/i/001000/001785.png' // WVR Focused Synth
+
+	protected moduleLink = <a
+		style={{cursor: 'pointer'}}
+		onClick={() => this.parser.scrollTo(Weaving.handle)}><NormalisedMessage message={Weaving.title}/>
+	</a>
+	protected suggestionContent: ReactNode = <Trans id="core.weaving.content">
```
nonowazu, line 55: I'd suggest a comment much like buffwindow that lets people know they can override this – this is me suggesting something non-core dev-facing though.

src/parser/core/modules/Weaving.tsx
```diff
@@ -0,0 +1,241 @@
...
+	static override handle = 'weaving'
+
+	@dependency protected castTime!: CastTime
+	@dependency private data!: Data
+	@dependency private invulnerability!: Invulnerability
+	@dependency private suggestions!: Suggestions
+
+	static override title = t('core.weaving.title')`Weaving Issues`
+
+	protected suggestionIcon: string = 'https://xivapi.com/i/001000/001785.png' // WVR Focused Synth
```
nonowazu, line 49: might as well go ahead and replace this with a `this.data` call. Also a comment letting people know they can override this would be nice.

nonowazu (changes requested):
As you said, lift and shift, but I figured we could make 1-2 things slightly better during the cleanup part.

ackwell (approved):

PR #1214, opened 2021-06-18
BLM - Port RotationWatchdog to Analyser

AkairyuGestalter:
Along with simply converting the Watchdog to the Analyser system, I've enhanced the rotation tracking to improve the analysis therein.

A Cycle is now broken up into four distinct phases:
- Unaspected Phase: This phase is only present if the cycle in some kind of opener or re-opener after a forced Enochian drop, or death.
- Umbral Ice Phase: Self-explanatory, the Ice phase of the cycle. This is generally considered the beginning of each cycle.
- Astral Fire Phase: The Fire phase of the cycle. This will terminate either when the cycle concludes, or Manafont is used.
- Manafont Phase: The additional Fire phase after Manafont is used. Tracking this separately from the main Astral Fire phase allows us to discretely examine spell usage before and after Manafont.

The expected Fire 4 count now starts from the baseline of 4, and increments based on the gauge state going into the Astral Fire phase or other factors, rather than assuming 6, and decrementing when needed. This should now prevent the alternate-playstyle 4-5 F4 with no B4/F1 cycles from getting dinged in the rotation table. I have not fully implemented support for how to improve based on that playstyle, nor do I intend to, but this should keep us from being overly snippy about it.

Excess T3 hardcast tracking now tracks separately for the Astral Fire and Manafont phases, and uses Actual MP Data (tm) to determine when the hardcasts actually lost the player a Fire 4.

PR also includes some other minor fixes, adjustments, and housekeeping work.


src/parser/jobs/blm/modules/Procs.tsx
```diff
@@ -58,7 +58,7 @@ export default class Procs extends CoreProcs {
 		const droppedThunderClouds: number = this.getDropCountForStatus(this.data.statuses.THUNDERCLOUD.id)
 		if (droppedThunderClouds > 0) {
 			this.suggestions.add(new Suggestion({
-				icon: this.data.actions.THUNDER_III_PROC.icon,
+				icon:  process.env.PUBLIC_URL + '/icon/action/t3p.png',
```
nonowazu, line 61: since you use this twice, wouldn't it be better to define this as a const somewhere else?

nonowazu (changes requested):
It's a lot of code, but I didn't see anything glaring that smacked me in the face.

The only small thing I'd suggest is making a const for that icon.

ackwell (approved):
rotation watchdog is a cursed file, may i never look over it again

PR #1217, opened 2021-07-20
Port Interrupts to Analyser

nonowazu:
From what I can tell, this is 100% the same as before, *except* that I commented out a number of line(s) and added TODOs that have to do with adding GlobalCooldown back as a dependency once that gets ported as well. Until then, I have a tiny, *tiny* shim that does a `Math.min` between the interrupt timestamp and the prepare timestamp and 2600ms.


nonowazu:
Disregard those TODOs, with the help of Ack and Az I manged to overcome.

ackwell (approved):

PR #1224, opened 2021-08-13
[PLD] Atonement Checklist

Rjb99:
In an attempt to make XIVA better at telling me I'm bad, I've added a Atonement Checklist item.
Current target is 100% usage of Atonement. I know of no reason to drop any stacks for a dps gain.


Rjb99:
For the record. I now hate Visual Studio. Despite being on different branches my SAM i18n files mixed in with these and caused it to constantly fail.

Rjb99:
Changed Rule to TieredRule
100% = success
99% to 95% = warn
94% to 0% = Red/bad

src/parser/jobs/pld/modules/Atonement.tsx
```diff
@@ -0,0 +1,80 @@
+import {t} from '@lingui/macro'
+import {Trans} from '@lingui/react'
+import {ActionLink} from 'components/ui/DbLink'
+import {Analyser} from 'parser/core/Analyser'
+import Checklist, {TieredRule, TARGET, Requirement} from 'parser/core/modules/Checklist'
+import {Data} from 'parser/core/modules/Data'
+import React from 'react'
+import {dependency} from '../../../core/Injectable'
```
nonowazu, line 8: personal nit, maybe change this to:

```ts
import {dependency} from 'parser/core/Injectable'
```

src/parser/jobs/pld/modules/Atonement.tsx
```diff
@@ -0,0 +1,80 @@
...
+	protected actions: number = 0
+	protected buffs: number = 0
+
+	override initialise() {
+
+		this.addEventHook({
+			type: 'action',
+			source: this.parser.actor.id,
+			action: this.data.actions.ATONEMENT.id,
+		}, this.onAtonement)
```
nonowazu, line 34: weird personal thought, wouldn't this be better?

```ts

		this.addEventHook({
			type: 'action',
			source: this.parser.actor.id,
			action: this.data.actions.ATONEMENT.id,
		}, () => this.actions++)
```

it'd save you a whole function, anyway?

Same with the `onRoyal` part

src/parser/jobs/pld/modules/Atonement.tsx
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	private onRoyal() {
+		this.buffs = this.buffs + STACKS_GAINED
+	}
+
+	private onComplete() {
+
+		this.checklist.add(new TieredRule({
+			name: 'Use Atonements Generated By Royal Authority',
```
nonowazu, line 56: Shouldn't this have a `Trans` tag?

nonowazu (changes requested):
I think it looks good personally, just a few minor nits.

Rjb99:
Fixed Nono's Nits.

I think this is good to merge now.

AkairyuGestalter (approved):

PR #1230, opened 2021-08-28
DNC - Flourish precast synth bugfix 

hintxiv:
Adds the corresponding flourish statuses to the GCDs which generate them. Fixes a bug where flourish gets synthed at the start of the fight.


nonowazu (approved):
Very straightforward, so looks good to me.

ayyaruq (approved):

PR #1239, opened 2021-09-25
fix: add CastTime modifiers for melee classes

AzariahAeroborn:
Used a short term hack for NIN because Huton depends on Combo and can't go to Analyser yet.  This fixes hilariously high uptime for the melee classes that have traited or buff/job gauge based speed modifiers that are basically always up.


ayyaruq:
Does this remove the need for MNK to fab a fake GL status in the current `speedmod.ts` file? Might be worth just replacing that.

src/parser/jobs/sam/modules/Shifu.ts
```diff
@@ -0,0 +1,28 @@
...
+	private shifuIndex: number | null = null
+
+	override initialise() {
+		super.initialise()
+
+		const shifuFilter = filter<Event>()
+			.target(this.parser.actor.id)
+			.status(this.data.statuses.SHIFU.id)
+
+		this.addEventHook(shifuFilter.type('statusApply'), this.onApplyPresence)
```
ackwell, line 15: p...resence?

ackwell (approved):

PR #1248, opened 2021-10-18
NIN TypeScript migration

Toastdeib:
Getting the groundwork done for the eventual Endwalker changes. This PR migrates all of the NIN modules besides DISPLAY_ORDER (which gave me weird errors when I tried) and the Huton/Ninki modules (which rely on Combos and can't be migrated yet because of that). Functionality should be entirely unchanged for all affected modules, this is just a straight migration.


Toastdeib:
Closing this but keeping the branch for later, I'm going to redo this migration as pure ts instead of ts + Analyser and then get to Analyser later.

PR #1279, opened 2021-10-25
Migrate MNK to analyser

ayyaruq:
Big cleanup of the job in general, which also revealed some bugs (like PB hooking entirely the wrong thing since 5.4 and yet never noticed by anyone). Should mostly be fine, null-checks and timestamps could prob use a particularly close look since I think it's right but I'm not super confident with the new prop names. I do notice the dev server seems to load slower even with hooks being less heavy, but I'm not sure if that's just dev mode having extra stuff to do. This is likely the final PR for MNK in Shadowbringers, unless I get motivated to add 2 really complex edge-case handling fixes to Twin and RoF modules, or just missed something.

While I mostly also migrated to DataLink, some stuff still uses ActionLink (with the new syntax) since the module doesn't need any status links.


ackwell (dismissed):
badabing badaboom

ackwell (approved):

PR #1281, opened 2021-10-26
Total Atonement usage fix

Kedka:
Currently the website shows triple the amount of possible Atonement. The listener traces Sword Oath instead of Royal Authority, which I believe the game treats as separate 3 events for every Sword Oath on the buff bar. This should fix the bug, showing one possible Atonement cast for every Sword Oath gained.


ackwell:
closing in favor of #1299 due to inactivity

PR #1289, opened 2021-10-28
port Core Overheal to Analyser

AzariahAeroborn:
As is tradition, I fiddled with the calculations on this port as well...

- Currently, the overheal module was only hooking the FFLogs `heal` events (the effect events in the new structure)
- The game, and thus FFLogs, does not consistently report `heal` events for heals that fully overheal (in general, it looks like any full-overheal on another target doesn't report an effect event)
- This caused our overheal numbers to be consistently low (significantly so if the player was doing a substantial amount of full-overheal)
- FFLogs's data tables currently treat all unmatched `calculatedheal` events as fully overheal (there's a possibility they could be ghosts from the target dying before the heal resolved, but that's a fairly infrequent occurrence)
- After this port, we will now also treat all unmatched `calculatedheal` events as fully overheal.

This doesn't quite get us to exact mapping with FFLogs' overheal %, because we are not considering shield absorption as healing at all.  The shield data comes in on the `removebuff` event for the shield and we aren't currently synthing that into healing events in the adapter steps.  Even on those remove buff events, the data is a little rough, and certain shield types seem to not ever report as expired/overheal *cough*Galvanize*cough*, so I am not adding shield handling at this time.  

Converting AST to core overheal will be a follow-up PR.

### DIFF
--- a/src/parser/core/modules/Checklist/Requirement.js
+++ b/src/parser/core/modules/Checklist/Requirement.js
@@ -3,6 +3,7 @@ export default class Requirement {
 	_percent = null
 	value = null
 	target = 100
+	_weight = 1
 	overrideDisplay = null
 
 	get content() {
@@ -19,6 +20,12 @@ export default class Requirement {
 	}
 	set percent(value) {
 		this._percent = value
+	}
+	get weight() {
+		return this._weight
+	}
+	set weight(value) {
+		this._weight = value
 	}
 
 	constructor(options) {
--- a/src/parser/core/modules/Overheal.tsx
+++ b/src/parser/core/modules/Overheal.tsx
@@ -311,12 +311,14 @@ export class Overheal extends Analyser {
 				requirements.push(new InvertedRequirement({
 					name: this.overhealName,
 					percent:  this.direct.percentInverted,
+					weight: 0,
 				}))
 
 				for (const trackedHeal of this.trackedOverheals) {
 					requirements.push(new InvertedRequirement({
 						name: trackedHeal.name,
 						percent: trackedHeal.percentInverted,
+						weight: 0,
 					}))
 				}
 			}
--- a/src/parser/core/modules/Overheal.tsx
+++ b/src/parser/core/modules/Overheal.tsx
@@ -1,8 +1,11 @@
 import {Trans} from '@lingui/react'
 import ACTIONS from 'data/ACTIONS'
-import {HealEvent} from 'fflogs'
-import Module, {dependency} from 'parser/core/Module'
+import {Event, Events} from 'event'
+import {Analyser} from 'parser/core/Analyser'
+import {filter, oneOf} from 'parser/core/filter'
+import {dependency} from 'parser/core/Injectable'
 import Checklist, {Requirement, TARGET, TieredRule} from 'parser/core/modules/Checklist'
+import {Data} from 'parser/core/modules/Data'
 import {DataSet, PieChartStatistic, Statistics} from 'parser/core/modules/Statistics'
 import Suggestions, {SEVERITY, TieredSuggestion} from 'parser/core/modules/Suggestions'
 import React from 'react'
@@ -57,7 +60,7 @@ export class TrackedOverheal {
 	 * Get current overheal as a percentage
 	 */
 	get percent(): number {
-		if (this.heal > 0) { return 100 * (this.overheal) / (this.heal + this.overheal) }
+		if (this.heal > 0) { return 100 * (this.overheal / this.heal) }
 		return 0
 	}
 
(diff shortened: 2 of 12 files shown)


src/reportSources/legacyFflogs/__tests__/__snapshots__/eventAdapter.ts.snap
```diff
@@ -158,7 +158,7 @@ Array [
     "targets": Array [
       Object {
         "amount": 36194,
-        "overheal": 0,
+        "overheal": 36194,
```
ackwell, line 161: to confirm; this is counting as overheal because it's a single-event test, and ergo is effectively ghosted, as such, is expected?

src/parser/core/modules/Overheal.tsx
```diff
@@ -93,16 +96,18 @@ export class TrackedOverheal {
...
-		this.heal += event.amount
-		this.overheal += event.overheal || 0
+	pushHeal(event: Events['heal']) {
+		this.heal += event.targets.reduce((total, target) => total + target.amount, 0)
+		this.overheal += event.targets.reduce((total, target) => total + target.overheal, 0)
 	}
 }
 
-export class CoreOverheal extends Module {
+export class Overheal extends Analyser {
```
ackwell, line 105: Ah, thanks for the rename

src/parser/core/modules/Overheal.tsx
```diff
@@ -93,16 +96,18 @@ export class TrackedOverheal {
...
 	static override handle: string = 'overheal'
+	static override debug = true
```
ackwell, line 107: d-d-debug

src/parser/core/modules/Overheal.tsx
```diff
@@ -194,8 +199,12 @@ export class CoreOverheal extends Module {
 			this.trackedOverheals.push(new TrackedOverheal(healCategoryOpts))
 		}
 
-		this.addEventHook('heal', {by: 'player'}, this.onHeal)
-		this.addEventHook('heal', {by: 'pet'}, this.onPetHeal)
+		this.addEventHook(filter<Event>().type('heal').source(this.parser.actor.id), this.onHeal)
+
+		const actorPets = this.parser.pull.actors
```
ackwell, line 204: nts; add a helper for this

src/reportSources/legacyFflogs/eventAdapter/assignOverheal.ts
```diff
@@ -0,0 +1,70 @@
+import {Event, Events} from 'event'
+import {FflogsEvent, HealEvent} from 'fflogs'
+import {AdapterStep} from './base'
+
+/**
+ * FFLogs models damage or healing events as "calculateddamage" or "calculatedheal" events at the time of cast (e.g. Character prepares Action on Target Amount)
```
ackwell, line 6: Should probably either amend the comment to just talk about healing, or add logic to (and rename) the adapter to handle both.

src/reportSources/legacyFflogs/eventAdapter/assignOverheal.ts
```diff
@@ -0,0 +1,70 @@
...
+	private storeHealEvent(adaptedEvents: Event[]) {
+		const healEvent = adaptedEvents.find((e): e is Events['heal'] => e.type === 'heal')
+		if (healEvent != null && healEvent.sequence != null) {
+			this.healEvents.set(this.buildIdentifierFromMultiTarget(healEvent), healEvent)
+		}
+	}
+
+	private populateOverheal(baseEvent: HealEvent, adaptedEvents: Event[]) {
+		const executeEvent = adaptedEvents.find((e): e is Events['execute'] => e.type === 'execute')
+		if (executeEvent != null) {
```
ackwell, line 43: can invert yada yada. same with conditional on 46.

src/parser/core/modules/Checklist/Requirement.js
```diff
@@ -3,6 +3,7 @@ export default class Requirement {
 	_percent = null
 	value = null
 	target = 100
+	_weight = 1
```
ackwell, line 6: i don't _like_ it, but i don't really like checklist, either. will fix later with a nuke i guess.

src/reportSources/legacyFflogs/__tests__/__snapshots__/eventAdapter.ts.snap
AzariahAeroborn: Correct across the board.

src/parser/core/modules/Overheal.tsx
AzariahAeroborn: Gfdi

src/reportSources/legacyFflogs/eventAdapter/assignOverheal.ts
AzariahAeroborn: Whoops. Missed that when I changed course to not worry about overkill after all since nothing wants it at this point. And the logic/signature looked a lot uglier trying to deal with both.

src/parser/core/modules/Checklist/Requirement.js
AzariahAeroborn: Seemed like the path of least resistance to make the bar (and thus the severity tiers) accurate for Overheal. I'd be fine with publicizing an unwritten rule that nothing new gets to use the weighting until we implement a replacement component.

ackwell (approved):

PR #1292, opened 2021-10-29
MCH - Port Reassemble to Analyser

hintxiv:


src/parser/jobs/mch/modules/Reassemble.tsx
```diff
@@ -1,86 +1,108 @@
...
 // These are the only GCDs that should be reassembled under normal circumstances
-const REASSEMBLE_GCDS: Array<keyof ActionRoot> = [
+const REASSEMBLE_GCDS: ActionKey[] = [
 	'AIR_ANCHOR',
 	'AUTO_CROSSBOW',
 	'DRILL',
 	'SPREAD_SHOT',
+	// 6.0:
+	//'CHAIN_SAW',
+	//'SCATTER_GUN',
```
ayyaruq, line 20: That's a bap.

src/parser/jobs/mch/modules/Reassemble.tsx
```diff
@@ -1,86 +1,108 @@
...
 
-	@dependency private combatants!: Combatants
 	@dependency private data!: Data
 	@dependency private suggestions!: Suggestions
 
 	private badReassembles = 0
 	private droppedReassembles = 0
-	private lastUse = 0
+	private lastGcd = 0
+	private reassembleActive = false
```
ayyaruq, line 40: Prefer `hasReassemble` or something, because I like to nitpick consistency.

src/parser/jobs/mch/modules/Reassemble.tsx
```diff
@@ -1,86 +1,108 @@
...
 
 	private REASSEMBLE_GCDS: number[] = []
 	private OTHER_GCDS: number[] = []
 
-	protected override init() {
+	override initialise() {
 		this.REASSEMBLE_GCDS = REASSEMBLE_GCDS.map(key => this.data.actions[key].id)
 		this.OTHER_GCDS = OTHER_GCDS.map(key => this.data.actions[key].id)
```
ayyaruq, line 47: Not actually part of the diff, but prefer camel-casing here so it's clearer if you're using the const or the local var, and also I'm impressed linter didn't rage at this tbh.

src/parser/jobs/mch/modules/Reassemble.tsx
```diff
@@ -1,86 +1,108 @@
...
+	private onReassemble() {
+		if (this.reassembleActive) {
+			this.droppedReassembles++
+		}
+		this.reassembleActive = true
+	}
+
+	private onCast(event: Events['action']) {
+		if (this.reassembleActive && !this.REASSEMBLE_GCDS.includes(event.action)) {
+			this.badReassembles += 1
```
ayyaruq, line 73: Any reason not to `this.badReassembles++`? I guess future proofing if you get stacks later or something, but for now that's probably clearer unless something breaks.

src/parser/jobs/mch/modules/Reassemble.tsx
```diff
@@ -1,86 +1,108 @@
...
 			content: <Trans id="mch.reassemble.suggestions.bad-gcds.content">
-				On single targets <ActionLink {...this.data.actions.REASSEMBLE}/> should only ever be used on <ActionLink {...this.data.actions.DRILL}/> and <ActionLink {...this.data.actions.AIR_ANCHOR}/>, as they're your strongest GCDs by a large margin.
+				On single targets <DataLink action="REASSEMBLE" /> should only ever be used on <DataLink action="DRILL" /> and <DataLink action="AIR_ANCHOR" />, as they're your strongest GCDs by a large margin.
 			</Trans>,
 			tiers: {
-				1: SEVERITY.MEDIUM,
-				2: SEVERITY.MAJOR,
+				1: SEVERITY.MINOR,
+				2: SEVERITY.MEDIUM,
+				3: SEVERITY.MAJOR,
```
ayyaruq, line 94: This could realistically be moved to a const for legibility.

src/parser/jobs/mch/modules/Reassemble.tsx
```diff
@@ -1,86 +1,108 @@
-import {Trans, Plural} from '@lingui/react'
-import {ActionLink} from 'components/ui/DbLink'
-import {ActionRoot} from 'data/ACTIONS/root'
-import STATUSES from 'data/STATUSES'
-import {CastEvent, BuffEvent} from 'fflogs'
-import Module, {dependency} from 'parser/core/Module'
-import Combatants from 'parser/core/modules/Combatants'
+import {Plural, Trans} from '@lingui/react'
+import {DataLink} from 'components/ui/DbLink'
```
ayyaruq, line 2: Since you only ever use actions, could prob leave this as `ActionLink`. The new syntax works there too, so you can still `<ActionLink action="REASSEMBLE"/>` as needed.

src/parser/jobs/mch/modules/Reassemble.tsx
```diff
@@ -1,86 +1,108 @@
...
+			.source(this.parser.actor.id)
+			.status(this.data.statuses.REASSEMBLED.id)
+
+		const gcdFilter = filter<Event>()
+			.source(this.parser.actor.id)
+			.type('action')
+			.action(oneOf([...this.REASSEMBLE_GCDS, ...this.OTHER_GCDS]))
+
+		this.addEventHook(reassembleFilter.type('statusApply'), this.onReassemble)
+		this.addEventHook(gcdFilter, this.onCast)
```
ayyaruq, line 59: Is it really worth hooking every GCD cast for quite a large list? Consider adding the hook in `onReassemble` and removing it in `onRemove`. Functionality inside the callback shouldn't need to change, but you'll be hooking way less junk events since `lastGcd` only matters inside an active Reassemble status. Since it's consumed by GCDs, timing derpery between pressing Reassemble and getting the status shouldn't be an issue.

src/parser/jobs/mch/modules/Reassemble.tsx
```diff
@@ -1,86 +1,108 @@
...
+		}
+		this.reassembleActive = true
+	}
+
+	private onCast(event: Events['action']) {
+		if (this.reassembleActive && !this.REASSEMBLE_GCDS.includes(event.action)) {
+			this.badReassembles += 1
 		}
-		this.lastUse = event.timestamp
+		this.lastGcd = event.timestamp
```
ayyaruq, line 75: Maybe include time in here, like `lastGcdTimestamp` or something, hypernit tho, I just think it's easier to read and tell you're operating on a timestamp rather than the last GCD event itself, although you could make it the actual Event so you have direct timestamp access as well as the action itself if you ever need it.

ayyaruq (commented):
Nits, mostly style, but I'd rethink that `onCast` as an ephemeral hook for performance.

src/parser/jobs/mch/modules/Reassemble.tsx
ackwell: There's no need to do so, though. Why use individual ones when the data link works for everything.

src/parser/jobs/mch/modules/Reassemble.tsx
ackwell: Ja, please keep 6.0 stuff out of the 5.x branch. I'll be splitting branches shortly for additions 👍

src/parser/jobs/mch/modules/Reassemble.tsx
ackwell: I'm not going to push hard for this, but yes. Adding/removing hooks is cheap, but every active hook has a slight overhead. If the data used by a hook is not used outside defineable windows, it's preferable to only hook it during those windows.

ackwell (approved):
generally lgtm; just keen to keep 6.0 out of the 5.x snapshot

src/parser/jobs/mch/modules/Reassemble.tsx
abodak: `+= 1` supremacy, there's no world where `++`  is clearer than `+= 1`.

(this is what happens when you spend too much time with python)

src/parser/jobs/mch/modules/Reassemble.tsx
```diff
@@ -1,94 +1,128 @@
...
+			this.history.badUses += 1
+		}
+		this.state.lastGcdTime = event.timestamp
+	}
+
+	private onRemove(event: Events['statusRemove']) {
+		if (event.timestamp !== this.state.lastGcdTime) {
+			this.history.droppedUses += 1
+		}
+		if (!(this.state.gcdHook == null)) {
```
ayyaruq, line 98: I think you can just use `if (this.state.gcdHook != null)` here, since the state var is always defined, it's only gcdHook that can be undef, and `!=` against null for undef will still behave as expected. In general, xiva style frowns upon negating falsey checks (but booleans are fine).

ayyaruq (approved):

PR #1293, opened 2021-10-30
migrate Gauge to analyser

AzariahAeroborn:
What it says on the tin.  And I didn't even change logic this time!  Should be a 1:1 migration, only PLD Oath Gauge and WAR Beast Gauge are currently consuming core Gauge, from what I could find.

### DIFF
--- a/src/parser/core/modules/Gauge/CounterGauge.ts
+++ b/src/parser/core/modules/Gauge/CounterGauge.ts
@@ -1,6 +1,5 @@
 import {ChartDataSets} from 'chart.js'
 import Color from 'color'
-import {Event} from 'event'
 import _ from 'lodash'
 import {AbstractGauge, AbstractGaugeOptions} from './AbstractGauge'
 
@@ -10,8 +9,6 @@ interface CounterHistory {
 	minimum: number
 	maximum: number
 }
-
-export type CounterGaugeModifier = Partial<Record<Event['type'], number>>
 
 export interface CounterGaugeOptions extends AbstractGaugeOptions {
 	/** Initial value of the gauge. Defaults to the minimum value of the gauge. */
--- a/src/parser/core/modules/Gauge/CounterGauge.ts
+++ b/src/parser/core/modules/Gauge/CounterGauge.ts
@@ -1,5 +1,6 @@
 import {ChartDataSets} from 'chart.js'
 import Color from 'color'
+import {EventTypeRepository} from 'event'
 import _ from 'lodash'
 import {AbstractGauge, AbstractGaugeOptions} from './AbstractGauge'
 
@@ -9,6 +10,9 @@ interface CounterHistory {
 	minimum: number
 	maximum: number
 }
+
+type EventKeys = keyof EventTypeRepository
+export type CounterGaugeModifier = Partial<Record<EventKeys, number>>
 
 export interface CounterGaugeOptions extends AbstractGaugeOptions {
 	/** Initial value of the gauge. Defaults to the minimum value of the gauge. */
--- a/src/parser/core/modules/Gauge/CounterGauge.ts
+++ b/src/parser/core/modules/Gauge/CounterGauge.ts
@@ -1,6 +1,6 @@
 import {ChartDataSets} from 'chart.js'
 import Color from 'color'
-import {EventTypeRepository} from 'event'
+import {Event} from 'event'
 import _ from 'lodash'
 import {AbstractGauge, AbstractGaugeOptions} from './AbstractGauge'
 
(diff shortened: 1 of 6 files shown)


src/parser/jobs/war/modules/Gauge.tsx
```diff
@@ -51,46 +30,60 @@ export class Gauge extends CoreGauge {
...
-				this.cooldowns.reduce('INFURIATE', INFURIATE_CDR)
-			},
-		)
 		this.addEventHook('complete', this.onComplete)
 	}
 
-	private onGaugeModifier(event: LegacyComboEvent | CastEvent) {
-		const modifiers = BEAST_MODIFIERS.get(event.ability.guid) || {}
+	private onGaugeModifier(event: Events['action' | 'combo']) {
+		const modifiers = this.beastGaugeModifiers.get(event.action)
```
ayyaruq, line 70: This should be renamed to `modifier`, since it's always going to be a unique single object anyway.

src/parser/core/modules/Gauge/index.ts
```diff
@@ -1,3 +1,4 @@
 export {Gauge} from './Gauge'
 export {CounterGauge} from './CounterGauge'
+export type {CounterGaugeModifier} from './CounterGauge'
```
ayyaruq, line 3: Does this need the type specifier?

ayyaruq (commented):
Couple of Q's but not requesting changes. It may be worth making `overcap` in CounterGauge a getter since users shouldn't be setting it and it's easier to read in consuming modules than destructuring the object.

src/parser/core/modules/Gauge/index.ts
AzariahAeroborn: Yes, our typescript settings make this required.

src/parser/jobs/war/modules/Gauge.tsx
ayyaruq: I'll sort this out later then.

ayyaruq:
I'm happy with this in the current state, but considering it's gauge, may be worth ack taking a look too.

src/parser/jobs/pld/modules/Gauge.tsx
```diff
@@ -1,55 +1,57 @@
...
+	])
+
+	override initialise() {
+		super.initialise()
+
+		this.addEventHook(filter<Event>()
+			.type('action')
+			.source(this.parser.actor.id)
+			.action(oneOf(Array.from(this.oathModifiers.keys()))),
+		this.onOathModifying)
```
ackwell, line 36:
```suggestion
		this.addEventHook(
			filter<Event>()
				  .type('action')
				  .source(this.parser.actor.id)
				  .action(oneOf(Array.from(this.oathModifiers.keys()))),
			this.onOathModifying,
		)
```

src/parser/core/modules/Gauge/CounterGauge.ts
```diff
@@ -10,6 +11,8 @@ interface CounterHistory {
 	maximum: number
 }
 
+export type CounterGaugeModifier = Partial<Record<Event['type'], number>>
```
ackwell, line 14: Not super keen on this. It's not actually a counter gauge modifier - that'd just be a number. It doesn't actually represent any type in this file, but rather, the implementation details of a few consumers.

Given it's as trivial as it is, it should probably be inlined into the files that use it - you might be able to narrow the key type, too, which would potentially let you drop the partial

src/parser/jobs/pld/modules/Gauge.tsx
```diff
@@ -1,55 +1,57 @@
...
 
 	private oathGauge = this.add(new CounterGauge({
 		chart: {label: 'Oath Gauge', color: JOBS.PALADIN.colour},
 	}))
-
-	protected override init() {
-		super.init()
-
-		this.addEventHook('cast', {by: 'player', abilityId: Array.from(OATH_MODIFIERS.keys())}, this.onOathModifying)
+	private oathModifiers = new Map<number, CounterGaugeModifier>([
```
ackwell, line 22: nts; seeing a bunch of this cropping up. i should probably set up some helpers to cover common data key mapping tasks on module init

src/parser/jobs/war/modules/Gauge.tsx
```diff
@@ -51,46 +30,60 @@ export class Gauge extends CoreGauge {
...
 
-	protected override init() {
-		super.init()
+	override initialise() {
+		super.initialise()
+
+		const playerFilter = filter<Event>().source(this.parser.actor.id)
+		const beastActions = Array.from(this.beastGaugeModifiers.keys())
+		this.addEventHook(playerFilter.type('action').action(oneOf(beastActions)), this.onGaugeModifier)
+		this.addEventHook(playerFilter.type('combo').action(oneOf(beastActions)), this.onGaugeModifier)
```
ackwell, line 61: can combine these with `.type(oneOf(['action', 'combo']))`

ackwell (changes requested):
mostly interested in the ded hook. not super keen on the type hoist.

otherwise; minor nits

src/parser/jobs/pld/modules/Gauge.tsx
ayyaruq: 👀

ackwell (approved):

ackwell:
thankeethankee

PR #1295, opened 2021-10-30
Excise legacy functions - Part 1

AkairyuGestalter:
Deleting a few legacy support functions that are no longer necessary due to ongoing Analyser port work:
- Gauge state accessor in BLM Gauge
- Feather consumption accessor in DNC Gauge
- Proc usage accessor in core Procs


ayyaruq (approved):
Only deletions? It's like all our legacy namedays have come at once.

PR #1299, opened 2021-10-30
Fix: PLD - Deal with stacking Atonement buff

AkairyuGestalter:
Since the fix to have `statusApply` events come through as stacking buffs build or decay, PLD's Atonement analyser has been unintentionally tripling the number of Atonements expected to be used. PR switches from incrementing gains on stack application, to checking for successful hits from Royal Authority.

I also went ahead and added tracking for overcap in the event that someone Royal Authorities again without having used all of the Atonement stacks, just in case we want to do something with it in the future.

Supercedes PR #1281

### DIFF
--- a/src/parser/jobs/pld/modules/Atonement.tsx
+++ b/src/parser/jobs/pld/modules/Atonement.tsx
@@ -47,7 +47,8 @@ export default class Atonement extends Analyser {
 	}
 
 	private onGenerateStacks(): void {
-		this.overcap += Math.max(this.currentStacks + STACKS_GAINED - MAX_STACKS, 0) // Tracking overcap in case someone wants to do something with it later
+		// Tracking overcap in case someone wants to do something with it later
+		this.overcap += Math.max(this.currentStacks + STACKS_GAINED - MAX_STACKS, 0)
 		this.potentialStacks += STACKS_GAINED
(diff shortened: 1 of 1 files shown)


src/parser/jobs/pld/modules/Atonement.tsx
```diff
@@ -34,14 +38,20 @@ export default class Atonement extends Analyser {
...
+			type: 'combo',
+			source: this.parser.actor.id,
+			action: this.data.actions.ROYAL_AUTHORITY.id,
+		}, this.onGenerateStacks)
 
 		this.addEventHook('complete', this.onComplete)
 	}
 
+	private onGenerateStacks(): void {
+		this.overcap += Math.max(this.currentStacks + STACKS_GAINED - MAX_STACKS, 0) // Tracking overcap in case someone wants to do something with it later
```
ackwell, line 50:
```suggestion
		// Tracking overcap in case someone wants to do something with it later
		this.overcap += Math.max(this.currentStacks + STACKS_GAINED - MAX_STACKS, 0)
```

ackwell (commented):
would kinda like to get this using status data (is it available, actually?), but prio fix over niceties

ackwell (approved):

PR #1306, opened 2021-10-31
Use actual GCD accounting instead of trusting status in PB drop math

ayyaruq:
Stacks remaining at the end of a window are 1 rather than 0, however we can't just subtract because it's possible for a legit stack to be left. Realistically it goes from 5->0 but the game prefers to do 6->1 and then just delete it. I have no idea how consistent this will even be, so I've changed the drop accounting to track the used GCDs and compare with the max stacks instead.

The constant is kinda silly, but `stacksApplied` is an optional field. Open to a cleaner way to do this tho, since I'm not sure if a null check is actually better in this case. This module is being rewritten in 6.0 since the entire skill is changing, so I'm not too fussed about expecting future changes in the 5.x branch anyway.


ackwell (approved):
would be nice to use status data if possible, in future

PR #1307, opened 2021-10-31
Fix over-eager weaving warnings on MNK

ackwell:
two lines code, eight lines comment. nice.

acc'll hate this.

pls2read the comment :)


ayyaruq (approved):
I hate this but weaving is about count rather than clipping. If users are still unhappy with it, they can provide feedback in discord but it should prob be sorted out in core weaving anyway.

PR #1308, opened 2021-10-31
Clean: PLD - Add & use Sword Oath stacksApplied data for Atonement

AkairyuGestalter:
- Apparently we hadn't included the `stacksApplied` data in PLD's status file for Sword Oath, that is added now.
- Updated stack tracking to return to using `statusApply` events for tracking generation/overcap and `statusRemove` for dropped stacks.
- Tinkered with the checklist wording to use the new DataLink elements, and programmatically include the `stacksApplied` value in the message.
- Not done: presenting overcap/dropped stack information anywhere. The value of doing this is left up to the PLD people :)


src/parser/jobs/pld/modules/Atonement.tsx
```diff
@@ -38,34 +36,52 @@ export default class Atonement extends Analyser {
...
+		}
+
+		this.currentStacks = event.data
+	}
+
+	private onRemoveSwordOath(): void {
+		// If any stacks were left when the status fell off, keep track of those
+		if (this.currentStacks > 0) {
+			this.droppedStacks += this.currentStacks
+		}
```
AzariahAeroborn, line 68: This counts ALL removals of the last stack as drops, whether they were expired or used.  We don't get an Apply stacks: 0 event on the last usage to update the currentStacks value.

This log http://127.0.0.1:3000/fflogs/gqaw9kRchv67W2xP/11/2 has a timeout at 2:19 for testing/comparison, but should only show 3 dropped stacks (2 timed out at 2:19 and 1 left over at end of fight)

AzariahAeroborn (changes requested):

src/parser/jobs/pld/modules/Atonement.tsx
AkairyuGestalter: Oop, you're right. I don't feel like having leftover stacks at the end of the fight should really count as a drop, but I corrected the stack usage/statusRemoval tracking to make sure we're tracking only the unused stacks at the time the status falls off.

ayyaruq (approved):

PR #1309, opened 2021-10-31
Resource graph for core gauge

AkairyuGestalter:
PR moves graphing output for core gauge analysers from the former standalone TimelineChart interface, into the timeline via the ResourceGraph:
- ResourceGraph public API simplified, if creating a custom dataGroup, the expectation is now that you'll create the group first. Attempting to add data to a non-existent group will now no-op instead of making the group
- Converted BLM's call to addDataGroup for the AF/UI grouper to use the simplified signature
- Swapped WAR/PLD to use the new graphing options
- Removed the old charting option from core Gauge
- Fixed a legacy time issue that snuck past review of the initial Analyser port of core Gauge


AkairyuGestalter:
After a *lot* of conversation in the Discord, I'm returning the charting option for CounterGauges and putting WAR's Beast Gauge back on the charting method. We'll explore improvements to the ResourceGraph timeline version in the lead up to/during Endwalker, with the intent to standardize onto the timeline in the near future.

src/parser/core/modules/Gauge/CounterGauge.ts
```diff
@@ -120,16 +124,36 @@ export class CounterGauge extends AbstractGauge {
...
+			this.resourceGraphs.addGauge(graphData, collapse)
+		}
+	}
+
 	/** @inheritdoc */
 	override generateDataset() {
-		// If there's no chart options, provide nothing
-		if (!this.chartOptions) {
+		// If there's no chart options, or if there are graph options, provide nothing (prefer graph)
+		if (this.chartOptions == null || this.graphOptions != null) {
```
ayyaruq, line 150: Why are these mutually exclusive? If a job module wants both, just show both. I realise it's duplicating information but the context of how it's presented is the important bit.

ackwell: No. One or the other.

AzariahAeroborn (approved):

PR #1311, opened 2021-10-31
migrate AST overheal to core

AzariahAeroborn:
As with the core Overheal port, this isn't going to result in a 1:1 mapping of overheal %, because we're now picking up and properly measuring full-overheal events.  So, logs across the board should see a higher overheal % (and closer to the amount FFLogs displays), though there will still be some difference between us and FFLogs depending on how much shielding the AST did.


src/parser/jobs/ast/modules/Overheal.tsx
```diff
@@ -1,111 +1,35 @@
...
-		return 100 - this.percent
-	}
-
-	override get content() {
-		if (this._percent !== null || this.value === null) { return `${this.percentInverted.toFixed(2)}%` }
-		return `${this.value.toFixed(0)}/${this.target.toFixed(0)}` // avoid weird floating point shit
-	}
+		},
+		{
+			name: <Trans id="ast.overheal.earthlystar.name">Earthly Star</Trans>,
```
ayyaruq, line 27: Does this really need translation over something like `<ActionLink showIcon=false action='EARTHLY_STAR'/>`? That way it's auto-localised, iirc can also meme `showTooltip=false` to just keep the name localised without having any mouseover. If it needs to be a React.JSX object, you can use a Fragment I think.

ayyaruq (commented):
Overall looks good, much simpler now that Overheal handles pets properly, but I'd rethink using a skill name as a translation string since it's kinda obnoxious to translate something that we already have i18n data for. Not forcing it, but that's my thoughts.

src/parser/jobs/ast/modules/Overheal.tsx
AzariahAeroborn: Mostly did it this way to keep the current behavior of AST and match the structure of SCH/WHM. Though I couldn't keep the i18n key because it's a little different format now (no trailing colon) so I updated the context in the key.

ayyaruq: Fair enough.

ayyaruq (approved):

PR #1312, opened 2021-10-31
[AST] analyser port for CI, update undraw to tsx, update combust to DataLink

otocephaly:
for undraw, I used this log to test https://www.fflogs.com/reports/pz94wPA3QMjfcJxC and noted in prod that undraw didn't show up in the checklist and now does with this converted file for some reason


src/parser/jobs/ast/modules/ArcanaUndrawUsage.tsx
```diff
@@ -0,0 +1,62 @@
...
+import {dependency} from 'parser/core/Injectable'
+import {Data} from 'parser/core/modules/Data'
+import Suggestions, {Suggestion, SEVERITY} from 'parser/core/modules/Suggestions'
+import React from 'react'
+
+export default class ArcanaUndrawUsage extends Analyser {
+	static override handle = 'arcanaundraws'
+	static override dependencies = [
+		'suggestions',
+	]
```
AzariahAeroborn, line 15: Remove this, in favor of the `@dependency private suggestions` import below - for TypeScript, we require using the `@dependency` decorator so that you get the correct typing on the imports when you reference them.

src/parser/jobs/ast/modules/ArcanaUndrawUsage.tsx
```diff
@@ -0,0 +1,62 @@
...
+			.action(this.data.actions.UNDRAW.id), this._onCast)
+		this.addEventHook('complete', this._onComplete)
+		this._badUndraws = 0
+	}
+
+	_onCast(event: Events['action']) {
+
+		this._saveIfBadUndraw(event)
+
+	}
```
AzariahAeroborn, line 35: Clean up extra whitespace (also on initialise above and onComplete below)

src/parser/jobs/ast/modules/CelestialIntersection.tsx
```diff
@@ -43,7 +48,7 @@ export default class CelestialIntersection extends Module {
 	onComplete() {
 		const holdDuration = this.uses === 0 ? this.parser.currentDuration : this.totalHeld
 		const usesMissed = Math.floor(holdDuration / (this.data.actions.CELESTIAL_INTERSECTION.cooldown))
-		const fightDuration = this.parser.fight.end_time - this.parser.fight.start_time
+		const fightDuration = this.parser.pull.duration
 		const maxUses = (fightDuration / this.data.actions.CELESTIAL_INTERSECTION.cooldown) - 1
```
AzariahAeroborn, line 52:
```suggestion
		const maxUses = (this.parser.pull.duration / this.data.actions.CELESTIAL_INTERSECTION.cooldown) - 1
```

Condense variable since there's no calculation going on now.

src/parser/jobs/ast/modules/CelestialIntersection.tsx
```diff
@@ -56,12 +61,12 @@ export default class CelestialIntersection extends Module {
...
-					Use <ActionLink {...this.data.actions.CELESTIAL_INTERSECTION} /> more frequently. Frequent uses can heal or mitigate a large amount of damage over the course of a fight, potentially resulting in fewer required healing GCDs.
+					Use <DataLink action="CELESTIAL_INTERSECTION" /> more frequently. Frequent uses can heal or mitigate a large amount of damage over the course of a fight, potentially resulting in fewer required healing GCDs.
 				</Trans>,
 				tiers: WASTED_USE_TIERS,
 				value: this.uses === 0 ? 100 : usesMissed,
 				why: <Trans id="ast.celestial-intersection.suggestion.why">
-					About {usesMissed} uses of <ActionLink {...this.data.actions.CELESTIAL_INTERSECTION} /> were missed by holding it for at least a total of {this.parser.formatDuration(holdDuration)}.
+					About {usesMissed} uses of <DataLink action="CELESTIAL_INTERSECTION" /> were missed by holding it for at least a total of {this.parser.formatDuration(holdDuration)}.
 				</Trans>,
 			}))
```
AzariahAeroborn, line 71: Not going to block on this comment, since you're porting a pre-existing condition, but this module could use some additional review for EW to update how it calculates missed / max uses.  It looks like the logic here is basically going for "use it whenever it's available", but the logic doesn't account for fight downtime (e.g. you can't use the skill during the giant lazor scene in e1).  If the goal really should be to use this skill on cooldown, it might be a candidate to add to a module that extends core CooldownDowntime, since that core module does account for fight downtime.  I know that certain other heal skills (including Divine Benison) intentionally don't want to use CooldownDowntime because there are cases where you don't want to spam them (aligning with fight damage, weaving considerations, etc.), so if that applies here as well then maybe we just look at getting this module to account for fight downtime in a later PR / for EW.

AzariahAeroborn (changes requested):

src/parser/jobs/ast/modules/CelestialIntersection.tsx
otocephaly: yeah I saw that so definitely something to add to TODO haha
I left a comment near the calculation in question as it definitely needs to be fixed in some sort of manner
thank you for the comment!

otocephaly:
updated based on comments above. Thanks Azar

AzariahAeroborn (approved):

PR #1318, opened 2021-11-03
Properly type IDs in MNK

ayyaruq:
Fairly simple, except for Fists. Weirdly while trying to get Fistless to play nice, I ended up fixing 3 bugs by accident. Unknown Fist detection is no longer needed since this handles bork statuses well enough, relying on the assumption of FoF if prepull statuses don't resolve.

I'm not much of a fan of the chart colours stuff, and I think `getFistUptimePercent` could maybe be a bit cleaner, so open to ideas there.


src/parser/jobs/mnk/modules/Fists.tsx
```diff
@@ -28,33 +30,39 @@ const FIST_SEVERITY = {
...
+	constructor(fistId: FistId, start: number) {
 		this.id = fistId
 		this.start = start
 	}
 }
 
-interface ChartColours { [fistId: number]: string}
-
 export class Fists extends Analyser {
+	static override debug = true
```
AzariahAeroborn, line 52:
```suggestion
	static override debug = false
```

src/parser/jobs/mnk/modules/Fists.tsx
```diff
@@ -190,52 +196,68 @@ export class Fists extends Analyser {
...
 		return this.fistory
 			.filter(fist => fist.id === fistId)
 			.reduce((total, current) => total + current.gcdCounter, 0)
 	}
 
-	getFistUptimePercent(fistId: number, unknownFist: boolean): string {
-		const status = this.data.getStatus(fistId)
-		if (status == null) { return '0' }
+	getFistUptimePercent(fistId: FistId): string {
+		const fightUptime = this.parser.currentDuration - this.invulnerability.getDuration({types: ['invulnerable']})
```
AzariahAeroborn, line 221: Typically Downtime is the preferred method to do uptime refunds, since it also accounts for time spent dead (so that uptime calcs don't double-ding against the "don't die" recommendation).

src/parser/jobs/mnk/modules/Fists.tsx
```diff
@@ -151,7 +157,7 @@ export class Fists extends Analyser {
 			why: <Trans id="mnk.fists.suggestions.stanceless.why">
 				<Plural value={this.getFistGCDCount(FISTLESS)} one="# GCD" other="# GCDs"	/> had no Fists buff active.
 			</Trans>,
-			tiers: FIST_SEVERITY.DOUSED,
+			tiers: SUGGESTION_TIERS.DOUSED,
```
AzariahAeroborn, line 160: This FISTLESS suggestion can never get hit, since we're unhooking the fistHook whenever a fist is removed, and only re-adding it on applying a fist status (though it looks like that's a pre-existing condition too... has this ever worked?)

Also, it doesn't look like we're doing anything with the GCD count under fists of fire, so we should probably adjust the onCast hook lifetime such that it's active for FISTLESS, Earth, and Wind, but not Fire. It might also make sense to just make the fist GCD counters buckets that always exist and you only have one bucket per relevant fist type, to simplify the getFistGCDCount logic (replace that array filtering & reducing with just accessing the buckets).

src/parser/jobs/mnk/modules/Forms.tsx
```diff
@@ -21,16 +23,16 @@ export class Forms extends Analyser {
...
 
 	private formless: number = 0
 	private resetForms: number = 0
 	private skippedForms: number = 0
 	private droppedForms: number = 0
 
-	private lastFormChanged?: number
-	private lastFormDropped?: number
+	private lastFormChanged: number | undefined
+	private lastFormDropped: number | undefined
```
AzariahAeroborn, line 35: There's no difference in effective type with these changes.  Personally, I would prefer the ?: format, especially for things that don't exist initially, eventually get set, and then keep that value or get updated to a new value but never back to unset/undefined.

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -82,7 +82,7 @@ export class RiddleOfFire extends Analyser {
 	@dependency private timeline!: Timeline
 
 	private history: Riddle[] = []
-	private riddle?: Riddle
+	private riddle: Riddle | undefined
```
AzariahAeroborn, line 85: See also comment on Forms

src/parser/jobs/mnk/modules/TwinSnakes.tsx
```diff
@@ -45,8 +46,8 @@ export class TwinSnakes extends Analyser {
 	@dependency private suggestions!: Suggestions
 
 	private history: TwinState[] = []
-	private ignoredGcds: number[] = []
-	private twinSnake?: TwinState
+	private ignoredGcds: Array<Action['id']> = []
+	private twinSnake: TwinState | undefined
```
AzariahAeroborn, line 50: See also comment on Forms

AzariahAeroborn (changes requested):
We should also probably look at lifting the fillActions and fillStatuses into helpers in the data module (like the Cause helpers that just got added), since many jobs are doing similar things for their hooks and could use the sugar as well.  Plus if it's a helper in the data module, we won't have to pass around the dependency to the helper.

src/parser/jobs/mnk/modules/Forms.tsx
ayyaruq: It's clearer about the separation between optional and potentially undefined. Due to how Forms work, these vars aren't optional.

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
ayyaruq: Same deal as above although I guess technically using RoF is optional, even if not using it is a terrible idea.

src/parser/jobs/mnk/modules/TwinSnakes.tsx
ayyaruq: Same deal here, it's not optional, just potentially undefined.

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
ayyaruq: Making this one optional afterall.

src/parser/jobs/mnk/modules/Fists.tsx
ayyaruq: File removed in 6.0 anyway since Fists are gone, leaving this alone for 5.x since it's about status uptime, not if that uptime was useful. Realistically this chart returns 100% pretty much exclusively, since there's no reason to use other fists anymore.

ayyaruq:
Helpers can move in another PR, prob get a rename too. I've left some of the undef vars as is because I want to separate them being optional from potentially undefined. I use it for flow control so I think it's best to not mark it as optional. Leaving Fists uptime as is because it realistically doesn't matter.

src/parser/jobs/mnk/modules/Fists.tsx
ayyaruq: We don't bother with GCDs under FoF because that's what we want to see usually. FISTLESS looks unused because it'll usually only happen if they forgot to apply a fist at the start. Previously it was mostly a thing in Stormblood post-RoE but we still see it sometimes. Alternatively, could invert the suggestion to count GCDs under FoF and then matchClosestHigher, but we'd need a global GCD count for that which seems excessive to hook.

Bucketing is interesting, but since fists are being removed in 6.0 I'm not sure it's worth the effort. Current MNK gameplay is basically perma-FoF so for the most part we don't see suggestions from this module often anyway.

AzariahAeroborn (approved):

PR #1319, opened 2021-11-03
Migrated TrickAttackUsage to ts and Analyser

Toastdeib:
What it says on the tin. This is the last major file that was pending ts migration for NIN, so that's all sorted now.


src/parser/jobs/nin/modules/TrickAttackUsage.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+		const playerFilter = filter<Event>().source(this.parser.actor.id)
+		this.castHook = this.addEventHook(playerFilter.type('action'), this.onCast)
+		this.addEventHook(playerFilter.type('action').action(this.data.actions.TRICK_ATTACK.id), this.onTrickAttack)
+		this.addEventHook('complete', this.onComplete)
+	}
+
+	private onCast(event: Events['action']) {
+		const action = this.data.getAction(event.action)
+		if (event.timestamp >= this.parser.fight.start_time && action && action.onGcd && this.mudraActions.indexOf(action.id) === -1) {
```
ayyaruq, line 50:
```suggestion
		if (event.timestamp >= this.parser.fight.start_time && action?.onGcd && !this.mudraActions.includes(action.id)) {
```
or
```suggestion
		if (event.timestamp >= this.parser.fight.start_time && action?.onGcd && this.mudraActions.indexOf(action.id) === -1) {
```

src/parser/jobs/nin/modules/TrickAttackUsage.tsx
```diff
@@ -0,0 +1,126 @@
...
+		}
+
+		if (this.taCasts.length > 0) {
+			const lastCast = this.taCasts[this.taCasts.length - 1]
+			const taAvailable = lastCast + this.data.actions.TRICK_ATTACK.cooldown
+			const downtime = this.downtime.getDowntime(
+				this.parser.fflogsToEpoch(taAvailable),
+				this.parser.fflogsToEpoch(event.timestamp),
+			)
+			this.lostTime += Math.max((event.timestamp - taAvailable) - downtime, 0)
```
ayyaruq, line 69: With a history of windows you could shift this to the onComplete callback and do it all at once, I'd expect that's faster. Noted the `indexOf` rather than `includes` above so I figure speed is of some concern here.

src/parser/jobs/nin/modules/TrickAttackUsage.tsx
```diff
@@ -0,0 +1,126 @@
...
+		if (this.taCasts.length > 0) {
+			const lastCast = this.taCasts[this.taCasts.length - 1]
+			// lostTime is only the time they were actually holding it off CD, but we want to add in the CD time of the final cast for
+			// calculating how many theoretical casts were lost. For example, 20s of holding + last cast 40s before the end of the fight
+			// would mean that they could've squeezed in an extra cast with perfect timing.
+			const lostCasts = Math.floor((this.lostTime + (this.parser.currentEpochTimestamp - lastCast)) / this.data.actions.TRICK_ATTACK.cooldown)
+			this.suggestions.add(new TieredSuggestion({
+				icon: this.data.actions.TRICK_ATTACK.icon,
+				content: <Trans id="nin.ta-usage.suggestions.missed.content">
+					Avoid holding <ActionLink {...this.data.actions.TRICK_ATTACK}/> for extended periods of time. It's typically ideal to use it as close to on cooldown as possible in order to keep it aligned with all the other raid buffs and personal burst windows, as well as maximizing the number of uses per fight.
```
ayyaruq, line 85:
```suggestion
					Avoid holding <ActionLink action="TRICK_ATTACK"/> for extended periods of time. It's typically ideal to use it as close to on cooldown as possible in order to keep it aligned with all the other raid buffs and personal burst windows, as well as maximizing the number of uses per fight.
```

src/parser/jobs/nin/modules/TrickAttackUsage.tsx
```diff
@@ -0,0 +1,126 @@
...
+				why: <Trans id="nin.ta-usage.suggestions.missed.why">
+					You delayed Trick Attack for a cumulative {this.parser.formatDuration(this.lostTime)}, costing you <Plural value={lostCasts} one="# potential use" other="# potential uses"/>.
+				</Trans>,
+			}))
+
+			const distanceFromOptimal = Math.abs(OPTIMAL_GCD_COUNT - this.gcdCount)
+			this.suggestions.add(new TieredSuggestion({
+				icon: this.data.actions.TRICK_ATTACK.icon,
+				content: <Trans id="nin.ta-usage.suggestions.opener.content">
+					Avoid unconventional timings for your first <ActionLink {...this.data.actions.TRICK_ATTACK}/> of the fight in order to line it up with all the other raid and personal buffs. In most openers, Trick Attack should be weaved in approximately 8-9 seconds into the fight.
```
ayyaruq, line 101:
```suggestion
					Avoid unconventional timings for your first <ActionLink action="TRICK_ATTACK"/> of the fight in order to line it up with all the other raid and personal buffs. In most openers, Trick Attack should be weaved in approximately 8-9 seconds into the fight.
```

src/parser/jobs/nin/modules/TrickAttackUsage.tsx
```diff
@@ -0,0 +1,126 @@
...
+				why: <Trans id="nin.ta-usage.suggestions.opener.why">
+					Your first Trick Attack was <Plural value={this.gcdCount} one="# GCD" other="# GCDs"/> into your opener.
+				</Trans>,
+			}))
+		} else {
+			// WHY ARE YOU EVEN PLAYING THIS JOB
+			this.suggestions.add(new Suggestion({
+				icon: this.data.actions.TRICK_ATTACK.icon,
+				content: <Trans id="nin.ta-usage.suggestions.none.content">
+					<ActionLink {...this.data.actions.TRICK_ATTACK}/> is the single most powerful raid buff in your kit and should be used on cooldown, or as close to it as possible depending on the flow of the fight.
```
ayyaruq, line 117:
```suggestion
					<ActionLink action="TRICK_ATTACK"/> is the single most powerful raid buff in your kit and should be used on cooldown, or as close to it as possible depending on the flow of the fight.
```

src/parser/jobs/nin/modules/TrickAttackUsage.tsx
```diff
@@ -0,0 +1,126 @@
...
+	private taCasts: number[] = []
+	private lostTime: number = 0
+	private gcdCount: number = 0
+	private castHook?: EventHook<Events['action']>
+
+	override initialise() {
+		this.mudraActions = MUDRAS.map(k => this.data.actions[k].id)
+
+		const playerFilter = filter<Event>().source(this.parser.actor.id)
+		this.castHook = this.addEventHook(playerFilter.type('action'), this.onCast)
```
ayyaruq, line 43: A comment here explaining why it's initialised before the TA cast might be useful. I assume it's for some kind of pre-pull memes.

ayyaruq (changes requested):
Couple of things, made suggestions for most - new DataLink syntax. I wonder if this could just be done by CDDT since it does basically the same math, but I figure if someone warranted that final condition it probably does need a suggestion. In the interests of assassinating (hah!) unnecessary indentation, you could lose the `else`, change the suggestion to tiered, and set the matcher to matchClosestHigher.

KelosPelmand:
> I wonder if this could just be done by CDDT 

I agree with this for the lostTime -> lostUsages part at least.  Some drift might get "refunded", so a straightforward calculation like this will miss some edge cases.  (See https://gist.github.com/KelosPelmand/88296d2135982b59c8123151ba7d8c4c#accumulated-drift-refunding for an explanation with tables).

AzariahAeroborn:
The TA module here is accounting for downtime at the "next expected use" time the same way that CDDT does, so it's lost time calculation is consistent with other jobs. As to whether it would save boilerplate or duplication, that's an exercise left to the job maintainer.

KelosPelmand:
It's counting use-to-use downtime correctly, but not applying refunds.  From what I see, it looks like the module will do this:

| Usage Number | Usage Time | Current time lost | Accumulated lostTime |
|--------------|------------|-------------------|----------------------|
| 1 | 0:00 | - | - |
| 2 | 1:02 | 0:01 | 0:02 |
| 3 | 2:10 | 0:08 | 0:10 |
| 4 | 3:30 | 0:20 | 0:30 |
| boss downtime | 3:55 to 4:35 |
| 5 | 4:35 | 0:00 (because of downtime adjustment, this is done correctly) | 0:30 (this is where it is wrong) |

The lostTime should be "refunded" at this point since there were only 3 possible uses and the drifting he did so far has no affect on what he may potentially get the rest of the fight.

CDDT only needs the lost time calculation and doesn't have to "apply refunds" because it just counts total possible uses and reports actual / possible.

src/parser/jobs/nin/modules/TrickAttackUsage.tsx
Toastdeib: No memes, this is for opener tracking purposes. There's a suggestion that dings you if your first Trick isn't weaved on the nth (currently 5th, may change in EW) GCD. I can add a comment tho.

src/parser/jobs/nin/modules/TrickAttackUsage.tsx
Toastdeib: Will adjust this, I'm still getting used to ts syntax.

src/parser/jobs/nin/modules/TrickAttackUsage.tsx
Toastdeib: TIL this syntax exists. I'll update to it, and use it going forward in my other Analyser migration branch that's currently in the works.

src/parser/jobs/nin/modules/TrickAttackUsage.tsx
Toastdeib: I can look at moving this around. I'm not sure it'll be meaningfully faster but it might be nice for logical separation to have the math happen at the end and have the event hooks just worry about the specific event in question.

Toastdeib:
Made a note in one of my replies to the file-specific comments, but I'm going to hold off on a CDDT port for this PR. Right now I'm just trying to get everything migrated over to ts/Analyser and I'd like to keep the ports as direct as possible. I'll definitely revisit the CDDT discussion in a future PR when I'm updating this module for EW.

src/parser/jobs/nin/modules/TrickAttackUsage.tsx
ayyaruq: Sounds good to me.

ayyaruq (approved):
Nicely done.

PR #1323, opened 2021-11-04
Nin analyser migration

Toastdeib:


src/parser/jobs/nin/modules/Ninjutsu.tsx
```diff
@@ -28,27 +30,28 @@ export class Ninjutsu extends Module {
...
-		this.addEventHook('cast', {by: 'player', abilityId: [this.data.actions.DOTON.id, this.data.actions.DOTON_TCJ.id]}, this.onDotonCast)
-		this.addEventHook('normaliseddamage', {by: 'player', abilityId: this.data.statuses.DOTON.id}, event => { this.current?.ticks.push(event.hitCount || 0) })
-		this.addEventHook('removebuff', {by: 'player', abilityId: this.data.statuses.DOTON.id}, this.finishDotonWindow)
+	override initialise() {
+		const playerFilter = filter<Event>().source(this.parser.actor.id)
+		this.addEventHook(playerFilter.type('action').action(oneOf([this.data.actions.HYOTON.id, this.data.actions.HYOTON_TCJ.id])), () => { this.hyotonCount++ })
+		this.addEventHook(playerFilter.type('action').action(this.data.actions.RABBIT_MEDIUM.id), () => { this.rabbitCount++ })
+		this.addEventHook(playerFilter.type('action').action(oneOf([this.data.actions.DOTON.id, this.data.actions.DOTON_TCJ.id])), this.onDotonCast)
+		this.addEventHook(playerFilter.type('damage').cause(this.data.matchCauseStatus(['DOTON'])), event => { this.current?.ticks.push(event.targets.length || 0) })
```
AzariahAeroborn, line 38:
```suggestion
		this.addEventHook(playerFilter.type('damage').cause(this.data.matchCauseStatus(['DOTON'])), event => { this.current?.ticks.push(event.targets.length ?? 0) })
```

We should move toward preferring ?? for nullish coalescing, rather than relying on falsey (granted, when we're coalescing a number to 0 it doesn't make any functional difference, but for making the most use of the language's type features...)

src/parser/jobs/nin/modules/Ninki.tsx
```diff
@@ -1,178 +1,108 @@
...
-interface DataPoint {
-	t: number,
-	y: number,
+// Constants
+const GCD_GAIN = 5
+const FINISHER_GAIN = 10
+const MUG_GAIN = 40
+const MEISUI_GAIN = 50
+const BUNSHIN_GAIN = 5
+const SPENDER_COST = -50
```
AzariahAeroborn, line 21: It's your preference, but from the perspective of the lint gods, these constants aren't required when using the GaugeModifier Map pattern.

src/parser/jobs/nin/modules/Ninki.tsx
```diff
@@ -1,178 +1,108 @@
...
-	))
-
-	private ninkiGcds: number[] = NINKI_GCDS.map(key => this.data.actions[key].id)
+	private ninkiGauge = this.add(new CounterGauge({
+		chart: {label: 'Ninki', color: JOBS.NINJA.colour},
+	}))
+
+	private ninkiModifiers = new Map<number, GaugeModifier>([
+		// Builders
+		[this.data.actions.SPINNING_EDGE.id, {action: GCD_GAIN}],
```
AzariahAeroborn, line 40: Builders on basic weaponskills that don't require a combo effect do still require a successful hit, and should be checking `damage` (and verifying that the damage event represents a successful hit via the isSuccessfulHit utility).  See also Shadow Fang, Death Blossom, Throwing Dagger, and Mug below.  (Though I'm not 100% sure on Mug since it's an Ability... the easiest way to test it ingame currently is to attack the wrong target in a firewall situation like Hades fire/ice adds or O12S M/F).

src/parser/jobs/nin/modules/Ninki.tsx
```diff
@@ -1,178 +1,108 @@
...
+	override initialise() {
+		super.initialise()
 
-	private onBuilderCast(event: CastEvent | LegacyComboEvent) {
-		const abilityId = event.ability.guid
-		// The .get() should never be undefined but we must appease the ts lint gods
-		this.addNinki(this.actionNinkiGain.get(abilityId) ?? 0, abilityId)
+		const playerFilter = filter<Event>().source(this.parser.actor.id)
+		const pets = this.parser.pull.actors.filter(actor => actor.owner === this.parser.actor).map(actor => actor.id)
+		this.addEventHook(playerFilter.type(oneOf(['action', 'combo'])).action(oneOf(Array.from(this.ninkiModifiers.keys()))), this.onGaugeModifier)
```
AzariahAeroborn, line 63: I kind of mentioned this in passing in one of the dev channels in a conversation with acc, but this really isn't a great pattern, since you wind up hooking too many events for builders that can only build on action.  Especially once you start accounting for builders that require a successful damage event, it's probably better to split these hooks to make sure the  hooks only get the needed event type for each builder.

AzariahAeroborn (changes requested):

src/parser/jobs/nin/modules/Ninki.tsx
Toastdeib: I did see the other Gauge implementations were doing this without constants. I suppose I'll swap over for consistency and to cut down on line count.

src/parser/jobs/nin/modules/Ninki.tsx
Toastdeib: I'll split them out, since I'll be changing `action` to `damage` anyway. At least for the damaging builders, the spenders and Meisui can stay as `action`.

src/parser/jobs/nin/modules/Ninki.tsx
Toastdeib: I'm going to operate on the assumption that Mug requires a successful hit for now; it's very likely the correct assumption, and I'm not in a position to test it at the moment anyway. We can revisit for EW (or if someone is able to do that test for me real quick).

src/parser/jobs/nin/modules/Ninki.tsx
```diff
@@ -35,22 +31,44 @@ export class Ninki extends CoreGauge {
 		chart: {label: 'Ninki', color: JOBS.NINJA.colour},
 	}))
 
+	private ninkiFilters = {
```
AzariahAeroborn, line 34: I like this pattern, very nice idea.

AzariahAeroborn (approved):

PR #1324, opened 2021-11-05
DNC - Port gauge to core

AkairyuGestalter:
Title, changes:
- Moves underflow backtrack correction from DNC gauge to core CounterGauge as an optional feature
- Ports DNC Gauge to use Core/CounterGauges instead of roll-my-own handling


src/parser/core/modules/Gauge/CounterGauge.ts
```diff
@@ -50,8 +69,10 @@ export class CounterGauge extends AbstractGauge {
 		super(opts)
 
 		this.minimum = opts.minimum || 0
-		this._value = opts.initialValue || this.minimum
+		this.initialValue = opts.initialValue || this.minimum
+		this._value = this.initialValue
 		this.maximum = opts.maximum || 100
+		this.correctHistory = opts.correctHistory || false
```
AzariahAeroborn, line 75: Especially now that `false` is a real default value for one of these options, we should use ?? coalescing rather than falsey to determine that we need to use the default.

src/parser/core/modules/Gauge/CounterGauge.ts
```diff
@@ -66,25 +87,81 @@ export class CounterGauge extends AbstractGauge {
...
+			this.correctGaugeHistory(delta, this._value)
+		}
+
+		this._value = newValue
+		this.pushHistory(reason, delta)
+	}
+
+	private correctGaugeHistory(spenderCost: number, currentGauge: number) {
+		// Get the last generation or initialisation event we've recorded
+		const lastGeneratorIndex = _.findLastIndex(this.history, event => event.reason === 'generate' || event.reason === 'init')
```
AzariahAeroborn, line 146: This approach is pretty specific to non-deterministic gauges; for a deterministic gauge it would be more correct to backtrack to the most recent init.  Could we reasonably give consumers an additional configuration option to choose which backtrack approach they want?  Otherwise you're going to get really funky looking gauge graphs if someone has some carry forward but not enough to immediately spend (e.g. a RDM comes in at 60 | 60 and builds up to a melee combo naturally before spending).

AzariahAeroborn (changes requested):

src/parser/core/modules/Gauge/CounterGauge.ts
AkairyuGestalter: Obviously I can't fully test the 'correcting but it's deterministic case' since DNC is most definitely not, but I've added the checks and logic for it in finding the last 'generator' index, and it's still working for DNC so... fingers crossed?

src/parser/core/modules/Gauge/CounterGauge.ts
```diff
@@ -66,25 +96,81 @@ export class CounterGauge extends AbstractGauge {
...
+		} else if (diff < 0 && this.correctHistory) {
+			this.correctGaugeHistory(delta, this._value)
+		}
+
+		this._value = newValue
+		this.pushHistory(reason, delta)
+	}
+
+	private correctGaugeHistory(spenderCost: number, currentGauge: number) {
+		// Get the initialisation event (or generation event if this gauge isn't deterministic) we've recorded
```
AzariahAeroborn, line 154:
```suggestion
		// Get the most recent initialisation event (or generation event if this gauge isn't deterministic) we've recorded
```

src/parser/core/modules/Gauge/CounterGauge.ts
```diff
@@ -66,25 +96,81 @@ export class CounterGauge extends AbstractGauge {
...
 	/** Set the current value of the gauge. Value will automatically be bounded to valid values. Value over the maximum will be tracked as overcap. */
-	set(value: number) {
-		this._value = Math.min(Math.max(value, this.minimum), this.maximum)
+	set(value: number, reason?: GaugeEventReason) {
+		const delta = Math.abs(value - this.value)
+		if (delta === 0 && reason != null) { return }
+
+		if (reason == null) {
+			reason = value > this._value ? 'generate' : 'spend'
+		}
```
AzariahAeroborn, line 138: I kind of agree with the comments ack made in #dev-general when we were first discussing this; it would be better to keep the +/- sign on the delta, so that a consumer can see what's happening with the history without needing to check two fields.  

That may also make the GaugeEventReason field obsolete, if we can treat a reset/init as delta = undefined and then have the backtrack look for those events instead.  But I could see an argument for not turning delta into a tri-state number, whichever works better for the backtrack logic there.

AkairyuGestalter: I'm going to leave the reason field there, since I'd rather keep delta as a simple numeric, rather than `number?` but yeah, we can make the delta that goes into the history be signed. Still have to Math.abs it for the history correction, but there's no reason that can't be done inside that function.

AzariahAeroborn (approved):

PR #1325, opened 2021-11-05
Scaffold BLM and DNC data for Endwalker

AkairyuGestalter:
Initial action/status data scaffolding for BLM and DNC:
- Entries for new actions and statuses
- Updates to existing actions and statuses (proc status durations, sharpcast charges, recasts, etc)
- Moved Sleep to caster Role actions

Readded avatar/contributor, and reassigned as BLM/DNC maintainer

Updated DNC AoEUsages due to changes to DNC proc structure


AkairyuGestalter:
Marking as draft since I'm making conflicting changes in BLM Gauge with it's port to core. Since we've got more time before Endwalker now I'd rather pull those in here and resolve them before PR review, rather than give ack a headache merging that into the Endwalker branch

AkairyuGestalter:
Caught up with recent changes to Gauge & resolved the merge conflicts. Ready for review.

src/data/ACTIONS/root/DNC.ts
```diff
@@ -8,41 +8,71 @@ export const DNC = ensureActions({
 		name: 'Single Standard Finish',
 		onGcd: true,
 		cooldown: 1500,
+		statusesApplied: [
+			'STANDARD_FINISH',
+			'STANDARD_FINISH_PARTNER',
+			'ESPRIT',
+		],
```
AzariahAeroborn, line 15: As much as it's technically wrong data, it's probably better if you only put the statusesApplied on the "got all steps" finishes - these are basically here to feed precast synthing, and the adapter for prepull statuses implicitly assumes that only one action can add a given status -- behavior isn't guaranteed deterministic if multiple actions have the same statusesApplied (though it will generally pick the "first" one in the status array, which is generally going to be the first one defined in the data files).

src/parser/jobs/dnc/modules/Procs.tsx
```diff
@@ -1,37 +1,41 @@
...
+			consumeActions: [this.data.actions.TILLANA],
 		},
 	]
 
 	override showDroppedProcSuggestion = true
 	override droppedProcIcon = this.data.actions.FOUNTAINFALL.icon
 	override droppedProcContent =
 		<Trans id="dnc.procs.suggestions.drops.content">
-			Avoid dropping your procs unless absolutely necessary. If you have to drop one to keep your Esprit from overcapping, <ActionLink {...this.data.actions.RISING_WINDMILL} /> or <ActionLink {...this.data.actions.REVERSE_CASCADE} /> will lose the least DPS overall.
+			Avoid dropping your procs unless absolutely necessary. If you have to drop one to keep your Esprit from overcapping, <DataLink action="RISING_WINDMILL" /> or <DataLink action="REVERSE_CASCADE" /> will lose the least DPS overall.
```
AzariahAeroborn, line 38: With the consolidation to two shared procs, should we update this copy to use the name of the shared proc instead of the actions that can cast it?  Will we even still need this module, since proc duration should more than cover using up two procs?

AzariahAeroborn (changes requested):

src/data/ACTIONS/root/DNC.ts
AkairyuGestalter: Fair enough, I wasn't _entirely_ sure what the statusesApplied field was actually being used for so I was trying to be as complete as I could. If that's all it's mainly for then I agree we can dispense with those (and the data for Tillana too).

src/parser/jobs/dnc/modules/Procs.tsx
AkairyuGestalter: Yeah, we can change the copy. I'd rather leave the suggestion present because while it's significantly less likely now, it's still technically possible to wind up in a situation where Tech finish's Esprit spam makes it difficult to get all the procs in. I'll re-evaluate the necessity of the suggestion after we actually get our hands on the post-expac build though.

AzariahAeroborn (approved):

PR #1326, opened 2021-11-05
core: let BuffWindow accept Status[]

ebshimizu:
drg module maintainer was on the floor during the original PR review whoops
Let BuffWindow accept a Status[] again and updated the Dragon Sight module to use the BuffWindow module better.

### DIFF
--- a/src/parser/core/modules/ActionWindow/windows/BuffWindow.ts
+++ b/src/parser/core/modules/ActionWindow/windows/BuffWindow.ts
@@ -1,7 +1,9 @@
 import {Status} from 'data/STATUSES'
 import {Event, Events} from 'event'
+import _ from 'lodash'
 import {TimestampHook, TimestampHookArguments} from 'parser/core/Dispatcher'
-import {filter, noneOf} from '../../../filter'
+import {ensureArray} from 'utilities'
+import {filter, noneOf, oneOf} from '../../../filter'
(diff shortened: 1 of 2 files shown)


src/parser/core/modules/ActionWindow/windows/BuffWindow.ts
```diff
@@ -47,15 +49,15 @@ export abstract class BuffWindow extends ActionWindow {
...
+			.status(oneOf(ensureArray(this.buffStatus).map(s => s.id)))
 
 		this.addEventHook(buffFilter.type('statusApply'), this.startWindowAndTimeout)
 		this.addEventHook(buffFilter.type('statusRemove'), this.endWindowByStatus)
 	}
 
 	private startWindowAndTimeout(event: Events['statusApply']) {
 		this.onWindowStart(event.timestamp)
-		const duration = this.buffStatus.duration
+		const duration = _.max(ensureArray(this.buffStatus).map(s => s.duration))
```
KelosPelmand, line 60: Since we know which status is being applied here, this should probably use the actual status rather than max duration of all possible.

```suggestion
		const duration = this.data.getStatus(event.status)?.duration
```

KelosPelmand (approved):

AzariahAeroborn (approved):

PR #1329, opened 2021-11-07
BLM -  Port Gauge to core

AkairyuGestalter:
Pls2ignore stupid revision history. Changes herein:
- Fixes TimerGauge graphing output (linear curve, data array construction, tooltip 'current' value calculation)
- Adds Dowtime window API to TimerGauge to find out when the timer was allowed to expire
- Port BLM's Gauge to Core & assorted coding style cleanup
- Adds display for BLM's AF/UI/Polyglot timers
- Adds tracking/core hooks for statuses that pause TimerGauges (Currently just TEA's timestop, fixes all the BrokenLogs we get from BLM Gauge thinking AF/UI/Polyglot dropped)
- Due to above, and the likelihood that just about all Gauges will need it, the Data dependency for Gauges is now protected in Core, instead of private in subclasses


src/parser/jobs/blm/modules/Gauge.tsx
```diff
@@ -40,99 +49,135 @@ export interface BLMGaugeState {
...
-}
-
-export default class Gauge extends Analyser {
+/** Graph colors/fade settings */
+const STANCE_FADE = 0.5
+const GAUGE_FADE = 0.25
+const TIMER_FADE = 0.75
+/* eslint-disable @typescript-eslint/no-magic-numbers */
+const ICE_COLOR = Color.rgb(47, 113, 177)
+const FIRE_COLOR = Color.rgb(210, 62, 38)
```
ayyaruq, line 68: Can convert these to hex and use the string form to avoid the linter override.

src/parser/jobs/blm/modules/Gauge.tsx
```diff
@@ -2,36 +2,45 @@
...
 import {FIRE_SPELLS, ICE_SPELLS_TARGETED, ICE_SPELLS_UNTARGETED} from './Elements'
 
+/** Configuration */
 const ENOCHIAN_DURATION_REQUIRED = 30000
 export const ASTRAL_UMBRAL_DURATION = 15000
 export const MAX_ASTRAL_UMBRAL_STACKS = 3
 export const MAX_UMBRAL_HEART_STACKS = 3
 const MAX_ASTRAL_UMBRAL_CAST_SCALAR = 0.5
 const FLARE_MAX_HEART_CONSUMPTION = 3
 const MAX_POLYGLOT_STACKS = 2
```
ayyaruq, line 27: Can consider renaming these imo. Lead with ASTRAL_UMBRAL, ENOCHIAN, POLYGLOT, etc and use the suffix to indicate what it's used for. Mostly just nicer for intellisense, but it's easier to read.

src/parser/jobs/blm/modules/Gauge.tsx
```diff
@@ -40,99 +49,135 @@ export interface BLMGaugeState {
...
+		maximum: ENOCHIAN_DURATION_REQUIRED,
+		onExpiration: this.onGainPolyglot.bind(this),
+		graph: {
+			label: <Trans id="blm.gauge.resource.polyglot-timer">Polyglot Timer</Trans>,
+			color: POLYGLOT_COLOR.fade(TIMER_FADE),
+		},
+	}))
+	private enochianGauge = this.add(new CounterGauge({
+		maximum: 1,
+	}))
```
ayyaruq, line 159: I'd likely just track this as a boolean rather than gauge. It's either on or off.

src/parser/jobs/blm/modules/Gauge.tsx
```diff
@@ -383,140 +413,88 @@ export default class Gauge extends Analyser {
...
-				refundTime += endOfDowntime - downtime.start // If the end of this enochian downtime occurred during an unableToAct time frame that lasted longer than the AF/UI timeout, refund that downtime
-			}
-		})
-		return Math.floor(Math.max(0, time - refundTime)/ENOCHIAN_DURATION_REQUIRED)
+		this.polyglotTimer.getDowntimeWindows().filter(window => window.start - this.parser.pull.timestamp > this.data.actions.FIRE_IV.castTime)
+			.forEach(downtime => {
+				if (this.unableToAct.getWindows({
+					start: downtime.start,
+					end: downtime.start,
+				}).filter((uta) => Math.max(0, uta.end - uta.start) >= ASTRAL_UMBRAL_DURATION).length > 0) {
```
ayyaruq, line 482: Might not need the extra parentheses on uta here.

ayyaruq (commented):
A few points here, not marking as required. Since you fab the gauge events, maybe consider having a separate module hook that to manage speed changes, instead of having all the value checks in here. Not sure how complex that would be tho.

With the `onCast` function, you might find it simpler to check the event type rather than if the field is in it, since we know an Action event always has an action member, while a cause might have an action. Could be clearer to read. You already type check slightly further down the function anyway.

I'd also consider adding a getter to CounterGauge as a boolean for if the gauge is capped, at least for a given time. Would simplify a lot of your checks since you can just `this.astralFireGauge.capped` rather than comparing against max everywhere. Give it another name if need be.

src/parser/core/modules/Gauge/TimerGauge.ts
```diff
@@ -250,4 +261,54 @@ export class TimerGauge extends AbstractGauge {
 	setRemoveTimestampHook(value: Analyser['removeTimestampHook']) {
 		this._removeTimestampHook = value
 	}
+
+	private internalDowntime(start = this.parser.pull.timestamp, end = this.parser.currentEpochTimestamp) {
```
AzariahAeroborn, line 265: I know that "downtime" makes sense when talking about time a timer based job gauge isn't running, but the term is also overloaded with unable to act time, and we're going to need to be referring to both when computing how long the job gauge wasn't running for that it should have been, so is there an alternate term we could use here to describe "time the gauge wasn't running or paused"?

src/parser/core/modules/Gauge/TimerGauge.ts
```diff
@@ -250,4 +261,54 @@ export class TimerGauge extends AbstractGauge {
...
+		})
+
+		return finalDowntimes
+	}
+
+	public isDowntime(when = this.parser.currentEpochTimestamp) {
+		return this.internalDowntime(when, when).length > 0
+	}
+
+	public getDowntime(start = this.parser.pull.timestamp, end = this.parser.currentEpochTimestamp) {
```
AzariahAeroborn, line 295: This needs to consider and exclude UTA time from the total gauge downtime.

src/parser/core/modules/Gauge/Gauge.tsx
```diff
@@ -48,16 +67,34 @@ export class Gauge extends Analyser {
...
+	protected onDeath() {
 		this.pauseGeneration = true
 		this.gauges.forEach(gauge => gauge.reset())
 	}
 
-	private onRaise() {
+	protected onRaise() {
 		this.pauseGeneration = false
 		this.gauges.forEach(gauge => gauge.raise())
 	}
```
AzariahAeroborn, line 78: I don't see any users being updated in this PR to use these - what's the reasoning behind exposing them?  If we are going to expose these, we should move the pauseGeneration infrastructure out to specifically the dancer gauges and make that behavior an implementation detail of gauge modules that need it.

src/parser/core/modules/Gauge/TimerGauge.ts
```diff
@@ -182,57 +192,58 @@ export class TimerGauge extends AbstractGauge {
...
 				// with its state at this point in time - we'll end up with two points showing the update on
 				// this timestamp.
-				const prev = data[length - 1]
-				prev.current = (prev.current || this.minimum / 1000) + ((prev.time - relativeTimestamp) / 1000)
-				prev.time = relativeTimestamp
+				prev.current = (prev.current || this.minimum / 1000) + ((prev.time - entry.timestamp) / 1000)
+				prev.time = entry.timestamp - 1
 
-			} else {
+			} else if (length > 0) {
```
AzariahAeroborn, line 210: I commented a little more on discord about an alternate approach for history tracking that would streamline a lot of this graph generation code, but if we keep the current approach, we don't need this if (length > 0) here - no harm done by adding a 0 at the timestamp of the first generation event.

AzariahAeroborn (changes requested):

AzariahAeroborn (approved):

PR #1330, opened 2021-11-10
RDM Analyser and TS Migration

Cephid:
This finishes migrating all of RDM onto Analyser and into TS.  MeleeCombos and Gauge were the big ports in this PR; the logic for them was left pretty much as is.  A few minor things were added the RDM TCs had asked for, ordering, Embolden info.  We couldn't decide if we wanted to keep CAC/Engagement so for now we've elected to disconnect the modules until we decide if they should stay or not while leaving the files in place.

I tried to find and fix all spacing issues since lint formatting isn't functional, I'm not sure I caught them all though.


KelosPelmand (commented):
Just commenting about History usage, didn't look through all the other logic.

src/parser/jobs/rdm/modules/MeleeCombos.tsx
```diff
@@ -0,0 +1,624 @@
...
+						})
+					}
+				</Table.Body>
+			</Table>
+		)
+	}
+
+	private startCombo(event: Events['action']) {
+		this._meleeCombos.openNew(event.timestamp)
+		const current = this._meleeCombos.getCurrent()
```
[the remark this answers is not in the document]
Cephid: Updated with suggestion.

src/parser/jobs/rdm/modules/MeleeCombos.tsx
```diff
@@ -0,0 +1,624 @@
...
+				this.startCombo(event)
+			} else {
+				if (this._meleeCombos.getCurrent() == null) {
+					return
+				}
+
+				if (action.combo.from) {
+					const fromOptions = Array.isArray(action.combo.from) ? action.combo.from : [action.combo.from]
+					const current = this._meleeCombos.getCurrent()
+					if (current) {
```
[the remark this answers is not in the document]
Cephid: Updated with suggestion.

src/parser/jobs/rdm/modules/DisplayOrder.ts
```diff
@@ -0,0 +1,9 @@
+export const DisplayOrder = {
+	DualCast: 1,
+	ComboIssues: 2,
+	InterruptedCasts: 3,
+	WeavingIssues: 4,
+	MeleeCombo: 5,
+	Embolden: 6,
+	Tinctures: 7,
```
AzariahAeroborn, line 8: style nit: For consistency with other jobs, these should use the CONSTANT_VALUE_ALL_CAPS naming pattern

src/parser/jobs/rdm/modules/Gauge.tsx
```diff
@@ -0,0 +1,307 @@
...
+}
+
+export class Gauge extends CoreGauge {
+	static override title = t('rdm.gauge.title')`Mana Gauge Usage`
+
+	@dependency private data!: Data
+	@dependency private suggestions!: Suggestions
+	@dependency private statistics!: Statistics
+
+	private _whiteManaGauge = this.add(new CounterGauge({
```
AzariahAeroborn, line 27: TypeScript style rule: Do not use _ for private properties; we use that convention for JavaScript because there's no accessibility modifier, but it's not necessary or desired in TypeScript files.

src/parser/jobs/rdm/modules/Gauge.tsx
```diff
@@ -0,0 +1,307 @@
...
+	}
+	private _severityLostMana = {
+		1: SEVERITY.MINOR,
+		20: SEVERITY.MEDIUM,
+		80: SEVERITY.MAJOR,
+	}
+	public readonly _manaDifferenceThreshold = 30
+	private readonly _manaLostDivisor = 2
+	private readonly _manaficationMultiplier = 2
+	public readonly _manaCap = 100
```
AzariahAeroborn, line 77: These would be preferred as non-exported `const` values outside of the analyser class definition (typically right after the import declarations).  Use the `const MANA_DIFFERENCE_THRESHOLD` naming convention.

src/parser/jobs/rdm/modules/Gauge.tsx
```diff
@@ -0,0 +1,307 @@
...
+		const modifier = this._gaugeModifiers.get(event.cause.action)
+
+		if (modifier == null) {
+			return
+		}
+
+		const amount = modifier
+		const penalized = this.isOutOfBalance()
+		const whiteModified = penalized.white ? Math.ceil(amount.white/this._manaLostDivisor) : amount.white
+		const blackModified = penalized.black ? Math.ceil(amount.black/this._manaLostDivisor) : amount.black
```
AzariahAeroborn, line 131: The game floors the mana gained when out of balance, not ceiling (using an 11 mana skill gains 5 mana of the low mana)
```suggestion
		const whiteModified = penalized.white ? Math.floor(amount.white/this._manaLostDivisor) : amount.white
		const blackModified = penalized.black ? Math.floor(amount.black/this._manaLostDivisor) : amount.black
```

src/parser/jobs/rdm/modules/Gauge.tsx
```diff
@@ -0,0 +1,307 @@
...
+
+		this._whiteManaGauge.modify(whiteModified)
+		this._blackManaGauge.modify(blackModified)
+		//Statistics Gathering
+		if (penalized.white) {
+			this._manaStatistics.white.imbalanceLoss = whiteModified
+		}
+		if (penalized.black) {
+			this._manaStatistics.black.imbalanceLoss = blackModified
+		}
```
AzariahAeroborn, line 148: This is not adding the correct amount of loss due to imbalance, you need to look at the difference between base amount and the amount actually gained.  Also, you really could get rid of the penalized checks here, since amount.white - whiteModified will be 0 when in balance.
```suggestion
		this._manaStatistics.white.imbalanceLoss = amount.white - whiteModified
		this._manaStatistics.black.imbalanceLoss = amount.black - blackModified
```

src/parser/jobs/rdm/modules/Gauge.tsx
```diff
@@ -0,0 +1,307 @@
...
+			icon: this.data.actions.VERHOLY.icon,
+			content: <Fragment>
+				<Trans id="rdm.gauge.suggestions.white-mana-wasted-content">Ensure you don't overcap your White Mana before a combo, overcapping White Mana indicates your balance was off; and you potentially lost out on Enchanted Combo damage.  You should look to execute at 80/80 or as close to it as possible.</Trans>
+			</Fragment>,
+			tiers: this._severityWastedMana,
+			value: this._whiteManaGauge.overCap,
+			why: <Fragment>
+				<Trans id="rdm.gauge.suggestions.white-mana-wasted-why">You lost {this._whiteManaGauge.overCap} White Mana due to capped Gauge resources</Trans>
+			</Fragment>,
+		}))
```
AzariahAeroborn, line 214: I know that this is somewhat a pre-existing condition, but it would probably be better to collapse these suggestions to consider both the aggregate white and black mana losses and output a single suggestion per type (e.g. based on total overcap, make a suggestion and show the two overcaps in the why).  It would help cut down on suggestion bloat from this module and allow other modules to be more relatively important, rather than this module potentially outputting 8 suggestions for what's really only 4 problems.

src/parser/jobs/rdm/modules/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+import {DualCast} from './Dualcast'
+import {Embolden} from './Embolden'
+import {Gauge} from './Gauge'
+import {GeneralCDDowntime} from './GeneralCDDowntime'
+import {Interrupts} from './Interrupts'
+import {MeleeCombos} from './MeleeCombos'
+import {Procs} from './Procs'
+import {Weaving} from './Weaving'
+
+export {DualStatistic} from 'parser/jobs/rdm/statistics/DualStatistic'
```
AzariahAeroborn, line 13: Are there other jobs consuming this component?  If so, we should probably move the component to core instead of having it be exported out of the job in a future update (I realize it's pre-existing and thus out of scope for this update).

AzariahAeroborn (changes requested):

src/parser/jobs/rdm/modules/MeleeCombos.tsx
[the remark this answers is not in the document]
AzariahAeroborn: Now that you're checking and early returning on line 140, this if check should be removed, it's redundant now.

AzariahAeroborn (changes requested):
Thanks github for not rendering large diffs by default.

src/parser/jobs/rdm/modules/MeleeCombos.tsx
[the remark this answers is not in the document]
Cephid: updated with suggestion.

src/parser/jobs/rdm/modules/DisplayOrder.ts
Cephid: updated with suggestion.

src/parser/jobs/rdm/modules/Gauge.tsx
Cephid: updated with suggestion.

src/parser/jobs/rdm/modules/Gauge.tsx
Cephid: updated with suggestion.

src/parser/jobs/rdm/modules/Gauge.tsx
Cephid: updated with suggestion.  Also I didn't realize I could have just clicked commit here heh.

src/parser/jobs/rdm/modules/Gauge.tsx
Cephid: updated slightly differently than suggested, it wasn't doing += so it was only capturing the last penalty of the set.  Fixed the other manaStatistic because it had the same problem.

src/parser/jobs/rdm/modules/Gauge.tsx
Cephid: Guidance I got was that they still wanted the individual types of ways to lose mana to appear separate but were fine with white and black collapsing together.  Collapsed them and threw on an icon ternary based on loss.  It was simple so I went ahead and did it.

src/parser/jobs/rdm/modules/index.tsx
Cephid: I do not recall if I was asked to keep it with RDM when I wrote it at the time; this was when some TS was first being added in iirc.  If Ackwell would like it moved to core I'll be happy to do that either in this PR or in a separate one after this one goes through.

AzariahAeroborn (approved):

PR #1331, opened 2021-11-11
feat: split ResourceSimulator into a blood Gauge analyser and a MPUsage analyser

AzariahAeroborn:
Migrates one of two remaining DRK job modules to analyser.  Also separates concerns between the blood gauge and MP overcap tracking (removing MP gauge tracking in favor of the automated graph on the timeline).

In order to track Dark Arts procs / TBN expiration, we need the optional absorbed property from statusRemove events to infer whether or not the shield popped.  For FFLogs data sources, continue using FFLogs' inference as we have been - other data sources will need to simulate shield consumption and report an `absorbed` amount accordingly.


src/parser/jobs/drk/modules/Gauge.tsx
```diff
@@ -0,0 +1,107 @@
...
+		if (action != null && action.onGcd) {
+			this.bloodGauge.modify(BLOOD_WEAPON_GAIN)
+		}
+	}
+
+	private onComplete() {
+		this.suggestions.add(new TieredSuggestion({
+			icon: this.data.actions.DELIRIUM.icon,
+			content: <Trans id="drk.gauge.suggestions.loss.content">
+					Avoid letting your Blood Gauge overcap - the wasted resources may cost you uses of your spenders over the course of the fight.
```
AkairyuGestalter, line 98: Can you fix this spacing here, since lint isn't doing it for us?

AkairyuGestalter (changes requested):
One minor nit

AkairyuGestalter (approved):

PR #1343, opened 2021-11-15
feat: convert WHM oGCD tracking to use core CooldownDowntime

AzariahAeroborn:
Two minor changes necessary in core CDDT to expose a way to get the number of actual usages of a tracked action and provide implementers a way to add their own suggestions for things that aren't covered by the checklist.


ackwell (approved):
ngl this is kind of a skim read but hey what could go wrong

PR #1345, opened 2021-11-16
feat: port Requiescat to core BuffWindow

AzariahAeroborn:
This required a moderate change to TrackedActions to allow for a group of interchangeable actions (since Holy Circle can be substitued for Holy Spirit during Requiescat if there are multiple targets to hit).  Updated consumers that had overrides that considered the TrackedAction so they work with the new schema.


src/parser/core/modules/ActionWindow/evaluators/ExpectedActionsEvaluator.tsx
```diff
@@ -77,7 +78,7 @@ export class ExpectedActionsEvaluator implements WindowEvaluator {
 	}
 
 	protected countUsed(window: HistoryEntry<EvaluatedAction[]>, action: TrackedAction) {
-		return window.data.filter(cast => cast.action.id === action.action.id).length
+		return window.data.filter(cast => ensureArray(action.action).map(a => a.id).includes(cast.action.id)).length
```
ackwell, line 81:
```suggestion
		return window.data.filter(cast => ensureArray(action.action).some(action => action.id === cast.action.id)).length
```

src/parser/core/modules/ActionWindow/evaluators/LimitedActionsEvaluator.tsx
```diff
@@ -52,7 +53,7 @@ export class LimitedActionsEvaluator implements WindowEvaluator {
 	public output(): undefined { return undefined }
 
 	private countUsed(window: HistoryEntry<EvaluatedAction[]>, action: TrackedAction) {
-		return window.data.filter(cast => cast.action.id === action.action.id).length
+		return window.data.filter(cast => ensureArray(action.action).map(a => a.id).includes(cast.action.id)).length
```
ackwell, line 56:
```suggestion
		return window.data.filter(cast => ensureArray(action.action).some(action => action.id === cast.action.id)).length
```

src/parser/jobs/brd/modules/RagingStrikes.tsx
```diff
@@ -56,7 +56,7 @@ class BarrageEvaluator extends ExpectedActionsEvaluator {
 	}
 
 	override countUsed(window: HistoryEntry<EvaluatedAction[]>, action: TrackedAction) {
-		if (action.action.id === this.barrageId) {
+		if (ensureArray(action.action)[0].id === this.barrageId) {
```
ackwell, line 59: i'm not keen on shovelling off this array->first operation onto every single consumer, especially given by and large they have no realistic reason to need to deal with it.

i think there's a few ways to go about reducing the impact of this change, the main ones i'd consider are
- updating `TrackedAction::action` to just be `TrackedActions::actions` and be done with it - this doesn't make it less impact, but it does make it much more predictable
- Using some generics and suchforth to allow using wider types of `TrackedAction` that incorporate extra info.

src/parser/jobs/pld/modules/Requiescat.tsx
```diff
@@ -23,223 +19,109 @@ const SEVERITIES = {
...
 
-		const lastRequiescat = this.lastRequiescat
+		this.addEventHook({type: 'action', source: this.parser.actor.id, action: this.data.actions.REQUIESCAT.id}, () => this.requiescatUsages.casts++)
+		this.addEventHook({type: 'statusApply', target: this.parser.actor.id, status: this.data.statuses.REQUIESCAT.id}, () => this.requiescatUsages.buffs++)
 
-		// If we're still in the considered window, log our actions to it
-		if (lastRequiescat != null && lastRequiescat.end == null) {
-			lastRequiescat.rotation.push(event)
-		}
+		this.addEvaluator(new RequiescatUsageEvaluator(this.requiescatUsages, this.data.actions.REQUIESCAT.icon))
```
ackwell, line 95: mutable cross-object references aaaaaaaaaaaaaaaa
thanks i hate it

not sure i know a better way without thinking about it more tho

src/parser/jobs/pld/modules/Requiescat.tsx
```diff
@@ -23,223 +19,109 @@ const SEVERITIES = {
...
+		const expectedGcds = Math.floor(windowDuration / this.globalCooldown.getEstimate()) + 1
+		if (expectedGcds < EXPECTED_HOLY_SPIRITS + 1 && ensureArray(action.action)[0] === this.data.actions.HOLY_SPIRIT) {
+			return (expectedGcds - 1) - EXPECTED_HOLY_SPIRITS
 		}
 
-		const lastRequiescat = this.lastRequiescat
-
-		if (lastRequiescat != null) {
-			lastRequiescat.end = event.timestamp
+		if (expectedGcds === 0 && ensureArray(action.action)[0] === this.data.actions.CONFITEOR) {
```
ackwell, line 109: looking at this adjustment function, i'm really starting to question having >1 tracked action per evaulator. it just leads to handling that looks like

```
if it's skill one
	do logic
	return

if it's skill two
	do logic
	return

...
``` 
which seems silly. out of scope for this pr but possibly worth refactoring evaluators down to single responsibility

ackwell (commented):
hm, this pr really shows some weaknesses in the design here. i think some stuff needs looking at before we get too neck deep with lots of consumers

PR #1350, opened 2021-11-21
[GNB] Convert and Split Ammo.tsx to CoreGauge and AoEUsages

Rjb99:
Converted Ammo.tsx to CoreGauge and moved all AoE related functions to a proper AoE checking to its own file.
Changed Ammo to be shown on the timeline and not as a separate chart.


src/parser/jobs/gnb/modules/AoE.tsx
```diff
@@ -0,0 +1,55 @@
...
+			- You wish to end cartridge neutral
+			- You were not mid combo:
+		650p in 2 GCDs
+		Burst Strike(500p), Demon Slice(150p), Demon Slaughter(0p Ghosts)
+		Versus
+		500p in 2 GCDs
+		Keen Edge(200p), Brutal Shell(300p), Solid Barrel(0p Ghosts)
+		If you would get all 3 GCDs with no ghosting then Both combos are 900 potency
+	I consider this to be high level enough that if you are doing this, you know what you are doing.
+	*/
```
ayyaruq, line 25: Thanks I hate it. Idle comment and not something you need to really change, but if you're using expanded comment syntax, consider keeping the left asterisks the whole way down so it's slightly easier to read (and more consistent with the rest of the codebase).

src/parser/jobs/gnb/modules/AoE.tsx
```diff
@@ -0,0 +1,55 @@
...
+	]
+
+	protected override adjustMinTargets(event: Events['damage'], minTargets: number): number {
+		if (event.cause.type !== 'action') {
+			return minTargets
+		}
+
+		const action = this.data.getAction(event.cause.action)
+
+		// How in the fuck did we even get here tbh
```
ayyaruq, line 47: Respecc. I recognise this comment.

src/parser/jobs/gnb/modules/AoE.tsx
```diff
@@ -0,0 +1,55 @@
...
+			minTargets: 3,
+		},
+		{
+			aoeAction: this.data.actions.FATED_CIRCLE,
+			stActions: [this.data.actions.BURST_STRIKE],
+			minTargets: 2,
+		},
+	]
+
+	protected override adjustMinTargets(event: Events['damage'], minTargets: number): number {
```
ayyaruq, line 40: You only ever return minTargets here, so this function is kinda pointless? I'd double check but since this seems to be cloned from MNK, consider if the filename being `.ts` is good enough too.

src/parser/jobs/gnb/modules/Ammo.tsx
```diff
@@ -42,371 +19,93 @@ const LEFTOVER_AMMO_SEVERITY_TIERS = {
...
-		this.dumpRemainingResources()
-	}
+	private ammoGauge = this.add(new CounterGauge({
+		maximum: MAX_AMMO,
+		/* We'll see how people like it being in the timeline over a seperate chart.
+		I've grown to prefer it on the timeline myself
+		chart: {label: 'Ammo', color: JOBS.GUNBREAKER.colour}, */
+		graph: {
+			handle: 'ammo',
+			label: <Trans id="gnb.gauge.resource.ammoLabel"> Ammo </Trans>,
```
ayyaruq, line 36: Prob don't need these extra spaces if it's going in the timeline, worth checking but nbd.

src/parser/jobs/gnb/modules/Ammo.tsx
```diff
@@ -42,371 +19,93 @@ const LEFTOVER_AMMO_SEVERITY_TIERS = {
...
+			color: JOBS.GUNBREAKER.colour,
+		},
+	}))
+
+	// Used for Checklist as CounterGauge seems to lack total tracking at this time.
+	private totalGeneratedAmmo = 0
```
ayyaruq, line 42: I feel like this can be calculated from history but I haven't the faintest idea how.

ayyaruq (changes requested):
Not sure I like the mixing and matching of ammo with cartridges but whatever. AoE needs a bit of cleanup for the unnecessary function, all my other comments are just nits and can be taken as you will.

I explicitly didn't write anything for your modifier map or hooks, but consider splitting costs into gains vs spends and having separate hooks (so you don't need to have negative costs and can invert as necessary), or otherwise multi-lining the hook filter since you only use the playerFilter const the one time so it's a relatively pointless assignment. Totally up to you tho, works fine as it is.

src/parser/jobs/gnb/modules/AoE.tsx
Rjb99: I did indeed borrow the Punch boy code. mostly cause MNK > SAM in file order.
minTargets is gone now.

src/parser/jobs/gnb/modules/Ammo.tsx
Rjb99: Possibly looping thru history after the fact and adding based on skill. but I think just keeping a rolling tally at the same time is the cleanest way at the moment

src/parser/jobs/gnb/modules/Ammo.tsx
```diff
@@ -42,371 +19,93 @@ const LEFTOVER_AMMO_SEVERITY_TIERS = {
...
+
+	])
+
+	override initialise() {
+		super.initialise()
+
+		const ammoActions = Array.from(this.ammoModifiers.keys())
+
+		this.addEventHook(filter<Event>().source(this.parser.actor.id).type(oneOf(['action', 'combo']))
+			.action(oneOf(ammoActions)), this.onGaugeModifier)
```
ayyaruq, line 62:
```suggestion
		this.addEventHook(
				filter<Event>()
						.source(this.parser.actor.id)
						.type(oneOf(['action', 'combo']))
						.action(oneOf(ammoActions)),
				this.onGaugeModifier,
		)
```

ayyaruq (changes requested):
One quick change, did it as a suggestion for you. Looks good!

ayyaruq (approved):

PR #1351, opened 2021-11-21
NIN Endwalker data updates

Toastdeib:
Starting small here; this just removes the now-defunct Shadow Fang and a couple of statuses, plus adds four new actions. It also includes small updates to two modules to account for the loss of SF.


AkairyuGestalter (changes requested):
Changes themselves look good, but iirc Ack wanted us to start having changelogs for anything that touches job code. Probably want to re-add yourself as a contributor anyways, since those got nuked :)

Toastdeib:
shit u rite. I'll add that in this evening.

AkairyuGestalter (approved):

PR #1353, opened 2021-11-22
Stub RPR data

ayyaruq:
Attributes might be incorrect, I went off of weaponskill vs spell. Combos are stubbed as we don't know the IDs yet.

### DIFF
--- a/src/data/ACTIONS/root/RPR.ts
+++ b/src/data/ACTIONS/root/RPR.ts
@@ -1,5 +1,320 @@
+import {Attribute} from 'event'
 import {ensureActions} from '../type'
 
 export const RPR = ensureActions({
-	// TODO: Add actions here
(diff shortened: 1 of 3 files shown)


AkairyuGestalter (changes requested):
Actions/Statuses look fine as a first pass given that we won't know icons/ids/cdgs until later. Should probably have yourself an actual changelog though, even if it's just something like 'Initial data...' etc.

ayyaruq:
I opted to not write a changelog for the data as it's just stubbed and there's no analyser code yet. Once we have something to analyse, or at least confirmed data, I'll write one. I don't think there's any benefit to a data-only changelog when the data isn't useful in any way yet.

AkairyuGestalter (approved):

PR #1354, opened 2021-11-22
Scaffold SGE data for Endwalker

AkairyuGestalter:
Just some data entry and stubbing


src/data/ACTIONS/root/SGE.ts
```diff
@@ -1,5 +1,298 @@
...
+		gcdRecast: 1500,
+		mpCost: 1000,
+		statusesApplied: ['EUKRASIAN_DIAGNOSIS', 'DIFFERENTIAL_DIAGNOSIS'],
+	},
+	HAIMA: {
+		id: SGE_STUB + 3,
+		name: 'Haima',
+		icon: '',
+		cooldown: 120000,
+		statusesApplied: ['HAIMATINON'],
```
Toastdeib, line 31: NBD right now but I'm assuming that Haimatinon will be a distinct buff from the actual shield, so once we have the name of that this'll probably want updating.

src/data/ACTIONS/root/SGE.ts
```diff
@@ -1,5 +1,298 @@
...
+		gcdRecast: 1500,
+		mpCost: 1100,
+		statusesApplied: ['EUKRASIAN_PROGNOSIS'],
+	},
+	PANHAIMA: {
+		id: SGE_STUB + 7,
+		name: 'Panhaima',
+		icon: '',
+		cooldown: 120000,
+		statusesApplied: ['PANHAIMATINON'],
```
Toastdeib, line 67: Same note here as on Haima.

src/data/ACTIONS/root/SGE.ts
```diff
@@ -1,5 +1,298 @@
...
+		icon: '',
+		cooldown: 15000,
+	},
+
+	/** Utility */
+	KARDIA: {
+		id: SGE_STUB + 12,
+		name: 'Kardia',
+		icon: '',
+		cooldown: 5000,
```
Toastdeib, line 101: Probably want to include Kardia/Kardion as `statusesApplied` in here.

src/data/ACTIONS/root/SGE.ts
```diff
@@ -1,5 +1,298 @@
...
+	},
+	EUKRASIA: { // TODO: Apparently this is on the GCD? Will need confirmation after launch
```
Toastdeib, line 103: Looks like, based on its classification as a Spell. If I had to guess, I'd say this will probably work a lot like the Mudra system - e.g. Eurkasia (1s) > Diagnosis (1.5s) for a 2.5s cast, much like Ten (0.5s) > Chi (0.5s) > Raiton (1.5s) is a 2.5s cast. Also my gut says this will probably apply a buff that Dosis/Diagnosis/Prognosis consumes but we'll see when EW launches I guess.

src/data/ACTIONS/root/SGE.ts
```diff
@@ -1,5 +1,298 @@
...
+		name: 'Kerachole',
+		icon: '',
+		cooldown: 30000,
+		statusesApplied: ['KERACHOLE', 'KERACHOLE_REGEN'],
+	},
+	TAUROCHOLE: {
+		id: SGE_STUB + 22,
+		name: 'Taurochole',
+		icon: '',
+		cooldown: 30000,
```
Toastdeib, line 168: Cooldown should be `45000` for this one.

src/data/ACTIONS/root/SGE.ts
```diff
@@ -1,5 +1,298 @@
...
+	},
+
+	/** AoE DPS Spells */
+	TOXIKON: {
+		id: SGE_STUB + 29,
+		name: 'Toxikon',
+		icon: '',
+		onGcd: true,
+		speedAttribute: Attribute.SPELL_SPEED,
+		gcdRecast: 2500,
```
Toastdeib, line 235: I assume it doesn't hurt anything but do we need to specify a 2.5s GCD recast on this and the three spells below it?

Toastdeib (changes requested):
Couple of small things that should be tweaked but overall looks good.

src/data/ACTIONS/root/SGE.ts
AkairyuGestalter: Sure enough, there's both a `Haima` and a `Haimatinon` buff getting applied in some footage I checked. Added/fixed.

src/data/ACTIONS/root/SGE.ts
AkairyuGestalter: Same changes as Haima

src/data/ACTIONS/root/SGE.ts
AkairyuGestalter: Yeah, on re-watching footage/re-checking Balance notes, it's applying a buff, and it's a Eukrasia (Instant, 1s GCD) > [Eukrasia Consumer] (Instant, 1.5s GCD) for a total of a 2.5s GCD (notably unaffected by Spell Speed though). Added the status and tagged it in Eukrasia's `statusesApplied`

src/data/ACTIONS/root/SGE.ts
AkairyuGestalter: Sure 'nuff

src/data/ACTIONS/root/SGE.ts
AkairyuGestalter: This was just my brain melting by the time I got here. I think it was meant for the Phlegma spells, since those are on the GCD, affected by spell speed, but also have a cooldown _and_ charges  (:blobsweat:). 'Tis fixed now.

Toastdeib (approved):

PR #1358, opened 2021-11-25
Combos - invulns don't break combo, apparently?

hintxiv:


nonowazu (approved):
Not that my approval means anything, but the logic looks right.

AzariahAeroborn (approved):

PR #1359, opened 2021-11-28
Endwalker - BLM - Gauge and RotationWatchdog updates

AkairyuGestalter:
Prelminary gauge changes for Endwalker BLM:
- Track Paradox stacks
- Update AF/UI state based on new spells/traits
- Treats Enochian as a calculated value in the gauge state when either AF/UI aspect is active
- Adds suggestion not to overcap Paradox
- Some cleanup and const renaming for clarity, plus updating the places those consts are referenced

Edit:
Couldn't detangle RotationWatchdog from this the way I had hoped. PR now also includes the following for RotationWatchdog:
- Updated the Cycle Endpoints list based on revised gauge behavior for the AoE Ice spells
- Counts Paradox usage in the Astral Fire phase as a F1 usage (for the "don't use extra F1s" check)
- Added ensureRecord type safety for the CYCLE_ERRORS const to prevent an issue we saw in production

Edit 2:
Also needed to put the 'Paradox only consumes Sharpcast in AF' change here since I changed the getGaugeState signature. The rest of the Sharpcast changes are on #1360


AkairyuGestalter:
Will need to make sure the bugfix on #1369 merges well with this whenever both are in the endwalker branch

ackwell (approved):

PR #1362, opened 2021-11-28
Endwalker - DNC - Updated Esprit gauge simulation

AkairyuGestalter:
- Improvisation no longer generates Esprit. Vastly simpler code, yay.
- Dancer now always generates five Esprit on Weaponskills, or ten if it's a proc action. Party members are still a ~20% chance to generate ten as far as we know.
- Tillana will presumably not generate Esprit since it's an Ability (side note: there has to be a better way to know whether an action is an Ability vs. Spell or Weaponskill than just hard-coding the list...)


AkairyuGestalter:
Should wait for #1392 to merge/pull into this branch before reviewing, will need to remove the old IMPROVISATION_HEALING status stub that PR is forced to leave behind

AzariahAeroborn (approved):

PR #1363, opened 2021-11-28
Endwalker - DNC - Combo updates

AkairyuGestalter:
- Removed allowance for dropping combo if both dances are used, the timer is long enough that won't be an issue anymore.
- Removed helper function in DirtyDancing used for above check


ackwell (approved):

ackwell:
@AkairyuGestalter cawnfleects

PR #1364, opened 2021-11-28
Endwalker - SGE - Preliminary Gauge and Overheal tracking

AkairyuGestalter:
Changes:
- Adds core Overheal tracking for SGE with additional HoT, oGCD, and Kardia effect buckets
- Adds preliminary implementation for tracking SGE gauge state.
- No suggestions are currently provided for SGE gauge action overcap/usage.

Addersgall Gauge:
- Starts at 3 stacks on instant start/reset. Not sure how to correctly deal with pre-pull uses of it's actions, but since they're only healing abilities this probably won't matter too much in practice
- Timer restarts immediately on death, it does not wait until the player is raised (at least in the media tour build)

Addersting Gauge:
- Generated when a Eukrasian Diagnosis shield applied by the player is broken.
- We don't currently have precise data about how much damage a shield absorbed and whether it broke or not from FFLogs, so this is working under the assumption that a shield that wasn't overwritten, lasted less than the full duration, and absorbed some amount of damage was a break.
- Shields overwritten by other SGEs/SCHs (ASTs with Neutral Sect?) have their status removed before the overwriting shield is applied (at the same timestamp though), so we have to resolve the shield one ms after the actual statusRemove event to be sure it wasn't overwritten.


AkairyuGestalter:
Should wait for #1393 to merge/pull into this branch before reviewing

src/parser/jobs/sge/changelog.tsx
```diff
@@ -1,7 +1,15 @@
+import CONTRIBUTORS from 'data/CONTRIBUTORS'
+import React from 'react'
+
 export const changelog = [
 	// {
 	// 	date: new Date('2020-04-20'),
 	// 	Changes: () => <>The changes you made</>,
-	// 	contrubutors: [CONTRIBUTORS.YOU],
+	// 	contributors: [CONTRIBUTORS.YOU],
```
ackwell, line 8: woops

src/parser/jobs/sge/modules/Gauge.tsx
```diff
@@ -0,0 +1,156 @@
...
+	applyTimestamp?: number,
+	removeTimestamp?: number,
+	absorbed?: number,
+	consumedOrOverwritten?: boolean
+}
+
+export class Gauge extends CoreGauge {
+	@dependency private actors!: Actors
+
+	private diagnoses: {[key: string]: DiagnosisData | undefined} = {}
```
ackwell: I'd err to typing this as
```suggestion
	private diagnoses: Partial<Record<Actor['id'], DiagnosisData>> = {}
```

same overall effect but cleaner partial handling & semantic type for the key

src/parser/jobs/sge/modules/Gauge.tsx
```diff
@@ -0,0 +1,156 @@
...
+		if (diagnosis == null) { return }
+		diagnosis.consumedOrOverwritten = true
+	}
+
+	// Keep track of shield removals, and set a timestamp hook to resolve the shield (overwrites happen 'after' the removal event but at the same timestamp, so we can't know for sure if the shield was overwritten yet)
+	private onShieldRemove(event: Events['statusRemove']) {
+		const diagnosis = this.diagnoses[event.target]
+		if (diagnosis == null) { return }
+		diagnosis.removeTimestamp = event.timestamp
+		diagnosis.absorbed = event.absorbed
```
ackwell, line 152: 😬 is there a way we can avoid checking explicit absorbs? they're very loose guesses at best, and if we start consuming them in xiva events, we'll need to make sure that every single possible event source is capable of simulating them to some degree of accuracy.

---

i get the feeling i've had this conversation before lmao. idek

ackwell (commented):
generally lookin good. comment on the shields but as mentioned i feel like this is a conversation i've had and later forgotten about

src/parser/jobs/sge/modules/Gauge.tsx
AzariahAeroborn: At the moment, the behavior of the `absorbed` field is (inherited from FFLogs):
1. A shield applies to the target
2. If the shielded target is the target of any damage events before the shield is removed, the value of the shield is presented as `absorbed` (FFLogs infers the shield got popped by the damage event)
3. If the shielded target is NOT the target of ANY damage events before the shield is removed, the value of the shield is presented as `amount`, which we ignore.

Per discussions with SCH/SGE mentors on the Balance, our current approach with SCH is to track whether a shield got popped or not, but to not consider shields one way or the other with overheal calculations.  So, in the context of this module and whether or not the SGE got a Toxicon stack, the check of whether absorbed is > 0 is necessary.

AkairyuGestalter: I agree that we should look into a more permanent/log source-agnostic way to detect shield breaks, but that's well above my head at the moment and this was the best "for now" solution I could figure out how to implement. If/when we figure out the core change we'll need, I will _happily_ switch to it in a heartbeat.

AzariahAeroborn (approved):

PR #1366, opened 2021-11-29
feat: port Requiescat to Core BuffWindow

AzariahAeroborn:
An alternate approach to #1345, to contain the specialized logic to the Requiescat module specifically so that we don't change the core interfaces for current consumers.  

Closes #1345


ackwell (approved):
nice, i think this is a good resolution, and gives us something stable until we've got time (post ew aaaaaaa-ness) to rethink the bw interface a little

sorry about the slow rtts on this changeset

PR #1367, opened 2021-11-29
feat: Port WHM DoTs to analyser and fix Core DoTs clipping calculation

AzariahAeroborn:
While porting WHM DoTs to core, I observed that the core clipping calculation was never tracking clips.  I fixed the clipping calculation, but did not re-implement handling for clipping to refresh before downtime as it's unclear whether that should be handled globally or per-job.  Deferring implementation for that until if/when EW fights have boss downtime where the boss does take damage from DoTs and a refresh would make sense.

Also adjusted AST and SCH thresholds to match the thresholds for WHM, per discussion in #dev-healer


src/parser/core/modules/DoTs.ts
```diff
@@ -7,15 +7,12 @@ import {Data} from 'parser/core/modules/Data'
...
+const MILLISECONDS_PER_MINUTE = 60000
 
-/** Mapping of a status ID to the duration it was up. */
-export interface DotDurations {
-	[statusID: number]: number
-}
-
-interface DotTargets {
-	[targetID: string]: DotDurations
+type DotTracking = Map<number, Map<string, DotTargetTracking>>
```
ayyaruq, line 12: Should this not be more explicit to `Status['id']` and `Actor['id']` or something?

src/parser/jobs/mnk/modules/Demolish.tsx
```diff
@@ -1,8 +1,8 @@
 import {Trans} from '@lingui/react'
-import {ActionLink} from 'components/ui/DbLink'
+import {ActionLink, DataLink} from 'components/ui/DbLink'
```
ayyaruq, line 2: Just use DataLink exclusively?

src/parser/jobs/mnk/modules/Demolish.tsx
```diff
@@ -44,16 +44,18 @@ export class Demolish extends DoTs {
...
 			icon: this.data.actions.DEMOLISH.icon,
 			content: <Trans id="mnk.demolish.suggestion.content">
 				Avoid refreshing <ActionLink action="DEMOLISH"/> significantly before its expiration. Unnecessary refreshes risk overwriting buff snapshots.
 			</Trans>,
 			tiers: SUGGESTION_TIERS.CLIPPING,
-			value: this.getClippingAmount(this.data.statuses.DEMOLISH.id),
+			value: demolishClipPerMinute,
 			why: <Trans id="mnk.demolish.suggestion.why">
-				You lost {this.parser.formatDuration(clip[this.data.statuses.DEMOLISH.id] ?? 0)} of Demolish to early refreshes.
+				An average of {this.parser.formatDuration(demolishClipPerMinute, 1)} seconds of <DataLink status="DEMOLISH" /> per minute lost to early refreshes.
```
ayyaruq, line 58: Prefer to keep the structure of the old text, something like this:
```suggestion
				You lost {this.parser.formatDuration(demolishClipPerMinute, 1)} seconds of <DataLink status="DEMOLISH" /> per minute on average to early refreshes.
```

src/parser/core/modules/DoTs.ts
```diff
@@ -79,74 +64,63 @@ export abstract class DoTs extends Analyser {
 
 	private onApply(event: Events['statusApply']) {
 		const status = this.data.getStatus(event.status)
-		const statusID = status?.id
-		if (statusID == null) { return }
+		// Cannot track for statuses that are not defined with a duration
+		if (status == null || status.duration == null) { return }
```
ayyaruq, line 68: More of a nit but I wonder if `status?.duration  == null` is cleaner since it accounts for both and is clear that the duration is what we care about. Up to you.

ayyaruq (changes requested):
I mostly reviewed the core and MNK changes since I think the rest is largely up to the job devs. Happy to do a pass of the whole lot tho. Core change looks good other than some type stuff, but I vaguely recall Map not liking certain types for keys too. Not sure if my thought for that is actually a better option or not. I'd like to change the MNK text so it's clearer that it's on the MNK for fucking up tho.

One question tho, is tracking by status ID with the value being a map of actors and clips a good idea when some statuses can overwrite others, or are we relying on DoTs being stackable and hoping SE doesn't change that?

src/parser/core/modules/DoTs.ts
ackwell: Wouldn't block on it, but yes - preference semantic lookup types like `Status['id']` over straight `number` or similar, as while they're _equivalent_, it'll help keep them up to date as types evolve

src/parser/core/modules/DoTs.ts
```diff
@@ -79,74 +64,63 @@ export abstract class DoTs extends Analyser {
...
-		const fightDurationMillis = (this.parser.currentDuration - this.invulnerability.getDuration({types: ['invulnerable']}))
+		const fightDurationMillis = (this.parser.pull.duration - this.invulnerability.getDuration({types: ['invulnerable']}))
 		if (fightDurationMillis <= 0) { return 0 }
 
-		const clipSecsPerMin = Math.round(((this.clips[statusID] ?? 0) * SECONDS_PER_MINUTE) / fightDurationMillis)
+		const statusApplications = this.statusApplications.get(statusId)
+		if (statusApplications == null) { return 0 }
+
+		const totalClipping = Array.from(statusApplications.values()).reduce((clip, target) => clip + target.totalClipping, 0)
+		const clipSecsPerMin = Math.round(totalClipping / (fightDurationMillis / MILLISECONDS_PER_MINUTE))
```
ackwell, line 123: I know, I know, it's a breaking change - but while you're in here (or perhaps as seperate PR) it'd be good to move this module to milliseconds, too. It's long term, but I'm trying to get _all_ time values in ms for the sake of consistency.

ackwell (commented):
just passing by with some comments

src/parser/jobs/mnk/modules/Demolish.tsx
AzariahAeroborn: There's been a mix of the two formats between jobs, and I saw more jobs with the "An average of..." format so I standardized on that.  Let me know if you would prefer to revert @ayyaruq

src/parser/jobs/mnk/modules/Demolish.tsx
AzariahAeroborn: There were a few jobs that were using ActionLink inside a React block and spreading the action in, which DataLink wasn't working with, so I kind of just got to a point where I stopped fiddling with ActionLinks.  Fixed these to DataLink only.

src/parser/core/modules/DoTs.ts
AzariahAeroborn: So good news and bad news on this.  It already was returning milliseconds per minute, so all the consumers were basically always returning major.  Fixed.

src/parser/jobs/mnk/modules/Demolish.tsx
ayyaruq: Applying the suggestion is fine imo. I know we like to standardise wording where possible but since this is in the job modules rather than the parent class itself I'd like to word it this way so it's clearer that the loss is the focus, especially for how important MNK's DoT is relative to other jobs.

src/parser/core/modules/DoTs.ts
```diff
@@ -79,74 +65,63 @@ export abstract class DoTs extends Analyser {
...
+		if (fightDuration <= 0) { return 0 }
+
+		const statusApplications = this.statusApplications.get(statusId)
+		if (statusApplications == null) { return 0 }
 
-		const clipSecsPerMin = Math.round(((this.clips[statusID] ?? 0) * SECONDS_PER_MINUTE) / fightDurationMillis)
-		return clipSecsPerMin
+		const totalClipping = Array.from(statusApplications.values()).reduce((clip, target) => clip + target.totalClipping, 0)
+		const clipMSPerMin = Math.round(totalClipping / (fightDuration / MILLISECONDS_PER_MINUTE))
+		return clipMSPerMin
```
ackwell, line 125: legend, thanks

ackwell (approved):

PR #1368, opened 2021-11-29
[AST] Batch port to analyser and convert to tsx

otocephaly:
Howdy
Nothing too crazy going on for this one
Lightspeed had the most changes because I largely lifted the code from DNC technicalities removing parts of the code that aren't used for lightspeed
I changed the format for Lightspeed because I like the ability to click on the time and see it in action -- if you prefer the previous accordion format, I can change it back. I leave it up to the reviewer


otocephaly:
Lightspeed before changes (production):
image

Lightspeed after changes (testing):
image

src/parser/jobs/ast/modules/Lightspeed.tsx
```diff
@@ -0,0 +1,231 @@
...
+		const castsMessage = <p><Trans id="ast.lightspeed.messages.num-casts"> There were a total of {actualCasts} out of a possible {totalPossibleCasts} <DataLink action="LIGHTSPEED" /> casts noted.</Trans></p>
+
+		const pre5_3message = <p><Trans id="ast.lightspeed.messages.explanation">
+		The main use of <DataLink action="LIGHTSPEED" /> should be for weaving card actions during <DataLink action="DIVINATION" /> and <DataLink action="SLEEVE_DRAW" /> windows.<br />
+		It can also be used for MP savings on heavy healing segments, keeping casts up while on the move and other specific scenarios.<br />
+		Each fight calls for a different strategy, but try to utilize it as much as possible.<br />
+		Unless it's being used for <DataLink action="ASCEND" />, lightspeed should fit at least 6 GCDs.<br />
+		</Trans></p>
+
+		const message = <p><Trans id="ast.lightspeed.messages.explanation.ost5_3">
```
ayyaruq, line 181:
```suggestion
		const message = <p><Trans id="ast.lightspeed.messages.explanation.post5_3">
```

Unless this actually is an original soundtrack, in that case carry on. It's a bit weird that we invert the localisation key from the variable name tho, consider renaming these so message gets explanation and pre5_3message gets pre5_3explanation or something. You'll likely need to edit these in the locale files manually.

src/parser/jobs/ast/modules/Lightspeed.tsx
```diff
@@ -0,0 +1,231 @@
...
+
+	override initialise() {
+		const lightspeedFilter = filter<Event>().status(this.data.statuses.LIGHTSPEED.id)
+
+		this.addEventHook(lightspeedFilter.type('statusApply')
+			.target(this.parser.actor.id), this.tryOpenWindow)
+		this.addEventHook(lightspeedFilter.type('statusRemove')
+			.target(this.parser.actor.id), this.tryCloseWindow)
+
+		this.addEventHook(filter<Event>().source(this.parser.actor.id).type('action'), this.onCast)
```
ayyaruq, line 58: Better to create this hook inside your statusApply callback and undefine it in your statusRemove callback. You'll need another var to assign the hook to tho, so you can safely reference it when undefining.

src/parser/jobs/ast/modules/Lightspeed.tsx
```diff
@@ -0,0 +1,231 @@
...
+	}
+
+	private tryOpenWindow(event: Events['statusApply']): LIGHTSPEED_Window {
```
ayyaruq, line 61: Callbacks shouldn't return things. Prefer to use a var to hold the current window, or otherwise a function to retrieve the window you want and share it between the diff callbacks as required.

src/parser/jobs/ast/modules/Lightspeed.tsx
```diff
@@ -0,0 +1,231 @@
...
+		const lastWindow: LIGHTSPEED_Window | undefined = _.last(this.history)
+
+		// If that happens, re-open the last window and keep tracking
+		if (lastWindow != null) {
+			if (!lastWindow.end) {
```
ayyaruq, line 66: End isn't a boolean. Prefer to explicitly null check this. I think there's an easier way to write this whole block but it's not coming to me atm.

src/parser/jobs/ast/modules/Lightspeed.tsx
```diff
@@ -0,0 +1,231 @@
...
+			lastWindow.end = event.timestamp
+		}
+	}
+
+	// Make sure all applicable statuses have fallen off before the window closes
+	private isWindowOkToClose(window: LIGHTSPEED_Window): boolean {
+		if (!window.buffsRemoved.includes(this.data.statuses.LIGHTSPEED.id)) {
+			return false
+		}
+		return true
```
ayyaruq, line 100: Easier to read, nbd.

```suggestion
		return !window.buffsRemoved.includes(this.data.statuses.LIGHTSPEED.id)) ? false : true
```

src/parser/jobs/ast/modules/LucidDreaming.tsx
```diff
@@ -0,0 +1,81 @@
...
+					</Trans>
+				</Fragment>,
+				severity: this.uses === 0 || _usesMissed > WASTED_USES_MAX_MEDIUM ? SEVERITY.MAJOR : SEVERITY.MEDIUM,
+				why: <Fragment>
+					<Trans id="ast.lucid-dreaming.suggestion.why">
+						<Plural value={_usesMissed} one="# use" other="# uses" /> of Lucid Dreaming were missed by holding it for at least a total of {this.parser.formatDuration(holdDuration)}.
+					</Trans>
+				</Fragment>,
+			}))
+		}
```
ayyaruq, line 77: Using tiered suggestions here would allow you to kill the conditional, outdent the block, and simplify the code.

src/parser/jobs/ast/modules/Synastry.tsx
```diff
@@ -21,24 +24,26 @@ const GCD_ST_HEAL: Array<keyof ActionRoot> = [
 ]
```
ayyaruq, line 24: GH being dumb, we no longer recommend using `Array<keyof ActionRoot>` since it's been replaced by a proper type, `ActionKey` which you can use like `ActionKey[]` for the array.

ayyaruq (changes requested):
Nice work! Lightspeed is the main standout here, particularly the way you store history and windows. I feel like it'd just be way simpler to store the current window in a var and push it to history when it closes, instead of constantly iterating the arrays with lodash. Another thing to note is there's a bit of weirdness in the i18n keys, specifically random uppercase usage of `ast.LIGHTSPEED.whatever` and the patch 5.3 bits.

You also seem to use classes a lot, in places where it just... doesn't have any benefit? Looks like it's so you can predefine a start time but that's relatively trivial with an interface instead, and less overhead. Considering you don't use history for anything other than checking last window and then the iterator at the end, I can't see any reason to use a class since there's nothing in there that would even benefit having class methods. GCD count is also kinda pointless, since you can loop the rotation for the GCD count anyway (or add a getter to the class for it and justify having a class I guess).

As an aside, prefixing things with `_` (other than lodash) is an JS-ism for private vars/functions, but TS has an actual private keyword, so you don't really need to do that unless it's to mask an internal class variable from a getter/setter with the same name.

src/parser/jobs/ast/index.tsx
```diff
@@ -35,8 +34,33 @@ export default new Meta({
...
+						<li>Sect,</li>
+						<li><DataLink action="SYNASTRY"/>,</li>
+						<li>and other modules.</li>
+					</ul>
+					<li><DataLink action="LIGHTSPEED"/> now shows actual casts and total possible uses.</li>
+					<li>Sect now has an updated suggestion to show possibility of using sect very close leading to unpreparedness. (i.e. using sect at the same time as the pull would be the same as using an oGCD which wastes time that could be spent starting your <DataLink action="MALEFIC_III" /> or other action cast.)</li>
+				</ul>
+			</>,
+			contributors: [CONTRIBUTORS.OTOCEPHALY],
+		},
```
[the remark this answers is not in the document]
otocephaly: I think I misinterpreted this then
image

lemme know if I should still reduce changelog

src/parser/jobs/ast/index.tsx
```diff
@@ -1,8 +1,7 @@
 import {t} from '@lingui/macro'
 import {Trans} from '@lingui/react'
-import {ActionLink} from 'components/ui/DbLink'
+import {DataLink} from 'components/ui/DbLink'
```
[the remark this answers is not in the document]
otocephaly: sounds good
I was just being consistent in changing them in case there will be other links in the future (not that it matters since this part is sanitized for EW anyway, but still)

src/parser/jobs/ast/index.tsx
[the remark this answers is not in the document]
otocephaly: I'll reduce anyway tbh

src/parser/jobs/ast/modules/Lightspeed.tsx
otocephaly: updated both this and the relating locale files
yarn extract was ran after doing that, but probably good to review quickly just in case

src/parser/jobs/ast/modules/Lightspeed.tsx
otocephaly: updated based on MNK fists like you recommended
works nicely
thanks!

src/parser/jobs/ast/modules/Lightspeed.tsx
otocephaly: updated

src/parser/jobs/ast/modules/Lightspeed.tsx
otocephaly: it has been updated to be simpler, since it always returned the same thing as the end, I just removed whatever we were originally looking for anyway which was if end exists, we undefine, otherwise return

src/parser/jobs/ast/modules/Lightspeed.tsx
otocephaly: I just removed this function and just replaced the section that was used with the literal check since it just checks false/true anyway

src/parser/jobs/ast/modules/LucidDreaming.tsx
otocephaly: done and updated with a new severities section
lemme know if you agree with the percentages

src/parser/jobs/ast/modules/Synastry.tsx
otocephaly: updated

src/parser/jobs/ast/index.tsx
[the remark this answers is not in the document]
ayyaruq: This only applies to `endwalker`, but your PR is to `master`. The last bit of the changelog is fine tho, the rest just isn't relevant to an actual user.

src/parser/jobs/ast/modules/Lightspeed.tsx
```diff
@@ -0,0 +1,234 @@
...
+					.type('action'),
+				this.onCast,
+			)
+		}
+
+		// If that happens, re-open the last window and keep tracking
+		if (this.currentWindow.end != null && this.currentWindow.end === event.timestamp) {
+			this.currentWindow.end = undefined
+		}
+		return this.currentWindow
```
ayyaruq, line 83: Function doesn't return anything, and `this.currentWindow` is a class var so there's no need to return after setting it.

```suggestion
```

src/parser/jobs/ast/modules/Lightspeed.tsx
```diff
@@ -0,0 +1,234 @@
...
+					.source(this.parser.actor.id)
+					.type('action'),
+				this.onCast,
+			)
+		}
+
+		// If that happens, re-open the last window and keep tracking
+		if (this.currentWindow.end != null && this.currentWindow.end === event.timestamp) {
+			this.currentWindow.end = undefined
+		}
```
ayyaruq, line 82: What's the conditions where this can even happen?

src/parser/jobs/ast/modules/Lightspeed.tsx
```diff
@@ -0,0 +1,234 @@
...
+
+		if (this.currentWindow == null) {
+			return
+		}
+
+		// Cache whether we've seen a buff removal event for this status, just in case they happen at exactly the same timestamp
+		this.currentWindow.buffsRemoved.push(event.status)
+
+		// Make sure all applicable statuses have fallen off before the window closes
+		if (this.currentWindow.buffsRemoved.includes(this.data.statuses.LIGHTSPEED.id)) {
```
ayyaruq, line 96: This function is called when Lightspeed falls off, so what's the point in even tracking this?

src/parser/jobs/ast/modules/Lightspeed.tsx
```diff
@@ -0,0 +1,234 @@
...
+
+		const action = this.data.getAction(event.action)
+
+		// Can't do anything else if we didn't get a valid action object
+		if (action == null) {
+			return
+		}
+
+		// If this window isn't done yet add the action to the list
+		if (this.currentWindow.end == null) {
```
ayyaruq, line 122: Current window will be undefined when the end timestamp is set, so this is always true.

ayyaruq (changes requested):
Bit of cleanup on Lightspeed still, otherwise gucci. I think you're expecting edge cases that don't exist, so it's really just removing unnecessary safety checks that make the code harder to read. I'd prob avoid calling stuff "try open" in future tho, since it's really just "onGain, open a window, it'll always succeed".

src/parser/jobs/ast/index.tsx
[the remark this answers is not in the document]
otocephaly: sounds good

src/parser/jobs/ast/modules/Lightspeed.tsx
otocephaly: removed

src/parser/jobs/ast/modules/Lightspeed.tsx
otocephaly: removed

otocephaly: also removed relating variable since now unused

src/parser/jobs/ast/modules/Lightspeed.tsx
otocephaly: removed

src/parser/jobs/ast/modules/Lightspeed.tsx
otocephaly: I think it was a situation that arose probably before?
I left it because while I wasn't sure of the situation doesn't mean it couldn't happen (whatever this situation was)
but anyway, I removed it

src/parser/jobs/ast/modules/Lightspeed.tsx
```diff
@@ -0,0 +1,230 @@
...
+		this.currentWindow.rotation.push(event)
+		if (action.onGcd) {
+			this.currentWindow.gcdCount++
+		}
+		if (this.parser.patch.before('5.3') && action.mpCost != null) {
+			this.currentWindow.mpSavings = this.currentWindow.mpSavings + action.mpCost/2
+		}
+
+		// If we haven't recorded a trailing GCD event for this closed window, do so now
+		if (this.currentWindow.end && !this.currentWindow.trailingGcdEvent && action.onGcd) {
```
ayyaruq, line 118: Should really be explicitly checking these, but also I think this whole block can be reordered for simplicity:

```ts
this.currentWindow.rotation.push(event)

if (this.parser.patch.before('5.3') && action.mpCost != null) {
	this.currentWindow.mpSavings = this.currentWindow.mpSavings + action.mpCost/2
}

if (action.onGcd) {
	this.currentWindow.gcdCount++

  // If we haven't recorded a trailing GCD event for this closed window, do so now
  if (this.currentWindow.end != null && this.currentWindow.trailingGcdEvent == null) {
    this.currentWindow.trailingGcdEvent = event
  }
}
```

ayyaruq (changes requested):

src/parser/jobs/ast/modules/Lightspeed.tsx
otocephaly: updated

ayyaruq (approved):

PR #1376, opened 2021-12-02
WAR 6.0 support

ayyaruq:
Still waiting on real IDs/URLs, and I'm not really happy with the Inner Release module yet since still wrapping my head around writing Evaluators. PR still a WIP anyway so yolo. I reordered the data files so... sorry for that.


src/parser/jobs/war/modules/OGCDDowntime.ts
```diff
@@ -19,16 +19,19 @@ export class OGCDDowntime extends CooldownDowntime {
 
 		{
 			cooldowns: [this.data.actions.INNER_RELEASE],
-			// IR can sit for up to 4 GCDs, enough to reapply Eye and avoid overcapping gauge
-			allowedAverageDowntime: 10000,
+			// IR can sit for up to 5 GCDs, enough to reapply Eye and avoid overcapping gauge
+			allowedAverageDowntime: 12500,
```
AzariahAeroborn: If your goal is to allow 0-5 GCDs of drift, you should specify an allowedAverageDowntime of half of 5 GCDs (unless you expect that it will often be held for multiple GCDs in normal rotation).

The higher you have the allowedAverageDowntime, the more likely you will show 9/8 uses and such.

src/parser/jobs/war/modules/PrimalChaos.tsx
```diff
@@ -0,0 +1,117 @@
...
+		const playerFilter = filter<Event>().source(this.parser.actor.id)
+
+		this.addEventHook(playerFilter.type('statusApply').status(oneOf(dropsyStatuses)), this.onGain)
+		this.addEventHook(playerFilter.type('statusRemove').status(oneOf(dropsyStatuses)), this.onDrop)
+		this.addEventHook('complete', this.onComplete)
+	}
+
+	private onGain(event: Events['statusApply']) {
+		switch (event.status) {
+		case this.data.statuses.NASCENT_CHAOS.id:
```
AzariahAeroborn, line 44: Personally, I would find separate functions for onGainNascentChaos and onGainPrimalRend rather than these switches, for readability reasons.  I don't know how much of a performance different it makes between either approach.

AzariahAeroborn (changes requested):

src/parser/jobs/war/modules/OGCDDowntime.ts
ayyaruq: It's to allow a 1-2-3 combo and one FC. Not sure why I bumped this higher than that or the comment changed to 5, prob a mistake while thinking about PRR which isn't relevant to average downtime. Does 9000 sound alright to deal with weaving latency and 4 GCDs?

src/parser/jobs/war/modules/PrimalChaos.tsx
ayyaruq: It was to reduce hooks mostly. Maybe an ack question.

src/parser/jobs/war/modules/OGCDDowntime.ts
ayyaruq: I just realised I've already fixed this but forgot to commit it lmao.

AzariahAeroborn: If it "should" be 0, but could be up to 4, I would set 5000 -- (0 * 2500 + 4 * 2500)/2.  9000ms average allowed downtime would basically imply that every time IR comes up you expect to need 3 GCDs of delay and maybe 4.  

Also, since IR changed in late 5.x to extend Eye on application, and IR is still doing that for Tempest, do we even need to allow for reapplying Tempest?  We'll already be dinging them if they let Tempest drop, does it really make sense to allow grace on the IR cooldown?

AzariahAeroborn (approved):

PR #1377, opened 2021-12-02
Added 3 additional actions to the Ninki gauge simulation

Toastdeib:
Small fix now that the official tooltips are out. Ninki wasn't included on these 3 skills in the media tour versions of their tooltips.


ayyaruq (approved):

ayyaruq:
Not sure if we're allowed to merge yet so... yolo?

PR #1388, opened 2021-12-08
MCH - 6.0 update

hintxiv:
Other than the WF rewrite that needs to happen, this should be everything for MCH EW support.


ackwell (changes requested):
code looks sane, but that icon is a 13.9mb png which is about 13.85mb too large - if you downsize it to a 100x100 jpg and run it through an optimiser (plenty online) it should be a bit better.
I'd like to automate the above but it's Not Done Yet:tm:

ackwell (approved):
12kb is comf

PR #1391, opened 2021-12-09
[AST] 6.0 initial updates

### LINKED ISSUE #1379: [AST 6.0] New seal logic
1. Change the consumption trigger to Astrodyne instead of Div
2. Delete sleeve draw. 
3. Delete the minor arcana conversion logic.

### LINKED ISSUE #1380: [AST 6.0] Astrodyne
1. Players can nom their seals anytime they want for the buff. 
2. We should be encouraging them to nom it on every minute mark, at least to match Div timings and keep seals coming
3. We can leverage our card data to see if they had a chance to get the 3-seal Astrodyne on that trick window. 
4. We should call out overcapping this gauge.

### LINKED ISSUE #1381: [AST 6.0] Actions & Status Data

### LINKED ISSUE #1382: [AST 6.0] Sleeve + Sect: kill
- Remove sleeve from all AST modules
- Remove Sect from all AST modules

otocephaly:
This PR includes
   - adding AST actions/statuses
   - removing sect / sleeve draw references
   - updating divination to astrodyne where appropriate
   - removing divination counting from arcana tracking
   - removing lord/lady of crown statuses where appropriate
   - adding self as contributor
   - other automatic commits (yarn, json something)


src/parser/jobs/ast/modules/Draw.tsx
```diff
@@ -18,9 +18,7 @@ import ArcanaTracking from './ArcanaTracking/ArcanaTracking'
 // TODO: THINGS TO TRACK:
 // Track them using Draw when they still have a minor arcana (oopsie) or a card in the spread
 
-const CARD_DURATION = 15000
-const SLEEVE_DRAW_PLAYS_GIVEN_530 = 1
-
+const oGCD_ALLOWANCE = 2000 //used in case the last draw comes up in the last second of the fight. takes another second to cast PLAY and therefore an AST would not DRAW if they couldn't even PLAY
```
suShirow: 2s is too short. Typically they would single weave draw with something else. We should set this to at least 5s since they would draw and then play in separate weaves. But even then they wouldn't get any value out of the last play as it wouldn't be on for even a single gcd.

src/parser/jobs/ast/modules/Draw.tsx
```diff
@@ -18,9 +18,7 @@ import ArcanaTracking from './ArcanaTracking/ArcanaTracking'
 // TODO: THINGS TO TRACK:
 // Track them using Draw when they still have a minor arcana (oopsie) or a card in the spread
```
suShirow: While we're here could you do me a favor and remove this comment it doesn't have any use anymore :x Sorry and thanks

package.json
```diff
@@ -24,6 +24,7 @@
     "@visx/pattern": "^1.7.0",
     "@visx/shape": "^1.7.0",
     "@xivanalysis/tooltips": "^0.1.3",
+    "caniuse-lite": "^1.0.30001285",
```
suShirow: @otocephaly I suppose you added it to get rid of the warnings? 

@ackwell Just thought I'd let you know as this is a package addition, but this PR will add the `caniuse-lite` package which is totally unrelated to the PR contents but I don't see any problem with this.

src/parser/jobs/ast/modules/ArcanaTracking/ArcanaTracking.tsx
```diff
@@ -425,12 +416,7 @@ export default class ArcanaTracking extends Analyser {
 
 			// Modify log, they were holding onto this card since index
 			// Differenciate depending on searchLatest
-			let arcanaStatus: number | undefined
-			if (this.lastDrawnBuff && this.crownPlays.includes(cardActionId)) {
-				arcanaStatus = this.lastDrawnBuff.data
-			} else {
-				arcanaStatus = this.arcanaActionToStatus(cardActionId)
-			}
+			const arcanaStatus: number | undefined = this.arcanaActionToStatus(cardActionId)
```
suShirow, line 419: is this due to the analyzer change? So we don't need to do worry about it being dropped from the lastDrawnBuff?

suShirow (approved):
Thanks for setting this up! Just had a few queries for my own understanding.

I do have a concern about the draw/play allowance time on fight end. We'll probably end up debating on it more in discord.

package.json
otocephaly: I have reverted the commit since I don't see why this package was brought up
I re-run the PR in the test environment and no errors came up so this is good to remove

src/parser/jobs/ast/modules/Draw.tsx
otocephaly: updated! I should've removed irrelevant comments anyway lol

src/parser/jobs/ast/modules/Draw.tsx
otocephaly: Very good point! Thanks!
I'll update the time and the comment

src/parser/jobs/ast/modules/ArcanaTracking/ArcanaTracking.tsx
otocephaly: I see the intention is to pick up statuses that were /statusoff or when the player died, but it seems this specific portion only applied for crown plays
I think the dropped-off statuses portions need to be rewritten a bit because it also doesn't seem to also be picking up death events appropriately

here's an example log with a death event that wasn't appropriately captured
https://xivanalysis.com/fflogs/L3mJVpA1rc4hGgKY/2/2
see timestamp 1:51 under arcana tracking
I think this is going to happen every time there is a death event when someone holds a card, but I could be wrong

when I have time, I'll look into the /statusoff event and see what we can do to improve this part

src/parser/jobs/ast/modules/Draw.tsx
otocephaly: I think based on the above, I set it to 7500 = 3 * 2500 since approx 2 GCDs for an AST to Draw and Play and then 1 GCD for the target to use one move

PR #1395, opened 2021-12-12
[GNB] Semi 6.0 Support 

Rjb99:
GNB 6.0 Support includes:
-New actions and statutes
-AmmoCombo restored from the Shadow Realm -- I mean Shadowbringers branch
-OGCD cooldowns renamed to Cooldowns as it also includes GCD with long recast and the name was bothering me

To Do:
-Update No Mercy windows
-Confirm cooldown offsets for first use

GNB is still not marked as supported officially in this PR.


src/parser/jobs/gnb/modules/AmmoCombo.tsx
```diff
@@ -0,0 +1,250 @@
...
+	public get actualTalon(): number {
+		return this.rotation.filter(event => event.action === ACTIONS.WICKED_TALON.id).length
+	}
+
+	public get actualEye(): number {
+		return this.rotation.filter(event => event.action === ACTIONS.EYE_GOUGE.id).length
+	}
+}
+
+export default class AmmoCombo extends Analyser {
```
AzariahAeroborn: No default exports.
```suggestion
export class AmmoCombo extends Analyser {
```

src/parser/jobs/gnb/modules/AmmoCombo.tsx
```diff
@@ -0,0 +1,250 @@
...
+		return this.rotation.filter(event => event.action === ACTIONS.ABDOMEN_TEAR.id).length
+	}
+
+	public get actualTalon(): number {
+		return this.rotation.filter(event => event.action === ACTIONS.WICKED_TALON.id).length
+	}
+
+	public get actualEye(): number {
+		return this.rotation.filter(event => event.action === ACTIONS.EYE_GOUGE.id).length
+	}
```
AzariahAeroborn: This is on the edge of the type of getters we try to avoid using.  From a raw performance standpoint, it would be a lot better to just have `jugUsed: number = 0`, `savageUsed: number = 0`, etc. buckets (or even just booleans, since it's a 0 or 1 thing for these skills) and then just track when the skills get used, rather than filtering the rotation array 5 times.  Also, RotationTable winds up calling the getters multiple times per row when you pass the getter to the target property, so it compounds the problem.  

Alternatively, since Gnashing Fang combos are strictly linear and the player doesn't have a choice to go back or use in special orders, could we just drop all of the individual skill target columns in favor of showing the total X/6 skills used and the actual rotation used, and then remove these getters altogether?

AzariahAeroborn (changes requested):

src/parser/jobs/gnb/modules/Cooldowns.tsx
```diff
@@ -1,21 +1,22 @@
...
-const FIRST_USE_OFFSET_BOWSHOCK = 10000 // Current Openers have Bowblast being used somewhere between after 2nd GCD to after 4th GCD, set to maximum
+const FIRST_USE_OFFSET_DOUBLE_DOWN = 12500 //Curent Opener has double being used as 5th GCD. *GCD Skill*
 
-const FIRST_USE_OFFSET_DIVIDE = 7500 // Current Openers have rough divide at 3rd GCD or literally on pull. set to 3rd GCD
+const FIRST_USE_OFFSET_DIVIDE = 15000 //Current Opener has rough divide set to be after 6th GCD
 
-const FIRST_USE_OFFSET_SONIC_BREAK = 12500 // Current Openers have the longest delay for Sonic break being on 5th GCD (Raid Buffer Opener)
+const FIRST_USE_OFFSET_SONIC_BREAK = 15000 //Current Opener has Sonic Break as 6th GCD *GCD Skill*
 
 export default class AbilityDowntime extends CooldownDowntime {
```
AzariahAeroborn: Can you also please remove the default export from this class while editing it?

Rjb99: Went and removed all the default exports from everything in the GNB modules

src/parser/jobs/gnb/modules/AmmoCombo.tsx
Rjb99: I'm not opposed to the 6/6 option, as even though you can add extenders to this combo, they aren't counted. So it'll just be the end user that has to see what's missing instead of us explicitly saying it

src/parser/jobs/gnb/modules/AmmoCombo.tsx
Rjb99: Done

AzariahAeroborn (approved):

PR #1397, opened 2021-12-13
[AST] timeline, divination, CDDT, overheal

### LINKED ISSUE #1383: [AST 6.0] CI - Capture 2 stack charges

### LINKED ISSUE #1384: [AST 6.0] Div to CDDT
Chain strat but without the RNG on the RNG job

otocephaly:
Updated the timeline to be more methodical and fixed minor arcana/neutral sect
Updated divination to be like lightspeed with some tweaks
Updated overheals to include delayed heals to show heals that are usually planned in advance
Consolidated CDDT to include some healing abilities and some dps abilities to show as a snapshot how the AST is managing their oGCDs
   - please note the weights and target are arbitrary so if you don't like them, lemme know and they can be adjusted fairly easily to what you think would be better
Updated changelog

For divination, since there aren't many logs that would contain this situation, I used this log to test overwriting divination
https://www.fflogs.com/reports/gGxcLpvwYC9FJVHf

Otherwise, it seems there is a bug with some of ASTs heals for exaltation, horoscope, and macrocosmos -- so while delayed heals seems to be functioning, it may look like it's not because fflogs hasn't attributed some of the heals of those sorts


src/parser/jobs/ast/modules/Divination.tsx
```diff
@@ -0,0 +1,297 @@
...
+import Suggestions, {SEVERITY, Suggestion} from 'parser/core/modules/Suggestions'
+import {Timeline} from 'parser/core/modules/Timeline'
+import React, {Fragment} from 'react'
+import {Icon, Message} from 'semantic-ui-react'
+import DISPLAY_ORDER from './DISPLAY_ORDER'
+
+const BASE_GCDS_PER_WINDOW = 6
+const PLAYERS_BUFFED_TARGET = 8
+
+interface DIVINATION_Window {
```
AzariahAeroborn: style nit:
```suggestion
interface DivinationWindow {
```

src/parser/jobs/ast/modules/Divination.tsx
```diff
@@ -0,0 +1,297 @@
...
+				SUGGESTION: Overwriting other AST's divination
+			*/
+			this.suggestions.add(new Suggestion({
+				icon: this.data.actions.DIVINATION.icon,
+				content: <Trans id="ast.divination.suggestion.usage.content">
+					Consider coordinating with your co-Astrologian to maximize the amount of time <DataLink action="DIVINATION" /> is up so the party can benefit from extra damage for longer.
+				</Trans>,
+				severity: SEVERITY.MAJOR,
+				why: <Trans id="ast.divination.suggestion.usage.why">
+					About {this.otherAst.countOverwritten} <Plural value={this.otherAst.countOverwritten} one="cast" other="casts" /> of <DataLink action="DIVINATION" /> were overwritten resulting in a loss of at least {this.parser.formatDuration(this.otherAst.timeOverwritten)}.
```
AzariahAeroborn: It would be better to remove the "About" here.  Even if there is a little bit of fuzziness with respect to possible overwrites, we should tune the sensitivity for counting to err on the side of not counting it as an overwrite unless we're sure so that we can be confident with the number of overwrites we're reporting here.

src/parser/jobs/ast/modules/Divination.tsx
```diff
@@ -0,0 +1,297 @@
...
+	private tryCloseWindow(event: Events['statusRemove']) {
+		//note: closing other AST window handled in open window since closing the window doesn't catch what we need because status remove happens before status apply
+
+		//to stop using function if window isn't even open. used when splicing
+		if (this.currentWindow == null) {
+			return
+		}
+
+		//close window and push to history
+		if (event.source === this.parser.actor.id) {
```
AzariahAeroborn: I think it would be better to limit this function to checking if we need to close the window here based on the source, and then passing the timestamp to a second function that actually does close the window.  Then the "close when your window gets overwritten" handler up in tryOpenWindow could just pass the timestamp to the actual close function as well and you don't have to synth a full event for it, while being a little more clear about why you're only trying to close certain windows.  (And similarly onComplete wouldn't have to synth an event either.  You could also just use this.parser.currentTimestamp in the actual close function and never pass around the event.timestamp).

src/parser/jobs/ast/modules/Divination.tsx
```diff
@@ -0,0 +1,297 @@
...
+				</Trans>,
+				severity: SEVERITY.MAJOR,
+				why: <Trans id="ast.divination.suggestion.usage.why">
+					About {this.otherAst.countOverwritten} <Plural value={this.otherAst.countOverwritten} one="cast" other="casts" /> of <DataLink action="DIVINATION" /> were overwritten resulting in a loss of at least {this.parser.formatDuration(this.otherAst.timeOverwritten)}.
+				</Trans>,
+			}))
+		}
+	}
+
+	// just output, no suggestions for now.
```
AzariahAeroborn: Comment is incorrect, as we are generating suggestions.

src/parser/jobs/ast/modules/Divination.tsx
```diff
@@ -0,0 +1,297 @@
...
+		})
+
+		const actualCasts = tableData.length
+		const noCastsMessage = <p><span className="text-error"><Trans id="ast.divination.messages.no-casts"> There were no casts recorded for <DataLink action="DIVINATION" />.</Trans></span></p>
+
+		//for now, the message is copied directly from DivinationDowntime
+		const message = <p><Trans id="ast.ogcd-downtime.divination.description">
+			<DataLink action="DIVINATION" /> provides Astrologian with a strong amount of raid DPS when stacked together with arcanum.
+			Damage percentage bonuses stack multiplicatively, so it's most optimal to stack it with cards from <DataLink action="DRAW" /> and an Astrologian's <DataLink action="ASTRODYNE" /> when 3 unique seals are obtained. <br/>
+			Try to time the usage to match raid buffs and high output phases of other party members - it's more important to use it on time rather than hold it.
```
AzariahAeroborn: Somewhat contradictory copy here, since on the one hand you're saying "use Divination when you have draw cards out and 3 Astrodyne seals" and on the other hand you're saying "use Divination with other raid buffs and party burst windows".  Maybe just remove the line about "damage percentage bonuses stack multiplicatively..."?

src/parser/jobs/ast/modules/Divination.tsx
```diff
@@ -0,0 +1,297 @@
...
+	trailingGcdEvent?: Events['action']
+	playersBuffed: Array<Actor['id']>
+	overlap: boolean
+}
+
+// in this module we only want to track Divination windows opened by
+// the character selected for analysis. windows that clip into
+// AST Divination will be marked.
+// Used DNC Technicalities as basis for this module. Rewritten from previous module for consistency purposes
+export default class Divination extends Analyser {
```
AzariahAeroborn: No default export.
```suggestion
export class Divination extends Analyser {
```

src/parser/jobs/ast/modules/oGCDs.tsx
```diff
@@ -0,0 +1,125 @@
...
+	DPS: 1,
+}
+
+//type for action
+const TYPES = {
+	HEALS: 'heals',
+	DPS: 'dps',
+}
+
+export default class oGCDs extends CooldownDowntime {
```
AzariahAeroborn: No default export
```suggestion
export class oGCDs extends CooldownDowntime {
```

src/parser/jobs/ast/modules/oGCDs.tsx
```diff
@@ -0,0 +1,125 @@
...
+			content: <Fragment><Trans id="ast.neutral-sect.suggestion.content">
+				Consider using <DataLink action="NEUTRAL_SECT" /> more frequently.
+				Frequent uses can mitigate a large amount of damage over the course of a fight, potentially resulting in fewer required healing GCDs.
+			</Trans></Fragment>,
+		},
+	]
+	override trackedCds = this.COOLDOWNS_TRACKED
+
+	override addJobSuggestions() {
+		this.COOLDOWNS_TRACKED.forEach(cooldownGroup => {
```
AzariahAeroborn: I'm not a huge fan of this approach, since it can potentially generate 6 different suggestions if the user is missing targets on all of the skills separately.  Generating multiple suggestions implies to the user that using cooldowns is 6 times more important than something else that generates a single suggestion, and especially to the extent that we usually allow for a lot of not-use on healing cooldowns it seems like a poor signal to send.

I would also prefer to see the suggestion calculations be based on percentUsed (invert the thresholds, basically), in order to match the way the checklist threshold is handled, so that the calculations are the same in both halves of the module, for future maintainer ease of mind.

src/parser/jobs/ast/modules/oGCDs.tsx
```diff
@@ -0,0 +1,125 @@
...
+const TYPES = {
+	HEALS: 'heals',
+	DPS: 'dps',
+}
+
+export default class oGCDs extends CooldownDowntime {
+	@dependency private suggestions!: Suggestions
+
+	override checklistName =  <Trans id="ast.ogcds.name">Use your OGDs</Trans>
+	override checklistDescription = <Trans id="ast.ogcds.description">Consider using the below actions to help support your party with damage buffs and reduce the amount of GCD healing necessary. </Trans>
```
AzariahAeroborn: "Consider using..." is pretty squishy.  Can this copy be improved to be more assertive?

AzariahAeroborn (changes requested):

src/parser/jobs/ast/modules/oGCDs.tsx
otocephaly: I get this error when I do that
image

src/parser/jobs/ast/modules/Divination.tsx
otocephaly: I get this error when I do that
image

src/parser/jobs/ast/modules/oGCDs.tsx
AzariahAeroborn: You will also need to update the class index to do a named import (import {oGCDs} from ...) instead of a default import without the braces.

otocephaly: works beautifully
thanks!

src/parser/jobs/ast/modules/Divination.tsx
otocephaly: pls ignore this -- resolved in other conversation

src/parser/jobs/ast/modules/Divination.tsx
otocephaly: updated both this one and lightspeed with the same comment

src/parser/jobs/ast/modules/Divination.tsx
otocephaly: removed about
the only doubt I had was the upper limit, not the lower so I thought it was conservative anyway

src/parser/jobs/ast/modules/Divination.tsx
otocephaly: added function to do as described and updated the relating splices as a result

src/parser/jobs/ast/modules/Divination.tsx
otocephaly: removed oop

src/parser/jobs/ast/modules/Divination.tsx
otocephaly: Originally with the wording I was hoping it would come across as aim to use arcana, astrodyne, and divination along with party burst windows, but I see that it did not come across as such
Removed line as suggested for clarity

src/parser/jobs/ast/modules/oGCDs.tsx
otocephaly: Updated to consolidate healing suggestions since subjective anyway
Thresholds updated to match

Updated suggestion calculation using percentUsed

src/parser/jobs/ast/modules/oGCDs.tsx
otocephaly: Updated along with healing suggestions

otocephaly:
Please see draft checklist with split healing / dps

image

otocephaly:
I have a few more changes coming because someone pointed out that Macrocosmos is actually GCD not oGCD oop

otocephaly:
should be good to review again

src/parser/jobs/ast/modules/CastTime.tsx
```diff
@@ -0,0 +1,28 @@
...
+import {filter} from 'parser/core/filter'
+import CoreCastTime from 'parser/core/modules/CastTime'
+
+export class CastTime extends CoreCastTime {
+	private pomIndex: number | null = null
+
+	override initialise() {
+		super.initialise()
+
+		const presenceOfMindFilter = filter<Event>()
```
suShirow: var naming nitpick, can we modify this filter from WHM to reflect the AST's Harmony of ___ by name?

suShirow (dismissed):

src/parser/jobs/ast/modules/CastTime.tsx
otocephaly: lol not nitpicky at all
I should use my eyes when copying instead of committing and praying lol
it's been updated

src/parser/core/modules/CooldownDowntime_copy.tsx
```diff
@@ -0,0 +1,350 @@
+//direct copy of CooldownDowntime (CDDT) to allow for a second use specific to secondary role actions (healing/dps, etc.)
```
AzariahAeroborn: #1451 added a way to do these defensive/healing only role actions, can you either update this PR to use those changes to core CDDT or separate this change into a future PR that uses the updated core CDDT?

AzariahAeroborn: @otocephaly No, sorry, I'm asking to remove this copy and adapt your CDDT usage in light of the updates to core.

AzariahAeroborn (changes requested):

src/parser/core/modules/CooldownDowntime_copy.tsx
otocephaly: done! at least it was easy to do
thanks for that!

AzariahAeroborn (approved):

PR #1401, opened 2021-12-13
DRK - Initial Endwalker data support

AzariahAeroborn:


src/parser/jobs/drk/modules/MultiHitSkills.ts
```diff
@@ -1,23 +1,27 @@
...
 		{
-			aoeAction: ACTIONS.UNLEASH,
-			stActions: [ACTIONS.HARD_SLASH],
+			aoeAction: this.data.actions.ABYSSAL_DRAIN,
+			stActions: [this.data.actions.CARVE_AND_SPIT],
+			minTargets: 4,
+		},
+		{
+			aoeAction: this.data.actions.UNLEASH,
+			stActions: [this.data.actions.HARD_SLASH],
```
ayyaruq, line 24: I believe this is 3 now as well. 1-2-3 combo is 770 potency, AoE combo is 290, so you'd need 3 targets to overtake it.

src/parser/jobs/drk/modules/OGCDDowntime.ts
```diff
@@ -11,12 +11,15 @@ export class OGCDDowntime extends CooldownDowntime {
 		},
 		{
 			cooldowns: [this.data.actions.DELIRIUM],
-			firstUseOffset: 15000,
+			firstUseOffset: 5000,
```
ayyaruq: While English openers use Delirium early and sit on the stacks, Japanese don't, so I'd leave this one alone or set it to like 7500ms (JP seems to do it after 3rd GCD).

ayyaruq (commented):
Not requesting changes but 2 things to double check.

src/parser/jobs/drk/modules/MultiHitSkills.ts
AzariahAeroborn: 1-2-3 combo is 770 potency (plus 600MP and 20 blood) over 3 hits, while AoE is 290 per target (plus 600MP and 20 blood) over 2 hits. 

So over 6 hits, you've made 1540 potency + 1200 MP + 40 blood on the single target vs 870 potency per target + 1800 MP + 60 blood on the AoE, making 2 the appropriate break point.

src/parser/jobs/drk/modules/OGCDDowntime.ts
AzariahAeroborn: Updated.

ayyaruq (approved):

PR #1405, opened 2021-12-14
SMN - Endwalker data

KelosPelmand:
Also touches SCH file enough to get rid of ACN data files.


src/parser/jobs/smn/modules/index.js
```diff
@@ -1,11 +1,11 @@
 import {Interrupts} from 'parser/core/modules/Interrupts'
+import {PetTimeline} from 'parser/core/modules/PetTimeline'
```
ayyaruq, line 2: What's the point of having a SMN specific PetTimeline module if using core now?

src/data/ACTIONS/root/SCH.ts
```diff
@@ -6,6 +6,31 @@ export const SCH = ensureActions({
...
+	RESURRECTION: {
+		id: 173,
+		name: 'Resurrection',
+		icon: 'https://xivapi.com/i/000000/000511.png',
+		onGcd: true,
+		speedAttribute: Attribute.SPELL_SPEED,
+		castTime: 8000,
+	},
+
+	SCH_PHYSICK: {
```
ayyaruq, line 25: Does SCH code handle this rename?

src/data/ACTIONS/root/SMN.ts
```diff
@@ -1,58 +1,87 @@
 import {Attribute} from 'event'
-import {PETS} from '../../PETS'
 import {ensureActions} from '../type'
 
 // use action id of a skill id in the group to avoid potential duplications
 export const SMN_COOLDOWN_GROUP = {
```
ayyaruq: This was only ever imported by ACN, is it still needed?

ayyaruq (commented):
Couple of Q's.

src/parser/jobs/smn/modules/index.js
KelosPelmand: I.. umm.. don't remember changing that to use core. (maybe vs did it?)  Either way, I'll remove the apparently unused SMN specific one after work and double check that nothing breaks.

src/data/ACTIONS/root/SCH.ts
KelosPelmand: I tested one SCH log that had a SCH Physick and it seemed ok.  I'll look for a few more when I'm testing again tonight for further verification.

src/data/ACTIONS/root/SMN.ts
KelosPelmand: To clarify, are you asking if the Pets data file should be removed?  I hadn't thought about that, but I can try removing it too and see if nothing breaks.  There are SCH and MCH pets in that file as well, so I'll run some of those parses as well to ensure there's not some hidden reference to them.

KelosPelmand:
Updated as requested.  Checked some SMN, MCH, and SCH parses, and all seem fine as related to skill renames and removal of Pets file.

src/parser/jobs/smn/modules/index.js
ayyaruq: If the local one is relevant just change the import here.

src/data/ACTIONS/root/SMN.ts
ayyaruq: `SMN_COOLDOWN_GROUP` specifically, which isn't part of the diff, but since it was only imported by the ACN data I don't think it's needed anymore? GH gets weird with how it shows comments outside the actual diff lines.

ayyaruq (approved):

PR #1406, opened 2021-12-15
RDM - Endwalker data

thatcsguy:
This fixes most of the obvious issues with what's currently live. Acceleration-related modules will still need fixing later.

- All spell/ability changes
- Renaming Embolden buffs due to rework
- Changing to Veraero III/Verthunder III in Dualcast suggestions.


src/data/ACTIONS/root/RDM.ts
```diff
@@ -34,37 +34,37 @@ export const RDM = ensureActions({
...
+		potency: 180,
 		cooldownGroup: 10,
 	},
 	CORPS_A_CORPS: {
 		id: 7506,
 		name: 'Corps-a-corps',
 		icon: 'https://xivapi.com/i/003000/003204.png',
-		cooldown: 40000,
+		cooldown: 35000,
 		potency: 130,
```
AzariahAeroborn, line 62: Need to add charges for the movement skills.

`charges: 2`

AzariahAeroborn (changes requested):

src/data/ACTIONS/root/RDM.ts
thatcsguy: Good catch.

AzariahAeroborn (approved):

PR #1409, opened 2021-12-15
RDM - Initial EW module updates

thatcsguy:
- Updated gauge (mana values and Manafication changes)
- Removed Acceleration logic from melee combo module
- Moving Corps-a-corps and Engagement/Displacement into GCCD as they no longer reset with Manafication (also removed their previous modules)
- Forgot Acceleration charges and Magick Barrier in original RDM data update
- Ignoring non-damaging utility oGCDs in Embolden display
- Adjusted various outdated suggestion text


src/parser/jobs/rdm/modules/MeleeCombos.tsx
```diff
@@ -531,9 +483,9 @@ export class MeleeCombos extends Analyser {
 					// Verify that using the finisher of higherMana won't put us out of balance at the end
 					const finisherManaGain = (this.gauge.gaugeModifiers.get(higherManaState.actions.finisher)?.white ?? 0) || (this.gauge.gaugeModifiers.get(higherManaState.actions.finisher)?.black ?? 0)
 					if (!((newHigherMana + finisherManaGain - newLowerMana) > MANA_DIFFERENCE_THRESHOLD)) {
-						// This is a net gain whether or not acceleration would be available - we can now fish for an additional proc of higherMana, push onto stack
+						// This is a net gain - we can now fish for an additional proc of higherMana, push onto stack
 						possibleDelays.push({
-							finisher: [lowerManaState.actions.proc, lowerManaState.actions.dualcast, this.data.actions.ACCELERATION, higherManaState.actions.finisher],
```
thatcsguy, line 536: This is the case I'm least sure about. The old comment mentioned that the acceleration state was irrelevant, but then the finisher included it regardless (without checking if it was even up).

Cephid: This specific case was the state wherein you had say 95|99 going into your combo, with a VerStone proc and no Acceleration.  If you did your combo and finished with VerFlare you'd potentially gain a proc without risking the problem of exceeding the mana different threshold, and therefore even if Acceleration wasn't available the correct choice here was to VerFlare even if the proc wasn't promised.  The inclusion of the Acceleration data was for the display.

thatcsguy: Okay then I think my change is fine. Thanks.

Cephid:
From a RDM Logic standpoint the changes here look to be a good EW start.

AzariahAeroborn (approved):

PR #1413, opened 2021-12-16
MCH - Update reassemble suggestion copy

hintxiv:


src/parser/jobs/mch/modules/Reassemble.tsx
```diff
@@ -106,7 +106,7 @@ export class Reassemble extends Analyser {
 		this.suggestions.add(new TieredSuggestion({
 			icon: this.data.actions.REASSEMBLE.icon,
 			content: <Trans id="mch.reassemble.suggestions.bad-gcds.content">
-				On single targets <DataLink action="REASSEMBLE" /> should only ever be used on <DataLink action="DRILL" /> and <DataLink action="AIR_ANCHOR" />, as they're your strongest GCDs by a large margin.
+				On single targets <DataLink action="REASSEMBLE" /> should only ever be used on <DataLink action="CHAIN_SAW" />, <DataLink action="DRILL" />, or <DataLink action="AIR_ANCHOR" />, as they're your strongest GCDs. On multiple targets, you can also use it on <DataLink action="SCATTERGUN" />.
```
ayyaruq: I feel like this needs a bit of rephrasing, you're effectively telling users off rather than suggesting what they can do better. See https://docs.xivanalysis.com/composing-job-analysis#suggestions.

ayyaruq (changes requested):

ayyaruq (approved):

PR #1415, opened 2021-12-17
Updated ACTIONS and STATUSES with new PLD information.

Dynme:
Adds new abilities and status effects at 80+, updates potencies and durations of various skills.


ayyaruq (approved):
Data looks good, I'll note that PLD doesn't seem to use potency data anywhere so it's probably not even necessary and adds extra maintenance effort in future patch layers.

src/data/ACTIONS/root/PLD.ts
```diff
@@ -238,7 +276,7 @@ export const PLD = ensureActions({
 		icon: 'https://xivapi.com/i/002000/002512.png',
 		onGcd: false,
 		cooldown: 10000,
-		statusesApplied: ['INTERVENTION'],
+		statusesApplied: ['INTERVENTION', 'KNIGHTS_RESOLVE', 'KNIGHTS_BENEDICTION'],
```
AzariahAeroborn: Having multiple actions with the same `statusesApplied` may result in incorrect actions being synthesized if the actions are being used pre-pull.  If either Intervention or Holy Sheltron can reasonably be used pre-pull, we should exclude these duplicated statusesApplied (and specifically for Intervention, we're probably going to need to add some special handling to synthesize an action by the PLD for a status applied to another actor).

At least as of now, `statusesApplied` is only used for that data fix-up process of synthesizing pre-pull actions, so it's not super critical to keep the data exactly in line with the actual job actions.

AzariahAeroborn (changes requested):

AzariahAeroborn (approved):

PR #1416, opened 2021-12-17
RDM - New CastTime module to account for Acceleration instant casts

thatcsguy:
- Adding a new CastTime module for RDM that is responsible for any cast time adjustments
- Added logic for Acceleration instant casts to new module
- Moved existing DualCast cast adjustments to this module (leaving the DualCast module just for finding player mistakes)

This is what the weaving issue and timeline looked like prior to this change. Weaving logic couldn't determine that the Veraero was instant, it was assuming the full cast time:
image
Afterwards, this weaving issue is no longer present.


Cephid:
Looks good from a RDM standpoint.  The only thing we might need to consider later is tracking losses of Acceleration; however we're not really certain it's a concern at this time so not tracking it is fine for now.

AzariahAeroborn (approved):

PR #1418, opened 2021-12-17
NIN - Update Huton analyser for 6.0

Toastdeib:
One of these days I'd love to rewrite this nightmare of an analyser from scratch, but for now, it's updated to account for the changes in Endwalker (60s timer, new extension actions). I also fixed a small bug in the Ninki sim while I was at it, since it related to the way Phantom Kamaitachi is handled under the hood (it was only generating 5 in the sim, not 10).

### DIFF
--- a/src/data/ACTIONS/root/NIN.ts
+++ b/src/data/ACTIONS/root/NIN.ts
@@ -116,6 +116,14 @@ export const NIN = ensureActions({
 		speedAttribute: Attribute.SKILL_SPEED,
 	},
 
+	PHANTOM_KAMAITACHI_BUNSHIN: {
+		id: 25775,
+		name: 'Phantom Kamaitachi',
+		icon: 'https://xivapi.com/i/002000/002929.png',
+		onGcd: true,
+		speedAttribute: Attribute.SKILL_SPEED,
+	},
+
 	TEN: {
 		id: 2259,
 		name: 'Ten',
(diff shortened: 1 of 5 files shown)


src/data/ACTIONS/root/NIN.ts
```diff
@@ -116,6 +116,14 @@ export const NIN = ensureActions({
 		speedAttribute: Attribute.SKILL_SPEED,
 	},
 
+	PHANTOM_KAMAITACHI_BUNSHIN: {
+		id: 25775,
+		name: 'Phantom Kamaitachi',
+		icon: 'https://xivapi.com/i/002000/002929.png',
+		onGcd: true,
```
KelosPelmand, line 123: nit, but if this is the pet skill, it's not really onGcd or affected by a speedAttribute

KelosPelmand (approved):
one nit, but otherwise looks good

src/data/ACTIONS/root/NIN.ts
Toastdeib: It seemed correct to me to keep the attributes here because this is kind of a weird case. We get `action` events with the player skill ID (25774) but the `damage` events come through as 25775 instead, as two halves of the same whole. It's not a thing the "pet" here does independently, it's only ever triggered manually when the player hits the Phantom Kamaitachi button, and it respects SkS/is a GCD.

KelosPelmand: SMN pets have some similar cases (the actual summons this expansion and egi assaults last expansion).  Not something I'd recommend blocking over either way though.

AzariahAeroborn (approved):

PR #1419, opened 2021-12-17
MCH - Wildfire (it's the remix)

hintxiv:
Detects debuff stack events instead of the old method counting damaging GCDs in WF, so hopefully it's a bit more accurate. Additionally, it will now show GCDs which were cast inside the WF window but "ghosted" the damage snapshot for clarity.

image


src/parser/jobs/mch/modules/Wildfire.tsx
```diff
@@ -0,0 +1,204 @@
...
+import {Timeline} from 'parser/core/modules/Timeline'
+import React, {Fragment} from 'react'
+import {Message} from 'semantic-ui-react'
+
+// We always want 6 GCDs in WF
+const EXPECTED_GCDS = 6
+
+const SEVERITIES = {
+	BAD_WILDFIRE: {
+		1: SEVERITY.MINOR,
```
Toastdeib, line 22: Not a math nerd so take this with a pinch of salt, but I would personally think the tiers should be more like `MEDIUM` for 1 and `MAJOR` for 3. WF is a pretty important part of the 2-minute burst cycle, so screwing up even one feels like it should be more than `MINOR`.

src/parser/jobs/mch/modules/Wildfire.tsx
```diff
@@ -0,0 +1,204 @@
...
+	)
+
+	private actionHook?: EventHook<Events['action']>
+
+	private actionFilter = filter<Event>()
+		.source(this.parser.actor.id)
+		.type('action')
+
+	override initialise() {
+		this.addEventHook(filter<Event>()
```
Toastdeib: Minor nit but it'd be a little cleaner to do a `const playerFilter = filter<Event>().source(this.parser.actor.id)` and then reuse that in each of the 3 hooks.

src/parser/jobs/mch/modules/Wildfire.tsx
```diff
@@ -0,0 +1,204 @@
...
+		this.history.closeCurrent(event.timestamp)
+
+		if (this.actionHook != null) {
+			this.removeEventHook(this.actionHook)
+			this.actionHook = undefined
+		}
+	}
+
+	private onAction(event: Events['action']) {
+		this.history.doIfOpen(current => current.actionEvents.push(event))
```
Toastdeib: Is there any filtering going on to  make sure that the target of your actions is the one afflicted with WF in multi-target situations? Looking through this, it seems like it's all based purely on the WF buff on the actual player. Same goes for the compound damage tracking. Basically what I'm wondering is if this will give weird results in a 2-target scenario (e.g. TEA) if you were to apply WF to one target and then unload on the other.

src/parser/jobs/mch/modules/Wildfire.tsx
hintxiv: I'm okay with minor here only because it's very easy to miss a single GCD in WF if you have high ping or just needed to clip a second wind or something.

src/parser/jobs/mch/modules/Wildfire.tsx
hintxiv: 👍

src/parser/jobs/mch/modules/Wildfire.tsx
hintxiv: The logic is counting WF debuff stacks themselves, so the GCD count should always be correct. Somehow I missed that this will throw off the "ghosting" logic though, and I can't think of a good way to resolve it so I'm opting to just leave it out. I updated the module description to hopefully make it clear why the GCD count may be lower than the # of GCDs shown in the rotation column.

Toastdeib (approved):

PR #1420, opened 2021-12-17
BRD - 6.0 support

hintxiv:
There's a lot of room to grow here, but Bard analysis is functional and covers all the basics so I'm calling it supported.


src/parser/jobs/brd/modules/MultiHitSkills.ts
```diff
@@ -4,20 +4,24 @@ import DISPLAY_ORDER from './DISPLAY_ORDER'
 
 export default class MultiHitSkills extends AoEUsages {
```
AkairyuGestalter: We're trying to move away from `export default` so this should just be `export class` (and update the index to import it as `import {MultiHitSkills}`)

src/parser/jobs/brd/modules/OGCDDowntime.ts
```diff
@@ -7,24 +7,24 @@ export default class OGCDDowntime extends CooldownDowntime {
 	override trackedCds = [
```
AkairyuGestalter, line 7: Ditto the `export default`  comment here

src/parser/jobs/brd/modules/index.ts
```diff
@@ -15,5 +14,4 @@ export default [
 	OGCDDowntime,
 	RagingStrikes,
 	RefulgentProcs,
```
AkairyuGestalter, line 16: I know you weren't otherwise touching Raging Strikes and Refulgent Procs on this PR, but it would be nice if we could fix up the `export default`s on those too while we're at it

AkairyuGestalter (approved):

PR #1422, opened 2021-12-17
NIN - Raiju analyser bugfix

Toastdeib:
Turns out I did a stupid with my initial Raiju analyser implementation. This one correctly handles stacks being built up and spent, so it won't ding people erroneously anymore.


KelosPelmand (approved):
Looks good overall.  For what it's worth, this looks like it will not ding for stacks lost when dying since you should get a statusRemove event on death.

Toastdeib:
I can confirm you're correct on that; one of the parses I was testing with only got dinged for 2 dropped stacks but they had at least 2-3 others caused by their 6 deaths that it didn't count. I _think_ that should be the desired behavior for this analyser; AFAIK, we usually don't ding on other things like that (e.g. lost gauge due to death), since we're already yelling at them for dying in the first place.

KelosPelmand:
Sounds fine to me.  I was pointing it out in case you hadn't considered it one way or the other.

AzariahAeroborn (approved):

PR #1423, opened 2021-12-17
DRG: Add Firstminds' Focus Gauge, mark supported for 6.0

ebshimizu:
Dependent on #1370, will take this PR out of draft mode once that PR is merged.

This PR adds the Firstminds' Focus gauge and tracks uses of Wyrmwind Thrust. It also adjusts the expected GCDs for Battle Litany from 8 to 6 in accordance with the reduced buff duration for BL.

This, and #1370, should be all that DRG needs for a base level of support in 6.0.


src/data/ACTIONS/root/DRG.ts
```diff
@@ -74,7 +74,7 @@ export const DRG = ensureActions({
 		icon: 'https://xivapi.com/i/000000/000317.png',
 		onGcd: true,
 		speedAttribute: Attribute.SKILL_SPEED,
-		potency: 140,
+		potency: 250,
```
KelosPelmand: There is a potency here and a combo potency, should this one be the non-comboed 140?

src/parser/jobs/drg/modules/Gauge.tsx
```diff
@@ -0,0 +1,78 @@
...
+			content: <Trans id="drg.gauge.suggestions.overcapped-fmf.content">
+				Make sure to use <DataLink action="WYRMWIND_THRUST" /> before <DataLink action="RAIDEN_THRUST" /> or <DataLink action="DRACONIAN_FURY" /> when you already have two charges of Firstminds' Focus to prevent losing uses of <DataLink action="WYRMWIND_THRUST" /> by overcapping.
+			</Trans>,
+			tiers: {
+				1: SEVERITY.MEDIUM,
+				2: SEVERITY.MAJOR,
+			},
+			value: fmfOvercap,
+			why: <Trans id="drg.gauge.suggestions.overcapped-fmf.why">
+				<Plural value={fmfOvercap} one="# Firstminds' Focus charge" other="# Firstminds' Focus charges" /> were lost.
```
KelosPelmand: Naming consideration: A 'charge' is used as a game term refering to having a use of an ability charged.  Perhaps another term such as stack would be more appropriate here and in the content message?

src/data/ACTIONS/root/DRG.ts
ebshimizu: yes thanks for catching!

src/parser/jobs/drg/modules/Gauge.tsx
ebshimizu: Stack works for me, the guide just calls it "Firstminds' Focus" and i definitely forgot charge is used in-game.

KelosPelmand (approved):
Looks good to me now

AzariahAeroborn (approved):

PR #1424, opened 2021-12-17
SCH - EW data updates

thatcsguy:
- PR #1405 moves everything out of the ACN action file, but we made the same spell changes. Shouldn't matter who merges first.


src/data/ACTIONS/root/SCH.ts
```diff
@@ -101,7 +110,29 @@ export const SCH = ensureActions({
 		name: 'Art of War',
 		icon: 'https://xivapi.com/i/002000/002819.png',
 		onGcd: true,
-		speedAttribute: Attribute.SPELL_SPEED,
+	},
```
KelosPelmand, line 139: Art of War and Art of War II appear to be affected by spell speed in game.

KelosPelmand (changes requested):
The SMN data pr straight up deletes ACN.ts after moving the 3 skills still used by max level SCH out of it.  If you want to keep the remaining sub-90 skills, I would personally recommend putting them into SCH.ts instead going forward.

src/data/ACTIONS/root/SCH.ts
thatcsguy: Sorry, was thinking that SpS reduced recasts by default but I see now that's not the case. Added.

> The SMN data pr straight up deletes ACN.ts after moving the 3 skills still used by max level SCH out of it. If you want to keep the remaining sub-90 skills, I would personally recommend putting them into SCH.ts instead going forward.

Yep, I'll move this back to draft until you can get that merged and I'll deal with any conflicts. I don't believe anything other than what you ported to SCH is needed.

AzariahAeroborn (approved):

PR #1425, opened 2021-12-18
RPR - Communio is 2.5s recast

hintxiv:


AkairyuGestalter (changes requested):
Why do we even have a cooldown on this? It doesn't have a long cooldown like the dances/air anchor, etc. and we don't generally need to specify the recast for spells, just the cast time. I think it's technically more appropriate to remove the cooldown entirely.

AkairyuGestalter (approved):

PR #1429, opened 2021-12-18
RPR - Tincture

hintxiv:


src/parser/jobs/rpr/modules/Tincture.tsx
```diff
@@ -0,0 +1,60 @@
...
+		this.addEvaluator(new ExpectedActionsEvaluator({
+			expectedActions: [
+				{
+					action: this.data.actions.COMMUNIO,
+					expectedPerWindow: 2,
+				},
+			],
+			suggestionIcon: this.data.actions.INFUSION_STR.icon,
+			suggestionContent: <Trans id="rpr.tincture.suggestions.trackedActions.content">
+				Try to fit two <DataLink action="COMMUNIO"/>s while under the effects of a Tincture.
```
KelosPelmand: How will this look when the action link translates?  Would "Try to use <Communio> two times while under the effects of a Tincture." or ".. two times during each Tincture" be better?  Maybe this doesn't matter since the whole line should get translated?

src/parser/jobs/rpr/modules/Tincture.tsx
```diff
@@ -0,0 +1,60 @@
...
+}
+
+// Remove support actions from the table to avoid clutter
+const IGNORED_ACTIONS: ActionKey[] = [
+	'ARCANE_CREST',
+	'ARMS_LENGTH',
+	'BLOODBATH',
+	'FEINT',
+	'HELLS_EGRESS',
+	'HELLS_INGRESS',
```
ayyaruq, line 22: Maybe also Regress?

ayyaruq (commented):
This will need to merge `endwalker` to unfuck CI. Q about adding Regress to ignored actions tho.

AkairyuGestalter (approved):

PR #1431, opened 2021-12-18
DRK 6.0 initial support

AzariahAeroborn:
What it says on the tin.  

- Added module to track Salt and Darkness usage: should be used 1 time per Salted Earth, and warn if it does not do damage to at least 1 target each use.
- Adjusted Delirium, as it's no longer a buff window but simply 3 stacks to use on free Blood attacks before the stacks expire
- Updated the list of functional tests with assistance from Nikroulah on the Balance
- Fixed new skills not being added to the ActionTimeline order
- Fixed wrong ID added for Salt and Darkness
- Fixed Unmend so it now is marked as not breaking combo


src/parser/jobs/drk/modules/ActionTimeline.tsx
```diff
@@ -9,17 +9,20 @@ export class ActionTimeline extends CoreActionTimeline {
...
 		// oGCD Damage
 		'LIVING_SHADOW',
+		'SHADOWBRINGER',
 		'EDGE_OF_SHADOW',
 		'FLOOD_OF_SHADOW',
 		'CARVE_AND_SPIT',
 		'ABYSSAL_DRAIN',
 		'PLUNGE',
 		'SALTED_EARTH',
+		'SALT_AND_DARKNESS',
```
ayyaruq, line 19: Could probably group these but whatevs. Consider C&S and Abyssal too since they're on a shared CD anyway.

src/parser/jobs/drk/modules/ActionTimeline.tsx
```diff
@@ -9,17 +9,20 @@ export class ActionTimeline extends CoreActionTimeline {
...
 		// Personal Mitigation
 		'LIVING_DEAD',
 		'SHADOW_WALL',
 		'RAMPART',
 		'DARK_MIND',
+		'OBLATION',
```
ayyaruq: Oblation is also part mitigation, you can put it on anyone, just like TBN.

src/parser/jobs/drk/modules/SaltAndDarkness.tsx
```diff
@@ -0,0 +1,80 @@
...
+	private onSaltDarknessDamage(event: Events['damage']) {
+		if (isSuccessfulHit(event)) {
+			this.saltAndDarknessDamaged++
+		}
+	}
+
+	private onComplete() {
+		const unused = this.saltedEarthUsed - this.saltAndDarknessUsed
+		this.suggestions.add(new TieredSuggestion({
+			icon: this.data.actions.SALT_AND_DARKNESS.icon,
```
ayyaruq, line 56: I'd maybe swap the icon for one of these, just so it's easier to see when both are triggered.

ayyaruq (commented):
LGTM, nice idea on checklist for Delirium. Smol questions on ActionTimeline but that's really display preference. Consider adding a thing to changelog that 6.0 is explicitly supported tho, prob as a separate entry.

Some things missing tho - Unmend no longer breaks combo, Esteem has its own Shadowbringer (missing in actions), and Plunge is missing the cooldown reducer in OGCDDowntime and a module to actually adjust the cooldown.

src/parser/jobs/drk/modules/ActionTimeline.tsx
AzariahAeroborn: Moved below TBN.

src/parser/jobs/drk/modules/SaltAndDarkness.tsx
AzariahAeroborn: I don't really want to use the Salted Earth icon for "didn't use S&D", and it doesn't make sense to use it for "S&D didn't hit anything", so I'd rather keep the duplicated icon.

AzariahAeroborn:
Unmend is being marked as does not break combo in this PR already.

src/parser/jobs/drk/modules/ActionTimeline.tsx
AzariahAeroborn: I think I prefer having S&D separated since we're making separate recommendations about using that specific skill.  Similarly for keeping AD separated - it's more clear that there's an incorrect skill used if that timeline row shows up on a fight that doesn't have relevant AoE opportunities.

src/parser/jobs/drk/modules/OGCDDowntime.ts
```diff
@@ -13,7 +13,13 @@ export class OGCDDowntime extends CooldownDowntime {
 			cooldowns: [this.data.actions.DELIRIUM],
 			firstUseOffset: 7500,
 		},
-		{cooldowns: [this.data.actions.PLUNGE]},
+		{
+			cooldowns: [this.data.actions.PLUNGE],
+			resetBy: {
+				actions: [this.data.actions.UNMEND],
+				refundAmount: 5,
```
ayyaruq: This is in milliseconds.
```suggestion
				refundAmount: 5000,
```

src/data/ACTIONS/root/DRK.ts
```diff
@@ -86,6 +86,7 @@ export const DRK = ensureActions({
 		name: 'Carve And Spit',
 		icon: 'https://xivapi.com/i/003000/003058.png',
 		cooldown: 60000,
+		cooldownGroup: 1,
```
ayyaruq: CDG here is 14.
```suggestion
		cooldownGroup: 14,
```

src/data/ACTIONS/root/DRK.ts
```diff
@@ -99,6 +100,7 @@ export const DRK = ensureActions({
 		name: 'Abyssal Drain',
 		icon: 'https://xivapi.com/i/003000/003064.png',
 		cooldown: 60000,
+		cooldownGroup: 1,
```
ayyaruq:
```suggestion
		cooldownGroup: 14,
```

ayyaruq (changes requested):

ayyaruq (approved):
LGTM, can you merge latest `endwalker` branch so CI unfucks itself?

AzariahAeroborn:
Added translation tag as requested in #1440 

Resolves #1440

ayyaruq (approved):
LGTM, complete aside but at some point might be worth using TimerGauge for Darkside.

PR #1432, opened 2021-12-18
Rdm minor fixes

Cephid:
Minor fixes for areas not yet referencing Thunder_III and Aero_III.  Updated the Interrupts suggestion to include Acceleration and state that Enchanted Reprise should only be used a last resort skill.  Marked supported for 6.0 as there's no other changes that have been requested at this time.


thatcsguy:
LGTM, and I agree with marking RDM as supported.

src/parser/jobs/rdm/index.tsx
```diff
@@ -18,19 +19,19 @@ export const RED_MAGE = new Meta({
...
 	changelog: [
-		// {
-		// 	date: new Date('2021-11-19'),
-		// 	Changes: () => <>The changes you made</>,
-		// 	contrubutors: [CONTRIBUTORS.YOU],
-		// },
+		{
+			date: new Date('2021-12-18'),
+			Changes: () => <>Minor fixes to suggestion text and shown abilities</>,
+			contributors: [CONTRIBUTORS.LEYLIA],
```
AzariahAeroborn: Since this is the initial support commit, the changelist should probably also include indications of support for Endwalker.  Apparently I did a poor job of checking earlier reviews for a changelog... can we retcon some of the earlier PR changes in here, or at least include them in the 12-18 changelog here?

AzariahAeroborn (changes requested):

src/parser/jobs/rdm/index.tsx
thatcsguy: I can make a changelog PR for my earlier changes. One min.

AzariahAeroborn (approved):

PR #1434, opened 2021-12-18
SGE - Gauge suggestions & ABC bugfix

AkairyuGestalter:
- Added suggestion to not allow the Addersgall gauge to sit at full for too long
- Added suggestion to not waste Addersgall timer due to capping the gauge with Rhizomata
- Added suggestion to not waste Addersting by overcapping the gauge
- Added gcdRecast property to Pneuma's data definition to fix the Always Be Casting metric
- Reordered the gauge object definitions to bring Addersting to the 'back' of the gauge graph, better visual clarity when collapsed


src/parser/jobs/sge/modules/Gauge.tsx
```diff
@@ -177,4 +187,54 @@ export class Gauge extends CoreGauge {
...
+		const addersgallLeniency = ADDERSGALL_TIME_REQUIRED / 2
+		const forceGainUtaWindows = this.unableToAct.getWindows().filter(uta => Math.max(0, uta.end - uta.start) >= ADDERSGALL_TIME_REQUIRED)
+		const addersgallExpirationTime = this.addersgallTimer.getExpirationTime(addersgallLeniency, this.parser.currentEpochTimestamp, forceGainUtaWindows, addersgallLeniency)
+		const lostAddersgall = Math.floor(addersgallExpirationTime / ADDERSGALL_TIME_REQUIRED)
+
+		if (lostAddersgall > 0) {
+			this.suggestions.add(new Suggestion({
+				icon: this.data.actions.KERACHOLE.icon,
+				content: <Trans id="sge.gauge.suggestions.lost-addersgall.content">
+					You lost Addersgall due to capping the gauge and letting the timer stop. Your Addersgall actions are your primary healing and mitigation tools, so you should try to use another one before regaining your third stack.
```
AzariahAeroborn: The WHM copy for their lily gauge is:
```
Use <DataLink action="AFFLATUS_RAPTURE" /> or <DataLink action="AFFLATUS_SOLACE" /> before using other GCD heals. It's okay to cap your lilies if you don't need to heal, move, or weave with them.
```

Should this copy be similar?  Addersgall winds up being part of MP management instead of having a DPS refund, but otherwise it seems similar in function to a lily to me - first choice for GCD heals, but not worth prioritizing running over the cap if you don't need the GCD heal.

AkairyuGestalter: Sage is arguably in a slightly different position from WHM: Addersgall actions are all oGCDs, and GCD healing gets further punished by not procing Kardia, so I feel like the language as it stands is probably appropriate.

src/parser/jobs/sge/modules/Gauge.tsx
```diff
@@ -177,4 +191,65 @@ export class Gauge extends CoreGauge {
...
+				5: SEVERITY.MAJOR,
+			},
+			value: this.adderstingGauge.overCap,
+			why: <Trans id="sge.gauge.suggestions.addersting-overcap.why">
+				<Plural value={this.adderstingGauge.overCap} one="# Addersting stack" other="# Addersting stacks"/> lost due to overcap.
+			</Trans>,
+		}))
+
+		// Calculate the number of possible addersgall stacks the player could have used: Initial max stacks, plus the maximum possible stacks if they kept the timer rolling the entire fight (aside from first-use leniency), and maximum usage of Rhizomata
+		const potentialAddersgalls = ADDERSGALL_MAX_STACKS + Math.floor((this.parser.pull.duration - addersgallLeniency)  / ADDERSGALL_TIME_REQUIRED) + this.cooldownDowntime.calculateMaxUsages({cooldowns: [this.data.actions.RHIZOMATA]})
```
AzariahAeroborn: Fights with a forced downtime window >1 minute aren't totally uncommon (e1s, e8s, most ultimates)... should we make some effort to consider that downtime when calculating maximum possible addersgall, similar to what we do with forceGainUtaWindows for the overcap suggestion?

For that matter, it looks like this will result in a `used / possible` value that doesn't match the number lost in the suggestion if there's any overcap during downtime, which seems like a poor UX.

AkairyuGestalter: Revised the potential stack calculation such that it should take the uta info into account now, by way of summing the actual amount generated with the stacks that were lost to timer downtime and the unused uses of Rhizomata.

AzariahAeroborn (approved):

PR #1435, opened 2021-12-19
RDM - Displaying recommended actions when suggesting a delayed combo

thatcsguy:
- I'm assuming this was the original intent of the module, as the actions were already being tracked, but just in a format that prevented their display.
- Finisher tracking was looking at the last action in the combo, which isn't very useful as it's always Resolution. This was causing every "correct" combo to still display the message to "See the suggestions section for finisher guidelines." Now tracking Verholy/Verflare, which displays the message only when appropriate.

Before:
image

After:
image


thatcsguy:
Per discussion in Discord, moved long suggestion text out of the main table and into footnotes which only appear when they are actually suggested.

Single suggestion:
image

Mutiple suggestions (there is a "delay combo" above the screenshot, they are ordered correctly):
image

AzariahAeroborn (approved):

PR #1443, opened 2021-12-20
Update WHM Actions&Statuses for 6.0

Dot-Inni:
Added EW actions to actions and corresponding statuses as well. Also changed all of the old actions that got changes with EW (unless I missed something) and deleted Fluid Aura.


ayyaruq (approved):
LGTM

PR #1445, opened 2021-12-20
Requiescat window changes

Dynme:
Update suggestions to cast 4 Holy Spirits/Circles before using Confiteor.
Update buff duration to 30s.
Remove tracking for the old buff that required 80% MP on use.


src/parser/jobs/pld/modules/Requiescat.tsx
```diff
@@ -15,27 +15,28 @@ const SEVERITIES = {
...
 	},
 }
-const EXPECTED_REQUIESCAT_CASTS = 4
+const EXPECTED_REQUIESCAT_CASTS = 5
 // When calculating rushing, adjust the start of the window by 1.5 seconds to allow for using Requiescat in the first weave slot
 const WINDOW_START_FORGIVENESS_FOR_RUSHING = 1500
-const REQUIESCAT_DURATION = 12000
+const REQUIESCAT_DURATION = 30000
 
 class RequiescatUsageEvaluator implements WindowEvaluator {
```
AzariahAeroborn: This evaluator should probably be removed completely.  At a minimum, the copy on the suggestion needs to be updated so that it's talking about not using Confiteor instead of missing the Requiescat buff due to not enough MP, but the stock ExpectedActionsEvaluator is probably a better fit for tracking Confiteor and reporting on missed usages.

It might also be worth looking at changing this analyser to not use core BuffWindow and just kind of do its own tracking, so that we can also track the post-Confiteor combo as well in the same module (can we just call that the afterparty module, since it comes after the Confetti is deployed? 🤔)

src/parser/jobs/pld/modules/Requiescat.tsx
```diff
@@ -171,10 +167,7 @@ export class Requiescat extends BuffWindow {
 	override output() {
 		return <Fragment>
 			<Message>
-				<Trans id="pld.requiescat.table.note">Each of your <DataLink status="REQUIESCAT" /> windows should contain 4 spells at minimum to maintain the alignment of your rotation.
-				Most of the time, a window should consist of 4 casts of <DataLink action="HOLY_SPIRIT" /> or <DataLink action="HOLY_CIRCLE" /> and end with a cast of <DataLink action="CONFITEOR" />.
-				However, under some circumstances, it is useful to drop one <DataLink action="HOLY_SPIRIT"/> per minute in order to better align your rotation with buffs or mechanics.
-				If you don't have a specific plan to do this, you should aim for 4 casts of <DataLink action="HOLY_SPIRIT" /> per <DataLink status="REQUIESCAT" /> window.</Trans>
+				<Trans id="pld.requiescat.table.note">Each of your <DataLink status="REQUIESCAT" /> windows should contain 5 spells, consisting of 4 casts of <DataLink action="HOLY_SPIRIT" /> or <DataLink action="HOLY_CIRCLE" /> and endING with a cast of <DataLink action="CONFITEOR" />.</Trans>
```
AzariahAeroborn: Unexpected capitalization of `endING`
```suggestion
				<Trans id="pld.requiescat.table.note">Each of your <DataLink status="REQUIESCAT" /> windows should contain 5 spells, consisting of 4 casts of <DataLink action="HOLY_SPIRIT" /> or <DataLink action="HOLY_CIRCLE" /> and ending with a cast of <DataLink action="CONFITEOR" />.</Trans>
```

(Don't forget to `yarn extract` after this change)

AzariahAeroborn (changes requested):

src/parser/jobs/pld/modules/Requiescat.tsx
Dynme: So, atm the Blades combo is being tracked by the core Combo module. If the PLD uses Confiteor and doesn't make it all the way to Valor, it flags as a broken combo. Would that need to be moved into this module to count as supported, or would that be post-supported improvement?

AzariahAeroborn: I guess my thinking was that the Confiteor combo was strong enough to justify having its own separate suggestion.  But if you're happy with just the regular combo tracking for it then I'm fine to go without separate tracking / suggestion.

Dynme: Yeah, I'm not opposed to making it its own suggestion, but since the information is already being presented if the player makes a mistake, I thought it better to work on other modules and cycle back around to this when the other bases are covered.

AzariahAeroborn (approved):

PR #1446, opened 2021-12-20
Update to Tinctures and small bug fix in Action data

Dynme:
Update Tincture module to track Expiacion instead of Spirits Within.
Update Shield Lob to not break combo.


Dynme:
This probably needs to be merged after #1445 since they both touch the ACTIONS file, and this one does so for actual changes vs just trying to pass the lint test.

AzariahAeroborn (dismissed):

AkairyuGestalter (approved):

PR #1448, opened 2021-12-20
Update No Mercy Window to Prioritise Double Down

jonnixs:
* Updated No Mercy window to prioritise Double Down (DD) over Blast Strike (BS) + Hypervelocity (HV). 
   - DD has a single target potency of 1200 (and cumulatively more over multiple targets) for 2 carts over two BS / HV combos which totals 1120.
   - DD is also a single GCD compared to BS 2 GCDs, meaning additional potency gained for that extra action.

Discuss.

### DIFF
--- a/src/parser/jobs/gnb/index.tsx
+++ b/src/parser/jobs/gnb/index.tsx
@@ -72,5 +72,14 @@ export const GUNBREAKER = new Meta({
 			</>,
 			contributors: [CONTRIBUTORS.EDEN],
 		},
+		{
+			date: new Date('2021-12-28'),
+			Changes: () => <>
+				<ul>
+					<li>Add Double Down to No Mercy window, and update Blast Shot GCD counts</li>
+				</ul>,
+			</>,
+			contributors: [CONTRIBUTORS.JONNIX],
+		},
 	],
 })
--- a/src/parser/jobs/gnb/modules/NoMercy.tsx
+++ b/src/parser/jobs/gnb/modules/NoMercy.tsx
@@ -24,6 +24,7 @@ const SEVERITIES = {
 
 const EXPECTED_USES = {
 	DOUBLE_DOWN: 1, //This is assuming that you enter NM with 3 carts: 1 cart: Gnashing, 2 carts: Double Down
+	BURST_STRIKE: 1,
 	GNASHING_FANG: 1,
 	SONIC_BREAK: 1,
 	ROUGH_DIVIDE: 1,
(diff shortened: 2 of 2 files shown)


Rjb99:
I don't think a one size fits all for No Mercy is a good idea considering GNB now has 3 different windows for it.

jonnixs:
What do you mean 3 different windows for it? As far as I'm aware in all windows (maybe minus the opener) you would still want to fit 9 GCDs in, you still want a Gnashing combo, Sonic Break, Blasting Zone, Bow Shock and 1 / 2 Rough Divides as before. The only difference I see is you want Double Down rather than Blast Strike, which is also on a 60 cooldown same as No Mercy.

Unsure how this is more or less one size fits all than any recommendation or how it worked previously.

Please do correct me if I'm wrong.

Rjb99:
image

Basically Bloodfest NM knocks out The first 3rd of a gnashing fang combo. Except on the opener which is a different one time only (probably) window for No Mercy

The other 2 non-Bloodfest windows are basically just hyper velocity or not though, so this approach would be fine for those.

jonnixs:
To be honest, I take that image with a large grain of salt. The opener for example is not one I expect to stay like that and if that changes, chances are so do the NM windows. If we were to assume that that image is the only way to use NM, that's trying to fit into a one size fits all. 

If we look at the analysis of some of the top DPS GNB on fflogs on the new extremes as examples, they for the most part still have full GF combos in all their NM windows and the one of two that don't seem to look like it was potentially a mistake or mechanic timing related.

Rjb99:
Looked over the data from the top 20 pulls from the normals, the trend does seem to be still shoving GF under every NM. 

Given it's just a 20 potency gain and is subject to fight timing, a general purpose one would work fine, I'm just overthinking things again.

Our General case though assuming 3 ammo to start and no combo started would have:
- 3 GF skills
- sonic break,
- double down
- 3 regular GCDs
-  1 burst strike (but no hyperVelocity in buffs)

Our general Bloodfest case though assuming 3 ammo to start and no combo started with Bloodfest off cooldown would have:
- 3 GF skills
- sonic break,
- double down
- 1-2 regular GCDs
- 3-4 burst strikes

I'm on mobile and have no idea how to leave a proper review,

But just re-add a burst strike to expected uses and adjust the Bloodfest evaluator to add 2 to the expected burst strikes so it expects 3 burst strikes and this should be good to merge in.

AzariahAeroborn (approved):
Approving per discussion and @Rjb99's verbal approval

PR #1450, opened 2021-12-21
[SAM] SAM 6.0 Data

Rjb99:
updated Data for SAM actions and statuses.
Removal of code for lost mechanics


AkairyuGestalter (dismissed):
First pass the data changes look ok, I'll give it a more thorough review when I'm on a PC instead of a phone. Since you're touching job module code though you should have a changelog entry

thatcsguy (commented):
I see that none of the potencies were touched at all, and it seems they aren't actually used anywhere in code. Can we just remove them entirely? If not, I'd argue they should be updated, but based on a question I asked in Discord it seems like removing them is reasonable.

Rjb99:
Potencies are a holdover I've never used nor did I have plans to use them.... though now that you mention monk using them I could do something similar for kaiten to show lost "free" potency from Higanbana and Midare instead of the current "X skills missed" from 5.0

I'll get a change log for fixing the status buffs once I get home from the holidays.

ayyaruq:
> though now that you mention monk using them

FWIW we only have potency on True and Twin to show the loss when calculating bad uses. I removed it from Boot because it's 2 separate Bootshine potencies that are relevant so the field wasn't of clear value. I've stripped potencies from the rest tho.

ayyaruq (commented):
Mostly looks good to me, I'd prob cleanup the comments a bit but - are you sure you wanna be using `hr1` icons? Extra load and the way they're displayed won't need the extra resolution anyway.

Might also be worth naming naming things for what they do rather than the skill, although that's mostly so it's easier to review, for example `Fuka.ts` as `SpeedBuff.ts` or something. NBD tho, just an observation since Fuko/Fuka/Fuga is confusing so without knowing SAM it's a bit iffy to read.

ayyaruq (approved):
Code looks good, might wanna double check those offsets at some point tho.

PR #1452, opened 2021-12-21
BLM - Sharpcast and Triplecast overwrite suggestions

AkairyuGestalter:
Added suggestions against overwriting Triplecast and Sharpcast while the statuses are still active. 2 charge problems


src/parser/jobs/blm/modules/Sharpcast.tsx
```diff
@@ -158,5 +162,22 @@ export class Sharpcast extends Analyser {
...
+		this.suggestions.add(new TieredSuggestion({
+			icon: this.data.actions.SHARPCAST.icon,
+			content: <Trans id="blm.sharpcast.suggestions.overwrote-sharpcasts.content">
+				You lost at least one guaranteed <DataLink status="THUNDERCLOUD" /> or <DataLink status="FIRESTARTER" /> proc by using <DataLink action="SHARPCAST" /> while the status was already active.
+			</Trans>,
+			tiers: {
+				1: SEVERITY.MINOR,
+				3: SEVERITY.MEDIUM,
+				5: SEVERITY.MAJOR,
+			},
```
AzariahAeroborn, line 176: These tiers seem fairly lenient for an active mistake (pressing the button too many times / too soon).  Is the potency gain from this skill small enough to "allow" wasting 2 charges before we complain by default?

src/parser/jobs/blm/modules/Triplecast.tsx
```diff
@@ -0,0 +1,50 @@
...
+		this.suggestions.add(new TieredSuggestion({
+			icon: this.data.actions.SHARPCAST.icon,
+			content: <Trans id="blm.triplecast.suggestions.overwrote-triplecasts.content">
+				You lost at least one instant cast spell by using <DataLink action="TRIPLECAST" /> while the status was already active.
+			</Trans>,
+			tiers: {
+				1: SEVERITY.MINOR,
+				3: SEVERITY.MEDIUM,
+				5: SEVERITY.MAJOR,
+			},
```
AzariahAeroborn: Same as comment on Sharpcast.

src/parser/jobs/blm/modules/Sharpcast.tsx
AkairyuGestalter: The tiers were cloned from the dropped Sharpcast version (probably should've made a const for it...). They're fairly lenient since dropping/overwriting Sharpcast might not have any effect, since you could wind up with the Firestarter/Thundercloud that Sharpcast was going to force anyways due to luck.

src/parser/jobs/blm/modules/Triplecast.tsx
AkairyuGestalter: This one's probably worth making tighter, since Triplecast is more of a DPS increase due to Fire 4's longer-than-the-GCD cast time, and the loss of mobility. Dropped the Medium/Major tiers down by 1 so it's a 1/2/4 tier.

AzariahAeroborn (approved):

PR #1454, opened 2021-12-22
RDM - Allowing proc losses in opener

thatcsguy:
- The current opener for RDM makes it very likely that you will overwrite a proc during the opener.
- This change will simply prevent the "delay" message from showing on any combo that starts in the first 15 seconds of a pull.

Before:
image

After:
image


Cephid:
From a RDM standpoint this looks good as all suggested openers discard procs currently.

AkairyuGestalter (approved):

PR #1455, opened 2021-12-22
RDM - Band-aid fix for Manafication usage mid-combo

thatcsguy:
Manafication states that it breaks all combos. However, what we currently model as the melee combo in code is actually two separate combos (the melee part, and the ranged part). We should eventually model this correctly (break up the combos, model mana stacks on the gauge, figure out how to merge all this into the UI), but the change will take time. Manafiction is safe to hit mid-combo as long as it is immediately before Verholy/Verflare, and this is quickly becoming a dominant strategy on top parses in order to fit multiple combos under buffs/pots. This is a band-aid fix to render the combos correctly until we can get around to the bigger rework of this module.

Timeline:
image

Before:
image

After:
image


Cephid:
This is a good stop gap fix; once all the PRs you have get merged in my intent is to start looking into getting the combos broken apart and adding a new enchanted gems gauge.

AkairyuGestalter (changes requested):
One more "needs changelog" comment for now 🙂

AkairyuGestalter (approved):

PR #1457, opened 2021-12-23
Update WHM ThinAir module for Endwalker

Dot-Inni:
Changed tracking to usage-based instead of time-window based, as Thin Air uses charges now and drops after the first spell. Also added some minor recommendations based on MP saved per Thin Air charge usage.


src/parser/jobs/whm/modules/Thinair.tsx
```diff
@@ -18,6 +18,12 @@ interface ThinAirRecord {
 	mpsaved: number,
 }
 
+const USAGE = {
+	BAD: 'red',
+	OKAY: 'yellow',
+	GOOD: 'green',
```
nonowazu: I would change these to some hex colors. I'm not sure what our standard colors are, but ackwell can probably tell you where to look.

src/parser/jobs/whm/modules/Thinair.tsx
```diff
@@ -106,17 +122,15 @@ export class Thinair extends Analyser {
...
 
 		const panels = this.history.map(record => {
-			const actions = record.casts.map(action => getDataBy(ACTIONS, 'id', action)).filter(x => !!x)
-			const gcds = actions.filter(action => action && action.onGcd)
-			const numGcds = gcds.length
-
 			return {
 				key: record.start,
 				title: {
 					content: <Fragment>
```
nonowazu: Since the "root" element is in a `<span>` tag, consider removing the `<Fragment>` tag for this.

nonowazu (changes requested):
I think this is mostly a good change, except for 2 minor considerations.

src/parser/jobs/whm/modules/Thinair.tsx
ayyaruq: Checklist uses the named colours so this is likely fine.

ayyaruq (changes requested):
Not sure about that span/fragment thing (I just use `<>thing</>` usually), but this module also includes a lot of old syntax that we no longer use:

- Are you able to move uses of `ACTIONS` and `STATUSES` to `this.data.actions` and `this.data.statuses`?
- We generally use `this.data.getStatus/getAction` over `getDataBy` now.
- In `ThinAirRecord`, `casts` should be of type `Array<Action['id']>` and may require importing `Action` from `data/ACTIONS`.

src/parser/jobs/whm/modules/Thinair.tsx
Dot-Inni: Tag has been removed.

Dot-Inni:
Did some minor cleanup, will try to get the old syntax fixed now.

ayyaruq (approved):
Noice. Could use some cleanup later for consistency with other modules, performance, and better null/state tracking tho.

PR #1458, opened 2021-12-24
RDM - Adding number of players buffed to Embolden module

thatcsguy:
- Logic was a bit too complicated to add to the base BuffWindow module since the self-buff is different from the party-buff.
- No additional Suggestions with this change, though it's obviously something we could consider.

Before:
image

After:
image


src/parser/jobs/rdm/modules/Embolden.tsx
```diff
@@ -0,0 +1,155 @@
...
+		// Count how many players we hit with the buff.
+		if (!lastWindow.playersBuffed.includes(event.target)) {
+			lastWindow.playersBuffed.push(event.target)
+		}
+	}
+
+	// Get the existing open buff window, or open a new one
+	private tryOpenWindow(event: Events['statusApply']): EmboldenWindow {
+		const lastWindow: EmboldenWindow | undefined = _.last(this.history)
+		if (lastWindow && !lastWindow.end) {
```
AkairyuGestalter: We're trying to switch to `!= null` as our standard pattern for figuring out if something is defined or no. Can you change this to be `if (lastWindow != null && lastWindow.end == null)` for this condition (and similar for the other places like it?

AkairyuGestalter (changes requested):
One coding style comment, and you'll need a changelog entry as well

AkairyuGestalter (approved):

PR #1459, opened 2021-12-24
DNC - Bugfix for bogus Technical Finish windows

thatcsguy:
It seems there were some non-player actors that were throwing off the logic and opening new Technical Finish windows when they shouldn't. You can see the issue on this parse (there is only 1 dancer): https://xivanalysis.com/fflogs/acFQK8Lm7jwJAVtN/1/4

This change is simply filtering out non-player-controlled actors when counting buffs.

Before:
image
image

After:
(no warning)
image


AkairyuGestalter (changes requested):
Fix looks ok. Does need a changelog though!

AkairyuGestalter (approved):

PR #1460, opened 2021-12-24
BLM - Bugfix - Paradox grants a stack of the active stance

AkairyuGestalter:
Minor fix. I observed in a dungeon run that Paradox's AF/UI timer refresh is actually done by granting a stack of the active timer, exactly as if it were the F1/B1 cast it takes the place of.


Toastdeib (approved):
Not super familiar with how new BLM works now but this looks fine, considering that the `onGain*Stacks` functions are (presumably) tried and true. I'm happy to push the button on this one once the conflicts are resolved.

PR #1463, opened 2021-12-25
PLD Update to FoF, Atonement, and Requiescat

Dynme:
Updated Flight or Fight module to require 3x Atonement casts and filer out Blade spells.
Updated Atonement module to allow dropping 1 cast per minute for buff alignment.
Updated Requiescat module to only display relevant abilities.
Updated ActionTimeline module to reflect upgraded abilities.


src/parser/jobs/pld/modules/Requiescat.tsx
```diff
@@ -79,6 +114,7 @@ export class Requiescat extends BuffWindow {
 
 	override initialise() {
 		super.initialise()
+		this.ignoreActions(EXCLUDED_ACTIONS.map(g => this.data.actions[g].id))
```
AzariahAeroborn: Consider using this.trackOnlyActions instead, to limit the skills displayed in the rotation list to a (much shorter) whitelist instead of maintaining a large blacklist.

If that whitelist winds up being only Holy [Spirit|Circle] + Requiescat, you can also change the evaluator to be a native ExpectedGcdCountEvaluator and remove the custom RequiescatGcdsEvaluator.

AzariahAeroborn (changes requested):

src/parser/jobs/pld/modules/Atonement.tsx
```diff
@@ -10,8 +10,8 @@ import React from 'react'
 
 // In seconds
 const USE_SEVERITY = {
-	95: TARGET.WARN,
-	100: TARGET.SUCCESS,
+	79: TARGET.WARN,
+	83: TARGET.SUCCESS,
```
ayyaruq, line 14: Any explanation for this change or?

ayyaruq (commented):
Smol q but otherwise nice, what Az said tho re Requiescat.

src/parser/jobs/pld/modules/Atonement.tsx
Dynme: PLD can now drop 1 Atonement out of the 6 per minute that we generate in order to do the 61s rotation and keep buffs aligned. So 83% would now be the old 100%, and 79% the old 95%.

AzariahAeroborn (approved):

PR #1464, opened 2021-12-26
[MNK] Initial changes to Action+Status and Endwalker updates

Malpp:
Remove Shoulder Tackle since it doesn't damage any more
Added Riddle of Wind to cooldowns
Removed all Fists since they don't exist anymore
Removed GreasedLighting since it becomes a passive
Updated Twin Snakes buff to Disciplined Fist
Disciplined Fist is now 15% increase up from 10% in translations
Updated potencies of abilities
Updated the cooldowns of the cooldowns
Updated TwinSnakes to track Disciplined Fist cooldown and removed suggestion for Four Point Fury since it applies the buff on use

Before (broken):
image
After:
image

Before:
image
After:
image

I didn't touch `Tornado Kick` since it changes into `Phanton Rush` at 90... unsure if we want to support lower levels or not, or if that's even possible.


src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -176,28 +167,14 @@ export class RiddleOfFire extends Analyser {
 		this.suggestions.add(new TieredSuggestion({
 			icon: this.data.actions.ELIXIR_FIELD.icon,
 			content: <Trans id="mnk.rof.suggestions.ogcd.content">
-				Aim to use <DataLink action="TORNADO_KICK"/>, <DataLink action="ELIXIR_FIELD"/>, and at least 1 <DataLink action="SHOULDER_TACKLE"/> during each <DataLink status="RIDDLE_OF_FIRE"/>.
+				Aim to use <DataLink action="TORNADO_KICK"/>, <DataLink action="ELIXIR_FIELD"/>, and at least 1 <DataLink action="THUNDERCLAP"/> during each <DataLink status="RIDDLE_OF_FIRE"/>.
```
AzariahAeroborn, line 170: This copy expects the user to use Thunderclap in each Riddle of Fire window, but we are not tracking it anywhere or displaying how many Thunderclaps they are actually using.  We should add Thunderclap tracking to replace the removed Shoulder Tackle information.

AzariahAeroborn (changes requested):

Please also see my comment regarding the Riddle of Fire update & Thunderclap.

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
Rjb99: We should just not? Thunderclap is no damage teleport skill.

AzariahAeroborn: Ah, yeah, then the copy should be fixed to not specify using Thunderclap

PR #1465, opened 2021-12-27
BRD 6.0 - Adding evaluator for RF/BV usage in RS window

jplippi:


src/parser/jobs/brd/modules/RagingStrikes.tsx
```diff
@@ -1,25 +1,37 @@
 import {t} from '@lingui/macro'
-import {Trans} from '@lingui/react'
-import {ActionLink, StatusLink} from 'components/ui/DbLink'
+import {Plural, Trans} from '@lingui/react'
+import {ActionLink, DataLink, StatusLink} from 'components/ui/DbLink'
```
hintxiv: Since you're using `DataLink` here we could just switch over the existing Links to use it instead of importing all three.

src/parser/jobs/brd/modules/RagingStrikes.tsx
```diff
@@ -218,4 +309,27 @@ export class RagingStrikes extends BuffWindow {
...
+		const buffAction = _.first(window.data.filter(it => it.action.id === this.data.actions[action].id))
+
+		if (buffAction) {
+			const gcdThreshold = _.last(
+				window.data
+					.filter(it => it.action.onGcd)
+					.slice(0, bestBeforeGCD)
+			)
+
+			if (gcdThreshold) {
```
hintxiv: See the comment below - explicit undefined check.

src/parser/jobs/brd/modules/RagingStrikes.tsx
```diff
@@ -218,4 +309,27 @@ export class RagingStrikes extends BuffWindow {
 		const firstGcd = gcdTimestamps[0]
 		return this.barrageRemoves.some(timestamp => firstGcd <= timestamp && timestamp <= (window.end ?? window.start))
 	}
+
+	private wasActionUsedInTime = (action: ActionKey, bestBeforeGCD: number, window: HistoryEntry<EvaluatedAction[]>): boolean => {
+		const buffAction = _.first(window.data.filter(it => it.action.id === this.data.actions[action].id))
+
+		if (buffAction) {
```
hintxiv: We've shifted to using explicit checks for undefined values recently rather than relying on their falsiness. Something like this:

```suggestion
		if (buffAction != null) {
```

src/parser/jobs/brd/modules/RagingStrikes.tsx
```diff
@@ -218,4 +309,27 @@ export class RagingStrikes extends BuffWindow {
...
+				// Used in time if it was used before specified GCD
+				return buffAction.timestamp < gcdThreshold.timestamp
+			}
+
+			// There's no limiting GCD in the window, so can only assume it was used in time
+			return true
+		}
+
+		// Buff was not used, thus it was not in time
+		return false
```
hintxiv: This is very nitpicky but to avoid the nesting you can invert the `buffAction` check to bail early instead of returning false at the very end. Something like this: 

```ts
if (buffAction == null) { return false }

const gcdThreshold = ...
```

hintxiv (changes requested):

hintxiv (approved):
lgtm

src/parser/jobs/brd/modules/RagingStrikes.tsx
```diff
@@ -1,25 +1,37 @@
...
 import {dependency} from 'parser/core/Injectable'
-import {BuffWindow, EvaluatedAction, ExpectedActionsEvaluator, ExpectedGcdCountEvaluator, TrackedAction, TrackedActionsOptions} from 'parser/core/modules/ActionWindow'
+import {
+	BuffWindow,
+	EvaluatedAction,
+	EvaluationOutput,
+	ExpectedActionsEvaluator,
+	ExpectedGcdCountEvaluator,
+	TrackedAction,
+	TrackedActionsOptions, WindowEvaluator,
```
AzariahAeroborn: With apologies for the nit, this should be separated onto two lines to be consistent with the rest of the import.

src/parser/jobs/brd/modules/RagingStrikes.tsx
```diff
@@ -63,6 +76,73 @@ class BarrageEvaluator extends ExpectedActionsEvaluator {
...
+	actions: ActionKey[]
+	predicate: (action: ActionKey, bestBeforeGCD: number, window: HistoryEntry<EvaluatedAction[]>) => boolean
+
+	suggestionIcon: string
+	severityTiers: SeverityTiers
+}
+
+class BuffEvaluator implements WindowEvaluator, BuffEvaluatorOptions {
+	actions: ActionKey[]
+	predicate: (action: ActionKey, bestBeforeGCD: number, window: HistoryEntry<EvaluatedAction[]>) => boolean
```
AzariahAeroborn: I think this is a good implementation, and will have value being generalized to core so Dancer (Tech Step/Devilment) and other jobs can use it.  To that end, I would like to see the predicate function be moved into the evaluator as a private logic function so that it can be easily shared.  

If you change actions to be an `Array<Action['id']>` type (requiring implementers to pass in the IDs of the actions being considered, rather than their ActionKey), then you don't wind up with a hidden dependency on Data and the `wasActionUsedInTime` function can be a function of the Evaluator directly instead of having to pass it in from the BuffWindow analyser.

AzariahAeroborn (changes requested):

AzariahAeroborn (approved):

PR #1466, opened 2021-12-27
DRG: Rewrite Battle Litany

ebshimizu:
Ported Battle Litany to Core BuffWindow.

An adjustment was also made to BuffWindow to only track windows based on the status applied to the parser actor and/or enemies (trick attack says hi). This eliminates buff mirroring issues for buffs that apply to the entire party (and their pets).

This module has two custom evaluators, both purely informational.
- Players Buffed: Displays the number of players affected by Battle Litany in each window
- "Double DRG": Displays a note if a window was overwritten by a different DRG using battle litany. This is mostly around to preempt questions about "why does this window look so short."


src/parser/jobs/drg/modules/BattleLitany.tsx
```diff
@@ -46,209 +97,123 @@ class BLWindow {
 // the character selected for analysis. windows that clip into or overwrite other
 // DRG litanies will be marked.
 // Used DNC technical step as basis for this module.
-export default class BattleLitany extends Analyser {
+export class BattleLitany extends BuffWindow {
```
AzariahAeroborn, line 97: Comment on this class needs cleaned up - legacy comment information from before the BuffWindow conversion.  (I tried to highlight the comment but GitHub was mean because nothing in the comment lines changed)

AzariahAeroborn (changes requested):

src/parser/jobs/drg/modules/BattleLitany.tsx
ebshimizu: ah yeah, thanks for the heads up, have updated that comment

AzariahAeroborn (approved):

PR #1467, opened 2021-12-28
Scope report store to legacy report source

ackwell:
Title. Was gonna do the new event types but got caught up doing this and it got big enough to be its own PR.

This moves the last of the stuff-that-touches-fflogs-directly into the legacy fflogs report source, effectively completing the abstraction of fflogs from xiva. Yay. With this, additional report sources can be added without any concern about fflogs. Unless you're adding a new fflogs report source. You'll need to think about fflogs if you're doing that.

This is just refactors, it should have no functional impact on the site.


ayyaruq (dismissed):
ErrorBoundary removed `core.find.refresh` from i18n, might wanna `yarn extract`. Everything else looks good tho, surprised I could follow it.

src/data/JOBS.ts
```diff
@@ -66,18 +65,17 @@ export type RoleKey = keyof typeof ROLES
 
 export interface Job {
 	name: MessageDescriptor
-	logType: ActorType
+	// logType: ActorType
```
ackwell: nts; delete

PR #1476, opened 2021-12-29
SGE - Refine Overheal buckets and bugfix for DoTs

AkairyuGestalter:
- SGE DoTs was getting the clipping amount for Dia instead of Eukrasian Dosis III. Copy-paste error indeed.
- Refine overheal categorization buckets based on feedback. Changing to condense the Addersgall actions (including the Kerachole regen) into one, and Holos/Physis into another. Leaving the Haima/Panhaima expiry bucket alone.


src/parser/jobs/sge/modules/Overheal.tsx
```diff
@@ -12,22 +12,22 @@ export class Overheal extends CoreOverheal {
...
 		},
 		{
 			name: <Trans id="sge.overheal.cooldowns.name">Cooldowns</Trans>,
 			color: SuggestedColors[2],
 			trackedHealIds: [
-				this.data.actions.DRUOCHOLE.id,
-				this.data.actions.IXOCHOLE.id,
-				this.data.actions.TAUROCHOLE.id,
 				this.data.actions.HOLOS.id,
+				this.data.statuses.PHYSIS.id,
```
Toastdeib, line 29: Probably don't need to include Physis here, since it upgrades directly to Physis II and shouldn't matter at level 60+. Looks fine though, and this doesn't hurt anything, so I'm happy to merge it if you want to leave this as-is.

Toastdeib (approved):

src/parser/jobs/sge/modules/Overheal.tsx
AkairyuGestalter: Let's just leave it for now. I originally added it for completeness since I wasn't paying attention to what level things came at. Might drop it later since it won't be relevant in any Ultimate, but it's fine for now.

PR #1477, opened 2021-12-30
Rdm combo split

Cephid:
Added a new Mana Stack Gauge for the combo resource for RDMs.  Modified the Combo such that the Melee and Finisher combo are separate but are handled similar as before so the display and finisher calculation remains unchanged for the time being.  Temporarily removed the suggestion on combo delays as it's currently causing confusion, we're not yet certain if it goes away completely or not.  Finally cleaned up MeleeCombos by getting rid of all the _ characters on property names.


src/parser/jobs/rdm/modules/ManaStackGauge.tsx
```diff
@@ -0,0 +1,214 @@
...
+	private manaStackGauge = this.add(new CounterGauge({
+		graph: {
+			label: <Trans id="rdm.manaStackGauge.resource.manaStacks">Mana Stacks</Trans>,
+			color: JOBS.RED_MAGE.colour,
+		},
+		maximum: MAXIMUM,
+		minimum: MINIMUM,
+	}))
+	public gaugeModifiers = new Map<number, ManaStackGaugeModifier>([
+		[this.data.actions.ENCHANTED_REPRISE.id, {manaStack: 1}],
```
thatcsguy: reprise doesn't give stacks

src/parser/jobs/rdm/modules/ManaStackGauge.tsx
```diff
@@ -0,0 +1,214 @@
...
+		[this.data.actions.ENCHANTED_RIPOSTE.id, {manaStack: 1}],
+		[this.data.actions.ENCHANTED_ZWERCHHAU.id, {manaStack: 1}],
+		[this.data.actions.ENCHANTED_REDOUBLEMENT.id, {manaStack: 1}],
+	])
+	public spenderModifiers = new Map<number, ManaStackGaugeModifier>([
+		[this.data.actions.VERHOLY.id, {manaStack: -1}],
+		[this.data.actions.VERFLARE.id, {manaStack: -1}],
+		[this.data.actions.SCORCH.id, {manaStack: -1}],
+		[this.data.actions.RESOLUTION.id, {manaStack: -1}],
+	])
```
thatcsguy: Verholy and Verflare just spend all 3.  Scorch and Resolution are just only available as combos after that but they dont technically touch mana stacks.

src/parser/jobs/rdm/modules/ManaStackGauge.tsx
```diff
@@ -0,0 +1,214 @@
...
+		1: SEVERITY.MAJOR,
+	}
+	manaStatistics = {
+		white: {
+			overcapLoss: 0,
+		},
+		black: {
+			overcapLoss: 0,
+		},
+	}
```
thatcsguy, line 80: I assume this is a carry-over from the other gauge file?

src/parser/jobs/rdm/modules/ManaStackGauge.tsx
```diff
@@ -0,0 +1,214 @@
...
+			this.onGaugeSpender
+		)
+		this.addEventHook('complete', this.onComplete)
+	}
+
+	private onGaugeModifying(event: Events['action']) {
+		const modifier = this.gaugeModifiers.get(event.action)
+
+		//Always check it we should break
+		if (this.onGaugeBreak(event)) {
```
thatcsguy: naming nitpick, but `onGauageBreak` sounds like it should only be called when the break happens, but you're calling it every time. Maybe `handleBreaks`?

src/parser/jobs/rdm/modules/ManaStackGauge.tsx
```diff
@@ -0,0 +1,214 @@
...
+			//We won't modify the gauge if this skill broke it.
```
thatcsguy: This comment is a bit misleading. You are obviously resetting the gauge, it's just happening as part of the `if` statement.

src/parser/jobs/rdm/modules/ManaStackGauge.tsx
```diff
@@ -0,0 +1,214 @@
...
+			return
+		}
+
+		if (modifier == null) {
+			return
+		}
+
+		if (this.manaStackGauge.capped) {
```
thatcsguy, line 123: I feel like this whole block should be removed. We aren't actually losing white/black mana by hitting an extra mana gauge builder.

src/parser/jobs/rdm/modules/ManaStackGauge.tsx
Cephid: No this is meant to track when you overcap your mana stack that mana you wasted in overcapping as that mana is lost, you gained nothing for it beyond not losing the chance to do the finisher combo.

src/parser/jobs/rdm/modules/ManaStackGauge.tsx
Cephid: I'd disagree with that reasoning, as it's also called by the Spender, but only if it needs to.  When we have Mana Stacks the vast majority of our abilities break the mana stack gauge, beyond that it's asking if there was a break on the gauge so it knows to return or proceed.

src/parser/jobs/rdm/modules/ManaStackGauge.tsx
Cephid: How is it not lost if I spent 70|70 for a finisher that should only cost 50|50?  That 20|20 isn't going to apply to my next Mana Stack Build, it's not mana I'm gonna get back, it's a total loss of 20|20 mana.

src/parser/jobs/rdm/modules/ManaStackGauge.tsx
Cephid: will rephrase it.

src/parser/jobs/rdm/modules/ManaStackGauge.tsx
Cephid: Thinking more on this one; I decided to break onGaugeBreak into onGaugeBreak and shouldGaugeBreak.  Changed the callers to ask if the gauge should break and if yes, then they'll call onGaugeBreak.

src/parser/jobs/rdm/modules/ManaStackGauge.tsx
thatcsguy: I suppose I understand. "Lost" is a strong word I guess, its not like you didn't do ANY damage with the mana (which is the case for Using Manafication over 50/50 e.g.). I'm fine with it and The Balance folks can decide if they like it or not.

src/parser/jobs/rdm/modules/ManaStackGauge.tsx
thatcsguy: Got it. Thanks.

src/parser/jobs/rdm/modules/ManaStackGauge.tsx
```diff
@@ -0,0 +1,214 @@
...
+			this.onGaugeSpender
+		)
+		this.addEventHook('complete', this.onComplete)
+	}
+
+	private onGaugeModifying(event: Events['action']) {
+		const modifier = this.gaugeModifiers.get(event.action)
+
+		//Always check it we should break
+		if (this.shouldGaugeBreak(event)) {
```
AzariahAeroborn: Should run the shouldGaugeBreak check first before we try to access the gaugeModifiers dictionary, as breakers will never exist in the Modifiers dictionary (and we don't need the modifier if we have a breaker since we're going to log the break and return).

src/parser/jobs/rdm/modules/ManaStackGauge.tsx
AzariahAeroborn: I do kind of agree that this function is confusing, and the first debug line's copy doesn't help the confusion either.  

Plus, this isn't going to generate the correct amount of overcap if the user does Moulinet > Riposte > Zwerchau > Redoublement - this is going to look up the 15|15 of Redoublement and add that to overcap, even though the user spent 70|70 (same scenario applies if user starts with 2x Riposte).

It seems to me that if we want to track overspend loss in the White/Black gauge, we're going to need the Mana Stack gauge to track state of each "group" of builders, and then evaluate those builders against what was necessary in order to get a Verholflare combo.

src/parser/jobs/rdm/modules/ManaStackGauge.tsx
```diff
@@ -0,0 +1,214 @@
...
+		this.manaStackGauge.modify(modifier.manaStack)
+	}
+
+	private onGaugeSpender(event: Events['action']) {
+		const modifier = this.spenderModifiers.get(event.action)
+
+		if (modifier == null) {
+			if (this.shouldGaugeBreak(event)) {
+				this.onGaugeBreak(event)
+			}
```
AzariahAeroborn: This condition can never return true, since breakers only call into the onGaugeModifier function.

src/parser/jobs/rdm/modules/ManaStackGauge.tsx
```diff
@@ -0,0 +1,214 @@
...
+				<Trans id="rdm.manaStackGauge.suggestions.manaStack-loss-content">Ensure that you don't use a non Enchanted GCD when you have ManaStacks, otherwise you lose them all</Trans>
+			</Fragment>,
+			tiers:  this.severity,
+			value:  manaStacksLost,
+			why: <Fragment>
+				<Trans id="rdm.manaStackGauge.suggestions.manaStack-loss-why">You lost { manaStacksLost} ManaStacks due to using a non Enchanted GCD</Trans>
+			</Fragment>,
+		}))
+
+		this.statistics.add(new DualStatistic({
```
AzariahAeroborn, line 187: Should we suppress this statistic if there was no overcap loss (and thus no suggestion being generated?  Alternatively, can we roll this statistic into the Why of the overcap suggestion?

src/parser/jobs/rdm/modules/ManaStackGauge.tsx
```diff
@@ -0,0 +1,214 @@
+import {t} from '@lingui/macro'
+import {Trans} from '@lingui/react'
+import {JOBS} from 'data/JOBS'
+import {Event, Events} from 'event'
+import {filter, oneOf} from 'parser/core/filter'
+import {dependency} from 'parser/core/Injectable'
+import {CounterGauge, Gauge as CoreGauge} from 'parser/core/modules/Gauge'
+import {Statistics} from 'parser/core/modules/Statistics'
+import Suggestions, {TieredSuggestion, SEVERITY} from 'parser/core/modules/Suggestions'
+import {Gauge} from 'parser/jobs/rdm/modules/Gauge'
```
AzariahAeroborn: I would prefer to see the Black/White Gauge module renamed appropriately to help further disambiguate it from the Mana Stack Gauge.

AzariahAeroborn (changes requested):

src/parser/jobs/rdm/modules/ManaStackGauge.tsx
Cephid: In the case of the example you gave, the correct answer is 15|15 lost.  The fact that potency was lost due to the bad management of enchanted skills is immaterial.  Any enchanted skill will generate a Mana Stack, no matter the order utilized except for Reprise so long as no other GCD skill was utilized between them.  This metric is only concerned with specifying the mana that was purely wasted on a mana stack over the 3 cap.  If I truly wanted to, I could do 3x Redoublement to go into the finisher combo, it would just be a horrible potency choice.

As for that debug line, that was from early in development and I just forgot to delete it once I organized the code better is all.

src/parser/jobs/rdm/modules/ManaStackGauge.tsx
Cephid: I'll rename it ManaGuage, and all dependency declarations of gauge to manaGauge!

src/parser/jobs/rdm/modules/ManaStackGauge.tsx
Cephid: We don't currently suppress the manafication version, and I haven't been asked to, so I treated this one the same; since if I suppress it here I should suppress it there.  We're currently not even sure if this particular metric will stay, but was deemed potentially useful to see for now.  

That said I don't really have an opinion on whether it should or shouldn't; would you like this one suppressed for now if it has a zero value pair, until I can get additional feedback?

AzariahAeroborn: I guess I don't have a strong opinion on it.  One thing to consider for the statistic as a whole is how bloated does it make the page on mobile to have this statistic (especially in the no waste case, but even if there is data here, is the data valuable enough to justify the extra space this takes up)?

Cephid: Lana's response when asked was that of no strong opinion and keeping it there would be more for consistency.  The usefulness of this and of the manafication metric are currently being debated, given the change to manafication.  We're honestly just not certain how much of a problem this particular item is, since this tracking of it is brand new.

src/parser/jobs/rdm/modules/ManaStackGauge.tsx
Cephid: I'll shift them around.

src/parser/jobs/rdm/modules/ManaStackGauge.tsx
Cephid: Indeed, removed the check entirely.

AzariahAeroborn (approved):

PR #1479, opened 2021-12-30
Change the double weave GCD threshold

furstblumier:
This PR aims to change the `REDUCE_MAX_WEAVES_RECAST_BELOW` from 2000ms to 1800ms. While it is definitely possible to double weave without clipping with a 2.0s gcd, it's def impossible to do so with a 1.8s GCD. Arguably the threshold should be realistically a bit higher but you cannot go below 3x anim lock of 0.6 which is basically the standard.


ayyaruq (changes requested):
Looks good, might wanna check jobs to see if any need additional overrides tho. Requesting a change but, you need a core changelog for this.

AzariahAeroborn (approved):

ayyaruq (approved):

Issue #1372, opened 2021-12-02
Reaper Endwalker TODO


ayyaruq:
Tracking progress for RPR implementation to make is easier for people to see when asking questions in Discord. If you feel like anything is incorrect or missing here, please feel free to leave a comment, or ping acchan#5683 in the #fb-reaper channel on the xivanalysis Discord server.

### Data Implementation
- [x] Actions #1353 
- [x] Statuses #1353 

### Analyser Implementation
- [x] Soul Gauge #1623
- [x] Shroud Gauge #1623 
- [x] Death Gauge #1619 
- [x] AoE #1430 
- [x] CDDT #1430 
- [x] Combos #1430 
- [x] Interrupts #1430 
- [x] Weaving #1430 
- [x] Buff Uptime (Death's Design) #1427
- [x] Buff Window (Enshrouded, Arcane Circle) #1553
- [x] Enhanced procs for Gibbet/Gallows #1551
- [x] Plentiful Harvest prepull/downtime #1482
- [x] Immortal Sacrifice charges not being wasted #1482
- [x] Harvest Moon charging #1559 
- [x] Soulsow should be flagged if not used in downtime (this should be done *after* the cleanup PR is done)
- [x] Soul Reaver stacks wasted or misused #1625 
- [x] Filtered tincture windows #1429 

### Misc
- [x] cleanup junk code and flag support #1638

hintxiv:
* Enshroud should be left out of CDDT since it's gauge-based

PR #1482, opened 2021-12-31
[RPR] Add Immortal Sacrifice stack consumption checklist item for Reaper

kennedyoliveira:
My first contribution to this repo, might have mistakes, feedback appreciated.

Addressing one of the TODO items from this issue #1372


src/parser/jobs/rpr/modules/ImmortalSacrifice.tsx
```diff
@@ -0,0 +1,147 @@
...
+		}
+
+		window.stackCount = event.data || window.stackCount + 1
+	}
+
+	private onPlentifulHarvest(event: Events['action']) {
+		const window = this.getOpenWindow()
+
+		if (!window) {
+			this.debug(`Plentiful Harvest executed but there were no immortal sacrifice window open: ${this.windows}`)
```
AzariahAeroborn: It's probably worth triggering Broken Log for this, so that we get notifications if it's happening (and so that the log displays a warning since other things might be mis-handled due to the bad data).

You can find an example in the `onNastrondCast` function in DRG's BloodOfTheDragon module.

src/parser/jobs/rpr/modules/ImmortalSacrifice.tsx
```diff
@@ -0,0 +1,147 @@
...
+		window.closed = true
+		window.endTime = event.timestamp
+		window.consumed = true
+	}
+
+	private onImmortalSacrificeExpiration(event: Events['statusRemove']) {
+		const window = this.getOpenWindow()
+
+		// valid case, means it was consumed
+		if (!window) {
```
AzariahAeroborn: For both this and the check above in `onPlentifulHarvest`: our style requirement is to use != null for null/undefined checking, rather than relying on falsey behavior.

AzariahAeroborn (dismissed):

I added a few comments for style things, I'm also going to wait and let someone more versed in Reaper review it for job accuracy, since I'm just a filthy caster/healer :)

src/parser/jobs/rpr/modules/ImmortalSacrifice.tsx
kennedyoliveira: Thanks, added.

src/parser/jobs/rpr/modules/ImmortalSacrifice.tsx
kennedyoliveira: Reworked all places with suggested style.

src/parser/jobs/rpr/modules/ImmortalSacrifice.tsx
```diff
@@ -0,0 +1,154 @@
...
+import {Trans} from '@lingui/react'
+import React from 'react'
+import {DataLink} from '../../../../components/ui/DbLink'
+import {Event, Events} from '../../../../event'
+import {Analyser} from '../../../core/Analyser'
+import {filter} from '../../../core/filter'
+import {dependency} from '../../../core/Injectable'
+import BrokenLog from '../../../core/modules/BrokenLog'
+import Checklist, {Requirement, Rule} from '../../../core/modules/Checklist'
+import {Data} from '../../../core/modules/Data'
```
ayyaruq: You can just use normal imports for this, no need for relative paths.

src/parser/jobs/rpr/modules/ImmortalSacrifice.tsx
```diff
@@ -0,0 +1,154 @@
...
+		this.addEventHook(playerFilter.type('statusRemove').status(this.data.statuses.IMMORTAL_SACRIFICE.id), this.onImmortalSacrificeExpiration)
+		this.addEventHook(playerFilter.type('action').action(this.data.actions.PLENTIFUL_HARVEST.id), this.onPlentifulHarvest)
+
+		this.addEventHook('complete', this.onComplete)
+	}
+
+	private onImmortalSacrificeStack(event: Events['statusApply']) {
+		let window = this.getOpenWindow()
+
+		if (window === null) {
```
ayyaruq: This will never be true.
```suggestion
		if (window == null) {
```

src/parser/jobs/rpr/modules/ImmortalSacrifice.tsx
```diff
@@ -0,0 +1,154 @@
...
+interface ImmortalSacrificeWindow {
+	startTime: number,
+	endTime?: number,
+	/**
+	 * If this window was closed, by consuming the stacks, or it expired
+	 */
+	closed: boolean,
+	stackCount: number,
+	consumed: boolean,
+}
```
ayyaruq: Is this really necessary? The times don't matter, neither does stack count of a window doesn't matter (which isn't even used), it's simpler to just count stacks gained vs used. If you want to use a window in some sort of output chart with timeline links then it makes more sense but in that case stick to consistent field names with the rest of the codebase. Consumed probably shouldn't be a boolean either.

src/parser/jobs/rpr/modules/ImmortalSacrifice.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+export default class ImmortalSacrifice extends Analyser {
```
ayyaruq: Default exports are not permitted.

src/parser/jobs/rpr/modules/ImmortalSacrifice.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+	static override handle = 'immortalsacrifice'
+	static override title = t('rpr.immortal-sacrifice.title')`Immortal Sacrifice stacks consumed`
```
ayyaruq: The title is for the navigation/panel, just `Immortal Sacrifice` is fine. I'd maybe rethink the i18n module name too so it's at least consistent with the module handle.

src/parser/jobs/rpr/modules/ImmortalSacrifice.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+	private windows: ImmortalSacrificeWindow[] = []
+
+	override initialise() {
+		super.initialise()
+
+		const playerFilter = filter<Event>().source(this.parser.actor.id)
+
+		this.addEventHook(playerFilter.type('statusApply').status(this.data.statuses.IMMORTAL_SACRIFICE.id), this.onImmortalSacrificeStack)
+		this.addEventHook(playerFilter.type('statusRemove').status(this.data.statuses.IMMORTAL_SACRIFICE.id), this.onImmortalSacrificeExpiration)
```
ayyaruq: Consider renaming these callbacks. Status application is not necessarily stack gain, and removal is a separate thing to expiry.

src/parser/jobs/rpr/modules/ImmortalSacrifice.tsx
```diff
@@ -0,0 +1,154 @@
...
+				startTime: event.timestamp,
+				closed: false,
+				consumed: false,
+				stackCount: 0,
+			}
+
+			this.windows.push(window)
+		}
+
+		window.stackCount = event.data ?? window.stackCount + 1
```
ayyaruq: This is unreliable, prefer to check for the explicit case or just ignore stack count since it's not used for anything.

src/parser/jobs/rpr/modules/ImmortalSacrifice.tsx
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	private onPlentifulHarvest(event: Events['action']) {
+		const window = this.getOpenWindow()
+
+		if (window === null) {
```
ayyaruq:
```suggestion
		if (window == null) {
```

src/parser/jobs/rpr/modules/ImmortalSacrifice.tsx
```diff
@@ -0,0 +1,154 @@
...
+			this.brokenLog.trigger(this, 'PH without immortal sacrifice stack',
```
ayyaruq, line 62: I feel like there's a better way to handle this, since you can just check if they have the status anyway.

src/parser/jobs/rpr/modules/ImmortalSacrifice.tsx
```diff
@@ -0,0 +1,154 @@
...
+		window.closed = true
+		window.endTime = event.timestamp
+		window.consumed = true
+	}
+
+	private onImmortalSacrificeExpiration(event: Events['statusRemove']) {
+		const window = this.getOpenWindow()
+
+		// valid case, means it was consumed
+		if (window === null) {
```
ayyaruq:
```suggestion
		if (window == null) {
```

src/parser/jobs/rpr/modules/ImmortalSacrifice.tsx
```diff
@@ -0,0 +1,154 @@
...
+			return
+		}
+
+		this.debug('Closing window that was expired before consumption', window)
+
+		window.closed = true
+		window.consumed = false
+		window.endTime = event.timestamp
```
ayyaruq: This is doubled up in PH, probably not what you want.

src/parser/jobs/rpr/modules/ImmortalSacrifice.tsx
```diff
@@ -0,0 +1,154 @@
...
+			const w = this.windows[this.windows.length - 1]
+
+			if (!w.closed) {
+				return w
+			}
+
+		}
+
+		return null
+	}
```
ayyaruq: This should have a proper return type, but realistically this pattern is awkward and error prone. You might wanna try having a class var for the current window and just have a stopAndSave function that closes and pushes it to a history var.

src/parser/jobs/rpr/modules/ImmortalSacrifice.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+				this.debug(`Start: ${start}, End: ${end}, Closed: ${w.closed}, Consumed: ${w.consumed}, Stack Count: ${w.stackCount}`)
+			})
+		})
+
+		const totalImmortalSacrifices = this.windows.length
+
+		if (totalImmortalSacrifices <= 0) {
+			return
+		}
```
ayyaruq: Why? If they got zero uses, that's a problem.

src/parser/jobs/rpr/modules/ImmortalSacrifice.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+		const consumedImmortalSacrifices = this.windows.filter(w => w.consumed).length
+		const percentUsed = (consumedImmortalSacrifices / totalImmortalSacrifices * 100).toFixed(2)
+
+		this.checklist.add(new Rule({
+			name: <Trans id="rpr.immortal-sacrifice.checklist.title">Uses all Immortal Sacrifices stack</Trans>,
```
ayyaruq:
```suggestion
			name: <Trans id="rpr.immortal-sacrifice.checklist.title">Use all Immortal Sacrifices stacks</Trans>,
```

src/parser/jobs/rpr/modules/ImmortalSacrifice.tsx
```diff
@@ -0,0 +1,154 @@
...
+			description: <Trans id="rpr.immortal-sacrifice.checklist.description">
+				<DataLink status="IMMORTAL_SACRIFICE"/> stacks will be generated when you
+				use <DataLink action="ARCANE_CIRCLE"/>, you can then consume those stacks by
+				using <DataLink action="PLENTIFUL_HARVEST"/>, not only it's a high potency skill, it also grants enough
+				Shroud gauge for a <DataLink action="ENSHROUD"/> usage.
```
ayyaruq:
```suggestion
				<DataLink status="IMMORTAL_SACRIFICE"/> stacks will be generated when you
				use <DataLink action="ARCANE_CIRCLE"/>, you can then consume those stacks by
				using <DataLink action="PLENTIFUL_HARVEST"/>. On top of being a high potency skill, it also grants enough
				Shroud gauge to immediately use <DataLink action="ENSHROUD"/>.
```

src/parser/jobs/rpr/modules/ImmortalSacrifice.tsx
```diff
@@ -0,0 +1,154 @@
...
+			</Trans>,
+			displayOrder: DISPLAY_ORDER.IMMORTAL_SACRIFICE,
+			requirements: [
+				new Requirement({
+					name: <Trans id="rpr.immortal-sacrifice.checklist.requirement.name">
+						<DataLink status="IMMORTAL_SACRIFICE"/> stacks consumed
+					</Trans>,
+					overrideDisplay: `${consumedImmortalSacrifices} / ${totalImmortalSacrifices} (${percentUsed}%)`,
```
ayyaruq: Not sure why you'd wanna override the display for this, but if you do may be worth forcing decimal places.

src/parser/jobs/rpr/modules/ImmortalSacrifice.tsx
```diff
@@ -0,0 +1,154 @@
...
+				<DataLink status="IMMORTAL_SACRIFICE"/> stacks will be generated when you
+				use <DataLink action="ARCANE_CIRCLE"/>, you can then consume those stacks by
+				using <DataLink action="PLENTIFUL_HARVEST"/>, not only it's a high potency skill, it also grants enough
+				Shroud gauge for a <DataLink action="ENSHROUD"/> usage.
+			</Trans>,
+			displayOrder: DISPLAY_ORDER.IMMORTAL_SACRIFICE,
+			requirements: [
+				new Requirement({
+					name: <Trans id="rpr.immortal-sacrifice.checklist.requirement.name">
+						<DataLink status="IMMORTAL_SACRIFICE"/> stacks consumed
```
ayyaruq:
```suggestion
						<DataLink status="IMMORTAL_SACRIFICE"/> stacks used
```

src/parser/jobs/rpr/modules/index.ts
```diff
@@ -4,6 +4,7 @@ import {AoE} from './AoE'
 import {Combos} from './Combos'
 import {CooldownDowntime} from './CooldownDowntime'
 import {DeathsDesign} from './DeathsDesign'
+import ImmortalSacrifice from './ImmortalSacrifice'
```
ayyaruq:
```suggestion
import {ImmortalSacrifice} from './ImmortalSacrifice'
```

src/parser/jobs/rpr/modules/index.ts
```diff
@@ -14,4 +15,5 @@ export default [
 	DeathsDesign,
 	Tincture,
 	Weaving,
+	ImmortalSacrifice,
```
ayyaruq: These should be in alphabetical order.

src/parser/jobs/rpr/changelog.tsx
```diff
@@ -3,6 +3,11 @@ import CONTRIBUTORS from 'data/CONTRIBUTORS'
 import React from 'react'
 
 export const changelog = [
+	{
+		date: new Date('2021-12-31'),
+		Changes: () => <>Added checklist for <DataLink status="IMMORTAL_SACRIFICE"/> consumption.</>,
```
ayyaruq:
```suggestion
		Changes: () => <>Add checklist for <DataLink status="IMMORTAL_SACRIFICE"/> consumption.</>,
```

ayyaruq (changes requested):
That avatar is going to need resizing, we expect images to be 100x100 and generally under 20kb.

Couple of questionable design choices but I think it's best to just fix the immediate issues and then do a cleanup pass later so we can get the functionality out there for users. For the suggestion style comments, you can just click "commit" to apply them.

Expanding on design choices - the window handling seems somewhat flaky and I'm not sure the value of the checklist when it's resolving against the number of Immortal Sacrifice windows rather than actual stacks used. The duplication in closing logic is also somewhat concerning, feels error prone. I think this could be simplified quite a lot as well as made more accurate, but as long as it works yolo for now.

src/parser/jobs/rpr/modules/ImmortalSacrifice.tsx
kennedyoliveira: Initially, I was going to offer more than just stack generated and consumed, I thought maybe checking the number of stacks you got and give a minor hint if you get low stacks, since they don't depend on you and are random, but in my experience, if everyone is keeping their gcd rolling and you hit all members with buff you should get 6+ more often then not, also would show the window for you to check when that happened what was going on, that's why I implemented this as a window at first. Later after thinking a bit about it, I'm not sure how useful this feedback to the user would be so I went for a simple generated/consumed but left the logic there to explore as I check more parses, etc.

Regarding the `consumed` I think it should be a boolean because you can't control the stack generation and when you are generating stacks you can't consume them, and when you are ready to consume you can't generate more, basically you can only consume all the stacks, so the amount consumed didn't make much sense to me, this ties to the same thing I was mentioning before, not sure how useful this info will be. I think it's easier to read something like you had the opportunity to use PH 5 times, but you only used 4 times, instead of you had 35 stacks but consumed only 28. Maybe it would make more sense to just rewrite the messages along these lines, focusing on the PH usage instead of stack.

What are you thoughts?

src/parser/jobs/rpr/modules/ImmortalSacrifice.tsx
kennedyoliveira: Can you elaborate on why immortal stack is not necessary a stack gain? Because the buff to generate it has a different name, also the skill buff is different, I think you get 3 different buffs + this status when you actually get a stack, but I'm totally new on this parse stuff could be missing something.

src/parser/jobs/rpr/modules/ImmortalSacrifice.tsx
kennedyoliveira: Since it is tied to the use of the buff I thought this message wouldn't make much sense, I'll add another message in the case it's 0 telling you should use the buff to generate stacks etc.

src/parser/jobs/rpr/modules/ImmortalSacrifice.tsx
ayyaruq: Regarding the window, I think it depends on the intent. Purely for a checklist, counting stacks gained vs spent is more efficient. It's simpler to implement and more accurate. If you wanted some sort of chart with each use of Arcane Crest showing number of people hit, stacks gained, stacks spent, and a link to the timeline so they can see the rest of their rotation then I think the window has a lot more value.

I think there's value in tracking actual consumptions, but you're probably right that the core thing here is the description text. We can adjust the message to reflect the intent, we just need to decide what we want to be showing.

src/parser/jobs/rpr/modules/ImmortalSacrifice.tsx
ayyaruq: I'm not super across the individual statuses yet (happy to look but I have RPR at 78 only lol) but the way stacking buffs work is kinda awkward. Each stack change counts as a new status apply even when it's *lower*, so normally we'd call this something like `onApply` (or `onGain` but it isn't really a gain here) and reconcile stack count against `event.data`. It gets a bit weird tho since your final stack will apply with 1 and then remove, so it never actually hits zero. Sequentially, it looks something like this:

```
statusApply stacks=3
statusApply stacks=2
statusApply stacks=1
statusRemove
```

src/parser/jobs/rpr/modules/ImmortalSacrifice.tsx
ayyaruq: Maybe consider a Suggestion for that.

src/parser/jobs/rpr/modules/ImmortalSacrifice.tsx
kennedyoliveira: Great, so I'll rework the code a bit to simplify it, but will go with the PH usage message, I think it works much better as a checklist.

src/parser/jobs/rpr/modules/ImmortalSacrifice.tsx
kennedyoliveira: Fixed

src/parser/jobs/rpr/modules/ImmortalSacrifice.tsx
kennedyoliveira: Ah got you, but since this one you consume all the stacks at once, I think it's a bit different from the others, this you just get the statusRemove, and thats it. At least I checked multiple of my own logs and they work like that, both for consuming via plentiful harvest and from letting them expire.

Here is an example from one of my own logs
image

Here is the link if you want to double-check or investigate something
https://www.fflogs.com/reports/qdYN1Q4B62RHvxM9#fight=9&type=damage-done&view=events

I did rename it as your suggestion but the logic seems good.

src/parser/jobs/rpr/modules/ImmortalSacrifice.tsx
kennedyoliveira: Fixed

src/parser/jobs/rpr/modules/ImmortalSacrifice.tsx
kennedyoliveira: Reworked

src/parser/jobs/rpr/modules/ImmortalSacrifice.tsx
kennedyoliveira: Poor choice on my words, by message I meant a suggestion, added it.

src/parser/jobs/rpr/modules/index.ts
kennedyoliveira: Fixed

src/parser/jobs/rpr/modules/ImmortalSacrifice.tsx
kennedyoliveira: Fixed, this was WebStorm auto path, not sure how to force it to use regular import, have to check that.

ayyaruq: There's prob some plugin for WebStorm that reads editorconfig files which will enforce most of our styles, but I don't think our linter checks for this either. General recommendation is to use Visual Studio Code (not related to Visual Studio itself) since we have a prebuilt plugin config for it.

src/parser/jobs/rpr/modules/ImmortalSacrifice.tsx
ayyaruq: Sounds good to me! There's no reason we can't have more than one display tho - if a checklist+table is useful, we can do both, and it doesn't have to be in this PR, can always add more features later.

src/parser/jobs/rpr/modules/ImmortalSacrifice.tsx
kennedyoliveira: changed the handle name

src/parser/jobs/rpr/modules/ImmortalSacrifice.tsx
kennedyoliveira: Could you elaborate a bit more why you say this is unreliable?

kennedyoliveira:
@ayyaruq I addressed most if not all your comments, could you do a second pass and let me know if there is anything pending?

lailleo (commented):
`/en/messages.json` seems to have a typo in it that could be fixed to make readability better.

>  not only it's a high potency skill

Should likely read as "not only is it a high potency skill".

kennedyoliveira:
> `/en/messages.json` seems to have a typo in it that could be fixed to make readability better.
> 
> > not only it's a high potency skill
> 
> Should likely read as "not only is it a high potency skill".

Maybe you checked outdated file, because we changed this message to `On top of being a high potency skill`.

src/parser/jobs/rpr/modules/ImmortalSacrifice.tsx
```diff
@@ -0,0 +1,158 @@
...
+			let stacksConsumed = 0
+
+			for (const closedWindow of this.closedWindows) {
+				totalStacks += closedWindow.stackCount
+
+				if (closedWindow.consumed) {
+					stacksConsumed += closedWindow.stackCount
+					plentifulHarvestUses += 1
+				}
+			}
```
ayyaruq, line 110: This could probably be rewritten as a reducer but eh.

ayyaruq (approved):
Merging as is, I think given the amount of RPR stuff and number of contributors I'm just gonna yolo changes for now and then do a cleanup pass later.

PR #1483, opened 2022-01-01
[AST] updating to 6.0 (Macrocosmos, Minor Arcana, Helios juggling, bug fixes)

otocephaly:
please see commits below for individual changes


src/parser/jobs/ast/modules/CrownPlay.tsx
```diff
@@ -0,0 +1,250 @@
...
+
+		this.addEventHook('complete', this.onComplete)
+	}
+
+	private onMinorArcana(event: Events['action']) {
+
+		this.minorArcana++
+
+		if (this.lastCrownTimestamp === 0) {
+			// The first use, take holding as from the start of the fight. assumes that crown play is not on CD
```
suShirow, line 102: AFAIK minor arcanas cannot be pulled outside of combat. So that should simplify our calcs here (which I understand were probably adapted from draw to some extent)

src/parser/jobs/ast/modules/CrownPlay.tsx
```diff
@@ -0,0 +1,250 @@
...
+		// update the last use
+		this.lastCrownTimestamp = event.timestamp
+		this.lastEventType = event.type
+	}
+
+	private onCrownAction(event: Events['action']) {
+		if (event.action === this.data.actions.LORD_OF_CROWNS.id) {
+			this.lordCasts++
+			//to catch prepulls
+			if (this.lordObtained === 0 && this.parser.pull.timestamp + this.data.actions.MINOR_ARCANA.cooldown < event.timestamp) { this.lordObtained++ }
```
suShirow, line 129: Same here, we don't need to look for prepulls

src/parser/jobs/ast/modules/CrownPlay.tsx
```diff
@@ -0,0 +1,250 @@
...
+		}
+
+		//if they overwrote a card but didn't play another, this will catch the last cast
+		if (this.lastEventType === 'action') {
+			if (this.lastCardType === this.data.statuses.LORD_OF_CROWNS_DRAWN.id) { this.lordObtained++ }
+			if (this.lastCardType === this.data.statuses.LADY_OF_CROWNS_DRAWN.id) { this.ladyObtained++ }
+		}
+
+		// Begin Theoretical Max Plays calc		//assumes that crown-play is not on cooldown at the start of the fight
+		const playsFromCrownPlay = Math.ceil(Math.max(0, (this.parser.pull.duration - oGCD_ALLOWANCE)) / this.data.actions.MINOR_ARCANA.cooldown)
```
suShirow, line 168: Will this formula change? since there is no way that minor arcana will be on CD before 0:00 on the fight

suShirow (changes requested):
I'll make another review for the other features as I get around to doing them >__<; Had a chance to look through the minor arcanas quickly. Only thing that sticks out is checking for prepull prep even though it's not possible to do that (?) 

Perhaps in a future update, we could also point out using using a GCD aoe heal despite having lady of crowns ready, though I'm not sure the potencies warrant this. (But that's for discussion in discord at a later date)

src/parser/jobs/ast/modules/HeliosSupport.tsx
```diff
@@ -21,50 +21,54 @@ const SEVERITIES = {
 	},
 }
 
-export class Horoscope extends Analyser {
+export class HeliosSupport extends Analyser {
```
suShirow: Maybe change the name of this module to just "Helios"?

IIRC my original intention was: if they ever cast an AOE GCD heal, we're going help track if they could have augmented it with neutral or horoscope, potentially saving them a heal further down the timeline thanks to the increased/prorated potency benefits.

It looks like these changes align with that, and so I support these changes. But let me know if I misunderstood it. Nice work!

src/parser/jobs/ast/modules/ArcanaTracking/ArcanaTracking.tsx
```diff
@@ -338,7 +338,7 @@ export default class ArcanaTracking extends Analyser {
 				...event,
 			},
 			drawState: undefined,
-			sealState: CLEAN_SEAL_STATE,
+			sealState: lastCardState.sealState,
```
suShirow, line 341: Doesn't dying remove all your seals? This looks like it would carry them over post-death... or did I miss something (._.)

suShirow:
Just need to get through the draw cap and macrocosmos changes, I've looked at the rest and added my comments

src/parser/jobs/ast/modules/ArcanaTracking/ArcanaTracking.tsx
otocephaly: It now carries them over post death (but not when changing instances)
The only reason I remembered this is because I saw a log where someone died and then immediately used astrodyne which wouldn't have been possible without retaining seals

image

otocephaly: Found an irl example to show that this was implemented this patch if you're curious
https://www.fflogs.com/reports/jtJwWN4aH9XRMYxG

https://xivanalysis.com/fflogs/jtJwWN4aH9XRMYxG/20/1104
Click the last zodiark

image

src/parser/jobs/ast/modules/CrownPlay.tsx
otocephaly: Yep, you're correct! The only reason I included this one is for dungeon purposes when they might have used it in the last pull and it's still on CD for the start of the boss
If you prefer, we could remove it for simplicity purposes
I'll leave that to your judgment though

src/parser/jobs/ast/modules/CrownPlay.tsx
otocephaly: I think this formula still holds since it would assume everything went perfectly and it wouldn't be on CD due to last pull before boss or something
I can remove the comment if you prefer based on discussion above
Lemme know!

src/parser/jobs/ast/modules/CrownPlay.tsx
suShirow: I've never thought of supporting dungeon runs before. It would help the log be less jank if someone did try to plug it in.
Let's leave it in!

src/parser/jobs/ast/modules/ArcanaTracking/ArcanaTracking.tsx
suShirow: Dear YoshiP, that's bonkers
- sushi

suShirow (approved):

PR #1485, opened 2022-01-01
Fix borked rushed IR suggestions

ayyaruq:
And an edge case that's fairly common, and an incorrect variable.


Toastdeib (approved):
Looks good to me.

PR #1489, opened 2022-01-02
Use combatantinfo events to provide accurate attribute updates

ackwell:
- Updates the speed constants, totally forgot to do that last time. It was fine until now as we were generating and consuming within our own ecosystem. Now we're using real values from external, these need to be accurate.
- Adds an initial patch checklist file so I don't forget this next time (good joke, I totally will)
- Actually does the title of the PR

Most of the changes are fixes for the actorUpdate dedupe adapter, because we kinda scuffed it last time because there was only one event per pull. Gnarly.

I've also made the speed stat adapter skip storing + calculating speed for actors we already know the values for. It's the small wins, I guess.

~~This does _not_ change the GCD statistic display. It's now more accurate, but will need a bit of a rework to take the `estimated` toggle into account and so on. It's midnight, I'll do it tomorrow or something.~~ Dunnit now.


src/reportSources/legacyFflogs/eventAdapter/translate.ts
```diff
@@ -359,6 +358,37 @@ export class TranslateAdapterStep extends AdapterStep {
 		}
 	}
 
+	private adaptCombatantInfoEvent(event: CombatantInfoEvent): Array<Events['actorUpdate']> {
+		// TODO: Use more info in here. We're currently extracting the speed attribute values for the logging player, but there's also player level, prepull statuses, and more in there.
+
+		const foo: Array<[number | undefined, Attribute]> = [
```
AzariahAeroborn: Rename variable to indicate what it's used for

AzariahAeroborn (dismissed):

src/reportSources/legacyFflogs/eventAdapter/translate.ts
ackwell: ah gdi

src/data/CONTRIBUTORS.ts
```diff
@@ -17,6 +17,15 @@ const CONTRIBUTORS = ensureRecord<Contributor>()({
 	// 		JOBS.SOME_JOB,
 	// 	],
 	// },
+	ACKWELL: {
+		name: 'ackwell',
+		avatar: 'https://secure.gravatar.com/avatar/383e78034167fdfe8303231a6dad33e5?s=100',
```
ackwell, line 22: i can't believe it took me 5 years to work out that i could use a gravatar for this shit

src/parser/core/modules/GlobalCooldown.tsx
```diff
@@ -18,20 +18,34 @@ export class GlobalCooldown extends Analyser {
...
 		this.statistics.add(new SimpleStatistic({
-			title: <Trans id="core.gcd.estimated-gcd">Estimated GCD</Trans>,
+			title: estimated
+				? <Trans id="core.gcd.estimated-gcd">Estimated GCD</Trans>
+				: <Trans id="core.gcd.gcd">GCD Recast</Trans>,
 			icon: this.data.actions.ATTACK.icon,
-			value: this.parser.formatDuration(this.getEstimate()),
-			info: (
+			value: this.parser.formatDuration(this.getDuration()),
+			info: estimated ? (
```
AzariahAeroborn: Should we update the copy to be "...not available from FF Logs for players other than the player who uploaded the log. ..." ?

ackwell: yes. yes it would. bleh.

ackwell: Done. Given the report source system, I've split the message into a generic everyone-gets-this bit, and a more-specific line that can be changed per-source. Obv, just fflogs there for now.

I've used a new (`.v2`) ID because I changed the params to the message, which will cause i18n crashes if there's a mismatch. I'm thinking this should be something we do more often but I really need to work out the i18n story going forward, the tooling is barely functional.

AzariahAeroborn (approved):

Issue #1303, opened 2021-10-31
Monk Endwalker TODO


ayyaruq:
Tracking progress for MNK updates to make is easier for people to see when asking questions in Discord. If you feel like anything is incorrect or missing here, please feel free to leave a comment, or ping acchan#5683 in the #fb-monk channel on the xivanalysis Discord server. If you'd like to work on any of the remaining tasks, *please* reach out to the #dev-melee channel in Discord so we can make sure people aren't duplicating effort for things that are already in progress.

For the most part, this expansion simplifies the code quite a lot, we can remove a lot of edge case handling and only Twin is even a little complex going forward.

### Data Updates
- [x] Update Action entries in root layer #1471 
- [x] Update Status entries in root layer #1471 
- [x] Merge PGL actions into MNK #1321 

### Analyser Updates
- [x] Delete Fists analyser #1471 
- [ ] New Gauge analyser for Beast Chakras and Nadis (in progress - Ay)
- [x] New Masterful Blitz analyser #1492 
- [ ] Replace PB analyser - chart of gauge state, GCDs used, and Blitz used (in progress - Ay)
- [x] Simplify Twin Snakes analyser - we can gain from 4PF now #1471 
- [x] ~~Improve Demolish clipping tracking -  adjust the checklist dynamically, and trying to allow some clips~~ (no longer necessary)
- [x] Update RoF analyser to contain a Blitz #1492 
  - [x] Celestial Revolution is always a mistake #1650 
  - [ ] Add RoW to RoF windows?
  - [ ] Rising Phoenix in any RoF window that contains BH, otherwise Elixir Field, this is mostly to ensure Leaden Fist spam happens under both buffs but maybe shouldn't be enforced?
  - [x] Migrate to core BuffWindow to better handle BH and drift?
- [x] Add Riddle of Wind to CDDT #1471 
- [x] Add Riddle of Wind to a stats box #1645 

### Other Updates
- [x] Split out the Blitz BuffWindow evaluator so the code is easier to read #1615 
- [ ] Add a new BuffWindow evaluator for Drift and track BH in it (in progress - Square)

Gauge is going to be... spicy? I'm still thinking through code structure of what the Gauge analyser should be doing relative to PB and Blitz. We can use Gauge as a data backend where PB/Blitz manage calls to it for building or spending, or we could inline checks directly to it, or we could hook everything as needed in the Gauge and have Blitz/PB just offer suggestions however we currently can't hook on gauge state. Open to ideas.

[...]

PR #1492, opened 2022-01-02
Migrate Riddle of Fire to BuffWindow

Malpp:
Update the RoF module to BuffWindow as suggested in #1303 

image
image


src/parser/jobs/mnk/changelog.tsx
```diff
@@ -2,6 +2,11 @@ import CONTRIBUTORS from 'data/CONTRIBUTORS'
 import React from 'react'
 
 export const changelog = [
+	{
+		date: new Date('2022-01-02'),
+		Changes: () => <>Migrated Riddle of Fire analyser to BuffWindow.</>,
```
hintxiv: Monk changelog uses imperative tense, could you rewrite this to say "migrate" instead of "migrated"? Also maybe just call it a "module" instead of "analyzer" since that's the leyperson's terminology for an xivanalysis module.

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -1,198 +1,199 @@
 import {t} from '@lingui/macro'
 import {Plural, Trans} from '@lingui/react'
-import {DataLink} from 'components/ui/DbLink'
-import {RotationTable} from 'components/ui/RotationTable'
+import {ActionLink, DataLink} from 'components/ui/DbLink'
```
hintxiv: No need to have both `ActionLink` and `DataLink` since the latter supercedes the former.

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -1,198 +1,199 @@
...
 
-const SUGGESTION_TIERS = {
-	1: SEVERITY.MINOR,
-	2: SEVERITY.MEDIUM,
-	5: SEVERITY.MAJOR,
-}
+class MasterfulBlitzEvaluator implements WindowEvaluator {
+	private _blitzSkills: Action[]
+	private _celestialRevolution: Action
+	private _masterfulBlitz: Action
```
hintxiv: The TypeScript `private` keyword replaces the need for prefixing variables with an underscore.

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -1,198 +1,199 @@
...
-		return this.casts.filter(event => event.action === this.data.actions.MEDITATION.id).length
+	private countBlitzSkills(window: HistoryEntry<EvaluatedAction[]>): number {
+		return window.data.filter(value => this._blitzSkills.includes(value.action)).length
 	}
 
-	get stars() {
-		return this.casts.filter(event => event.action === this.data.actions.SIX_SIDED_STAR.id).length
+	private numberOfBlitzExpectedInWindow(windowIndex: number): number {
+		// 2 on odds and 1 on evens
+		return 2 - windowIndex % 2
```
hintxiv: This logic isn't robust enough. If there's a long downtime, or the user just makes a mistake with one of their PBs, you might flip your cycle (i.e. 1 blitz in window 1, 1 blitz in window 2, then 2, 1, 2, 1, 2, 1... from there). This means that only one window missed the requirement, but now we're going to flag every odd window because it doesn't follow the pattern.

I think the easiest solution for this is to just check the # of blitzes in the previous window - if the previous window had 2 blitzes, require 1; if it had 1, require 2.

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -1,198 +1,199 @@
...
-			if (this.riddleHook != null) {
-				this.removeEventHook(this.riddleHook)
-				this.riddleHook = undefined
-			}
-		}
+	@dependency globalCooldown!: GlobalCooldown
 
-		this.riddle = undefined
-	}
+	blitzSkills = [this.data.actions.ELIXIR_FIELD, this.data.actions.RISING_PHOENIX, this.data.actions.PHANTOM_RUSH]
```
hintxiv: These are in `constants.ts` already, you probably want to just reuse that definition.

hintxiv (changes requested):

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
Malpp: In the constants it's `ActionKey` but Im trying to use it with comparing `Action`.
This is the only spot I use it:
```ts
private countBlitzSkills(window: HistoryEntry<EvaluatedAction[]>): number {
    return window.data.filter(value => this.blitzSkills.includes(value.action)).length
}
```
Is there a way to do it?

hintxiv:
```ts
const blitzActions = BLITZ_SKILLS.map(key => this.data.actions[key].id)
```

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
ayyaruq: I'm not sure that actually solves it, since downtime exists. It's possible user could have enough charges of PB that they can do 2 blitzes in 2 consecutive windows I think. Might be worth a time check since last PB or something.

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
ayyaruq: MNK also has a `utilities.ts` with a `fillActions` function, so you can do like
```ts
const blitzActions = fillActions(BLITZ_SKILLS, this.data)
```

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -1,198 +1,208 @@
...
+	LimitedActionsEvaluator,
+	WindowEvaluator,
+} from 'parser/core/modules/ActionWindow'
+import {HistoryEntry} from 'parser/core/modules/ActionWindow/History'
+import {GlobalCooldown} from 'parser/core/modules/GlobalCooldown'
+import {SEVERITY, Suggestion, TieredSuggestion} from 'parser/core/modules/Suggestions'
 import React from 'react'
 import {BLITZ_SKILLS} from './constants'
 import {DISPLAY_ORDER} from './DISPLAY_ORDER'
-import {fillActions} from './utilities'
```
ayyaruq, line 17: You probably want this so you can fill your `blitzActions`.

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -1,198 +1,208 @@
...
+	9: SEVERITY.MEDIUM,
+	6: SEVERITY.MAJOR,
+}
+
+const MISSED_BROTHERHOOD_SEVERITY = {
+	1: SEVERITY.MEDIUM,
+	2: SEVERITY.MAJOR,
+}
+
+const MEDITATION_DURING_ROF_SEVERITY = {
```
ayyaruq: I'd maybe make this "bad gcds" or something, for example using anatman or form shift as well. I'd clean up the names of these vars a bit in general tho.

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -1,198 +1,208 @@
...
+	1: SEVERITY.MEDIUM,
+	3: SEVERITY.MAJOR,
+}
+
 const EXPECTED = {
 	GCDS: 11,
-	BLITZES: 1,
```
ayyaruq, line 23: The only EXPECTED here is now GCDs, could prob just make this `const EXPECTED_GCDS = 11` and be done with it, although I'm not sure why this is needed since the expected GCD evaluator has the GlobalCooldown object yet also wants an expected field.

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -1,198 +1,208 @@
...
+		return 2
+	}
+}
 
-	override initialise(): void {
-		this.blitzActions = fillActions(BLITZ_SKILLS, this.data)
-		const playerFilter = filter<Event>().source(this.parser.actor.id)
-		this.addEventHook(playerFilter.type('statusApply').status(this.data.statuses.RIDDLE_OF_FIRE.id), this.onGain)
-		this.addEventHook(playerFilter.type('statusRemove').status(this.data.statuses.RIDDLE_OF_FIRE.id), this.onDrop)
+class BrotherhoodEvaluator implements WindowEvaluator {
```
ayyaruq: At a high level, I'm not sure the value of this. It's not really a matter of if they used BH or not, but if it was wrapped into the window properly without drift. I was thinking more along the lines of a generic DriftEvaluator, so we could track the difference in timing and make sure the full duration of BH was inside RoF. Might be worth talking to #fb-monk people for feedback on this.

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -1,198 +1,208 @@
...
+				each <ActionLink action="RIDDLE_OF_FIRE"/> window.</Trans>,
+			adjustCount: window => {
+				// 6SS counts as 2 GCDs
+				return -this.numberOfSixSidedStarInWindow(window)
+			},
+		}))
+		this.addEvaluator(new LimitedActionsEvaluator({
+			expectedActions: [
+				{
+					action: this.data.actions.MEDITATION,
```
ayyaruq, line 117: I'd add other non-damage GCDs to this, like Form Shift and Anatman.

ayyaruq (changes requested):
First pass, some thoughts on logic and what we're actually trying to track. Once this is sorted could do a bit of code cleanup (mostly layout/formatting). Very nice work tho!

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
Malpp: Could I have something like: if the time between window 1 and window 2 is greater than 80 seconds (2 PB CDs), than I expect 2 blitzed in that window?

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
Malpp: `DriftEvaluator` doesn't seem to exist? Unless it's something that would need to be created?

Right now since I'm tracking the action of doing BH inside the RoF window I guarantee the full during inside RoF.

I've noticed on my logs that I've been analyzing, sometimes I press BH before RoF because of timing difference, which shows up that I missed BH. I'm guessing the `DriftEvaluator` is what's going to handle that?

AzariahAeroborn:
@Malpp The lint error from the build is because you need to run `yarn extract` and commit the changes that will make to the locale files (due to changing the body of the suggestion content in commit 6d8df95).  

Also, if you can merge the endwalker branch back into your work branch (or rebase, whichever approach you prefer) and sort out the merge conflicts with the contributor & changelog files, we can poke the necessary reviewers to get the rest of it moving.

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -1,198 +1,257 @@
...
+				essentially wasting a GCD.
+			</Trans>,
+			suggestionWindowName,
+			severityTiers: BAD_GCD_DURING_ROF_SEVERITY,
+		}))
+	}
+
+	private numberOfSixSidedStarInWindow(window: HistoryEntry<EvaluatedAction[]>): number {
+		return window.data.filter(value => value.action.id === this.data.actions.SIX_SIDED_STAR.id).length
 	}
```
hintxiv: This is only called once and it's a one-liner, it doesn't need to be in its own function. You do this in a few places (e.g. `checkIfBrotherhoodIsInWindow`) - someone trying to contribute to this file in the future is going to have to jump around a lot to follow the logic.

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -1,198 +1,257 @@
...
+class MasterfulBlitzEvaluator implements WindowEvaluator {
+	private blitzSkills: Array<Action['id']>
+	private celestialRevolution: Action
+	private masterfulBlitz: Action
+	private brotherhood: Action
+
+	constructor(blitzSkills: Array<Action['id']>, celestialRevolution: Action, masterfulBlitz: Action, brotherhood: Action) {
+		this.blitzSkills = blitzSkills
+		this.celestialRevolution = celestialRevolution
+		this.masterfulBlitz = masterfulBlitz
```
hintxiv: `masterfulBlitz` is assigned but never accessed, this seems like dead code.

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -1,198 +1,257 @@
...
-// Expected GCDs and buffs
-// technically can get 2 tacles and should as much as possible, but don't ding for it in case it's for mechanics
-const EXPECTED = {
-	GCDS: 11,
-	BLITZES: 1,
+const TOTAL_GCD_DURING_ROF_SEVERITY = {
+	10: SEVERITY.MINOR,
+	9: SEVERITY.MEDIUM,
+	6: SEVERITY.MAJOR,
 }
```
hintxiv: A bit nitpicky, but you can combine all these severity consts into just one const:

```ts
const SEVERITIES = {
    TOTAL_GCD_DURING_ROF: { ... }
    ...
}

File: src/parser/jobs/mnk/modules/RiddleOfFire.tsx [hunk 14]
```diff
@@ -1,198 +1,257 @@
...
 	}
 
-	private stopAndSave(endTime: number = this.parser.currentEpochTimestamp): void {
-		if (this.riddle != null) {
-			this.riddle.end = endTime
+	private numberOfBrotherhoodsExpectedInWindow(windowIndex: number, windows: Array<HistoryEntry<EvaluatedAction[]>>): number {
+		// The first window should always have a Brotherhood
+		if (windowIndex === 0) {
+			return 1
+		}
```
hintxiv: If someone weaves BH > RoF in the opener, this will flag.

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -1,198 +1,257 @@
...
+
+		if (this.lastBrotherhoodActionIsGreaterThanCooldown(windowIndex, windows)) {
+			return 1
+		}
 
-			this.history.push(this.riddle)
+		return 0
```
hintxiv: This doesn't really capture the nature of the mistake here - if a person does BH > RoF they'll get a 0/0 even though a normal person would consider this to be a BH + RoF window. If a person completely misaligns BH and it drops before RoF even goes up they'll also get 0/0 when we should be calling it 0/1. 

I'm okay with axing this logic altogether and saving it for a separate module (or a separate PR) but I don't think it should be merged as-is.

hintxiv (changes requested):

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -1,198 +1,262 @@
...
+		6: SEVERITY.MAJOR,
+	},
+	BROTHERHOOD_OUTSIDE_ROF_SEVERITY: {
+		1: SEVERITY.MEDIUM,
+		2: SEVERITY.MAJOR,
+	},
+	BAD_GCD_DURING_ROF_SEVERITY: {
+		1: SEVERITY.MEDIUM,
+		2: SEVERITY.MAJOR,
+	},
```
ayyaruq: Hypernit - these names are redundant. You're effectively writing `SEVERITY.FOO_SEVERITY`, they're all relative to RoF too, so I think `MISSED_GCDS`, `BAD_GCDS`, and `DRIFTED_BH` or something is clear enough.

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -1,198 +1,262 @@
...
 }
 
-const SUGGESTION_TIERS = {
-	1: SEVERITY.MINOR,
-	2: SEVERITY.MEDIUM,
-	5: SEVERITY.MAJOR,
-}
+const DOUBLE_PERFECT_BALANCE_IN_SECONDS = 80
```
ayyaruq: This isn't needed.

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -1,198 +1,262 @@
...
-	casts: Array<Events['action']>
-	start: number
-	end?: number
-	rushing: boolean = false
-
-	constructor(start: number, data: Data) {
-		this.data = data
-		this.start = start
-		this.casts = []
+	constructor(blitzSkills: Array<Action['id']>, celestialRevolution: Action, brotherhood: Action) {
```
ayyaruq: To make calling this easier, might wanna take some sort of `opts` object with these params as members. Also allows for easier defaults.

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -1,198 +1,262 @@
...
+			return new Suggestion({
+				icon: this.celestialRevolution.icon,
+				content: <Trans id="mnk.rof.suggestions.celestialrevolution.content">Avoid using <ActionLink
+					action="CELESTIAL_REVOLUTION"/>.</Trans>,
+				severity: SEVERITY.MAJOR,
+				why: <Trans id="mnk.rof.suggestions.celestialrevolution.why"><ActionLink
+					action="RISING_PHOENIX"/> and <ActionLink action="ELIXIR_FIELD"/> do more potency along with
+					AoE.</Trans>,
+			})
+		}
```
ayyaruq: This can be a tiered severity. Since it's 1 use is major, if you set the tiers with `1: SEVERITY.MAJOR` it'll only show if they used CR, and less indentation is easier to read.

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -1,198 +1,262 @@
...
 	}
 
-	get meditations() {
-		return this.casts.filter(event => event.action === this.data.actions.MEDITATION.id).length
+	output(windows: Array<HistoryEntry<EvaluatedAction[]>>): EvaluationOutput {
+		return {
+			format: 'table',
+			header: {
+				header: <ActionLink showName={false} action={'MASTERFUL_BLITZ'}/>,
```
ayyaruq: Lets be consistent, unless there's a reason not to.
```suggestion
				header: <ActionLink showName={false} action="MASTERFUL_BLITZ"/>,
```

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -1,198 +1,262 @@
...
-	static override displayOrder = DISPLAY_ORDER.RIDDLE_OF_FIRE
 
-	@dependency private data!: Data
-	@dependency private suggestions!: Suggestions
-	@dependency private timeline!: Timeline
-
-	private history: Riddle[] = []
-	private riddle?: Riddle
-	private riddleHook?: EventHook<Events['action']>
+	private numberOfBlitzExpectedInWindow(windowIndex: number, windows: Array<HistoryEntry<EvaluatedAction[]>>): number {
```
ayyaruq: Hypernit - can we shorten some of these names?

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -1,198 +1,262 @@
...
+		// First window should always contain 2 blitz
+		if (windowIndex === 0) {
+			return 2
+		}
```
ayyaruq: Still not keen on this, I think it's an unsafe assumption. For example, user goes into a gatekeeper fight second half and with BH on cooldown. You care about time since previous BH, not number of windows.

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -1,198 +1,262 @@
...
+		if (isBrotherhoodInWindow) {
+			return 2
+		}
 
-	override initialise(): void {
-		this.blitzActions = fillActions(BLITZ_SKILLS, this.data)
-		const playerFilter = filter<Event>().source(this.parser.actor.id)
-		this.addEventHook(playerFilter.type('statusApply').status(this.data.statuses.RIDDLE_OF_FIRE.id), this.onGain)
-		this.addEventHook(playerFilter.type('statusRemove').status(this.data.statuses.RIDDLE_OF_FIRE.id), this.onDrop)
+		if (this.timeBetweenLastWindowInSeconds(windowIndex, windows) > DOUBLE_PERFECT_BALANCE_IN_SECONDS) {
```
ayyaruq: As noted above, we don't really operate on seconds, I think you can just use the times directly with the action cooldown. This also doesn't guarantee they have 2 charges technically. Might be worth a timer hook on the cooldown itself or something here. Not too sure, might need to think about it.

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -1,198 +1,262 @@
...
+			return 2
+		}
 
-		this.addEventHook('complete', this.onComplete)
+		const blitzSkillsInLastWindow = this.countBlitzSkills(windows[windowIndex - 1])
+		if (blitzSkillsInLastWindow === 2) {
+			return 1
+		}
```
ayyaruq: You can fold this into a single line, or at least lose the const, same above tbh. Hypernit - slap a newline before the final return tho.

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -1,198 +1,262 @@
...
 	}
 
-	onCast(event: Events['action']): void {
-		const action = this.data.getAction(event.action)
+	private timeBetweenLastWindowInSeconds(windowIndex: number, windows: Array<HistoryEntry<EvaluatedAction[]>>): number {
+		if (windowIndex === 0) {
+			return 0
+		}
+		return (windows[windowIndex].start - windows[windowIndex].start) / 1000
+	}
```
ayyaruq: I don't think this is needed.

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -1,198 +1,262 @@
...
-		if (action == null) { return }
+class BrotherhoodEvaluator implements WindowEvaluator {
+	private brotherhood: ActionRoot['BROTHERHOOD']
+	private brotherhoodStartHistory: number[]
 
-		// MNK mentors want the oGCDs :angryeyes:
-		if (this.riddle != null) {
-			this.riddle.casts.push(event)
-		}
+	constructor(brotherhood: ActionRoot['BROTHERHOOD'], brotherhoodStartHistory: number[]) {
```
ayyaruq: Same deal here, an options block might be better.

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -1,198 +1,262 @@
...
+		this.brotherhoodStartHistory = brotherhoodStartHistory
 	}
 
-	private onGain(event: Events['statusApply']): void {
-		if (this.riddle != null) { return }
+	suggest(windows: Array<HistoryEntry<EvaluatedAction[]>>) {
+		const possibleNumberOfBrotherhoods = Math.ceil(windows.length / 2)
+		const brotherhoodUsagesInsideRof = windows.map(previousValue => previousValue.data.filter(value => value.action === this.brotherhood))
+			.filter(value => value.length > 0).length
+		const missedBhInsideRof = possibleNumberOfBrotherhoods - brotherhoodUsagesInsideRof
```
ayyaruq: Can we rename these? Something like `maxPossibleBH`, `nestedBH`, and `missedBH` are prob good enough. The BH in RoF var should probably be broken into multiple lines more sanely so it's easier to read.

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -1,198 +1,262 @@
...
-			filter<Event>()
-				.source(this.parser.actor.id)
-				.type('action'),
-			this.onCast,
-		)
+	output(windows: Array<HistoryEntry<EvaluatedAction[]>>): EvaluationOutput {
+		return {
+			format: 'table',
+			header: {
+				header: <ActionLink showName={false} action={'BROTHERHOOD'}/>,
```
ayyaruq: Same deal here.
```suggestion
				header: <ActionLink showName={false} action="BROTHERHOOD"/>,
```

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -1,198 +1,262 @@
...
 
-		this.riddle = new Riddle(event.timestamp, this.data)
-		this.riddle.rushing = this.data.statuses.RIDDLE_OF_FIRE.duration >= (this.parser.pull.timestamp + this.parser.pull.duration) - event.timestamp
+		return new TieredSuggestion({
+			icon: this.brotherhood.icon,
+			tiers: SEVERITIES.BROTHERHOOD_OUTSIDE_ROF_SEVERITY,
+			value: missedBhInsideRof,
+			content: <Trans id="mnk.rof.suggestions.brotherhood.content">Missed {missedBhInsideRof} <ActionLink
+				action="BROTHERHOOD"/> inside <ActionLink action="RIDDLE_OF_FIRE"/> window.
+			</Trans>,
```
ayyaruq: This should use proper i18n Plural to format correctly, and try to avoid line-breaking HTML tags and their params.

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -1,198 +1,262 @@
...
+			why: <Trans id="mnk.rof.suggestions.brotherhood.why"><ActionLink
+				action="BROTHERHOOD"/> has a shorter duration than <ActionLink action="RIDDLE_OF_FIRE"/>.</Trans>,
```
ayyaruq: This message doesn't really explain why they're being dinged. Also, this is a bit iffy to read, avoid line-breaking mid tag, you can start the line with ActionLink outright, and pop the closing Trans tag on the next line aligned with the `why` field.

ayyaruq (changes requested):
Stopping my review here, insane variable names aside, we need to rethink this window indexing stuff. At the current state I think it's easier to not use BuffWindow, which means we've definitely overcomplicated things because it shouldn't be easier to roll custom logic than using a core module designed for this purpose. Maybe we need to have a concrete list of goals for this PR, it seems to be handling some things but not others, and the way BH is done atm is unreliable. Happy to discuss although I'm taking a break from discord for a bit.

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
ayyaruq: Sorry for the slow reply, yes I think that's a good approach.

ayyaruq (changes requested):
Just need another `yarn extract` and this is good to go. Thanks for your patience!

ayyaruq (approved):

PR #1496, opened 2022-01-03
6.0 PLD Support

Dynme:
Now with less page crashes and type errors!


src/parser/jobs/pld/modules/Goring.tsx
```diff
@@ -1,62 +1,231 @@
...
 
-	override trackedStatuses = [
-		STATUSES.GORING_BLADE.id,
-	]
+	private readonly statusDuration = {
+		[this.data.statuses.GORING_BLADE.id]: this.data.statuses.GORING_BLADE.duration,
+		[this.data.statuses.BLADE_OF_VALOR.id]: this.data.statuses.BLADE_OF_VALOR.duration,
+	}
+
+	private lastDotCast: Status['id'] = this.data.statuses.GORING_BLADE.id
```
AzariahAeroborn: After the other changes to the DotTargetTracker & clip calculation, this is no longer used and can be removed along with the onXCast hooks.

src/parser/jobs/pld/modules/OGCDDowntime.tsx
```diff
@@ -1,24 +1,17 @@
 import {CooldownDowntime} from 'parser/core/modules/CooldownDowntime'
 
-// Allowed downtime set to 4s to account for PLD's natural rotation drift.
-const ALLOWED_DOWNTIME_FOF = 4000
-const ALLOWED_DOWNTIME_REQ = 4000
+// Allowed downtime set to 2s to account for PLD's natural rotation drift.
+const ALLOWED_DOWNTIME_FOF = 3000
+const ALLOWED_DOWNTIME_REQ = 2000
```
AzariahAeroborn: Comment here indicates average allowed downtime of 2s, but the constant for Fight or Flight downtime is 3s.  Is that intended?

AzariahAeroborn (changes requested):

AzariahAeroborn (approved):

PR #1497, opened 2022-01-03
[SCH] Add Aetherflow to Tracked Cooldowns

dthomash:
Just a small changed to add Aetherflow to the tracked cooldowns section.


src/parser/jobs/sch/modules/ScholarCooldownDowntime.tsx
```diff
@@ -0,0 +1,23 @@
+import {Trans} from '@lingui/react'
+import {ActionLink} from 'components/ui/DbLink'
+import {CooldownDowntime} from 'parser/core/modules/CooldownDowntime'
```
ayyaruq: Prefer to rename the import and use a sane name for the class.

```suggestion
import {CooldownDowntime as CoreCooldownDowntime} from 'parser/core/modules/CooldownDowntime'
```

src/parser/jobs/sch/modules/ScholarCooldownDowntime.tsx
```diff
@@ -0,0 +1,23 @@
...
+import React, {Fragment} from 'react'
+
+export default class ScholarCooldownDowntime extends CooldownDowntime {
```
ayyaruq: We don't allow default exports.

```suggestion
export class CooldownDowntime extends CoreCooldownDowntime {
```

src/parser/jobs/sch/modules/ScholarCooldownDowntime.tsx
```diff
@@ -0,0 +1,23 @@
...
+	override trackedCds = [
+		{
+			cooldowns: [this.data.actions.CHAIN_STRATAGEM],
+			allowedAverageDowntime: 7500,
+			firstUseOffset: 10000,
```
ayyaruq: These don't seem to reflect actual usage in openers. The allowed average here also seems really high, although you don't actually need to configure it unless there's a good reason for an explicit setting.

src/parser/jobs/sch/modules/ScholarCooldownDowntime.tsx
```diff
@@ -0,0 +1,23 @@
...
+		},
+		{
+			cooldowns: [this.data.actions.AETHERFLOW],
+			allowedAverageDowntime: 5000,
+			firstUseOffset: 10000,
```
ayyaruq: Same deal as Chain Strat here.

src/parser/jobs/sch/modules/index.js
```diff
@@ -1,18 +1,18 @@
-import ChainStrat from './ChainStrat'
 import DoTs from './DoTs'
 import Interrupts from './Interrupts'
 import Overheal from './Overheal'
 import Pets from './Pets'
+import ScholarCooldownDowntime from './ScholarCooldownDowntime'
```
ayyaruq:
```suggestion
import {CooldownDowntime} from './ScholarCooldownDowntime'
```

As an aside, these imports are somewhat against convention and should be updated, not sure how this even made it into the endwalker branch with default exports.

src/parser/jobs/sch/modules/index.js
```diff
@@ -1,18 +1,18 @@
...
 import Tincture from './Tincture'
 import Weaving from './Weaving'
 
 export default [
-	ChainStrat,
 	DoTs,
 	Interrupts,
 	Overheal,
 	Pets,
+	ScholarCooldownDowntime,
```
ayyaruq: Adjust appropriately.

src/parser/jobs/sch/modules/ScholarCooldownDowntime.tsx
```diff
@@ -0,0 +1,23 @@
...
+		},
+	]
+	override checklistDescription = <Fragment>
+		<div><Trans id="sch.cooldown.chainstrat.description">Try to use <ActionLink {...this.data.actions.CHAIN_STRATAGEM}/> on cooldown throughout the fight, particularly when your team has other buffs up for maximum effect.</Trans></div>
+		<div><Trans id="sch.cooldown.aetherflow.description">Try to use <ActionLink {...this.data.actions.AETHERFLOW}/> on cooldown throughout the fight. If you have extra Aetherflow charges, cast Energy Drain to do additional damage before casting <ActionLink {...this.data.actions.AETHERFLOW}/>.</Trans></div>
+	</Fragment>
```
ayyaruq: Any reason this needs a div or explicit Fragment? Compared to say,
```suggestion
	override checklistDescription = <>
		<Trans id="sch.cooldown.chainstrat.description">Try to use <ActionLink action="CHAIN_STRATAGEM"/> on cooldown throughout the fight, particularly when your team has other buffs up for maximum effect.</Trans>
		<Trans id="sch.cooldown.aetherflow.description">Try to use <ActionLink action="AETHERFLOW"/> on cooldown throughout the fight. If you have excess charges, cast <ActionLink action="SCH_ENERGY_DRAIN"/> to do additional damage before casting <ActionLink showIcon={false} action="AETHERFLOW"/>.</Trans>
	</>
```

I'm not sure this is actually a good idea tho, the double message is awkward UX imo.

ayyaruq (changes requested):
I'd double check those CDDT numbers, and requesting fixing the imports. Also, a changelog entry is required for this change.

src/parser/jobs/sch/modules/ScholarCooldownDowntime.tsx
dthomash: It looked awkward all as one paragraph to me. So I added the explicit divs to force a line break. If there's a way to do that more idomatically in react I'm game. But to your other point maybe it is better to just change it to a generic message for the cooldown section, and then maybe add more specific messages as conditional suggestions.

src/parser/jobs/sch/modules/ScholarCooldownDowntime.tsx
dthomash: These were the original numbers from the chainstrat module in the endwalker branch. I just copied them over.

src/parser/jobs/sch/modules/ScholarCooldownDowntime.tsx
ayyaruq: I don't see any suggestions but that sounds like a more reasonable approach.

src/parser/jobs/sch/modules/ScholarCooldownDowntime.tsx
ayyaruq: Doesn't look like they were updated for 6.0 then, will need to handle that. If the PR is purely to add Aetherflow, it should *only* add Aetherflow tho. If you're doing a more general update for Cooldown Downtime then the numbers should be fixed. Up to you which you prefer.

src/parser/jobs/sch/modules/ScholarCooldownDowntime.tsx
dthomash: I left it loose because I wasn't sure why Chain Strat was left loose. I would think maybe 1 or 2 GCDs would be more appropriate. But if leaving it unset works that's probably what I'll do then since adding it is just to make sure you're casting it every minute.

ayyaruq: Average allowed downtime defaults to 1250ms, but note that this is the average, so you can have up to double that. With 7500, you're saying they can effectively leave it off cooldown for 15 seconds. I'm seeing Aetherflow used pretty consistently beyond 10s into the fight so this first use offset seems wrong. Chain seems to be around 5-6 seconds (3rd GCD?), so I think that might need tightening too, but compare openers across regions since they tend to differ, especially JP/KR.

dthomash: Ah, I didn't realize that the grace period effectively doubled. Yeah that would be way to long. I'll take a close look at openers tomorrow to make sure these values get set correctly.

AzariahAeroborn:
Hey, thanks for the commit! We require a change log entry in the `/parser/<JOB>/index.ts` file for any updates / bug fixes that change job logic, and you'll need to upload an avatar to /src/data/avatar and add yourself to the /src/data/CONTRIBUTORS.ts file.

src/parser/jobs/sch/modules/ScholarCooldownDowntime.tsx
dthomash: After looking at other CDDT modules it seems standard to not include suggestions in the CDDT module. So I just removed the custom checklistDescription altogether.

src/parser/jobs/sch/modules/ScholarCooldownDowntime.tsx
dthomash: After looking a some guides and some top logs it seems like Chain Strat always gets casted right before the 4th GCD (to match up buff windows), which at max GCD is 10s. So 10s for chain strat first use seems fine. The first Aetherflow cast should generally be casted before the first Chain Strat so that you can fit 5 energy drains in the Chain Strat window. So I picked 7.5s for that since Aetherflow would need to be casted before the 3rd GCD.

I think the averageAllowedDowntime should be fine at the default since it's just used to calculate the max number of casts, and the only time I saw Chain Strat being delayed looked like a specific raid strat.

ayyaruq (changes requested):
Looking good - final thing, could you resize your avatar to 100x100 and make sure the file size is more appropriate? We don't show these any larger and usually expect them to be 10-50kb at most.

src/parser/jobs/sch/modules/ScholarCooldownDowntime.tsx
ayyaruq: I'd maybe check some logs from other regions to make sure there's no different approach used in like, JP since they tend to have diff openers that are equally viable. Also worth checking some non-top logs, since the higher ranked players tend to have approaches highly optimised that don't reflect the general case. Usually 80-90th percentile is a good baseline for generalising.

dthomash: I checked a few more logs in the range you suggested, on both savage and normal, and in multiple regions. The opener seems to hold pretty steady. The variations that moved Chain Strat around only seemed to move it forwards, so those variations would not be penalized by the current setting.
1. Precast Broil
2. Biolysis
  a. Weave Aetherflow
3. Broil
  a. Optional Energy Drain Weave (you can fit a max of 5 Energy Drain casts in Chain Strat duration, so you do one early or one late)
4. Broil
  a. Weave Chain Strat
5. First burst window (5x Broil, 1x Dissipation, 5x Energy Drain) with slight variations

ayyaruq: Awesome, thanks for triple checking!

ayyaruq (approved):

PR #1503, opened 2022-01-04
[WHM] Adding Lilybell and Aquaveil to WHM oGCDs

psummers678:
Moved the list of WHM oGCDs to the defensive checklist, and added the new actions.
Bit of redundancy with checklist + suggestions, however did not want to remove the previous functionality either. If there is a preference let me know.

Also removed the default export on the file


suShirow (approved):
Thanks for your PR WHM was in need of love

PR #1515, opened 2022-01-07
[AST] 6.05 support

otocephaly:
nothing major here
reduced thresholds for lucid dreaming because ast is nice with mana this patch and the focus should be on astrodyne and draw
updated draw math because it still wasn't calculating correctly. This time I am confident that the calculation is actually right. If you're curious, it should line up with the white gaps in the timeline for draw which is what sparked me to investigate why it didn't


suShirow (approved):
Bug feex are gud

PR #1516, opened 2022-01-07
DRG: Track First Brood's Gaze with Core Gauge, End of Fight LotD Suggestions

ebshimizu:
Two changes:

- The First Brood's Gaze gauge (the thing you get by using mirage dive) is now being tracked with Core Gauge
- Two suggestions for end of fight life windows have been added. The reasoning for this is that the EW DRG rotation has us delaying the first LotD by 30 seconds to align with 1 minute buff timing for the rest of the fight. The problem with this is that it's very easy to automatically delay life at the end of a, say, 9 minute fight and lose out on a window (even through it's unbuffed). Two cases are being detected here, both are types of transposition errors (GSK before MD):
  - Case 1 (Major): LotD was improperly delayed. Detected by checking if the player has two eyes at the end of the fight as used GSK before they had two eyes (meaning they could've transposed the cast order and entered life). There is a 5s buffer for this check (two GCDs).
  - Case 2 (Medium): LotD was used but truncated (used late). If we are in life at the end of the fight, check to see if we could've used it earlier, and if it's more than a 10s difference (likely could've fit an extra Nastrond), flag it (see my log for an example: https://www.fflogs.com/reports/B4Q16j3WG9DFygNR#fight=13&source=124)


ebshimizu:
Rhy seconds the logic here so we're good to go with regard to job-specific logic.

src/parser/jobs/drg/modules/BloodOfTheDragon.tsx
```diff
@@ -245,17 +268,14 @@ export default class BloodOfTheDragon extends Analyser {
...
-
-		// TODO: check if you lose eyes on death
-		this.eyes = 0
 	}
 
-	onRaise(event: Event) {
+	onRaiseLotd(event: Event) {
 		// So floor time doesn't count against BotD uptime
 		this.lastEventTime = event.timestamp
 	}
```
AzariahAeroborn: Is this even still needed?  Since BotD is a passive now, there's no uptime tracking to maintain for that, correct?

AzariahAeroborn (changes requested):

src/parser/jobs/drg/modules/BloodOfTheDragon.tsx
ebshimizu: yep looks like it's unnecessary now, let me get that outta here

AzariahAeroborn (approved):

PR #1517, opened 2022-01-08
BRD - Added Drifting EA module and BL/RoD tracker in RS

jplippi:


src/parser/jobs/brd/modules/DriftingEa.tsx
```diff
@@ -0,0 +1,182 @@
+//                     Lv.89 DRIFTING EA
```
suShirow, line 1: Oh my gawd it's so cute

src/parser/jobs/brd/modules/DriftingEa.tsx
```diff
@@ -0,0 +1,182 @@
...
+		// Begin the next window
+		this.currentWindow = {
+			start: observedUseTime,
+			casts: [],
+		}
+	}
+
+	private onCast(event: Events['action']) {
+		this.currentWindow.casts.push(event)
+	}
```
AzariahAeroborn: It looks to me like we are not doing any analysis or display of the casts used in between EA casts, so can we remove this tracking and the associated hook for all player casts?

AzariahAeroborn (changes requested):

src/parser/jobs/brd/modules/DriftingEa.tsx
jplippi: Yep, I just kinda copied from MCH module and forgot to remove this part

AzariahAeroborn (approved):

PR #1519, opened 2022-01-08
fix: when tracking target deaths for invulnerability, also track raises

AzariahAeroborn:
This improves the behavior of the CDDT module against level 70+ striking dummies, since the dummy will be marked as targetable again once they raise.  

Extremely low level dummies (such that every hit one shots them and they never go back above 0 hp) will still have the original problems caused from being marked invulnerable for most of the duration of the pull.  I did some investigation into the possibility of checking the max hp of the enemies and flagging below a certain threshold.  However, many effect actors and certain trash mobs in dungeons have hp that would also flag for that sort of threshold check so I don't think that would be a good threshold check to use.


src/parser/core/modules/Invulnerability.tsx
```diff
@@ -110,7 +110,7 @@ const DEFAULT_ACTOR_FILTER = filter<Actor>().team(Team.FOE)
 
 export class Invulnerability extends Analyser {
 	static override handle = 'invulnerability'
-	static override debug = false
+	static override debug = true
```
suShirow: hey az did you intentionally leave debug on?

AzariahAeroborn: Zodiark dammit...

ayyaruq (approved):

PR #1520, opened 2022-01-09
Update Procs.tsx

lailleo:
Update procs module to make overwriting severity more in line with EW changes, per discussion with RDM TC group.


ayyaruq (changes requested):
Thanks for your contribution! As this changes job code, a changelog entry is required. Should be in either the index or changelog files depending which RDM devs preferred, just above the `rdm/modules` directory.

lailleo:
> Thanks for your contribution! As this changes job code, a changelog entry is required. Should be in either the index or changelog files depending which RDM devs preferred, just above the `rdm/modules` directory.

Should be updated. I also added an avatar/contributor info per a discussion with Leylia in the RDM channels.

ayyaruq (approved):

PR #1529, opened 2022-01-10
[SAM] 6.0 to 6.05 Support

Rjb99:
This PR contains:

- Higanbana
- Shoha
- Kaiten
- Ogi

Also contains changes to DisplayOrder, update info graph links, marks SAM as supported for 6.0 to 6.05.


src/parser/jobs/sam/modules/Kaiten.tsx
```diff
@@ -0,0 +1,149 @@
...
+				.action(oneOf(this.TheGoodOnes)),
+			this.onGoodCast
+		)
+
+		this.addEventHook('complete', this.onComplete)
+
+	}
+
+	private onKaiten() {
+		if (this.kaitenUp === true) {
```
ayyaruq, line 89: This is a boolean so no need to be explicit, more of a note here than requiring any changes.

src/parser/jobs/sam/modules/Kaiten.tsx
```diff
@@ -0,0 +1,149 @@
...
+			this.doubleKaitens++
+		}
+		this.kaitenUp = true
+	}
+
+	private onKaitenRemove() {
+		this.kaitenUp = false
```
ayyaruq, line 96: Could inline this but again, style pref more than a req.

src/parser/jobs/sam/modules/Kaiten.tsx
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	private onBadCast() {
+		if (this.kaitenUp === true) {
+			this.badKaitens++
+		}
+	}
+
+	private onGoodCast() {
+		if (this.kaitenUp === false) {
```
ayyaruq, line 106: Might be more reliable to just use your boolean for checking for kaiten overwrites, and then `this.actors.current.hasStatus(this.data.statuses.KAITEN.id)` for where you wanna see if it's up at all.

src/parser/jobs/sam/modules/Kaiten.tsx
```diff
@@ -0,0 +1,149 @@
...
+			this.missedKaitens++
+		}
+	}
+
+	private onComplete() {
+		const test1 = this.missedKaitens
```
ayyaruq: Vestigial test code?

src/parser/jobs/sam/modules/Kaiten.tsx
```diff
@@ -0,0 +1,149 @@
...
+		this.data.actions.JINPU.id,
+		this.data.actions.ENPI.id,
+		this.data.actions.SHIFU.id,
+		this.data.actions.FUKO.id,
+		this.data.actions.GEKKO.id,
+		this.data.actions.MANGETSU.id,
+		this.data.actions.KASHA.id,
+		this.data.actions.OKA.id,
+		this.data.actions.YUKIKAZE.id,
+	]
```
ayyaruq, line 49: I'd probably use a const `ActionKey[]` array for this, and the data module matcher functions in your hooks. Easier to edit/maintain later.

src/parser/jobs/sam/modules/Ogi.tsx
```diff
@@ -0,0 +1,72 @@
...
+	@dependency private checklist!: Checklist
+	@dependency private data!: Data
+
+	private OgiReadys = 0
+	private OgisDone = 0
+
+	Ogis = [
+		this.data.actions.OGI_NAMIKIRI.id,
+		this.data.actions.KAESHI_NAMIKIRI.id,
+	]
```
ayyaruq, line 23: Same as kaiten, might want an `ActionArray[]` and then `this.data.matchActionId(OGI_ACTIONS)` in the hook (which internally uses the oneOf stuff). I'd fix variable name casing and make it private tho if not.

src/parser/jobs/sam/modules/Ogi.tsx
```diff
@@ -0,0 +1,72 @@
...
+		this.addEventHook('complete', this.onComplete)
+	}
+
+	private onOgiReady() {
+		this.OgiReadys++
+	}
+
+	private onOgi()	{
+		this.OgisDone++
+	}
```
ayyaruq: You could inline this. Not requesting it, just noting it.

src/parser/jobs/sam/modules/Ogi.tsx
```diff
@@ -0,0 +1,72 @@
...
+import React from 'react'
+import DISPLAY_ORDER from './DISPLAY_ORDER'
+
+export class Ogi extends Analyser {
+	static override handle = 'ogi'
+	@dependency private checklist!: Checklist
+	@dependency private data!: Data
+
+	private OgiReadys = 0
+	private OgisDone = 0
```
ayyaruq:
```suggestion
	private ogiReadys = 0
	private ogisDone = 0
```

src/parser/jobs/sam/modules/Sen.tsx
```diff
@@ -219,7 +219,7 @@ export class Sen extends Analyser {
 
 		let lastSenState = this.lastSenState
 
-		if ((lastSenState == null) || (lastSenState.isDone === true)) {
+		if ((typeof lastSenState === 'undefined') || (lastSenState.isDone === true)) {
```
ayyaruq: Why?

src/parser/jobs/sam/modules/Shoha.tsx
```diff
@@ -0,0 +1,136 @@
...
+		[this.data.actions.SHOHA.id, {action: -3}],
+		[this.data.actions.SHOHA_II.id, {action: -3}],
+	])
+
+	meditateStart = -1
+
+	override initialise() {
+		super.initialise()
+
+		const MeditateActions = Array.from(this.MeditateGaugeModifiers.keys())
```
ayyaruq:
```suggestion
		const meditateActions = Array.from(this.MeditateGaugeModifiers.keys())
```

src/parser/jobs/sam/modules/Shoha.tsx
```diff
@@ -0,0 +1,136 @@
...
+		this.addEventHook(playerFilter.type('statusApply').status(this.data.statuses.MEDITATE.id), this.onApplyMeditate)
+		this.addEventHook(playerFilter.type('statusRemove').status(this.data.statuses.MEDITATE.id), this.onRemoveMeditate)
+
+		this.addEventHook('complete', this.onComplete)
+	}
+
+	private onGaugeModifier(event: Events['action']) {
+		const modifier = this.MeditateGaugeModifiers.get(event.action)
+
+		if (modifier !== undefined)	{
```
ayyaruq: Prefer null checks.
```suggestion
		if (modifier != null)	{
```

ayyaruq (changes requested):
Mostly just comments - I'd change the icons in Kaiten tho, you have 3 suggestions that are likely to display more than one at a time but with the same icon, are there alternatives you could use? Visual separation helps users a lot.

Other thing of note is the conditionals use a lot of explicit checks, typeOf in particular isn't exactly safe (iirc) but also these booleans can be simplified and we generally prefer null checks over tight undefined checks. This is the only change I'm actually requesting.

You could also fix some variable naming to be more consistent with style for the rest of the codebase but that's more of a nit. Overall, good PR, very little to change. As an aside, might be worth cleaning up remaining default exports while you're here.

src/parser/jobs/sam/modules/Sen.tsx
Rjb99: I merged in a copy of code that didn't get the updates from my last PR it appears.
That is Why

Rjb99:
Requested Changes have been made, was a mismerge from current codebase to the one I was working on before the PR.

Regarding Kaiten images, I don't have any offical icons that would fit the bill, though I am debating making a "blue" kaiten like Tsubame for double kaiten. PieChart stats will probably be added to both of those for feedback as well later down the line as our good vs bad options keep expanding as time goes on

src/parser/jobs/sam/modules/Sen.tsx
Rjb99: fixed

src/parser/jobs/sam/modules/Kaiten.tsx
Rjb99: removed

ayyaruq (approved):
Yolo for now, some code style issues to cleanup later tho.

PR #1532, opened 2022-01-10
6.05 housekeeping

ackwell:
Sanity checks invuln (nothing needed changing, yay), and regenerates UTA.

I've also added docs on how to do both of the above.

Out of scope for this PR, but I'll probably make core's support range effect the site and narrow it down so it needs to be manually bumped alongside the jobs.


ayyaruq (approved):
LGTM

Issue #1513, opened 2022-01-06
Tincture usage missing from Timeline module


nicholasthie:
In one of my logs, I used 3 tinctures in the fight with the last one being used near the end of the fight. My last tincture usage is recorded in the "Tinctures" module, but it's missing from the "Timeline" module (Medicated) . I'm guessing it's because the fight ended before the tincture ran out, but not sure. The job being used here is GNB.

Link to the log: https://xivanalysis.com/fflogs/aJ4A2GzHZXBCtjmy/82/1773

![image](https://user-images.githubusercontent.com/7281764/148433660-abb170f3-fcba-4177-ac27-409d12374278.png)
![image](https://user-images.githubusercontent.com/7281764/148433699-20fb9472-3056-4e14-b2e7-52f5a6041f60.png)

ayyaruq:
This was fixed about 10 days ago, if the issue is still present for you please comment and we'll reopen the issue.

PR #1533, opened 2022-01-10
Resolved issue with medication not being included 

mattmanuel90:
As mentioned here. https://github.com/xivanalysis/xivanalysis/issues/1513

Tested with this log https://www.fflogs.com/reports/aJ4A2GzHZXBCtjmy#fight=82&source=1773

medication

Btw, I just started raiding for the first time and static member referred us this. It's been extremely helpful ^^

### DIFF
--- a/src/parser/core/modules/Medicated.ts
+++ b/src/parser/core/modules/Medicated.ts
@@ -38,10 +38,11 @@ export class Medicated extends Analyser {
 		this.pot = {start: event.timestamp}
 	}
 
-	private stopAndSave(endTime: number) {
+	private stopAndSave(endTime: number = this.parser.currentEpochTimestamp) {
 		if (this.pot != null) {
 			this.pots.push({...this.pot, end: endTime})
 		}
+
 		this.pot = undefined
 	}
 
@@ -49,14 +50,16 @@ export class Medicated extends Analyser {
 		if (this.pot == null) {
 			throw new Error('potion instance not found')
 		}
+
 		this.stopAndSave(event.timestamp)
 	}
 
 	private onComplete() {
+		const status = this.data.statuses.MEDICATED
 
 		// Add a pot that is still up before end of pull
 		if (this.pot != null && this.pot.end == null) {
-			this.stopAndSave(this.parser.pull.timestamp + this.parser.pull.duration)
+			this.stopAndSave()
 		}
 
 		if (this.pots.length > 0) {
@@ -65,9 +68,8 @@ export class Medicated extends Analyser {
 				order: -99,
 			})
 
-			const status = this.data.statuses.MEDICATED
 			this.pots.forEach(pot => row.addItem(new StatusItem({
-				status,
+				status: status,
 				start: pot.start - this.parser.pull.timestamp,
 				end: pot.end != null ? pot.end - this.parser.pull.timestamp : this.parser.pull.duration,
 			})))
--- a/src/parser/core/modules/Medicated.ts
+++ b/src/parser/core/modules/Medicated.ts
@@ -64,7 +64,7 @@ export class Medicated extends Analyser {
 			})
 
 			this.pots.forEach(pot => row.addItem(new StatusItem({
-				status: status,
+				status,
 				start: pot.start - this.parser.pull.timestamp,
 				end: pot.end != null ? pot.end - this.parser.pull.timestamp : this.parser.pull.duration,
 			})))
--- a/src/parser/core/modules/Medicated.ts
+++ b/src/parser/core/modules/Medicated.ts
@@ -43,13 +43,19 @@ export class Medicated extends Analyser {
 			throw new Error('potion instance not found')
 		}
 
-		this.pots.push({end: event.timestamp, ...this.pot})
+		this.pots.push({ ...this.pot, end: event.timestamp })
 
 		this.pot = undefined
 	}
 
 	private onComplete() {
 		const status = this.data.statuses.MEDICATED
+
+		// Add a pot that is still up before end of pull
+		if (this.pot && !this.pot.end) {
+			this.pots.push(this.pot)
+			this.pot = undefined
+		}
 
 		if (this.pots.length > 0) {
 			const row = new SimpleRow({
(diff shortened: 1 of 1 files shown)


src/parser/core/modules/Medicated.ts
```diff
@@ -43,14 +43,20 @@ export class Medicated extends Analyser {
 			throw new Error('potion instance not found')
 		}
 
-		this.pots.push({end: event.timestamp, ...this.pot})
+		this.pots.push({ ...this.pot, end: event.timestamp })
```
ayyaruq: Why?

src/parser/core/modules/Medicated.ts
```diff
@@ -43,22 +43,28 @@ export class Medicated extends Analyser {
...
 
 		this.pot = undefined
 	}
 
 	private onComplete() {
 		const status = this.data.statuses.MEDICATED
 
+		// Add a pot that is still up before end of pull
+		if (this.pot && !this.pot.end) {
```
ayyaruq: We require explicit null checks rather than boolean coercion here, consider this instead:
```suggestion
		if (this.pot != null && this.pot.end == null) {
```

For legibility, might be better to make a stopAndSave method and call it in offMedication and here as well. Maybe also close out the window at the start of the onComplete function too.

src/parser/core/modules/Medicated.ts
```diff
@@ -43,22 +43,28 @@ export class Medicated extends Analyser {
...
+
 		if (this.pots.length > 0) {
 			const row = new SimpleRow({
 				label: 'Medicated',
 				order: -99,
 			})
 
 			this.pots.forEach(pot => row.addItem(new StatusItem({
-				status: status,
+				status,
```
ayyaruq: Also why?

ayyaruq (changes requested):
Thanks for your contribution! Not sure why some changes were made, main thing is to fix the boolean coercion.

src/parser/core/modules/Medicated.ts
mattmanuel90: Ahh was just a readability thing based on personal preference, but when spreading values its better for it to go on left hand side and assignments values on the right unless the developer intends to override the values with the spread. In this case, I assumed we never wanted to override the end value with the spread

ayyaruq: Yeah that's not a thing we need to do, it's just to merge the existing values rather than overriding anything.

mattmanuel90:
> Thanks for your contribution! Not sure why some changes were made, main thing is to fix the boolean coercion.

Thanks for the feedback! Sorry, some of my habits have crept in here where I do inline refactoring based on my judgement which is a bit rude on my part as it detracts from the PR's true changes. Will resolve the boolean coercion.

src/parser/core/modules/Medicated.ts
mattmanuel90: Sorry this one was pedantic. I think its easier to read a deferred object creation that's closer to where its used.

src/parser/core/modules/Medicated.ts
mattmanuel90: Ahh I referred to other examples of stopAndSave, but is this something you had in mind?
Some files rely on ```this.parser.currentEpochTimestamp``` as the end window time, but most seem to rely on ```this.parser.pull.timestamp + this.parser.pull.duration```. Guessing they'd produce the same value.

```
	private stopAndSave(endTime: number) {
		if (this.pot != null) {
			this.pots.push({ ...this.pot, end: endTime })
		}
		this.pot = undefined
	}

	private offMedication(event: Events['statusRemove']) {
	        // ...
		this.stopAndSave(event.timestamp);
	}

	private onComplete() {

		// Add a pot that is still up before end of pull
		if (this.pot != null && this.pot.end == null) {
			this.stopAndSave(this.parser.pull.timestamp + this.parser.pull.duration)
		}

               // ...
	}
```

ayyaruq: The old timestamp format stuff is just from it not being updated I guess. Epoch is the correct way. I'd default the `endTime` param to the current fight time (so you can call it without params if wanted), but basically yes that's the gist of it.

src/parser/core/modules/Medicated.ts
ayyaruq: I find it harder to read this way, destructuring isn't complex but it is arcane.

src/parser/core/modules/Medicated.ts
```diff
@@ -38,27 +38,36 @@ export class Medicated extends Analyser {
...
-
-		this.pots.push({end: event.timestamp, ...this.pot})
-
-		this.pot = undefined
+		this.stopAndSave(event.timestamp)
 	}
 
 	private onComplete() {
-		const status = this.data.statuses.MEDICATED
+
```
ayyaruq, line 59: Can prob lose the whitespace here.

src/parser/core/modules/Medicated.ts
```diff
@@ -38,27 +38,36 @@ export class Medicated extends Analyser {
 		this.pot = {start: event.timestamp}
 	}
 
+	private stopAndSave(endTime: number) {
```
ayyaruq: I'd set a default here to the current parser time.

src/parser/core/modules/Medicated.ts
```diff
@@ -38,27 +38,36 @@ export class Medicated extends Analyser {
...
+			this.stopAndSave(this.parser.pull.timestamp + this.parser.pull.duration)
+		}
 
 		if (this.pots.length > 0) {
 			const row = new SimpleRow({
 				label: 'Medicated',
 				order: -99,
 			})
 
+			const status = this.data.statuses.MEDICATED
```
ayyaruq: Thinking about it, this is easier to read if it's at the start, or otherwise inject it straight up where it's used.

src/parser/core/modules/Medicated.ts
mattmanuel90: Reverted. I've also added a newline between each block & statement.

src/parser/core/modules/Medicated.ts
mattmanuel90: Thanks for clarifying. Have opted to rely on Epoch as the default param. And also tidied up code to reflect true changes to source.

ayyaruq (approved):
Nicely done, thanks for your patience!

PR #1535, opened 2022-01-10
Shadowbringers branch: remove unused throw for unmatched effect packets

AzariahAeroborn:
See #1534 - same change for the Shadowbringers branch, to prevent unwanted Sentry errors.


ackwell (approved):
core support range is already set to 5.08 so 👍

PR #1545, opened 2022-01-14
Add RPR Enshroud and Arcane Circle windows.

KelosPelmand:
Enshroud is done via a standard BuffWindow.  Due to how relatively large the table will be and how relatively unlikely it is for people to need to look at it, I put it at the bottom.  A more compact view would be nice, but the alternate view I came up with had it's own problems, so in the interest of getting something up so we can get to supported, I used the standard BuffWindow.


hintxiv:
I think the evaluator is missing some rushing logic
image

ayyaruq:
Let's break this up into separate PRs for the new evaluators and BRD stuff. Up to you if you want each of the new evaluators in their own PR or not. Prob easier to cherry pick into new branches instead of removing from history here, so closing for now.

PR #1546, opened 2022-01-15
[BRD] Add song statuses, refactor ABC override, fix a bug with unclosed windows

hintxiv:


AzariahAeroborn (changes requested):
Needs a change log, but otherwise looks good.

AzariahAeroborn (approved):

PR #1547, opened 2022-01-15
fix: Invert checks for shield breaking

AzariahAeroborn:
This switches from looking for total amount absorbed to total amount remaining on a shield.  Currently, FFLogs is still showing 0 reamining for any shield that has been at least partially absorbed on new logs, but this gets us back to the state of things before the absorb event changes, and once they fix things for partial absorbs then new logs should work as expected.


ayyaruq (approved):
LGTM, is there a further pending change on the fflogs side you wanted to wait for or? I'm a bit unclear on what's going on there. Code is fine tho.

PR #1553, opened 2022-01-21
Add RPR Enshroud and Arcane Circle windows

KelosPelmand:
Enshroud is done via a standard BuffWindow. Due to how relatively large the table will be and how relatively unlikely it is for people to need to look at it, I put it at the bottom. A more compact view would be nice, but the alternate view I came up with had it's own problems, so in the interest of getting something up so we can get to supported, I used the standard BuffWindow.

This PR depends on #1552


hintxiv:
Looks good to me, I pulled it down and the buff windows behave as expected. Just needs conflicts fixed after the recent changes to master.

AzariahAeroborn (approved):

PR #1554, opened 2022-01-21
Update Ogi.tsx for proper naming convention.

kalittles:
Minor update to displayed text for using the proper number of Ogi Namikiris.


kalittles:
Accidently hit commit before the YARN extract finished, that's why the diff looks super weird. Should be fine now.

AzariahAeroborn (changes requested):

kalittles:
Thank you for the response!

I've made the requested changes, although I feel like I should definitely do some more in the future if being listed as a contributor is a requirement, haha. I just look at my XIVAnalysis a lot and kept seeing it, so I wanted to change it. I feel a little disingenuous being listed when all I did was fix a typo.

No actual logic was changed, so I didn't update /parser/SAMURAI/index.ts if that's okay.

src/parser/jobs/sam/index.js
```diff
@@ -26,6 +26,8 @@ export const SAMURAI = new Meta({
 
 	contributors: [
 		{user: CONTRIBUTORS.RYAN, role: ROLES.DEVELOPER},
+		{user: CONTRIBUTORS.ACKWELL},
```
AzariahAeroborn: This looks like an unintentional copy paste error.
```suggestion
```

AzariahAeroborn (changes requested):
Sorry for not being more specific with my original feedback, I need to update my template... if you would, please shrink the avatar to 100x100 px or smaller, preferably under 100KB in size.

In hindsight, I do agree that this small copy change is ok without a change log.

src/parser/jobs/sam/index.js
kalittles: Okay, I've made the changes.

And no problem! No need to apologize. I'm glad something else useful came from it! I appreciate you walking me through the process, and hopefully I can contribute in more meaningful ways in the future.

AzariahAeroborn (approved):

PR #1555, opened 2022-01-21
Reaper gauges

kennedyoliveira:


src/parser/jobs/rpr/modules/Gauge.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+	private soulGauge = this.add(new CounterGauge({
+		minimum: 0,
+		maximum: SOUL_MAX_VALUE,
+		graph: {
+			label: 'Soul Gauge',
+			color: SOUL_GAUGE_COLOR.fade(0.25),
+		},
+		correctHistory: true,
+		deterministic: false,
```
AkairyuGestalter: The `deterministic: false` flag is meant for gauges like DNC's Feathers where it's not guaranteed that a given action always produces a specific amount of that gauge. RPR's aren't like that, so you can skip setting that property and accept the default (it only affects the history correction code anyways)

src/parser/jobs/rpr/modules/Gauge.tsx
```diff
@@ -0,0 +1,183 @@
...
+const SHROUD_CONSUMERS: ActionKey[] = [
+	'ENSHROUD',
+]
+
+export class Gauge extends CoreGauge {
+	@dependency private suggestions!: Suggestions
+
+	/* eslint-disable @typescript-eslint/no-magic-numbers */
+	private soulGeneratorModifiers = new Map<number, number>([
+		[this.data.actions.SLICE.id, 10],
```
AkairyuGestalter: Since all of the generation/spend amounts are 10 or 50, you could probably define those as consts too, to avoid the no-magic-numbers override

src/parser/jobs/rpr/modules/Gauge.tsx
```diff
@@ -0,0 +1,183 @@
...
+			value: this.soulGauge.overCap,
+			why: <Trans id="rpr.gauge.suggestions.soul-gauge-overcap.why">
+				You lost {this.soulGauge.overCap} Soul due to overcapped gauge.
+			</Trans>,
+		}))
+
+		this.suggestions.add(new TieredSuggestion({
+			icon: this.data.actions.ENSHROUD.icon,
+			content: <Trans id="rpr.gauge.suggestions.shroud-gauge-overcap.content">
+				Try to not overcap Shroud gauge because it can lead to less usages of <DataLink action="ENSHROUD"/> over
```
AkairyuGestalter: Grammar note: `fewer uses` since this is a thing that can be counted

src/parser/jobs/rpr/modules/Gauge.tsx
```diff
@@ -0,0 +1,183 @@
...
+				the course of the fight.
+			</Trans>,
+			tiers: {
+				10: SEVERITY.MINOR,
+				30: SEVERITY.MEDIUM,
+				50: SEVERITY.MAJOR,
+			},
+			value: this.shroudGauge.overCap,
+			why: <Trans id="rpr.gauge.suggestions.shroud-gauge-overcap.why">
+				You lost {this.shroudGauge.overCap} Shroud due to overcapped gauge.
```
AkairyuGestalter: Food for thought: it might be more beneficial to users to translate this into lost uses of Enshroud, rather than a raw "how much gauge you lost" number. I've found that to be easier to understand. Won't block on this one if you'd rather keep it as-is though. Same deal with the soul gauge overcap suggestion too.

AkairyuGestalter (changes requested):
A couple coding style suggestions to take a look at, will check functionality later when I'm back at my computer.

hintxiv:
@kennedyoliveira hi, are you in the xivanalysis discord? Would be great to get these changes in asap since they're really important for RPR analysis. Let us know if there's anything you need help with.

src/parser/jobs/rpr/modules/Gauge.tsx
kennedyoliveira: fixed

src/parser/jobs/rpr/modules/Gauge.tsx
kennedyoliveira: fixed

src/parser/jobs/rpr/modules/Gauge.tsx
kennedyoliveira: thanks, English is not my first language, fixed.

src/parser/jobs/rpr/modules/Gauge.tsx
kennedyoliveira: Any suggestion on what would be said in case of not a single lost use, like you lost 40 shroud gauge?

ghost: Chiming in:
I personally would go for something more along the lines of
"You lost {this.shroudGauge.overCap} Shroud gauge due to overcapping, which cost you {this.shroudGauge.overCap / SHROUD_SPEND_AMOUNT} uses of Enshroud."

src/parser/jobs/rpr/modules/Gauge.tsx
```diff
@@ -0,0 +1,197 @@
...
+
+		this.addEventHook('complete', this.onComplete)
+	}
+
+	private onSoulGeneration(event: Events['action']) {
+		const generatedAmount = this.soulGeneratorModifiers.get(event.action)
+		if (generatedAmount) {
+			this.soulGauge.generate(generatedAmount)
+		}
+	}
```
ghost: For Soul Gauge generators and spenders, the amounts can be hardcoded.  
```suggestion
	private onSoulGeneration() {
            this.soulGauge.generate(BASE_GAUGE_GENERATION_AMOUNT)
	}
```

src/parser/jobs/rpr/modules/Gauge.tsx
```diff
@@ -0,0 +1,197 @@
...
+
+	private onSoulConsumption(event: Events['action']) {
+		const spentAmount = this.soulConsumptionModifiers.get(event.action)
+		if (spentAmount) {
+			this.soulGauge.spend(spentAmount)
+		}
+	}
```
ghost: same as above
```suggestion
	private onSoulConsumption(event: Events['action']) {
            this.SoulGauge.spend(BASE_GAUGE_CONSUMPTION_AMOUNT)
	}
```

src/parser/jobs/rpr/modules/Gauge.tsx
```diff
@@ -0,0 +1,197 @@
...
+	private shroudGeneratorModifiers = new Map<number, number>([
+		[this.data.actions.GIBBET.id, BASE_GAUGE_GENERATION_AMOUNT],
+		[this.data.actions.GALLOWS.id, BASE_GAUGE_GENERATION_AMOUNT],
+		[this.data.actions.GUILLOTINE.id, BASE_GAUGE_GENERATION_AMOUNT],
+		[this.data.actions.PLENTIFUL_HARVEST.id, PLENTIFUL_HARVEST_SHROUD_GENERATION_AMOUNT],
+	])
+
+	private shroudConsumerModifiers = new Map<number, number>([
+		[this.data.actions.ENSHROUD.id, ENSHROUD_GAUGE_CONSUMPTION],
+	])
```
ghost: I would personally replace all of this by adding a couple extra event hooks and altering your actionKey[]s, since there's a lot of repetition here, when what you have is mostly default behavior, with a few edge cases.

src/parser/jobs/rpr/modules/Gauge.tsx
```diff
@@ -0,0 +1,197 @@
...
+
+		const playerFilter = filter<Event>().source(this.parser.actor.id)
+
+		// soul gauge
+		this.addEventHook(playerFilter.type('action').action(this.data.matchActionId(SOUL_GENERATORS)), this.onSoulGeneration)
+		this.addEventHook(playerFilter.type('action').action(this.data.matchActionId(SOUL_CONSUMERS)), this.onSoulConsumption)
+
+		// shroud gauge
+		this.addEventHook(playerFilter.type('action').action(this.data.matchActionId(SHROUD_GENERATORS)), this.onShroudGeneration)
+		this.addEventHook(playerFilter.type('action').action(this.data.matchActionId(SHROUD_CONSUMERS)), this.onShroudConsumption)
```
ghost: I would set this up as:

- Hook for things that generate 10 soul gauge (default behavior)
- Hook for things that generate 50 soul gauge (edge case)
- Hook for consumers
// shroud gauge
- Hook for things that generate 10 shroud Gauge
- Hook for Plentiful Harvest (edge case)
- Hook for Enshroud

It's 2 extra hooks, but it'll make your code way cleaner than having all of the extra noise involved in accepting and handling the maps you're using right now.

src/parser/jobs/rpr/modules/Gauge.tsx
```diff
@@ -0,0 +1,197 @@
...
+				the course of the fight.
+			</Trans>,
+			tiers: {
+				10: SEVERITY.MINOR,
+				30: SEVERITY.MEDIUM,
+				50: SEVERITY.MAJOR,
+			},
+			value: this.shroudGauge.overCap,
+			why: <Trans id="rpr.gauge.suggestions.shroud-gauge-overcap.why">
+				You lost {this.shroudGauge.overCap} Shroud due to overcapped gauge.
```
ghost:
```suggestion
				You lost {this.shroudGauge.overCap} Shroud due to overcapped gauge, which cost you {this.shroudGauge.overCap / ENSHROUD_GAUGE_CONSUMPTION} uses of Enshroud.
```

src/parser/jobs/rpr/modules/Gauge.tsx
```diff
@@ -0,0 +1,197 @@
...
+
+const SOUL_GENERATORS: ActionKey[] = [
+	'SLICE',
+	'WAXING_SLICE',
+	'INFERNAL_SLICE',
+	'SPINNING_SCYTHE',
+	'NIGHTMARE_SCYTHE',
+	'SOUL_SLICE',
+	'SOUL_SCYTHE',
+]
```
ghost: I would spin Soul Slice and Soul Scythe out into their own ActionKey[], since it's usually better to handle defaults and edge cases, than to handle everything with special logic.

src/parser/jobs/rpr/modules/Gauge.tsx
```diff
@@ -0,0 +1,197 @@
...
+	'UNVEILED_GIBBET',
+	'UNVEILED_GALLOWS',
+]
+
+const SHROUD_GENERATORS: ActionKey[] = [
+	'GIBBET',
+	'GALLOWS',
+	'GUILLOTINE',
+	'PLENTIFUL_HARVEST',
+]
```
ghost: same with this and Plentiful Harvest, it could even just be on its own in a hook, since it's just the one action.

ghost (commented):
Saw some low-hanging fruit that I would personally change.

src/parser/jobs/rpr/modules/Gauge.tsx
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	private onComplete() {
+		this.suggestions.add(new TieredSuggestion({
+			icon: this.data.actions.SOUL_SLICE.icon,
+			content: <Trans id="rpr.gauge.suggestions.soul-gauge-overcap.content">
+				Try to not overcap Soul gauge since it's required for shroud generation for your highest damaging skills
+				via <DataLink action="ENSHROUD"/>.
+				Pay special attention when using <DataLink action="SOUL_SLICE"/> or <DataLink action="SOUL_SCYTHE"/> as
+				they increase the gauge by <b>50</b>, make sure you have <b>50</b> or less in your gauge before using.
```
eatmorehufu: You can use constants here instead of 50 (translations will need to be updated).
Probably SOUL_GAUGE_GENERATION_AMOUNT and SOUL_MAX_VALUE - SOUL_GAUGE_GENERATION_AMOUNT  respectively.

ghost:
lgtm (Looks good to me)
I especially like what you did with the callback names, they're much clearer now.

src/parser/jobs/rpr/modules/Gauge.tsx
```diff
@@ -0,0 +1,178 @@
...
+import {ActionKey} from 'data/ACTIONS'
+import {Event} from 'event'
+import {filter} from 'parser/core/filter'
+import {dependency} from 'parser/core/Injectable'
+import {CounterGauge, Gauge as CoreGauge} from 'parser/core/modules/Gauge'
+import Suggestions, {SEVERITY, TieredSuggestion} from 'parser/core/modules/Suggestions'
+import React from 'react'
+
+const SOUL_MAX_VALUE = 100
+const SHROUD_MAX_VALUE = 100
```
ayyaruq, line 13: Prefer to keep all the soul stuff together, then have a separate block for shroud, and slap a comment above them to indicate as such.

src/parser/jobs/rpr/modules/Gauge.tsx
```diff
@@ -0,0 +1,178 @@
...
+]
+
+/**
+ * Actions on cooldown that generate souls, those generate
+ * more souls than regular combo actions
+ */
+const SOUL_GENERATORS_CDS: ActionKey[] = [
+	'SOUL_SLICE',
+	'SOUL_SCYTHE',
+]
```
ayyaruq, line 33: There's no reason to keep these separate imo.

src/parser/jobs/rpr/modules/Gauge.tsx
```diff
@@ -0,0 +1,178 @@
...
+ * Base amount for the soul gauge consumption
+ */
+const BASE_GAUGE_CONSUMPTION_AMOUNT = 50
+const PLENTIFUL_HARVEST_SHROUD_GENERATION_AMOUNT = 50
+const ENSHROUD_GAUGE_CONSUMPTION = 50
+
+/**
+ * Amount generated for the "Soul" skills, which are cooldowns
+ */
+const SOUL_GAUGE_GENERATION_AMOUNT = 50
```
ayyaruq, line 68: Unnecessarily large comments, these can be reordered for sanity imo. Harvest is also clearly for shroud, so it doesn't really need it in the name. Not sure why Enshroud needs a special const either.

src/parser/jobs/rpr/modules/Gauge.tsx
```diff
@@ -0,0 +1,178 @@
...
+	override initialise() {
+		super.initialise()
+
+		const playerFilter = filter<Event>().source(this.parser.actor.id)
+
+		// soul gauge
+		this.addEventHook(playerFilter.type('action').action(this.data.matchActionId(SOUL_GENERATORS)), this.onSoulGeneration)
+		this.addEventHook(playerFilter.type('action').action(this.data.matchActionId(SOUL_GENERATORS_CDS)), this.onSoulGenerationCds)
+		this.addEventHook(playerFilter.type('action').action(this.data.matchActionId(SOUL_CONSUMERS)), this.onSoulConsumption)
+		this.addEventHook(playerFilter.type('action').action(this.data.actions.PLENTIFUL_HARVEST.id), this.onPlentifulHarvest)
```
ayyaruq, line 100: Consumption aside, there's no reason Plentiful or CDs need separate hooks here.

src/parser/jobs/rpr/modules/Gauge.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+		// shroud gauge
+		this.addEventHook(playerFilter.type('action').action(this.data.matchActionId(SHROUD_GENERATORS)), this.onShroudGeneration)
+		this.addEventHook(playerFilter.type('action').action(this.data.matchActionId(SHROUD_CONSUMERS)), this.onShroudConsumption)
+		this.addEventHook(playerFilter.type('action').action(this.data.actions.ENSHROUD.id), this.onEnshroud)
```
ayyaruq, line 105: Enshroud is also a consumer, doesn't need a separate hook. It's also already in your shroud consumers array, so this is doubling up.

src/parser/jobs/rpr/modules/Gauge.tsx
```diff
@@ -0,0 +1,178 @@
...
+		this.shroudGauge.generate(PLENTIFUL_HARVEST_SHROUD_GENERATION_AMOUNT)
+	}
+
+	private onEnshroud() {
+		this.shroudGauge.spend(ENSHROUD_GAUGE_CONSUMPTION)
+	}
+
+	private onShroudConsumption() {
+		this.shroudGauge.spend(BASE_GAUGE_CONSUMPTION_AMOUNT)
+	}
```
ayyaruq, line 136: These need to be restructured. Your base generation is the same, so realistically a switch for the +50 skills and defaulting to the +10 is fine. Consumers are all 50 as well, so they can effectively be a single hook using the action to determine which gauge.

src/parser/jobs/rpr/modules/Gauge.tsx
```diff
@@ -0,0 +1,178 @@
...
+	private onComplete() {
+		this.suggestions.add(new TieredSuggestion({
+			icon: this.data.actions.SOUL_SLICE.icon,
+			content: <Trans id="rpr.gauge.suggestions.soul-gauge-overcap.content">
+				Try to not overcap Soul gauge since it's required for shroud generation for your highest damaging skills
+				via <DataLink action="ENSHROUD"/>.
+				Pay special attention when using <DataLink action="SOUL_SLICE"/> or <DataLink action="SOUL_SCYTHE"/> as
+				they increase the gauge by <b>{SOUL_GAUGE_GENERATION_AMOUNT}</b>,
+				make sure you have <b>{SOUL_MAX_VALUE - SOUL_GAUGE_GENERATION_AMOUNT}</b> or
+				less in your gauge before using.
```
ayyaruq, line 147: Appreciate the formatting, but I think this could use a little bit less newlines for improved legibility of the code. Some consistency of capitalisation would be useful here too I think, and maybe make it sound a bit more natural. Happy to discuss this one in discord.

src/parser/jobs/rpr/modules/Gauge.tsx
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	private onShroudConsumption() {
+		this.shroudGauge.spend(BASE_GAUGE_CONSUMPTION_AMOUNT)
+	}
+
+	private onComplete() {
+		this.suggestions.add(new TieredSuggestion({
+			icon: this.data.actions.SOUL_SLICE.icon,
+			content: <Trans id="rpr.gauge.suggestions.soul-gauge-overcap.content">
```
ayyaruq, line 141: We already know it's a gauge, `soul-gauge-overcap` is extremely redundant. As your suggestions are only for overcap, `rpr.gauge.suggestions.soul.content` is more than adequate. Same goes for the other i18n keys.

ayyaruq (changes requested):
Could use a lot of structural cleanup, I think this is extremely noisy for what it's doing. Each hook has an overhead to it so we try to avoid unnecessary hooks. The Enshroud hook in particular is duplicating another hook already, leading to incorrect values for gauge consumption. The other big thing is the suggestion texts, they could use a grammar pass.

PR #1558, opened 2022-01-22
Add tracking for WAR's Nascent Chaos status timeouts/overwrites

ayyaruq:
This required a type fix for core Procs as well, so that's been rolled in. Changelog entries added for both WAR and core.


src/parser/jobs/war/modules/Infuriate.tsx
```diff
@@ -0,0 +1,57 @@
...
+			() => this.cooldowns.reduce('INFURIATE', INFURIATE_CDR),
+		)
+	}
+
+	override trackedProcs = [{
+		procStatus: this.data.statuses.NASCENT_CHAOS,
+		consumeActions: [this.data.actions.INNER_CHAOS, this.data.actions.CHAOTIC_CYCLONE],
+	}]
+
+	override droppedProcIcon = this.data.actions.INFURIATE.icon
```
AkairyuGestalter, line 51: You'll need to set `override show[Dropped/Overwrote]ProcSuggestions = true` if you want these to display. All of the default suggestions in core are hidden by default.

src/parser/jobs/war/modules/Infuriate.tsx
```diff
@@ -0,0 +1,57 @@
...
+	override droppedProcSeverityTiers = INFURIATE_SEVERITY
+	override droppedProcContent = <Trans id="war.infuriate.suggestions.dropped.content">Avoid letting <DataLink status="NASCENT_CHAOS"/> fall off without using it.</Trans>
```
AkairyuGestalter: The 'dropped nascent chaos' suggestion is duplicating one that's in `PrimalChaos.tsx`. Should the two files be merged? PrimalChaos looks like it's also essentially doing proc tracking for Primal Rend

AkairyuGestalter (dismissed):
Couple things. One minor, one that might need some thinks

src/parser/jobs/war/modules/Infuriate.tsx
ayyaruq: I'm gonna revert the changes to Infuriate, and port PrimalChaos to core Procs.

ayyaruq: Actually I lied, prefer to split these outright.

AkairyuGestalter (approved):

PR #1559, opened 2022-01-22
Harvest moon

ghost:
completes task "Harvest Moon charging" in #1372 

adds:
checklist item tracking number of Harvest Moon uses compared to forced downtime longer than Soulsow cast.


Rjb99:
This might be too naive of a check tbh. Soul sow is a 5 second channel and not all downtime will allow for a full channel even if the downtime is longer than 6 seconds.

Movement during downtime is still a possible requirement. 
P2S limit cut and p3s pairs come to mind as long enough, but I need to move sometimes based on my party members brain cell count to stay alive.

ghost:
My question on that is two-fold:
1) is it worth making the module fight gnostic just to track movement and make sure there's 5 seconds of time where the player can stand still? Also, would that even be possible with the data we have access to?
and
2) even then, is it still optimal to try to optimize movement in order to gain a soulsow channel where one would normally not be possible?

I know on P2S, when I was grabbing random purple logs to test the code, I was always getting people who had optimized in a soulsow cast during Kampeos Harma. I also was having similar experiences with grabbing random purple P3S logs, where people were getting 4 Harvest Moons in the fight, which matched with the number of expected uses calculated by my code.

EDIT:
I guess my worry though is what happens if we get another fight like E12S, where there's significant downtime that contains movement that cannot be optimized out.

Rjb99:
1) Probably not. I don't think movement is something we could easily track without manually doing it.

2) My examples are possible, and I have done them, but like I said, require a 3rd party to adjust.
p2s for example: Purple/head = easy soul. Blue/body? I'm forced to depend on my partner to move around me during the channel.  I tend to stick to the philosophy of "if the other 7 people have -2 brain cells, can you still do it?" for my code and analysis on Samurai. But don't let my skepticism stop you from trying to track soulsow/harvest moon, as it is an important thing to do on Reaper.

Another thing that pops to mind, does this take into account downtime with forced inactive time? I.e: zodiark normals table spin is long enough for a channel, but it locks your character up. Would events like that cause a false positive?

ghost:
It doesn't currently track time where you as a player are unable to take any actions, no. Granted, that wouldn't be terribly difficult to add, so I might as well add it to scope.  
RE: zodiark, that fight uses Bind, rather than DFTC or Stun, so it's an uptime window, because the boss doesn't go untargetable either, IIRC. (that might just be EX though.)

Currently, the code is set to expect the player to use soulsow before the pull, and then every time the boss goes either invincible (e.g. Titan Hard), or untargetable (e.g. Advanced Relativity).
Granted, with all of that, I think that having it be "If you at least used your pre-pull soulsow, you're doing acceptably, but there's space for optimization" in the form of a warn probably will end up working fine. The only way to get a Fail on this module is to never use harvest moon at all.

ghost:
@Rjb99 could you take a look at 0315ca1 and let me know what you think?

Rjb99:
Looks like it'll cover those downtimes without being able to act now, ltgm

ghost:
@Rjb99 Are there any changes that need to be made to this module at this point in time?

If not, I'd like to request a full review, if you have time.

ghost:
UPDATE: [6.08] No balance changes made to RPR, PR can continue unaltered.

src/parser/jobs/rpr/modules/HarvestMoon.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+const SOULSOW_CAST_TIME = 5000 //milliseconds
+const SOULSOW_CAST_GRACE = 1000
+const USAGE_PERCENTAGE_SEVERITY = {
+	0: TARGET.FAIL,
+	100: TARGET.SUCCESS,
+}
+
+export class HarvestMoon extends Analyser {
+	static override handle = 'HarvestMoon'
```
ayyaruq: Alternatively lowercase the whole thing.
```suggestion
	static override handle = 'harvestMoon'
```

src/parser/jobs/rpr/modules/HarvestMoon.tsx
```diff
@@ -0,0 +1,79 @@
...
+	@dependency private invulnerability!: Invulnerability
+	@dependency private unableToAct!: UnableToAct
+
+	private harvestMoonCasts = 0
+
+	override initialise() {
+		super.initialise()
+
+		this.addEventHook(filter<Event>().source(this.parser.actor.id).type('action')
+			.action(this.data.actions.HARVEST_MOON.id), this.onHarvestMoon)
```
ayyaruq: Hopefully GH doesn't mangle indentation characters here. If you're gonna use multiline filters, go all in. Also, slap a newline before the complete hook.
```suggestion
		this.addEventHook(
			filter<Event>()
				.source(this.parser.actor.id)
				.type('action')
				.action(this.data.actions.HARVEST_MOON.id),
			this.onHarvestMoon,
		)
```

src/parser/jobs/rpr/modules/HarvestMoon.tsx
```diff
@@ -0,0 +1,79 @@
...
+		this.addEventHook('complete', this.onComplete)
+	}
+
+	private onHarvestMoon() {
+		this.harvestMoonCasts++
+	}
```
ayyaruq: Honestly you could inline this. I'd prob shorten the name to just `moons` or something too.

src/parser/jobs/rpr/modules/HarvestMoon.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+	private getValidatedInvulnWindows() {
```
ayyaruq: If you're returning stuff, type the return. Can kill some of the redundant braces too.

src/parser/jobs/rpr/modules/HarvestMoon.tsx
```diff
@@ -0,0 +1,79 @@
...
+		const hmUsesPercent = (this.harvestMoonCasts / expectedUses * 100).toFixed(2)
+		this.checklist.add(new TieredRule({
+			name: <Trans id="rpr.harvest-moon.checklist.title">
+				Use <DataLink action="SOULSOW" /> during downtime.
+			</Trans>,
+			description: <Trans id="rpr.harvest-moon.checklist.description">
+				Harvest Moon is one of your highest potency abilities.
+				Aim to get at least one use each time the boss is targetable.
+			</Trans>,
+			tiers: USAGE_PERCENTAGE_SEVERITY,
```
ayyaruq: Your severity is effectively pointless here. Since you want success to be 100%, anything below that is an auto fail, so you can inline here with just the success target and remove the tiered rule. By setting fail to 0, you're telling users they won't ever fail unless they don't use it at all.

src/parser/jobs/rpr/modules/HarvestMoon.tsx
```diff
@@ -0,0 +1,79 @@
...
+import {Invulnerability} from 'parser/core/modules/Invulnerability'
+import {UnableToAct} from 'parser/core/modules/UnableToAct'
+import React from 'react'
+
+const SOULSOW_CAST_TIME = 5000 //milliseconds
+const SOULSOW_CAST_GRACE = 1000
+const USAGE_PERCENTAGE_SEVERITY = {
+	0: TARGET.FAIL,
+	100: TARGET.SUCCESS,
+}
```
ayyaruq: I think this can use some whitespace but the cast time doesn't need to be a const and the severity as noted on the checklist is doing nothing. I'd prob rename the grace const to like buffer or something tho.

ayyaruq (changes requested):
Couple of changes, notably adding the cast time to the soulsow data object. Also, while I prefer `harvestmoon` for the handle/i18n, You're currently mixing and matching `harvest-moon` and `harvestmoon`. I don't really care either way as long as you settle on one for consistency.

src/parser/jobs/rpr/modules/HarvestMoon.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+	override initialise() {
+		super.initialise()
+
+		this.addEventHook(
+			filter<Event>()
+				.source(this.parser.actor.id)
+				.type('action')
+				.action(this.data.actions.HARVEST_MOON.id),
+			() => { this.harvestMoonCasts++ })
```
ayyaruq: The curlies are unnecessary.

src/parser/jobs/rpr/modules/HarvestMoon.tsx
```diff
@@ -0,0 +1,84 @@
...
+import {Analyser} from 'parser/core/Analyser'
+import {filter} from 'parser/core/filter'
+import {dependency} from 'parser/core/Injectable'
+import Checklist, {Requirement, Rule} from 'parser/core/modules/Checklist'
+import {Data} from 'parser/core/modules/Data'
+import {Invulnerability} from 'parser/core/modules/Invulnerability'
+import {UnableToAct} from 'parser/core/modules/UnableToAct'
+import React from 'react'
+
+const BUFFER = 1000
```
ayyaruq: Context for clarity.
```suggestion
const SOULSOW_BUFFER = 1000
```

src/parser/jobs/rpr/modules/HarvestMoon.tsx
```diff
@@ -0,0 +1,84 @@
...
+	private getExpectedUses(): number {
+		const ADJUSTED_CAST = this.data.actions.SOULSOW.castTime + BUFFER
+		const invulnWindows = this.invulnerability.getWindows().filter((w) => { return w.end - w.start >=  ADJUSTED_CAST })
+
+		if (this.unableToAct.getDuration() > 0) {
+			return invulnWindows.filter((w) => {
+				const firstWindow = this.unableToAct.getWindows({start: w.start, end: w.end})[0]
+
+				return firstWindow == null ? false : w.start - firstWindow.start >= ADJUSTED_CAST || firstWindow.end - w.end >= ADJUSTED_CAST
+			}).length + 1
```
ayyaruq: This is kinda awkward to read, might be worth linebreaking it a little. You can keep the ternary tho. I thought the logic here needed updating for something?

src/parser/jobs/rpr/modules/HarvestMoon.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+		this.checklist.add(new Rule({
+			name: <Trans id="rpr.harvestmoon.checklist.title">
+				Use <DataLink action="SOULSOW" /> during downtime.
+			</Trans>,
+			description: <Trans id="rpr.harvestmoon.checklist.description">
+				Harvest Moon is one of your highest potency abilities.
+				Aim to get at least one use each time the boss is targetable.
+			</Trans>,
+
```
ayyaruq:
```suggestion
```

src/parser/jobs/rpr/modules/HarvestMoon.tsx
```diff
@@ -0,0 +1,84 @@
...
+						<DataLink action="HARVEST_MOON" /> uses
+					</Trans>,
+					overrideDisplay: `${this.harvestMoonCasts} / ${expectedUses} (${hmUsesPercent}%)`,
+					percent: hmUsesPercent,
+				}),
+			],
+
+			target: 100,
+
+		}))
```
ayyaruq:
```suggestion
			],
			target: 100,
		}))
```

ayyaruq (commented):
While I appreciate the whitespace, it's good to keep logically grouped things together. As discussed in discord, I thought there was an issue with the first window logic checks?

src/parser/jobs/rpr/modules/HarvestMoon.tsx
```diff
@@ -0,0 +1,84 @@
...
+		const expectedUses = this.getExpectedUses()
+		const hmUsesPercent = this.getUsedPercentage(expectedUses)
+
+		this.checklist.add(new Rule({
+			name: <Trans id="rpr.harvestmoon.checklist.title">
+				Use <DataLink action="SOULSOW" /> during downtime.
+			</Trans>,
+			description: <Trans id="rpr.harvestmoon.checklist.description">
+				Harvest Moon is one of your highest potency abilities.
+				Aim to get at least one use each time the boss is targetable.
```
ayyaruq:
```suggestion
				Harvest Moon is one of your highest potency abilities. Aim to get at least one use each time the boss is targetable.
```

src/parser/jobs/rpr/modules/HarvestMoon.tsx
```diff
@@ -0,0 +1,84 @@
...
+		const invulnWindows = this.invulnerability.getWindows().filter((w) => w.end - w.start >=  ADJUSTED_CAST)
+
+		if (this.unableToAct.getDuration() > 0) {
+			return invulnWindows.filter((invulnWindow) => {
+				const unableToActWindow = this.unableToAct.getWindows({start: invulnWindow.start, end: invulnWindow.end})[0]
+
+				return unableToActWindow == null
+					?false
+					:unableToActWindow.start - invulnWindow.start >= ADJUSTED_CAST
+						||unableToActWindow.end - invulnWindow.end >= ADJUSTED_CAST
```
ayyaruq: Is there a way to structure this better? Whitespace aside, the logic here is difficult to follow the way it's written.

ayyaruq (changes requested):
Minor fixes, otherwise looks good. Need to sort out the merge conflicts tho, let me know if you need assistance with this.

src/parser/jobs/rpr/modules/HarvestMoon.tsx
ghost: Honestly, I'm not sure. As is, the way I see the problem, there are two cases where you should be using HM during downtime, and a bunch where you shouldn't, so hard coding those two cases (5s before a UTA, or 5s after a UTA) is the way that makes sense to me
That said, I'd be happy to discuss it further in discord, brainstorm ideas to figure out a cleaner way to do it.

ghost:
I'll handle the merge to get the upstream changes dealt with btw

src/parser/jobs/rpr/modules/HarvestMoon.tsx
ghost: Regarding readability, I see that I made a minor mistake, whitespace-wise, with following the style of Analyser.ts's ternaries, so I'll fix that.
Do you have any other ideas regarding ways that this logic could be made clearer?

ghost:
words cannot describe how happy I am that we squash commit history from PRs.

src/parser/jobs/rpr/modules/HarvestMoon.tsx
dthomash: I ran into something similar when calculating drift in the Aetherflow module for Scholar. Using Aetherrflow and the core CDDT module as inspiration you could maybe do something like:

```
getExpectedUses(): number {
  let currentTimestamp = this.parser.pull.start
  let expectedUses = 0
  const ADJUSTED_CAST = this.data.actions.SOULSOW.castTime + SOULSOW_BUFFER
  const pullEndTimestamp = this.parser.pull.start + this.parser.pull.duration
  
  while (currentTimestamp < pullEndTimestamp) {
    expectedUses++
  
    currentTimestamp = currentTimestamp + ADJUSTED_CAST
  
    // This is example logic, I'm not super familiar with the edge cases so it's probably more complicated than this
    // The idea would be to adjust current timestamp to wherever the expected cast should be based on downtime edge cases
    if (this.unableToAct.isUnableToAct({ timestamp: currentTimestamp})) {
      const unableToActWindow = this.unableToAct.getWindows({start: currentTimestamp, end: currentTimestamp})[0]
      currentTimestamp = unableToActWindow.end + SOULSOW_BUFFER
    }
  }

  return expectedUses
}
```

Might need a little bit of sprucing up to get it to compile and cover any edge cases I missed.

ghost: I'll do some testing

src/parser/jobs/rpr/modules/HarvestMoon.tsx
```diff
@@ -0,0 +1,90 @@
...
+			filter<Event>()
+				.source(this.parser.actor.id)
+				.type('action')
+				.action(this.data.actions.HARVEST_MOON.id),
+			() => this.harvestMoonCasts++)
+
+		this.addEventHook('complete', this.onComplete)
+	}
+
+	private windowFilterCallback(inputWindow: {start: number, end: number}): boolean {
```
ayyaruq: I'd like to rename this if we can, something more self-explanatory.

src/parser/jobs/rpr/modules/HarvestMoon.tsx
```diff
@@ -0,0 +1,90 @@
...
+		const unableToActWindow = this.unableToAct.getWindows(inputWindow)[0]
+
+		if (unableToActWindow == null) {
+			return false
+		}
+
+		if (unableToActWindow.start - inputWindow.start >= this.data.actions.SOULSOW.castTime + SOULSOW_BUFFER) {
+			return true
+		}
```
ayyaruq, line 47: This is used a few times, may just wanna const the cast time with buffer. You can also invert this so it can be a single conditional with the previous one.
```suggestion
		if ((unableToActWindow == null || unableToActWindow.start - inputWindow.start < this.data.actions.SOULSOW.castTime + SOULSOW_BUFFER) {
			return false
		}
```

src/parser/jobs/rpr/modules/HarvestMoon.tsx
```diff
@@ -0,0 +1,90 @@
...
+		return invulnWindows.length + 1
+	}
+
+	private getUsedPercentage(expectedUses: number): string {
+		return (this.harvestMoonCasts / expectedUses * 100).toFixed(2)
+	}
+
+	private onComplete() {
+		const expectedUses = this.getExpectedUses()
+		const hmUsesPercent = this.getUsedPercentage(expectedUses)
```
ayyaruq, line 69: Expected uses is only used here, could probably change the function to just return the used percent for you outright.

ayyaruq (changes requested):
Requesting the `windowFilterCallback` function be renamed, the rest is just notes and we can resolve later.

src/parser/jobs/rpr/modules/HarvestMoon.tsx
ghost: the value is used on line 83 as well, for the display

ayyaruq: Ah you're right, my bad then.

ayyaruq (approved):

PR #1561, opened 2022-01-23
DRG: Update Buff Windows to track oGCDs

ebshimizu:
DRG optimization is all about stuffing every possible ~~bit of spaghetti~~ oGCD into buff windows (and only buff windows) so the Lance Charge and Dragon Sight modules have been adjusted accordingly.

Couple things to note here, starting with a change to core:
- A `DisplayedActionsEvaluator` was added to show Dragonfire Dive without requiring a use in a window (we expect one every other window, see below)

DRG specific things:
- oGCDs tracked are: High Jump, Gierskogul, Dragonfire Dive, Mirage Dive, Life Surge, Spineshatter Dive
- Severity tiers are set at 1/6/12 for LC and 1/3/6 for DS (DS cd is twice as long as LC). These are somewhat lenient, though we could arguably make them even more lenient (or not have a major tier?), as this is supposed to be optimized for after hitting them on CD.
- Dragonfire Dive should be used every other Lance Charge window, however trying to detect whether or not a window should have had a DFD has a number of corner cases that are difficult to resolve, so we opt to show it without penalty.

Lance Charge is special and has an expected action adjustment:
- Spineshatter Dive: Technically it's optimal to hold both charges for the two minute buff windows, so we don't want to yell at people for (correctly) holding charges. During analysis, we mark windows where the player used 0 SSDs and had at least one charge left. We then mark the next window as expecting two SSD uses. If both charges are used during a window where two uses are expected, we mark the prior window as a successful hold.

that's a lot of text but it's really not that much code i'm just documenting everything
also i fixed a changelog timestamp (drg be time traveling)


ebshimizu:
note: this PR is unaffected by 6.08

ebshimizu:
flipping this to draft while actively working on stuff based on discord discussions

ebshimizu:
ok putting this back up. took out linked list lookahead/lookback, put in some other logic just in lance charge. added a new evaluator for DFD (show actions, no expected uses)

AzariahAeroborn (approved):

PR #1562, opened 2022-01-23
[MCH] Harden Reassemble detection

hintxiv:
Fixes a bug where a reassemble was flagged as dropped because its statusRemove timestamp is ~1 millisecond later than the GCD cast timestamp.


AzariahAeroborn (changes requested):
This change should include a changelog.

AzariahAeroborn (approved):

PR #1563, opened 2022-01-24
[SCH] Aetherflow Module

dthomash:
Initial implementation of the Aetherflow module for Endwalker. Changes from Shadowbringers:

- Changed
  - Summary data moved from bottom row to it's own summary table at the top
  - Cleaned up how the cast times were displayed on the table
- Added
  - Total cast count of each Aetherflow ability add to summary table
  - Link to timeline added to cast times (shows from 5 seconds before the corresponding Aetherflow or Dissipation cast to 65 seconds after
    - This time window was picked because the cooldown on Aetherflow is 60 seconds


ghost (commented):
lgtm
I don't have commit permission, but ye

src/parser/jobs/sch/modules/Aetherflow.tsx
```diff
@@ -0,0 +1,231 @@
...
+import {Data} from 'parser/core/modules/Data'
+import {Timeline} from 'parser/core/modules/Timeline'
+import React from 'react'
+import {Button, Grid, Table} from 'semantic-ui-react'
+
+export class Aetherflow extends Analyser {
+	static override handle = 'aetherflow'
+	static override title = t('sch.aetherflow.title')`Aetherflow`
+
+	private recticationActive: boolean = false
```
AzariahAeroborn:
```suggestion
	private recitationActive: boolean = false
```

src/parser/jobs/sch/modules/Aetherflow.tsx
```diff
@@ -0,0 +1,231 @@
...
+		const end = timestamp - this.parser.pull.timestamp + this.AETHERFLOW_TIMELINE_END_PADDING
+		this.timeline.show(start, end)
+	}
+}
+
+interface AetherflowWindow {
+	aetherflowGenerateActionId: number
+	timestamp: number
+	drift: number
+	cummulativeDrift: number
```
AzariahAeroborn:
```suggestion
	cumulativeDrift: number
```

src/parser/jobs/sch/modules/Aetherflow.tsx
```diff
@@ -0,0 +1,231 @@
...
+	private onGenerateAetherflow(event: Events['action']) {
+		// Calculate any values that depend on the previous aetherflow window
+		const prevWindow = event.action === this.data.actions.AETHERFLOW.id ? this.prevAetherflowWindow : this.prevDissipationWindow
+		const cooldown = this.data.getAction(event.action)?.cooldown ?? this.data.actions.AETHERFLOW.cooldown
+		let drift = 0
+		let cummulativeDrift = 0
+		if (this.aetherflowWindows.length > 1 && prevWindow) {
+			drift = event.timestamp - prevWindow.timestamp - cooldown
+			cummulativeDrift = prevWindow.cummulativeDrift + drift
+		}
```
AzariahAeroborn: Should this calculation take into account Unable To Act downtime (e.g. where the player is incapacitated by a fight transition)?  The downtime module has a way to look up specifically windows where the player can't do anything (as opposed to just downtime where no enemies are targetable) so we don't count that drift time against the player.

Also, you could simplify your object tracking by simply doing all of the drift calculations at the end during the onComplete method, rather than needing to keep a hold of the last Aetherflow / Dissipation window as you go and then muddy that window information with a cumulative drift property that isn't really a part of just that specific window.

AzariahAeroborn (changes requested):
A couple of spelling nitpicks for property names, but I didn't try to find every usage and make suggestions, so you'll probably want to rename with your editor rather than applying the suggestion directly.

My main concern is making sure we have appropriate consideration for downtime forgiveness - considering Unable To Act time is a minimum, but we might also want to consider if we should forgive wasted stacks or drift if there is a longer period of untargetable/invulnerable downtime -- if AF comes off cooldown during the post-adds transition for Zodiark or Hydaelyn, but the user has 1 or more stacks left, what should the expected behavior be?

src/parser/jobs/sch/modules/Aetherflow.tsx
dthomash: Agreed on both counts.

To address your concern about what to do about wasted stacks around transition windows: A scholar needs a maximum of 10s (if they have no spell speed) to cast Aetherflow and spend all 3 stacks. Since Aetherflow's cooldown is 60s you have at least 50s of wiggle room to figure out when you want to cast your stacks. So any scripted downtime would have to last more than 50s to force the scholar into the situation where they have aetherflow stacks left and Aetherflow is off cooldown. The longest scripted downtime I saw was 20-25s, which leaves the scholar with 25-30s of wiggle room to figure out when to use their aetherflow stacks if it's going to come off cooldown during the scripted downtime. That gives the scholar at least 10-12 weave windows to spend their atherflow charges. To me that seems like more than enough time to spend aetherflow stacks to not be caught in a situation where you need to pick between drift and wasting stacks.

AzariahAeroborn: Longer scripted downtimes aren't uncommon (the Eden Prime laser transition, Shiva/Ryne Savage after adds, many of the TEA phase transitions). The TEA ones specifically aren't UTA downtime -- while we don't explicitly worry about supporting ultimates, if there is relatively simple logic that could handle forgiveness for long untargetable downtime, it would be nice to have the module handle that to avoid the per fight weirdness. If it's a real mess then at least handling drift during a UTA downtime should be ok.

dthomash: That seems fair. Something else I noticed too was that scholar can cast Aetherflow without a target. So it depends on if the downtime is a mini cutscene or if the boss is just untargetable.  So I'll crunch some more numbers and figure out a good minimum UTA downtime duration where any UTA downtime starts forgiving drifting the cooldown. Probably something around 50-60s where it wouldn't be unreasonable that a scholar not drifting their cooldowns might end up caught with extra stacks of aetherflow during a transition.

dthomash: So upon further analysis I came up with 12 different edge cases based on whether or not the downtime was UTA vs invuln and a bunch of different timings. But the logic to handle all of that would have been, in my opinion, rather cumbersome and hard to test and maintain. So I opted for a simple check that moves the desired cast time to the end of a UTA window and gives the scholar a bit of buffer to use any remaining aetherflow stacks.

I used this log to check that I was bumping up the expected cast time correctly: https://www.fflogs.com/reports/1mBNCMdhpJ3kALgX#fight=29&type=casts&source=3

src/parser/jobs/sch/modules/Aetherflow.tsx
```diff
@@ -0,0 +1,258 @@
...
+
+		// Calculate Dissipation drift
+		const dissipationWindows = this.aetherflowWindows
+			.filter(window => window.aetherflowGenerateActionId === this.data.actions.DISSIPATION.id)
+		dissipationWindows.forEach((currentWindow, index) => {
+			const previousWindow = dissipationWindows[index-1]
+			if (!previousWindow) {
+				currentWindow.drift = 0
+			} else {
+				currentWindow.drift = this.calculateDrift(previousWindow.timestamp, currentWindow.timestamp)
```
AzariahAeroborn: Since dissipation's cooldown is longer than aetherflow's, this is going to result in an incorrect drift calculation for dissipation.  Maybe change the calculateDrift function to expect a "skillCooldown" parameter and pass in as appropriate?

AzariahAeroborn (changes requested):
Looks good, I like the organization of the calculateDrift function and the downtime handling.  Just need to cleanup how downtime is referenced for the two skills and then we should be good to go.

src/parser/jobs/sch/modules/Aetherflow.tsx
dthomash: Yeah, I whiffed on that one. Forgot to double check the dissipation drift times. Good catch.

AzariahAeroborn (approved):

PR #1578, opened 2022-01-26
Update Faerie-Gauge to CoreGauge and add 6.0 Support

ghost:
Pree much what it says on the tin.


src/parser/jobs/sch/modules/Gauge.tsx
```diff
@@ -0,0 +1,98 @@
...
+		const playerFilter = filter<Event>().source(this.parser.actor.id)
+
+		//Faerie Summon
+		this.addEventHook(playerFilter.type('action').action(oneOf([this.data.actions.SUMMON_EOS.id, this.data.actions.SUMMON_SELENE.id])), this.onSummon)
+		//sanity check, your pets can't take actions if they're not out.
+		this.addEventHook(filter<Event>().source(oneOf(this.actorPets)), this.onSummon)
+		//Consumers
+		this.addEventHook(filter<Event>().type('heal').source(oneOf(this.actorPets))
+			.cause(this.data.matchCauseStatusId([this.data.statuses.FEY_UNION.id])), this.onGaugeSpend)
+		this.addEventHook(playerFilter.type('action').action(this.data.actions.SCH_FEY_BLESSING.id), this.onGaugeSpend)
```
AzariahAeroborn: Fey Blessing no longer consumes gauge as of 6.0

src/parser/jobs/sch/modules/Gauge.tsx
```diff
@@ -0,0 +1,98 @@
...
+		super.initialise()
+		const playerFilter = filter<Event>().source(this.parser.actor.id)
+
+		//Faerie Summon
+		this.addEventHook(playerFilter.type('action').action(oneOf([this.data.actions.SUMMON_EOS.id, this.data.actions.SUMMON_SELENE.id])), this.onSummon)
+		//sanity check, your pets can't take actions if they're not out.
+		this.addEventHook(filter<Event>().source(oneOf(this.actorPets)), this.onSummon)
+		//Consumers
+		this.addEventHook(filter<Event>().type('heal').source(oneOf(this.actorPets))
+			.cause(this.data.matchCauseStatusId([this.data.statuses.FEY_UNION.id])), this.onGaugeSpend)
```
AzariahAeroborn:
```suggestion
		this.addEventHook(
			filter<Event>()
				.type('heal')
				.source(oneOf(this.actorPets))
				.cause(this.data.matchCauseStatus(['FEY_UNION'])),
			this.onGaugeSpend
		)
```

For complicated filters, we prefer multi-line formatting.  Also, you can simplify the matchCauseStatus call by passing the status key directly rather than having to call this.data.statuses, but I won't block on the latter if you prefer that filter as-is.

src/parser/jobs/sch/modules/Gauge.tsx
```diff
@@ -0,0 +1,98 @@
...
+	private actorPets = this.parser.pull.actors.filter(actor => actor.owner != null && actor.owner.id === this.parser.actor.id).map(pet => pet.id)
+
+	override initialise() {
+		super.initialise()
+		const playerFilter = filter<Event>().source(this.parser.actor.id)
+
+		//Faerie Summon
+		this.addEventHook(playerFilter.type('action').action(oneOf([this.data.actions.SUMMON_EOS.id, this.data.actions.SUMMON_SELENE.id])), this.onSummon)
+		//sanity check, your pets can't take actions if they're not out.
+		this.addEventHook(filter<Event>().source(oneOf(this.actorPets)), this.onSummon)
```
AzariahAeroborn: This is going to hook every pet action for the entire duration of the fight and run the onSummon check.  Instead, you should save this hook to a property, and unset this hook as soon as we have marked the fairy as active within the pull.

src/parser/jobs/sch/modules/Gauge.tsx
```diff
@@ -0,0 +1,98 @@
...
+			this.data.actions.LUSTRATE.id,
+			this.data.actions.INDOMITABILITY.id,
+			this.data.actions.SACRED_SOIL.id,
+			this.data.actions.EXCOGITATION.id,
+		])), this.onGaugeGenerate)
+		//Death
+		this.addEventHook('death', () => this.fairyOut = false)
+		//Dissipation
+		this.addEventHook(playerFilter.type('statusApply').status(this.data.statuses.DISSIPATION.id), () => this.dissipationActive = true)
+		this.addEventHook(playerFilter.type('statusRemove').status(this.data.statuses.DISSIPATION.id), () => this.dissipationActive = false)
```
AzariahAeroborn: You could potentially simplify the module's interface a little bit and streamline the logic for gauge generation if you have dissipationApply -> this.fairyOut = false and dissipationRemove -> this.fairyOut = true

I won't block on this either way, if you prefer the current logic that's fine with me.

AzariahAeroborn (changes requested):

src/parser/jobs/sch/modules/Gauge.tsx
```diff
@@ -0,0 +1,105 @@
...
+	static override title = t('sch.gauge.title')`Faerie Gauge Usage`
+
+	@dependency private suggestions!: Suggestions
+	// Defaults
+	private gauge = this.add(new CounterGauge({
+		chart: {label: 'Faerie Gauge', color: GRAPH_COLOR},
+	}))
+	private fairyOut: boolean = false
+	private actorPets = this.parser.pull.actors.filter(actor => actor.owner != null && actor.owner.id === this.parser.actor.id).map(pet => pet.id)
+	private petHook?: EventHook<{ type: 'action'; action: number; source: string; target: string; timestamp: number; }>
```
AzariahAeroborn:
```suggestion
	private petHook?: EventHook<Event>
```

You should be able to just use type Event to match the type being filtered on line 44.

src/parser/jobs/sch/modules/Gauge.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+	override initialise() {
+		super.initialise()
+		const playerFilter = filter<Event>().source(this.parser.actor.id)
+
+		//Faerie Summon
+		this.petHook = this.addEventHook(playerFilter.type('action').action(oneOf([this.data.actions.SUMMON_EOS.id, this.data.actions.SUMMON_SELENE.id])), this.onSummon)
```
AzariahAeroborn: You added this to the wrong hook - we always need to check for re-summons (e.g. after death), but the sanity check for marking the pet out if it was out pre-pull (line 44) should only run once / until we see the pet is out.

src/parser/jobs/sch/modules/Gauge.tsx
```diff
@@ -0,0 +1,105 @@
...
+		this.addEventHook('complete', this.onComplete)
+	}
+
+	private onSummon() {
+		if (!(this.fairyOut)) {
+			this.fairyOut = true
+			if (! isUndefined(this.petHook)) {
+				this.removeEventHook(this.petHook)
+			}
+		}
```
AzariahAeroborn:
```suggestion
		this.fairyOut = true
		if (this.petHook != null) {
			this.removeEventHook(this.petHook)
		}
```

Don't really need the sanity check, since there's no case where changing from fairyOut = true to fairyOut = true would cause problems.  And style preference is to use != null for null/undefined checks rather than the lodash import.

src/parser/jobs/sch/modules/Gauge.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+		this.gauge.generate(GAUGE_GAIN_AMOUNT)
+	}
+	private onComplete() {
+		if (this.gauge.overCap > 0) {
+			this.suggestions.add(new TieredSuggestion({
+				icon: this.data.actions.SUMMON_SERAPH.icon,
+				content: <Trans id="sch.gauge.overcap.content">
+					Avoid using Aetherflow abilities when your Faerie Gauge is already at 100.
```
AzariahAeroborn: Sorry I didn't read this closely enough the first time - should this be expanded or reworded to suggest using Fey Union to consume the tether, rather than suggesting to just not use AF abilities if capped?

src/parser/jobs/sch/modules/Gauge.tsx
```diff
@@ -0,0 +1,105 @@
...
+const GAUGE_GAIN_AMOUNT = 10
+// Graph color
+const GRAPH_COLOR = Color(JOBS.SCHOLAR.colour)
+// Start at 30 because you can overcap when seraph is out and you can't drain the gauge
+const OVERCAP_SEVERITY = {
+	30: SEVERITY.MINOR,
+	50: SEVERITY.MEDIUM,
+	70: SEVERITY.MAJOR,
+}
+export default class FaerieGauge extends CoreGauge {
```
AzariahAeroborn: Project style guide - no default exports for new modules.
```suggestion
export class FaerieGauge extends CoreGauge {
```

src/parser/jobs/sch/modules/index.js
```diff
@@ -1,5 +1,6 @@
 import {CooldownDowntime} from './CooldownDowntime'
 import DoTs from './DoTs'
+import FaerieGauge from './Gauge'
```
AzariahAeroborn:
```suggestion
import {FaerieGauge} from './Gauge'
```
Project style guide - no default exports for new modules.

AzariahAeroborn (changes requested):
Sorry for the "oh by the ways" I missed the first time, was reading through this too quickly between tasks.  Overall looks good, just need to remove the correct hook for the summon tracking and a couple style cleanups and it'll be good to go.

src/parser/jobs/sch/modules/Gauge.tsx
ghost: Not sure when that got added, because I don't remember doing it myself.

src/parser/jobs/sch/modules/Gauge.tsx
ghost: Sounds like a good idea.

src/parser/jobs/sch/modules/Gauge.tsx
ghost: VSCode is saying that that's an error.

ghost: When I tried swapping it, it gave me 2000 characters or so of type ((Hook return value)) is unassignable to ((full union definition))

src/parser/jobs/sch/modules/Gauge.tsx
```diff
@@ -0,0 +1,103 @@
...
+import {filter, oneOf} from 'parser/core/filter'
+import {dependency} from 'parser/core/Injectable'
+import {CounterGauge} from 'parser/core/modules/Gauge'
+import {Gauge as CoreGauge} from 'parser/core/modules/Gauge/Gauge'
+import Suggestions, {SEVERITY, TieredSuggestion} from 'parser/core/modules/Suggestions'
+import React from 'react'
+
+const GAUGE_GAIN_AMOUNT = 10
+// Graph color
+const GRAPH_COLOR = Color(JOBS.SCHOLAR.colour)
```
ayyaruq: Not sure why this is needed, it's only used in one place and there's no reason it would need updating.

src/parser/jobs/sch/modules/Gauge.tsx
```diff
@@ -0,0 +1,103 @@
...
+// Start at 30 because you can overcap when seraph is out and you can't drain the gauge
+const OVERCAP_SEVERITY = {
+	30: SEVERITY.MINOR,
+	50: SEVERITY.MEDIUM,
+	70: SEVERITY.MAJOR,
+}
+export class FaerieGauge extends CoreGauge {
+	static override handle = 'fairieGauge'
```
ayyaruq: Worth keeping the spelling consistent.
```suggestion
	static override handle = 'faeriegauge'
```

src/parser/jobs/sch/modules/Gauge.tsx
ayyaruq: This should be just `EventHook<Events['action']>` if it's an action, or otherwise grab the actual action event.

src/parser/jobs/sch/modules/Gauge.tsx
```diff
@@ -0,0 +1,103 @@
...
+	private fairyOut: boolean = false
+	private actorPets = this.parser.pull.actors.filter(actor => actor.owner != null && actor.owner.id === this.parser.actor.id).map(pet => pet.id)
+	private petHook?: EventHook<{ type: 'action'; action: number; source: string; target: string; timestamp: number; }>
+
+	override initialise() {
+		super.initialise()
+		const playerFilter = filter<Event>().source(this.parser.actor.id)
+
+		//Faerie Summon
+		this.addEventHook(playerFilter.type('action').action(oneOf([this.data.actions.SUMMON_EOS.id, this.data.actions.SUMMON_SELENE.id])), this.onSummon)
```
ayyaruq: Make a proper ActionKey array for these, and then you can `action(this.data.matchActionId(SOME_CONST))`.

src/parser/jobs/sch/modules/Gauge.tsx
```diff
@@ -0,0 +1,103 @@
...
+		//sanity check, your pets can't take actions if they're not out.
+		this.petHook = this.addEventHook(filter<Event>().source(oneOf(this.actorPets)), this.onSummon)
```
ayyaruq: This assumes they summoned pre-pull. Not sure that's safe unless you're determining pet backwards in some way. I'd also make a petFilter like your playerFilter.

src/parser/jobs/sch/modules/Gauge.tsx
```diff
@@ -0,0 +1,103 @@
...
+		//Consumers
+		this.addEventHook(filter<Event>()
+			.type('heal')
+			.source(oneOf(this.actorPets))
+			.cause(this.data.matchCauseStatus(['FEY_UNION'])),
+		this.onGaugeSpend)
```
ayyaruq: Since you can only spend gauge while the correct pet is out, this hook should probably be ephemeral unless you have a reason not to.

src/parser/jobs/sch/modules/Gauge.tsx
```diff
@@ -0,0 +1,103 @@
...
+		//generators
+		//I only use this list of actions once, so as ugly as it is, I think it's still better to inline it
+		this.addEventHook(playerFilter.type('action').action(oneOf([
+			this.data.actions.SCH_ENERGY_DRAIN.id,
+			this.data.actions.LUSTRATE.id,
+			this.data.actions.INDOMITABILITY.id,
+			this.data.actions.SACRED_SOIL.id,
+			this.data.actions.EXCOGITATION.id,
+		])), this.onGaugeGenerate)
```
ayyaruq: Const this and do as above with an ActionKey[] and matchActionId. Also, if you're gonna multiline, multiline the whole thing.

src/parser/jobs/sch/modules/Gauge.tsx
```diff
@@ -0,0 +1,103 @@
...
+		//Death
+		this.addEventHook('death', () => this.fairyOut = false)
```
ayyaruq: Gauge doesn't reset on death?

src/parser/jobs/sch/modules/Gauge.tsx
```diff
@@ -0,0 +1,103 @@
...
+		//Dissipation
+		this.addEventHook(playerFilter.type('statusApply').status(this.data.statuses.DISSIPATION.id), () => this.fairyOut = false)
+		this.addEventHook(playerFilter.type('statusRemove').status(this.data.statuses.DISSIPATION.id), () => this.fairyOut = true)
```
ayyaruq: The faerie automatically comes back when Dissipation falls off?

src/parser/jobs/sch/modules/Gauge.tsx
```diff
@@ -0,0 +1,103 @@
...
+		this.gauge.spend(GAUGE_GAIN_AMOUNT)
+	}
+
+	private onGaugeGenerate() {
+		if (!this.fairyOut) {
+			// can't generate a guage without the fairy, so bail out
+			return
+		}
+
+		this.gauge.generate(GAUGE_GAIN_AMOUNT)
```
ayyaruq: You could invert this instead:
```suggestion
		if (this.fairyOut) {
				this.gauge.generate(GAUGE_GAIN_AMOUNT)
		}
```

src/parser/jobs/sch/modules/Gauge.tsx
```diff
@@ -0,0 +1,103 @@
...
+	}
+	private onComplete() {
+		if (this.gauge.overCap > 0) {
```
ayyaruq: This doesn't do anything. Your suggestion is tiered, so if the overcap is below 30 this won't show in the first place. If you remove the condition, the code is easier to read.

src/parser/jobs/sch/modules/Gauge.tsx
```diff
@@ -0,0 +1,103 @@
...
+			this.suggestions.add(new TieredSuggestion({
+				icon: this.data.actions.SUMMON_SERAPH.icon,
+				content: <Trans id="sch.gauge.overcap.content">
```
ayyaruq: Keep consistent with the module handle unless it doesn't make sense to.

```suggestion
				content: <Trans id="sch.faeriegauge.overcap.content">
```

src/parser/jobs/sch/modules/Gauge.tsx
```diff
@@ -0,0 +1,103 @@
...
+					Use <DataLink action="FEY_UNION" /> when your Faerie Gauge is high to spend gauge and avoid using Aetherflow abilities when your Faerie Gauge is full.
+				</Trans>,
+				tiers: OVERCAP_SEVERITY,
+				value: this.gauge.overCap,
+				why: <Trans id="sch.gauge.overcap.why">
```
ayyaruq:
```suggestion
				why: <Trans id="sch.faeriegauge.overcap.why">
```

ayyaruq (dismissed):
Mostly looks good, I'm a bit hm on some of those hooks tho but admittedly not super across how SCH works in EW. I think some more whitespace would help with legibility, but also more use of consts. Since this module only tracks the Faerie gauge, I'd also rename the file so it's clear that Aetherflow is separate entirely and fix up the handle/i18n.

src/parser/jobs/sch/modules/Gauge.tsx
ghost: NGL it was left in bc it was in the old version will remove

src/parser/jobs/sch/modules/Gauge.tsx
ghost: `EventHook<Events['action']>` doesn't work. I've tried it dozens of times. the existing code is the *only* way that my development environment doesn't scream at me about unassignable values.

src/parser/jobs/sch/modules/Gauge.tsx
AzariahAeroborn: I specifically asked for it to be this way, because we only need to check if the pet exists pre-pull until we see it.  This is the best way to do it without an adapter (which isn't really necessary for SCH pets because they don't do anything that complicated.

src/parser/jobs/sch/modules/Gauge.tsx
AzariahAeroborn: The pet is out all the time, it's the same action for either fairy (they're skins only), and this is a specific healing event for the fairy tether, there's no reason to make the hook ephemeral.

src/parser/jobs/sch/modules/Gauge.tsx
ghost: ...
OK, I finally figured it out by *copy pasting your code*. didn't realize there were *two different similarly named classes in the event package*.

src/parser/jobs/sch/modules/Gauge.tsx
ghost: yep

src/parser/jobs/sch/modules/Gauge.tsx
ghost: so... if I want to add anything to the death event, I have to completely rewrite the whole thing?

src/parser/jobs/sch/modules/Gauge.tsx
ghost: That entire list is only ever used once, seems like a massive waste to add a const for a throwaway value.

src/parser/jobs/sch/modules/Gauge.tsx
AzariahAeroborn: I think this should be fine - CounterGauges automatically reset on Death (or at least I didn't do any explicit resetting when I set up a CounterGauge for Blood Weapon).  @ayyaruq is there something else I'm not aware of that we need to consider here?

AzariahAeroborn (approved):

PR #1582, opened 2022-01-26
Tincture ast

ghost:
Yet another PR, very much like the rest.

Expected actions inside tincture:
1 Combust III (DoT)
0-1 Lord of Crowns (oGCD Damage, not always possible)
1 Astrodyne (MP Regen / Speed / Damage buff)


ayyaruq:
Do AST players actually want this? Asking because several jobs have opted to not show pots, which is why it's not a core analyser on all jobs. I have no idea for AST tho so asking before spending time on code review.

ghost:
That's honestly a good question. It was requested in [fb-astrologian](https://discord.com/channels/441414116914233364/725970639910993931/935786554197168180), and I didn't really stop to consider whether some jobs might not really care about getting analysis on tincture windows.  
That does make sense though, Especially jobs like AST and WHM, it's quite intuitive to put all of their most damaging abilities under a tincture.

ghost:
That said, AST is the odd-one-out when it comes to not having a tincture or Swiftcast module, at least in the healer role, all three other healers have both.

ayyaruq:
Fair enough. After a cursory look, I'd prob make the Lord of Crowns check properly dynamic (otherwise it'll show 1/0 for when it's not expected), and prob either filter out stuff that doesn't matter, or hugely limit what does. The chart will be massive with healing stuff that does no damage otherwise, unless pots also buff healing in a way that matters outside of fight specifics.

otocephaly:
I also recommend including divination since usually the tincture is aligned with burst windows

ghost:
Brain small, couldn't figure out how to reliably track LoC, so for now I've removed it.

suShirow:
> Brain small, couldn't figure out how to reliably track LoC, so for now I've removed it.

As a naive impl, we should in theory be able to load in ArcanaTracking and use getCardState to know whether they even HAD an LoC for the pot window but I realize checking needs to go a lot further than that. We'd also have to check if Minor is on CD, then track what they pull after they pulled it. 
So ArcanaTracking isn't up to the task just yet. 

So with that in mind I think it's fine to just drop that first with some comment underneath informing the user that it is planned in a future update? 

(edit - formatting)

ghost:
I'll add the note about LoC being added in a future update, and recommending that players attempt to get one inside their tincture if possible.

suShirow:
Did something go wonky when merging from endwalker branch? The bot's tagged the PR funny x_x

ghost:
> Did something go wonky when merging from endwalker branch? The bot's tagged the PR funny x_x

I had to rebase, and i think it forwarded the commit history. Maybe we should have @ackwell check on it?

src/parser/jobs/ast/modules/Tincture.tsx
```diff
@@ -4,8 +4,16 @@ import {ExpectedActionsEvaluator} from 'parser/core/modules/ActionWindow'
 import {SEVERITY} from 'parser/core/modules/Suggestions'
 import {Tincture as CoreTincture} from 'parser/core/modules/Tincture'
 import React from 'react'
+import {Message} from 'semantic-ui-react'
 
 export class Tincture extends CoreTincture {
+
+	override prependMessages = <Message>
+		<Trans id="sch.tincture.prepend.content">
```
suShirow: Gotta modify the i18n key here for ast

suShirow (changes requested):

src/parser/jobs/ast/modules/Tincture.tsx
suShirow:
```suggestion
		<Trans id="ast.tincture.prepend.content">
```

ghost: I may be the dumbest man to ever commit to this repo LMAO

suShirow (approved):
Will not merge yet til I confirm the rebase actions won't do anything weird x_x

PR #1583, opened 2022-01-26
Add Swiftcast for AST

ghost:
Mer crimus otocephaly.
More code for 6.08 bump


suShirow (approved):
Thanks for this. Was gonna suggest to get a display order for swift but the default should set it below Divination, which makes sense.

PR #1588, opened 2022-01-27
Allow hiding progress for checklist modules

Warchamp7:
I am still a React noob so let me know if I did this wrong

This makes the display of progress on a checklist entry hideable. By default it is shown (Maintaining current behaviour)

A second commit changes the defensive cooldown checklist to hide progress. Examples below.

image

image

### DIFF
--- a/src/parser/core/modules/CooldownDowntime.tsx
+++ b/src/parser/core/modules/CooldownDowntime.tsx
@@ -105,7 +105,7 @@ export abstract class CooldownDowntime extends Analyser {
 		when they are available, but do not clip your GCD to use them.</Trans>
 	protected checklistTarget = DEFAULT_CHECKLIST_TARGET
 
-	protected defenseChecklistName = <Trans id="core.cooldownDowntime.use-defense-cds">Use your defensive cooldowns</Trans>
+	protected defenseChecklistName = <Trans id="core.cooldownDowntime.use-defense-cds">Defensive cooldown usage</Trans>
 	protected defenseChecklistDescription = <Trans id="core.cooldownDowntime.defense-cd-metric">
 		Using your mitigation and healing cooldowns can help you survive mistakes, or relieve some stress on the healers and let them deal more damage.
 		While you shouldn't use them at the expense of your rotation or buff alignment, you should try to find helpful times to use them.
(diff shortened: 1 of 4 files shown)


src/parser/core/modules/CooldownDowntime.tsx
```diff
@@ -105,7 +105,7 @@ export abstract class CooldownDowntime extends Analyser {
 		when they are available, but do not clip your GCD to use them.</Trans>
 	protected checklistTarget = DEFAULT_CHECKLIST_TARGET
 
-	protected defenseChecklistName = <Trans id="core.cooldownDowntime.use-defense-cds">Use your defensive cooldowns</Trans>
+	protected defenseChecklistName = <Trans id="core.cooldownDowntime.use-defense-cds">Defensive cooldown usage</Trans>
```
ghost, line 108: We typically tell people what to do, rather than what they've done.
So leaving this as "Use your defensive cooldowns" would be fine, IMO.

src/parser/core/modules/CooldownDowntime.tsx
```diff
@@ -210,6 +210,7 @@ export abstract class CooldownDowntime extends Analyser {
 				name: this.defenseChecklistName,
 				description: this.defenseChecklistDescription,
 				requirements: defensiveRequirements,
+				showProgress: false,
```
ghost, line 213: This will turn off progress bars for all CDDT modules, which would be less than ideal, since most jobs use them for their offensive cooldowns, and it's preferred for them to still receive a progress bar.

ghost (changes requested):

src/parser/core/modules/CooldownDowntime.tsx
ghost: My mistake, this is actually fine.

src/parser/core/modules/CooldownDowntime.tsx
Warchamp7: The thought here was that since it's not necessarily a "mistake" that I wasn't sure if it made sense to use the improvement style phrasing anymore

AzariahAeroborn:
Per discussion on Discord, we are not going to proceed with this change at this time.  The current design & intent for the Checklist is not a good fit for optionally turning the % indicator off.

PR #1590, opened 2022-01-27
[Core] Add overridable prop for prepended messages on ActionWindow

suShirow:
Example screaming elements illustrated below can be added anytime you implement something that implements ActionWindow via overriding prependMessages

image


suShirow:
https://github.com/xivanalysis/xivanalysis/pull/1582 depends on this current PR as it needs to add its own messages to Tincture

suShirow:
If nothing is passed in, you just get this
image

ayyaruq (approved):

PR #1592, opened 2022-01-27
TableStatistic + core Defensives module

hintxiv:
This PR adds a new type of statistic which outputs a table. It also:

* Adds a new core module for tracking defensive CDs that leverages TableStatistic
* Rewrites RadiantAegis to leverage the core module
* Adds a RPR implementation

image
image


src/parser/core/modules/Statistics/TableStatistic.tsx
```diff
@@ -0,0 +1,46 @@
+import React from 'react'
+import {AbstractStatistic, AbstractStatisticOptions} from './AbstractStatistic'
+import styles from './TableStatistic.module.css'
+
+interface FixedLengthArray<T, L extends number> extends ReadonlyArray<T> {
+	0: T
+	length: L
+}
```
hintxiv, line 8: This is fully lifted from `PieChartStatistic`, not sure what the right approach is for sharing this code.

src/parser/core/modules/Defensives.tsx
```diff
@@ -0,0 +1,86 @@
...
+			<ActionLink key={index} {...defensive} />,
+			this.uses.get(defensive.id) ?? 0,
+			this.cooldownDowntime.calculateMaxUsages({cooldowns: [defensive]}),
+		])
+
+		this.statistics.add(new TableStatistic({
+			headings: [
+				'Defensive',
+				'Uses',
+				'Max',
```
AkairyuGestalter: Can /should we include a utilization percentage here, like radiant aegis and CDDT do today?

AkairyuGestalter (changes requested):
One question about the display contents for now, will give this a closer look tomorrow

src/parser/core/modules/Defensives.tsx
hintxiv: Would be nice to show that, but adding another column makes the formatting super awkward. It's definitely too big for `width = 1` but too small for `width = 2`. Open to any suggestions on how to fix it

AzariahAeroborn: Instead of separate columns for uses and max, could we just have a Uses/Max column and display `10/18 (55%)`?

hintxiv: Thanks that's a good idea, let me try

AkairyuGestalter (approved):

PR #1594, opened 2022-01-29
[SAM] Meikyo bugifx

Rjb99:
Bugfix for Meikyo because apparently, Samurais are unable to bend space time, my bad.

Short tl;dr EoF cursed, can return possible GCDs that exceed 3. Hard capping it to 3.


src/parser/jobs/sam/modules/Meikyo.tsx
```diff
@@ -94,6 +94,9 @@ export class Meikyo extends BuffWindow {
 			const fightTimeRemaining = (this.parser.pull.timestamp + this.parser.pull.duration) - window.start
 			const possibleGCDs = Math.floor(fightTimeRemaining / SAM_BASE_GCD_SPEED_BUFFED)
 
+			if (possibleGCDs > SEN_GCDS) { //No Samurai, you can not slash so fast you bend space-time and get 2 GCDs for the price of 1
+				return 0
+			}
 			return possibleGCDs - defaultEoFValue
```
ayyaruq: You return 0 later, so I think you can simplify this. If you return the `possibleGCDs - defaultEoFValue` with an inverted condition, you can keep the later `return 0` and make this simpler.

src/parser/jobs/sam/changelog.tsx
```diff
@@ -8,6 +8,10 @@ export const changelog = [
 	// 	contrubutors: [CONTRIBUTORS.YOU],
 	// },
 	{
+		date: new Date('2022-1-28'),
```
ayyaruq:
```suggestion
		date: new Date('2022-01-28'),
```

ayyaruq (changes requested):
Not requesting code changes, just noting the condition could be simplified. Changelog needs fixing tho. Surprised the older entry parses sanely since it has the same issue.

src/parser/jobs/sam/modules/Meikyo.tsx
hintxiv:
```suggestion
                        return Math.min(SEN_GCDs, possibleGCDs - defaultEoFValue)
```

Rjb99:
@ayyaruq requested changes made

AzariahAeroborn (approved):

ayyaruq (approved):

PR #1595, opened 2022-01-29
Fix duplicate searing light entries with multiple summoners

eatmorehufu:
Add the petsFilter to the statusApply and statusRemove event hooks.


AkairyuGestalter (commented):
Some food for thought:
Iirc this module was lifted/repurposed from DNC's Technical Finish. In that one, rather than hiding windows initiated by another dancer, I show a `N/A` for the `Players Buffed` column. That way you can still assess whether you were properly saving resources to burst in the window that lined up with other raid buffs. I don't know how applicable that is to SMN's gameplay loop, but it's something you might consider as an alternative solution.

eatmorehufu:
> 

That could work but I feel like that's covered by the Timeline module at the bottom, where you can visually see which GCDs you fit inside the other party buffs.

I think it's cleaner to treat another summoner's Searing Light as a generic party buff rather than give it special treatment in the module when there are multiple SMNs. Imagine the module with 4 SMNs when you have 100% Searing Light uptime (2m CD, 30s duration).

This is all small potatoes considering running identical classes is a niche/fun thing rather than a serious feedback thing.

KelosPelmand:
Since we track nothing in the Searing Light window anyway due to the effect Spell Speed (or mechanics) can have on how the window lines up with your rotation, there isn't a need to show the other SMN's windows like there is for DNC.

The more relevant thing to test here is how this change handles 2 SMNs casting SL at the same time instead of being well disciplined like in this example.  If you have already test that, let me know, otherwise I'll try to dig out the parses from when we had a previous issue involving multiple SMNs casting SL at the same time.

eatmorehufu:
Having multiple overriding searing light looks as expected:

- If your searing light got overridden, your buff window is shortened so you look like you have fewer GCDs during your window (which is giving some feedback that clipping/overriding searing light is wrong)
- If you clip someone else the module won't complain.

eatmorehufu:
Updated PR with conversation in discord: 
https://discord.com/channels/441414116914233364/470050747720138753/937086689497722980

- You will no longer get bad entries for other people's searing lights
- If yours gets overriden by other SMN, you get credit for what would have been your own duration

KelosPelmand (approved):
lgtm. Thanks for the contribution.

PR #1597, opened 2022-01-30
Fix Searing Light module with multiple SMNs

eatmorehufu:
- You will no longer get credit for other SMN's Searing Lights
- If they override your own Searing Light, you still get credit for what would have been the duration of your own buff.


KelosPelmand (approved):
lgtm (still)

AkairyuGestalter (approved):

PR #1598, opened 2022-01-31
[SCH] Add defensive cooldowns

dthomash:
Add scholar's defensive cooldowns


ayyaruq (approved):
Seems a little odd when you don't necessarily use these on cooldown but LGTM as far as code logic goes.

PR #1599, opened 2022-01-31
People die when they are killed

ackwell:
People die when they are killed


ackwell:
merging cus im an admin and i get to do what i want

people die when they are killed

PR #1601, opened 2022-02-01
RDM SwiftCast and Defensive Cooldowns

Cephid:
Added Defensive Cooldowns.
Added Swiftcast Validation, these rules are the same as Dualcast.


src/parser/jobs/rdm/modules/Swiftcast.tsx
```diff
@@ -0,0 +1,31 @@
...
+		//As long as we aren't in downtime, the cast time must exceed the base GCD to qualify.
+		if (!this.downtime.isDowntime() && castTime <= BASE_GCD) {
+			return false
+		}
+		//Then it had to be VerRaise, VerAero III, or VerThunder III or we were in downtime so it's valid
+		return true
+	}
+	//Use the default behavoir if we've gotten back no cast time
+	return super.considerSwiftAction(_action)
+}
```
AzariahAeroborn: Looks like the indentation on these got a bit janky, could you clean these up?

Also, the `_action` variable name in the base class is because the variable doesn't get used in the base implementation -- when overriding, you should go ahead and remove the underscore from the variable name and just use `action`

AzariahAeroborn (changes requested):

src/parser/jobs/rdm/modules/Swiftcast.tsx
Cephid: Oh I didn't realize you could change the textual name when overriding it like that.  Good to know.

AzariahAeroborn (approved):

PR #1607, opened 2022-02-03
[SCH] Defensive cooldown updates

dthomash:
Based on feedback added information message to checklist item in line with other healer recommendations. Removed the following abilities from defensive section:
- Deployment Tactics
- Dissipation


ayyaruq (approved):
LGTM. Generally prefer to avoid editing existing changelog entries but it's just punctuation so whatevs. I'd also consider the i18n key for the description text being more relevant - "defence-cd-metric" vs "description", one of these is clearer than the other. Approved since code-wise it's fine, no idea about the SCH logic so would be nice to have a second opinion here.

PR #1608, opened 2022-02-05
Add fflogs absorbed event

eatmorehufu:
And do nothing with it.

### DIFF
--- a/src/reportSources/legacyFflogs/__tests__/__snapshots__/eventAdapter.ts.snap
+++ b/src/reportSources/legacyFflogs/__tests__/__snapshots__/eventAdapter.ts.snap
@@ -1,5 +1,7 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`Event adapter individual events adapts absorbed 1`] = `Array []`;
+
 exports[`Event adapter individual events adapts applybuff 1`] = `
 Array [
   Object {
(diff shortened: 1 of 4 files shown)


ayyaruq:
Did fflogs change this again? I recall this event being removed due to a rename or something in early Endwalker.

eatmorehufu:
It's there now, fwiw. Here's an example:

```
{
    "timestamp": 452966,
    "type": "absorbed",
    "sourceID": 2,
    "sourceIsFriendly": true,
    "targetID": 6,
    "targetIsFriendly": true,
    "ability": {
        "name": "Eukrasian Diagnosis",
        "guid": 1002607,
        "type": 8,
        "abilityIcon": "012000-012954.png"
    },
    "fight": 3,
    "attackerID": 10,
    "attackerIsFriendly": false,
    "amount": 11372,
    "extraAbility": {
        "name": "attack",
        "guid": 872,
        "type": 128,
        "abilityIcon": "000000-000101.png"
    },
    "targetResources": {
        "hitPoints": 86098,
        "maxHitPoints": 86098,
        "mp": 10000,
        "maxMP": 10000,
        "tp": 0,
        "maxTP": 0,
        "x": 10004,
        "y": 9532,
        "facing": -158,
        "absorb": 52
    }
}
```

AzariahAeroborn (changes requested):
Lint failure is due to not adding a sample `absorbed` event to the fakeEvents record in the tests for the FFLogs adapter.  You should be able to drop your sample in there and then update the snapshot for the test file (`yarn test --updateSnapshot`) for the new test event -- the snapshot should result in no event being adapted from the test absorbed event.

AzariahAeroborn (approved):

PR #1609, opened 2022-02-05
Add WAR tincture module.

eatmorehufu:
Open to suggestions on tuning the required skills. The skill amounts listed are a conservative estimate of what you could put in a pot window, I think you could have 3 IC, 4FC if not 5, and 3 Onslaughts.

Text for the suggestion copied from other tincture text for consistency.


ayyaruq (changes requested):
This effectively duplicates the IR module output, other than the IC uses. Orogeny is also not handled at all here, same for Chaotic Cyclone and Decimate, even tho AoE is fairly rare in tincture windows for raids, but it's not like raids are the only content we support.

We intentionally don't have a tinctures module for WAR and I don't see that changing. What could be of value is a suggestion for missing 2-3x IC/CC under tinctures. Up to you for if you want to do that but as the job maintainer I won't approve a WAR tincture table being added.

PR #1615, opened 2022-02-10
Split out BlitzEvaluator from RoF

Malpp:
Also rename skills to actions

### DIFF
--- a/src/parser/jobs/mnk/modules/evaluators/BlitzEvaluator.tsx
+++ b/src/parser/jobs/mnk/modules/evaluators/BlitzEvaluator.tsx
@@ -1,0 +1,41 @@
+import {DataLink} from 'components/ui/DbLink'
+import {Action} from 'data/ACTIONS'
+import {EvaluatedAction, EvaluationOutput, WindowEvaluator} from 'parser/core/modules/ActionWindow'
+import {HistoryEntry} from 'parser/core/modules/ActionWindow/History'
+import React from 'react'
+
+interface BlitzEvaluatorOpts {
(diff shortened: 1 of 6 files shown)


src/parser/jobs/mnk/modules/evaluators/BlitzEvaluator.tsx
```diff
@@ -0,0 +1,41 @@
+import {DataLink} from 'components/ui/DbLink'
+import {Action} from 'data/ACTIONS'
+import {EvaluatedAction, EvaluationOutput, WindowEvaluator} from 'parser/core/modules/ActionWindow'
+import {HistoryEntry} from 'parser/core/modules/ActionWindow/History'
+import React from 'react'
+
+interface BlitzEvaluatorOpts {
+	blitzActions: Array<Action['id']>,
+}
```
hintxiv, line 9: Since there's only one parameter here, we can save a few lines of code by getting rid of this interface and having the constructor accept `blitzActions: Array<Action['id']>` directly.

hintxiv (changes requested):
Looks good, definitely improves readability for the RoF module.

src/parser/jobs/mnk/modules/evaluators/BlitzEvaluator.tsx
ayyaruq: That's bad design, I requested a change from that approach to this one in the previous PR.

hintxiv (approved):

ayyaruq (approved):
Nicely done.

PR #1617, opened 2022-02-11
Fix Swiftcast end of fight bug

eatmorehufu:
Forgiving a missed Swift at the end means 1 - 1 = 0, instead of 1 + 1 = 2.

https://discord.com/channels/441414116914233364/915957728155959356/934127115656388618

This logic is still slightly too permissive, instead of checking 1 GCD it can check something like 0.5 - 1.5s, but being more permissive at EoF is also fine.


hintxiv (approved):
👍

ayyaruq (approved):

PR #1618, opened 2022-02-12
Add patch KR 6.0

Ra-Workspace:
Included minor change.


ayyaruq:
For context, this is the 6.0 release however it includes the job changes from 6.05 and 6.08, just not the content. From an analyser perspective this makes sense, as long as fflogs doesn't get weird across patches that don't match their content. Would like to see what others think of how the date is managed for KR.

ayyaruq (approved):
LGTM as per ack's "yeah that sounds sane".

PR #1619, opened 2022-02-12
Add Death gauge to RPR

ayyaruq:
Draft due to slight issue with Timeline that's being investigated. Most of the weirdness in this module is to do with how Enshroud itself works, and just generally a lot of extra checking because this is effectively the only stack-type gauge that counts down from a maximum.


src/parser/jobs/rpr/modules/DeathGauge.tsx
```diff
@@ -0,0 +1,129 @@
...
+const MAX_STACKS = 5
+
+export class DeathGauge extends CoreGauge {
+	@dependency private actors!: Actors
+	@dependency private brokenLog!: BrokenLog
+
+	// Lemure's Shroud, the blue pips
+	private lemureShroud = this.add(new CounterGauge({
+		maximum: 0,
+		graph: {label: 'Lemure Shroud', color: JOBS.MACHINIST.colour, collapse: false},
```
ghost:
```suggestion
		graph: {label: 'Lemure Shroud', color: JOBS.PALADIN.colour, collapse: false},
```
more accurate colour

ghost (approved):
Job logic lgtm

src/parser/jobs/rpr/modules/DeathGauge.tsx
```diff
@@ -0,0 +1,126 @@
...
+	private onLemure(event: Events['action']) {
+		const action = this.data.getAction(event.action)
+
+		// Sanity check
+		if (action == null || !this.actors.current.hasStatus(this.data.statuses.ENSHROUDED.id)) { return }
+
+		// Communio gets special handling since it eats everything you have
+		if (action.id === this.data.actions.COMMUNIO.id && this.lemureShroud.value > 0) {
+			// Use set so we can lodge this reset as a spend rather than a reset
+			this.lemureShroud.set(0, 'spend')
```
eatmorehufu, line 102: Using Communio also wipes out your void shroud, you could call onDeshroud here since it's actually a reset of the void shroud and not a "spend".

src/parser/jobs/rpr/modules/DeathGauge.tsx
```diff
@@ -0,0 +1,126 @@
...
+	private onEnshroud() {
+		// Pre-fill Lemure stacks to max, using generate as that's what the tooltip says it does internally
+		this.lemureShroud.setMaximum(MAX_STACKS)
+		this.lemureShroud.generate(MAX_STACKS)
+
+		// Technically this can be 5 but your window would end if it did
+		this.voidShroud.setMaximum(MAX_STACKS - 1)
+	}
+
+	private onDeshroud() {
```
AkairyuGestalter, line 83: You may want to override the onDeath function to call onDeshroud to make sure the setMaximums get called so you don't have ghost 0/5 on the tooltips if you die during Enshroud until the next window starts (See DNC Gauge for an example of some job-specific death logic)

AkairyuGestalter: On second thought, dying will drop the status... lol, never mind

src/parser/jobs/rpr/modules/DeathGauge.tsx
AkairyuGestalter: That should get taken care of by the status removal that using Communio will result in as the Enshroud window ends. I don't think it's necessary to manually call onDeshroud here

AkairyuGestalter (approved):
I thought I had comments but then realized it was actually fine.

src/parser/jobs/rpr/modules/DeathGauge.tsx
ayyaruq: > That should get taken care of by the status removal that using Communio will result in as the Enshroud window ends.

That was the idea yeah.

PR #1623, opened 2022-02-13
Add Soul and Shroud gauges to RPR

ayyaruq:
The Death's Design handling was super aaaaaaa. That is all.


src/parser/jobs/rpr/modules/OtherGauges.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+const DESIGN_GAIN = 10
+const BASE_SHROUD_MOD = 10
+const HIGH_SHROUD_MOD = 50
+
+export class OtherGauges extends CoreGauge {
+	@dependency private actors!: Actors
+	@dependency private suggestions!: Suggestions
+
+	static override debug = true
```
Rjb99: Nice Debug is true

ayyaruq: Damn it Ryan

src/parser/jobs/rpr/modules/OtherGauges.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+	// Initialise our gauges - default max is 100 so yolo it is
+	private soulGauge = this.add(new CounterGauge({
```
Rjb99, line 41: Legit complaint though, add a fade to the graphs. Being able to see both resources in a collapsed view is super helpful.
Based on the screenshot, you can't see the Red soul gauge (which you made purple lol) if the Blue gauge is same or higher value.

Color(JOBS.BLUE_MAGE.colour).fade(.25) or something should be fine

ayyaruq: That's a good call, should prob also do in Death gauge.

src/parser/jobs/rpr/modules/OtherGauges.tsx
```diff
@@ -0,0 +1,142 @@
...
+			value: this.soulGauge.overCap,
+		}))
+
+		this.suggestions.add(new TieredSuggestion({
+			icon: this.data.actions.PLENTIFUL_HARVEST.icon,
+			content: <Trans id="rpr.gauge.shroud.suggestions.overcap.content">
+					Avoid letting your Shroud gauge overcap as it can cost you uses of <DataLink action="ENSHROUD"/> over the fight.
+			</Trans>,
+			why: <Trans id="rpr.gauge.shroud.suggestions.overcap.why">
+				{this.soulGauge.overCap} Shrouds lost to overcapping Shroud gauge.
```
ghost: problem: you're using the soul gauge value for shroud gauge
nit: your soul gauge overcap should really be `⌈this.shroudGauge.overCap / 50⌉` or the text should be changed to reflect that it's shroud gauge lost, instead of being full uses of enshroud.
```suggestion
				{this.shroudGauge.overCap} Shrouds lost to overcapping Shroud gauge.
```

src/parser/jobs/rpr/modules/OtherGauges.tsx
```diff
@@ -0,0 +1,142 @@
...
+	'GIBBET',
+	'GALLOWS',
+	'GUILLOTINE',
+	'PLENTIFUL_HARVEST',
+	'ENSHROUD',
+]
+
+const GAUGE_FADE = 0.25
+
+const DESIGN_GAIN = 10
```
ghost, line 32: nit: this name confused the shit out of me before I went down and looked for where you used it.

src/parser/jobs/rpr/modules/OtherGauges.tsx
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	private onComplete() {
+		this.suggestions.add(new TieredSuggestion({
+			icon: this.data.actions.SOUL_SLICE.icon,
+			content: <Trans id="rpr.gauge.soul.suggestions.overcap.content">
+					Avoid letting your Soul gauge overcap. The wasted resources cost you Shroud generation which you need for your highest damage skills.
+			</Trans>,
+			why: <Trans id="rpr.gauge.soul.suggestions.overcap.why">
+				{this.soulGauge.overCap} Souls lost to overcapping Soul gauge.
```
ghost, line 124: nit: this value should be /10 given the existing copy, IMO.

ghost (changes requested):
Overall, quite good, I have a few nitpicks, and one change that I think needs to be made.

src/parser/jobs/rpr/modules/OtherGauges.tsx
ayyaruq: Might be a bit of a disconnect here. The job guide and game itself call each point of gauge "souls", even tho you realistically consume it in much larger blocks. The idea here is to show how much total gauge was overcapped, not individual actions. RPR only has one real use for Soul gauge tho, so maybe assuming that use is fine. Open to ideas, the current approach is more in alignment with what other jobs do but I'm not married to that.

ghost: Ah, noted. Makes sense to me.

src/parser/jobs/rpr/modules/OtherGauges.tsx
ayyaruq: Good catch. Not sure about showing actual Enshrouds, the game itself calls each point of the gauge a shroud so it's kinda awkward naming here. Like Soul gauge, I'm open to changing this, just wanna see what users prefer.

AkairyuGestalter (approved):

PR #1625, opened 2022-02-14
[RPR] Add Soul Reaver checks to Reaper

Rjb99:
I can't keep getting away with being an idiot.
Added a Soul Reaver check to Reaper to check if we drop stacks.


src/parser/jobs/rpr/modules/Reaver.tsx
```diff
@@ -0,0 +1,137 @@
...
+		this.data.actions.SOULSOW.id, //Yes, even this takes it.
+		this.data.actions.HARPE.id,
+		this.data.actions.SHADOW_OF_DEATH.id,
+		this.data.actions.WHORL_OF_DEATH.id,
+		this.data.actions.SOUL_SCYTHE.id,
+		this.data.actions.SOUL_SLICE.id,
+		this.data.actions.SPINNING_SCYTHE.id,
+		this.data.actions.NIGHTMARE_SCYTHE.id,
+		this.data.actions.PLENTIFUL_HARVEST.id,
+	]
```
ayyaruq: Prefer these as consts and then in your hooks, use `this.data.matchActionId` to fill the IDs instead of `oneOf` in your action filter. Generally, we don't use leading capitals in names either, but with a const it's all caps so whatever.

src/parser/jobs/rpr/modules/Reaver.tsx
```diff
@@ -0,0 +1,137 @@
...
+const GLUTTONY_GAIN = 2 //Gluttony grants 2 reavers, also therotically max
+
+export class Reaver extends Analyser {
+	static override handle = 'reaver'
+
+	@dependency private data!: Data
+	@dependency private suggestions!: Suggestions
+
+	//Trackers
+	private haveReaver = false
```
ayyaruq: Is tracking this independently useful, or could you just `this.actors.current.hasStatus(this.data.statuses.SOUL_REAVER.id)`?

src/parser/jobs/rpr/modules/Reaver.tsx
```diff
@@ -0,0 +1,137 @@
...
+				.action(oneOf(this.TheMakers)),
+			this.onGenCast
+		)
+
+		this.addEventHook('complete', this.onComplete)
+
+	}
+
+	private onGenCast(event: Events['action']) : void {
+		if (this.haveReaver === true) {
```
ayyaruq: It's an actual boolean, so you don't need the value to compare against.

src/parser/jobs/rpr/modules/Reaver.tsx
```diff
@@ -0,0 +1,137 @@
...
+			this.currentReaverStacks = 0
+			this.haveReaver = false
+		}
+		if (event.action === this.data.actions.GLUTTONY.id)	{
+			this.currentReaverStacks = GLUTTONY_GAIN
+			this.haveReaver = true
+		} else {
+			this.currentReaverStacks = SOUL_GAIN
+			this.haveReaver = true
+		}
```
ayyaruq: A little whitespace goes a long way. Since these do the same thing, you can condense this a lot.
```suggestion
		this.stacks = event.action === this.data.actions.GLUTTONY.id)	? GLUTTONY_GAIN : SOUL_GAIN
		this.haveReaver = true
```

src/parser/jobs/rpr/modules/Reaver.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+		this.addEventHook('complete', this.onComplete)
+
+	}
+
+	private onGenCast(event: Events['action']) : void {
+		if (this.haveReaver === true) {
+			this.droppedReaver += this.currentReaverStacks
+			this.currentReaverStacks = 0
+			this.haveReaver = false
```
ayyaruq: This is the same as what your bad actions hook is doing, may as well just call that function in the first place.

src/parser/jobs/rpr/modules/Reaver.tsx
```diff
@@ -0,0 +1,137 @@
...
+		}
+	}
+
+	private onGoodCast() {
+		this.currentReaverStacks-- //Everything currently consumes 1 stack
+
+		if (this.currentReaverStacks <= 0) { //bug check + flag change, just in case they somehow start with stacks.
+			this.currentReaverStacks = 0
+			this.haveReaver = false
+		}
```
ayyaruq: I'm not sure a simpler way to do this off the top of my head, but some clamping could be useful. Sanity check the adjustment before you change the stack count instead of after.

src/parser/jobs/rpr/modules/Reaver.tsx
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	private onComplete() {
+		this.suggestions.add(new TieredSuggestion({
+			icon: this.data.actions.GLUTTONY.icon,
+			content: <Trans id = "rpr.reaver.suggestion.droppedreaver.content">
+				Avoid dropping <DataLink status="SOUL_REAVER"/> by using <DataLink action="GUILLOTINE"/> , <DataLink action="GALLOWS"/> , and <DataLink action="GIBBET"/>. These actions grant you more damage and 10 shroud gauge, which is important to maximize the amount of <DataLink action="ENSHROUD"/> over the course of a fight.
+			</Trans>,
+			tiers: SEVERITY_TIERS,
+			why: <Trans id ="rpr.suggestion.droppedreaver.why">
```
ayyaruq:
```suggestion
			why: <Trans id ="rpr.reaver.suggestion.dropped.why">
```

src/parser/jobs/rpr/modules/Reaver.tsx
```diff
@@ -0,0 +1,137 @@
...
+		if (this.currentReaverStacks <= 0) { //bug check + flag change, just in case they somehow start with stacks.
+			this.currentReaverStacks = 0
+			this.haveReaver = false
+		}
+	}
+
+	private onComplete() {
+		this.suggestions.add(new TieredSuggestion({
+			icon: this.data.actions.GLUTTONY.icon,
+			content: <Trans id = "rpr.reaver.suggestion.droppedreaver.content">
```
ayyaruq:
```suggestion
			content: <Trans id = "rpr.reaver.suggestion.dropped.content">
```

src/parser/jobs/rpr/modules/Reaver.tsx
```diff
@@ -0,0 +1,137 @@
...
+				Avoid dropping <DataLink status="SOUL_REAVER"/> by using <DataLink action="GUILLOTINE"/> , <DataLink action="GALLOWS"/> , and <DataLink action="GIBBET"/>. These actions grant you more damage and 10 shroud gauge, which is important to maximize the amount of <DataLink action="ENSHROUD"/> over the course of a fight.
```
ayyaruq:
```suggestion
				Avoid dropping <DataLink status="SOUL_REAVER"/> by using <DataLink action="GALLOWS"/>, <DataLink action="GIBBET"/>, or <DataLink action="GUILLOTINE"/>. These actions have high damage and grant you 10 Shroud gauge, giving you more chances to use <DataLink action="ENSHROUD"/> over the course of a fight.
```

ayyaruq (changes requested):
Requesting changes for the i18n keys so they're consistent, beyond that it's mostly thoughts. Some of the names can be fixed up, they're not very obvious in meaning. Can we not use the Reaver status data to get accurate stack counts? Given how Reaver works I can see how that might be spicy, mostly curious.

src/parser/jobs/rpr/modules/Reaver.tsx
Rjb99: Moved to Consts and tried to make names clearer with Stack_makers, Stack_droppers and Stack_Skills_To_Use, not fully sold on my brain's naming for the 3rd one while trying to keep Stack as the first word.

Rjb99:
Cleaned up the code abit and made the requested i18n changes. Changes rendered all comments either null or were implemented.

Made changes to the 3 functions I'm hooked to, and added a death function to drop the stacks, though I'm debating if I should just make it set the stacks to 0 instead of calling dropStacks and flagging the stacks as dropped.

src/parser/jobs/rpr/modules/Reaver.tsx
Rjb99: Due to code changes, it won't decrement unless we have the status of soul reaver, so I removed the error checking as it's not needed.

ayyaruq:
> a death function to drop the stacks, though I'm debating if I should just make it set the stacks to 0 instead of calling dropStacks and flagging the stacks as dropped.

I think death should count as drops, if you don't die you don't lose them.

Rjb99:
> > a death function to drop the stacks, though I'm debating if I should just make it set the stacks to 0 instead of calling dropStacks and flagging the stacks as dropped.
> 
> I think death should count as drops, if you don't die you don't lose them.

Fair Enough, not everyone's death is caused by your wave buddy in p2s killing themselves right before you pot and double enshroud

ghost:
Logic LGTM, style also seems pree good.

ayyaruq (approved):

PR #1626, opened 2022-02-14
Brotherhood module

Malpp:
image
image


src/parser/jobs/mnk/modules/Brotherhood.tsx
```diff
@@ -0,0 +1,64 @@
...
+import {filter, oneOf} from 'parser/core/filter'
+import {dependency} from 'parser/core/Injectable'
+import {BuffWindow, EvaluatedAction} from 'parser/core/modules/ActionWindow'
+import {HistoryEntry} from 'parser/core/modules/ActionWindow/History'
+import {GlobalCooldown} from 'parser/core/modules/GlobalCooldown'
+import React from 'react'
+import {DISPLAY_ORDER} from './DISPLAY_ORDER'
+import {PlayersBuffedEvaluator} from './evaluators/PlayersBuffedEvaluator'
+
+export interface BrotherhoodBuff {
```
ayyaruq: Does this need exporting? It doesn't seem to be used anywhere outside this file.

src/parser/jobs/mnk/modules/Brotherhood.tsx
```diff
@@ -0,0 +1,64 @@
...
+	start: number
+	target: string
+}
+
+export class Brotherhood extends BuffWindow {
+	static override debug = true
```
ayyaruq: This will need to be set to false or removed before merging. Doesn't look like there's any residual debug statements so may as well remove it.
```suggestion
```

src/parser/jobs/mnk/modules/Brotherhood.tsx
```diff
@@ -0,0 +1,64 @@
...
+			.type('statusApply')
+			.status(this.data.statuses.BROTHERHOOD.id)
+			.source(this.parser.actor.id)
+			.target(oneOf(playerCharacters)),
+		this.buffApplied)
+
+		this.addEvaluator(new PlayersBuffedEvaluator({
+			affectedPlayers: this.affectedPlayers.bind(this),
+			suggestionContent: <Trans id="mnk.bh.playersbuffed.content">
+				To maximise raid dps, make sure <DataLink action="BROTHERHOOD"/> hits all party members.
```
ayyaruq: I think this reads a bit less demandingly, and by having the skill name at the end it forces users to read the whole message.
```suggestion
				To maximise raid damage, try to hit all party members with <DataLink action="BROTHERHOOD"/>.
```

src/parser/jobs/mnk/modules/Brotherhood.tsx
```diff
@@ -0,0 +1,64 @@
...
+import {BuffWindow, EvaluatedAction} from 'parser/core/modules/ActionWindow'
+import {HistoryEntry} from 'parser/core/modules/ActionWindow/History'
+import {GlobalCooldown} from 'parser/core/modules/GlobalCooldown'
+import React from 'react'
+import {DISPLAY_ORDER} from './DISPLAY_ORDER'
+import {PlayersBuffedEvaluator} from './evaluators/PlayersBuffedEvaluator'
+
+export interface BrotherhoodBuff {
+	start: number
+	target: string
```
ayyaruq: Might need to update imports but targets are actors.
```suggestion
	target: Actor
```

src/parser/jobs/mnk/modules/evaluators/BlitzEvaluator.tsx
```diff
@@ -6,13 +6,16 @@ import React from 'react'
 
 interface BlitzEvaluatorOpts {
 	blitzActions: Array<Action['id']>,
+	excepted?: number
```
ayyaruq:
```suggestion
	expected?: number
```

src/parser/jobs/mnk/modules/evaluators/BlitzEvaluator.tsx
```diff
@@ -6,13 +6,16 @@ import React from 'react'
...
 }
 
 export class BlitzEvaluator implements WindowEvaluator {
+	private readonly excepted: number | undefined;
```
ayyaruq:
```suggestion
	private readonly expected: number | undefined;
```

src/parser/jobs/mnk/modules/evaluators/BlitzEvaluator.tsx
```diff
@@ -6,13 +6,16 @@ import React from 'react'
...
 	private blitzActions: Array<Action['id']>
 
 	constructor(opts: BlitzEvaluatorOpts) {
 		this.blitzActions = opts.blitzActions
+		this.excepted = opts.excepted
```
ayyaruq:
```suggestion
		this.expected = opts.expected
```

src/parser/jobs/mnk/modules/evaluators/BlitzEvaluator.tsx
```diff
@@ -29,7 +32,7 @@ export class BlitzEvaluator implements WindowEvaluator {
 			rows: windows.map(window => {
 				return {
 					actual: this.countBlitzes(window),
-					expected: undefined,
+					expected: this.excepted,
```
ayyaruq:
```suggestion
					expected: this.expected,
```

src/parser/jobs/mnk/modules/Brotherhood.tsx
```diff
@@ -0,0 +1,64 @@
...
+			suggestionContent: <Trans id="mnk.bh.playersbuffed.content">
+				To maximise raid dps, make sure <DataLink action="BROTHERHOOD"/> hits all party members.
+			</Trans>,
+			suggestionIcon: this.data.actions.BROTHERHOOD.icon,
+			status: this.brotherhood.id,
+		}))
+	}
+
+	private affectedPlayers(window: HistoryEntry<EvaluatedAction[]>): number {
+		return this.brotherhoodBuffs.filter(value => Math.abs(window.start - value.start) < this.brotherhood.duration).length
```
ayyaruq: Might be worth renaming `value` here so it's clearer what you're trying to do.

src/parser/jobs/mnk/modules/evaluators/PlayersBuffedEvaluator.tsx
```diff
@@ -0,0 +1,87 @@
...
+	/**
+	 * This is used to determine how many players were buffed in a given window.
+	 * @param window The window to determine how many players were buffed
+	 * @returns The amount of players buffed in the window
+	 */
+	affectedPlayers: (window: HistoryEntry<EvaluatedAction[]>) => number
+	/**
+	 * Defaults to full party size.
+	 */
+	expectedPlayerCount?: number
```
ayyaruq: If this is optional, it doesn't have a default. Additionally, not all content uses 8, so I'd check if there's a way to grab party size sanely.

src/parser/jobs/mnk/modules/evaluators/PlayersBuffedEvaluator.tsx
```diff
@@ -0,0 +1,87 @@
...
+	/**
+	 * Defaults to minor.
+	 */
+	severityTiers?: SeverityTiers,
```
ayyaruq: Same here.

src/parser/jobs/mnk/modules/evaluators/PlayersBuffedEvaluator.tsx
```diff
@@ -0,0 +1,87 @@
...
+	suggestionIcon: string
+}
+
+export class PlayersBuffedEvaluator implements WindowEvaluator {
+	private readonly affectedPlayers: (window: HistoryEntry<EvaluatedAction[]>) => number
+	private readonly expectedPlayerCount: number;
+	private readonly severityTiers: SeverityTiers;
+	private readonly status: Status['id'];
+	private readonly suggestionContent: JSX.Element;
+	private readonly suggestionIcon: string;
```
ayyaruq: Our code style doesn't use trailing semicolons.
```suggestion
	private readonly expectedPlayerCount: number
	private readonly severityTiers: SeverityTiers
	private readonly status: Status['id']
	private readonly suggestionContent: JSX.Element
	private readonly suggestionIcon: string
```

src/parser/jobs/mnk/modules/evaluators/PlayersBuffedEvaluator.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+	constructor(opts: PlayersBuffedEvaluatorOptions) {
+		this.affectedPlayers = opts.affectedPlayers
+		this.expectedPlayerCount = opts.expectedPlayerCount || DEFAULT_EXPECTED_PLAYER_COUNT
```
ayyaruq: if you remove the optional flag you can handle this more sanely:
```suggestion
		this.expectedPlayerCount = opts.expectedPlayerCount ?? DEFAULT_EXPECTED_PLAYER_COUNT
```

src/parser/jobs/mnk/modules/evaluators/PlayersBuffedEvaluator.tsx
```diff
@@ -0,0 +1,87 @@
...
+		this.status = opts.status
+		this.suggestionIcon = opts.suggestionIcon
+		this.severityTiers = opts.severityTiers || DEFAULT_PLAYERS_MISSED_SEVERITY
```
ayyaruq: Here too.
```suggestion
		this.severityTiers = opts.severityTiers ?? DEFAULT_PLAYERS_MISSED_SEVERITY
```

src/parser/jobs/mnk/modules/evaluators/PlayersBuffedEvaluator.tsx
```diff
@@ -0,0 +1,87 @@
...
+		this.suggestionContent = opts.suggestionContent
+	}
+
+	public suggest(windows: Array<HistoryEntry<EvaluatedAction[]>>) {
+		const missed = this.countAffectedPlayerPerWindow(windows).filter(value => value < this.expectedPlayerCount).length
+		return new TieredSuggestion({
+			icon: this.suggestionIcon,
+			content: this.suggestionContent,
+			why: <Trans id="core.buffwindow.suggestions.playersbuffed.why">
+				{missed} of your <StatusLink id={this.status}/> uses did not fully buff the party.
```
ayyaruq: Not sure this works, StatusLink doesn't have an `id` prop. Seems fine in your screenshots which is odd.

ayyaruq (commented):
Mostly some code cleanup stuff to resolve, I'm not sure BuffWindow is a good idea for Brotherhood tho. We don't care what skills were in there, just the number of players hit. Would possibly be better as either a checklist or info box, thoughts?

src/parser/jobs/mnk/modules/Brotherhood.tsx
```diff
@@ -0,0 +1,89 @@
...
+			.target(oneOf(playerCharacters)),
+		this.onBrotherhoodBuff)
+
+		this.addEventHook(filter<Event>()
+			.type('action')
+			.action(this.data.actions.BROTHERHOOD.id)
+			.source(this.parser.actor.id),
+		this.onBrotherhoodAction)
+
+		this.addEventHook(filter<Event>().type('complete'), this.onComplete)
```
ayyaruq: Nit; simpler form, nbd really, just shorter.
```suggestion
		this.addEventHook('complete', this.onComplete)
```

src/parser/jobs/mnk/modules/Brotherhood.tsx
```diff
@@ -0,0 +1,89 @@
...
+import {Event, Events} from 'event'
+import {Analyser} from 'parser/core/Analyser'
+import {filter, oneOf} from 'parser/core/filter'
+import {dependency} from 'parser/core/Injectable'
+import Checklist, {Requirement, Rule} from 'parser/core/modules/Checklist'
+import {Data} from 'parser/core/modules/Data'
+import React from 'react'
+import {DISPLAY_ORDER} from './DISPLAY_ORDER'
+
+const FULL_PARTY_SIZE = 8
```
ayyaruq: If you're getting the list of players for the hook, you don't need this.

src/parser/jobs/mnk/modules/Brotherhood.tsx
```diff
@@ -0,0 +1,89 @@
...
+	private brotherhoodBuffTimestamps: number[] = []
+	private brotherhoodCastTimestamps: number[] = []
+	private brotherhood: StatusRoot['BROTHERHOOD'] = this.data.statuses.BROTHERHOOD
+
+	override initialise() {
+		super.initialise()
+
+		const playerCharacters = this.parser.pull.actors
+			.filter(actor => actor.playerControlled)
+			.map(actor => actor.id)
```
ayyaruq: You can make a var for like `this.partySize` and initialise it to `playerCharacters.length`, or just turn `playerCharacters` itself into a class var. Should make the party size more appropriate to the actual content, and won't break odd party sizes.

src/parser/jobs/mnk/modules/Brotherhood.tsx
```diff
@@ -0,0 +1,89 @@
...
+			.source(this.parser.actor.id),
+		this.onBrotherhoodAction)
+
+		this.addEventHook(filter<Event>().type('complete'), this.onComplete)
+	}
+
+	private onComplete() {
+		this.checklist.add(new Rule({
+			name: <Trans id="mnk.bh.checklist.name">Buff your party</Trans>,
+			description: <Trans id="mnk.bh.checklist.desc">
```
ayyaruq:
```suggestion
			description: <Trans id="mnk.bh.checklist.description">
```

src/parser/jobs/mnk/modules/Brotherhood.tsx
```diff
@@ -0,0 +1,89 @@
...
+				To maximise raid damage, try to hit all party members with <ActionLink action="BROTHERHOOD"/>.
+			</Trans>,
+			displayOrder: DISPLAY_ORDER.BROTHERHOOD,
+			requirements: [
+				new Requirement({
+					name: <Trans id="mnk.bh.checklist.req.name"><ActionLink action="BROTHERHOOD"/> buffs</Trans>,
```
ayyaruq:
```suggestion
					name: <Trans id="mnk.bh.checklist.requirement.name"><ActionLink action="BROTHERHOOD"/> buffs</Trans>,
```

src/parser/jobs/mnk/modules/Brotherhood.tsx
```diff
@@ -0,0 +1,89 @@
...
+					value: this.totalAffectedPlayers(),
+					target: this.maxPossibleTargets(),
+				}),
+			],
+			target: 95,
+		}))
+	}
+
+	private maxPossibleTargets(): number {
+		return this.brotherhoodCastTimestamps.length * FULL_PARTY_SIZE
```
ayyaruq: This is where we can use that party size var that's calculated from the log.

src/parser/jobs/mnk/modules/Brotherhood.tsx
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	private totalAffectedPlayers(): number {
+		return this.brotherhoodCastTimestamps.map(bhTimestamp => this.affectedPlayers(bhTimestamp))
+			.reduce((previousValue, currentValue) => previousValue + currentValue, 0)
+	}
+
+	private affectedPlayers(castTimestamp: number): number {
+		return this.brotherhoodBuffTimestamps.filter(buffTimestamp => Math.abs(castTimestamp - buffTimestamp) < this.brotherhood.duration).length
```
ayyaruq: This feels kinda brittle. I'd try something like making an interface for like a BrotherhoodWindow and incrementing a field in that, either as a raw number for hits, or a map of actors. There's a secondary `MEDITATIVE_BROTHERHOOD` status that goes on the casting MNK, you can use the statusRemove event for this to close the window. Then we can do the accounting at the end.

src/parser/jobs/mnk/modules/Brotherhood.tsx
```diff
@@ -0,0 +1,89 @@
...
+	static override debug = false
+	static override handle = 'brotherhood'
+	static override title = t('mnk.bh.title')`Brotherhood`
+	static override displayOrder = DISPLAY_ORDER.BROTHERHOOD
+
+	@dependency private checklist!: Checklist
+	@dependency private data!: Data
+
+	private brotherhoodBuffTimestamps: number[] = []
+	private brotherhoodCastTimestamps: number[] = []
```
ayyaruq: I think these could use some renaming or otherwise rethink how we store hits. Probably like `brotherhoodUsed` and `brotherhoodApplied` or something. An interface approach is likely cleaner tho, see my comment on the `affectedPlayers` method.

src/parser/jobs/mnk/modules/Brotherhood.tsx
```diff
@@ -0,0 +1,89 @@
...
+import {dependency} from 'parser/core/Injectable'
+import Checklist, {Requirement, Rule} from 'parser/core/modules/Checklist'
+import {Data} from 'parser/core/modules/Data'
+import React from 'react'
+import {DISPLAY_ORDER} from './DISPLAY_ORDER'
+
+const FULL_PARTY_SIZE = 8
+
+export class Brotherhood extends Analyser {
+	static override debug = false
```
ayyaruq, line 21: I'm fine with trailing debugs as long as it's set to false, but it doesn't look like there's any debug statements in this file anymore?

ayyaruq (changes requested):
Sorry to request changes again, the i18n keys should be fixed and we should make the party size variable. The other comments are just my thoughts/ideas and are at your discretion.

src/parser/jobs/mnk/modules/Brotherhood.tsx
```diff
@@ -0,0 +1,89 @@
+import {t, Trans} from '@lingui/macro'
```
ayyaruq: Oops missed a spot - you'll want `Trans` imported from `@lingui/react` instead.

ayyaruq (changes requested):

src/parser/jobs/mnk/modules/Brotherhood.tsx
```diff
@@ -0,0 +1,89 @@
...
+			name: <Trans id="mnk.bh.checklist.name">Buff your party</Trans>,
+			description: <Trans id="mnk.bh.checklist.desc">
+				To maximise raid damage, try to hit all party members with <ActionLink action="BROTHERHOOD"/>.
+			</Trans>,
+			displayOrder: DISPLAY_ORDER.BROTHERHOOD,
+			requirements: [
+				new Requirement({
+					name: <Trans id="mnk.bh.checklist.req.name"><ActionLink action="BROTHERHOOD"/> buffs</Trans>,
+					value: this.totalAffectedPlayers(),
+					target: this.maxPossibleTargets(),
```
ayyaruq, line 68: TBH the target is probably 100% unless the other players are dead.

src/parser/jobs/mnk/modules/Brotherhood.tsx
Malpp: Didnt know this could be this simplified, nice

src/parser/jobs/mnk/modules/Brotherhood.tsx
Malpp: I dont think so since Brotherhood has a 15 yard range
image
I also think it's good to set a target so the person can look at the log and ask themselves why didn't they hit that target of the max possible.

ayyaruq: If the target is always 100%, then it's clear that someone was out of range. The target here sets at what point the checklist turns green, so if they see green with say 50% because all the ranged/casters are in Narnia, they won't necessarily notice the error and get their party to stay in range. Only reason I can see to adjust this is for death, since that's not something the player can control, although even then they could try telling their party to not die. Thaaaat said, your function for getting the max possible is gonna be 100% always anyway, so kinda defeats the purpose of having a function.

src/parser/jobs/mnk/modules/Brotherhood.tsx
```diff
@@ -0,0 +1,105 @@
...
+		}))
+	}
+
+	private maxPossibleTargets(): number {
+		return this.windows.length * this.playerCharacters.length
+	}
+
+	private totalAffectedPlayers(): number {
+		return this.windows.map(window => window.targetsHit)
+			.reduce((previousValue, currentValue) => previousValue + currentValue, 0)
```
ayyaruq, line 81: Nit
```suggestion
		return this.windows
			.map(window => window.targetsHit)
			.reduce((total, hits) => total + hits, 0)
```

src/parser/jobs/mnk/modules/Brotherhood.tsx
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	private onRemove(event: Events['statusRemove']) {
+		const lastWindow = _.last(this.windows)
+		if (lastWindow == null) {
+			return
+		}
+
+		lastWindow.end = event.timestamp
+	}
```
ayyaruq, line 104: Prefer the callback functions to be above `onComplete` so it's clearer where hooks are processed and where utility functions live. Generally fetching the last window and mutating a history is a real pain, a better way to handle this is to have a class var for the current window, then on remove you close the window and push it to the history.

ayyaruq (changes requested):
Apologies for taking so long to look at this. I don't recall why this moved to the window interface since it was fine before, but with this approach I think it's worth tracking a current window and then pushing it into the history on closure rather than mutating the history. Easier to read that way, and safer.

eatmorehufu:
If you use the RaidBuffWindow  (https://github.com/xivanalysis/xivanalysis/blob/a5ed3628f3594c35722cfbdadb2c905285a429d9/src/parser/core/modules/ActionWindow/windows/RaidBuffWindow.ts) class you can get all of this functionality for free without duplicating all of that code.

RaidBuffWindow currently doesn't generate a suggestion or a checklist item but you (or i) can extend it to have probably a suggestion and not a checklist. More jobs using this will mean that raidbuffs will have a more uniform appearance on the site across all of the jobs, instead of being implemented slightly differently every time it shows up.

ayyaruq:
> RaidBuffWindow currently doesn't generate a suggestion or a checklist item but you (or i) can extend it to have probably a suggestion and not a checklist. More jobs using this will mean that raidbuffs will have a more uniform appearance on the site across all of the jobs, instead of being implemented slightly differently every time it shows up.

The checklist isn't really a big deal, it's more for core metrics which this isn't really, but it's also a quick reference so eh. Not sure I like how RaidBuffWindow handles the end of a window, although in this case it's tracked by Meditative Brotherhood falling off the MNK player, which actually leaves it a little brittle in the case of 2 MNKs overwriting each other. That said, I'm not sure what RBW would add here either. I'd not call this code duplicated when it's handling quite differently, even if the result is basically the same but with less junk in the way.

AkairyuGestalter:
Baseline Brotherhood tracking superseded by #1694, and player buffed tracking ought be informational only and done by using RaidBuffWindow to be consistent with other jobs (technically partly done in #1694 as well, though only for windows that missed being in Riddle of Fire, so that'll probably be worth a second pass update). Closing this PR.

PR #1627, opened 2022-02-14
Masterful Blitz stats

Malpp:
Allows someone to quickly see if they messed up by doing a celestial revolution or if the balance between all blitz are off (maybe double casted one of them)
image


ayyaruq:
I'm on the fence for this one. The chart is super nice, but I wonder if the gauge module would just show this anyway? The breakdown *is* nice tho, might be worth asking users what they would like to see, thoughts?

Malpp:
> I'm on the fence for this one. The chart is super nice, but I wonder if the gauge module would just show this anyway? The breakdown _is_ nice tho, might be worth asking users what they would like to see, thoughts?

I went in The Balance discord and asked
image
image

AkairyuGestalter (approved):
Neat piece of UI. If people don't like it they can ignore it just fine, but if it's useful for someone then cool. Lets do it.

PR #1628, opened 2022-02-14
Enemy action timeline

ackwell:
**This is a proof of concept.** It should not be merged as-is - there's significant issues with the current output that need to be resolved before it's useable.

This PR implements an "action timeline" for enemies in the fight. Ideally, this would make it possible to see exact timing of mechanics as they occur, rather than needing to cross-reference other timeline sources.

The main blocker of this current implementation is the amount of data - a lot of bosses use a _lot_ of copies of themselves to cast things, and a lot of actions have a multitude of copies with different action IDs that are effectively the same in practice.

Any solution for this would need to be something that needs as _little_ boss-specific configuration as is reasonably possible - we don't want to be manually adding support for every single boss more than we have to already.

Image of current output showing what it can do - and the problem it faces:

image

### DIFF
--- a/src/parser/core/modules/EnemyTimeline.tsx
+++ b/src/parser/core/modules/EnemyTimeline.tsx
@@ -1,0 +1,122 @@
+import Tooltip from 'components/ui/DbLink'
+import {Action} from 'data/ACTIONS'
+import {Event, Events} from 'event'
+import _ from 'lodash'
+import React from 'react'
+import {Actor, Team} from 'report'
+import {Analyser} from '../Analyser'
+import {filter, oneOf} from '../filter'
+import {dependency} from '../Injectable'
+import {SimpleItem, SimpleRow, Timeline} from './Timeline'
+
+interface Hit {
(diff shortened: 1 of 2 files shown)


Kagu-chan:
I would rather suggest - besides maybe offering this view for detail - to limit those to the actual casts visible to the players. 
 - `attack` as an event may happen, but is normally not visible to the player.
 - `True Flare` may be happening as a part of `Pitiless Flail of Purgation`, but is not visible as a cast to the players.
 - Grouping them by name rather then event id may reduce the overhead here
 - I do not know if we have the information if something is a result of a cast or the actual cast itself visible to the player
 - If we have the information i would rather limit the data actually visible on the enemy list. Sure there are invisible entities, but these are the things players figure out during progress or actively looking for it in fflogs

That are my thoughts on the topic - limiting to the information we would have ingame as well, but offering this detailed, but extensive view to those interested in the actual details.
To figure out "what have i done on which mechanic" the information i would have ingame should be more then enough.

PR #1629, opened 2022-02-14
Timeline-wide tooltip

ackwell:
**This is a proof of concept.** It should not be merged as-is - there's significant issues with the current output that need to be resolved before it's useable.

This PR implements a tooltip that covers the entire gauge's content, replacing the currently row-specific implementation used on resource graphs.

The main blockers on this are
- it's currently an absolute fucking mess of code, i apologise to anyone who looks at the diff
- blm exists

Most of the work to be done is cleanup. And working out what the hell we should do about blm.

Image of current output showing what it can do - and blm:

image


AkairyuGestalter:
Hmm, I wonder if there's a way I can compress the AF/UI gauge state into a single counter/timer with an enum flag indicating which element is the active one (or none if it dropped). That could potentially clean up a decent chunk of that mess. Alternatively, perhaps an option on the resource for 'hide from tooltip if empty' since one of the two elements will always be 0?

Also a UI nit: I would have the whitespace gap before the title of the group, rather than between the title and the resources contained in that group.

PR #1631, opened 2022-02-14
Hotfix/harvest moon

ghost:
Logic for Harvest Moon tracking got borked.

This is what I get for coding at 3 AM.

### DIFF
--- a/src/parser/jobs/rpr/modules/HarvestMoon.tsx
+++ b/src/parser/jobs/rpr/modules/HarvestMoon.tsx
@@ -36,13 +36,14 @@ export class HarvestMoon extends Analyser {
 	}
 
 	private canChargeMoon(inputWindow: {start: number, end: number}): boolean {
+		const ADJUSTED_CAST_TIME = this.data.actions.SOULSOW.castTime + SOULSOW_BUFFER
 		const unableToActWindow = this.unableToAct.getWindows(inputWindow)[0]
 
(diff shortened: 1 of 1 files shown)


eatmorehufu (approved):
I think the logic is right, but I think this code should be refactored in a functional style to be less error-prone and more readable.

The added comments don't seem totally necessary, like:
```
// True if there are ADJUSTED_CAST_TIME milliseconds or more remaining in the inputWindow
return inputWindow.end - inputWindow.start >= ADJUSTED_CAST_TIME
```

ayyaruq (approved):
Nice docs, thanks!

ayyaruq:
> I think the logic is right, but I think this code should be refactored in a functional style to be less error-prone and more readable.
> 
> The added comments don't seem totally necessary, like:
> 
> ```
> // True if there are ADJUSTED_CAST_TIME milliseconds or more remaining in the inputWindow
> return inputWindow.end - inputWindow.start >= ADJUSTED_CAST_TIME
> ```

Preferring to have working code over nice code. Part of the TODO for RPR involves a code cleanup but I'd like to do that after we have the bulk of the functionality. RPR has had a lot of contributions from many different devs and the code reflects that, but trying to fix that in every PR just makes merges take longer and usually gets clobbered in subsequent PRs anyway. Keen to see how you think this should look in a more functional style tho.

PR #1633, opened 2022-02-15
Refactor Harvest Moon module

eatmorehufu:
Refactoring means writing longer code right.

### DIFF
--- a/src/parser/jobs/rpr/modules/HarvestMoon.tsx
+++ b/src/parser/jobs/rpr/modules/HarvestMoon.tsx
@@ -67,6 +67,14 @@ export class HarvestMoon extends Analyser {
 	// to calculate the actable Windows. Guarantees at least 2 windows.
 	private unableToActWindows(invulWindow: Window): Window[] {
 		const unactableWindows = this.unableToAct.getWindows(invulWindow) as Window[]
+
+		// Edge case where the unableToAct window is larger than the invulWindow.
+		if (this.unableToActWindows.length === 1
+			&& unactableWindows[0].start < invulWindow.start
+			&& unactableWindows[0].end > invulWindow.end) {
+			return unactableWindows.concat(this.dummyWindow(unactableWindows[0].end))
+		}
+
 		if (unactableWindows.length === 0 || unactableWindows[0].start > invulWindow.start) {
 			unactableWindows.unshift(this.dummyWindow(invulWindow.start))
 		}
--- a/src/parser/jobs/rpr/modules/HarvestMoon.tsx
+++ b/src/parser/jobs/rpr/modules/HarvestMoon.tsx
@@ -6,7 +6,7 @@ import {filter} from 'parser/core/filter'
 import {dependency} from 'parser/core/Injectable'
 import Checklist, {Requirement, Rule} from 'parser/core/modules/Checklist'
 import {Data} from 'parser/core/modules/Data'
-import {Invulnerability} from 'parser/core/modules/Invulnerability'
+import {Window, Invulnerability} from 'parser/core/modules/Invulnerability'
 import {UnableToAct} from 'parser/core/modules/UnableToAct'
 import React from 'react'
 
@@ -21,6 +21,7 @@ export class HarvestMoon extends Analyser {
 	@dependency private unableToAct!: UnableToAct
 
 	private harvestMoonCasts = 0
+	private ADJUSTED_CAST = this.data.actions.SOULSOW.castTime + SOULSOW_BUFFER
 
 	override initialise() {
 		super.initialise()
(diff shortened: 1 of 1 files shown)


src/parser/jobs/rpr/modules/HarvestMoon.tsx
```diff
@@ -35,37 +36,60 @@ export class HarvestMoon extends Analyser {
...
+	private unableToActWindows(invulWindow: Window): Window[] {
+		const unactableWindows = this.unableToAct.getWindows(invulWindow) as Window[]
 
-		// will only run the window filter if there is a non-zero amount of unableToAct time during the fight.
-		if (this.unableToAct.getDuration({start: this.parser.pull.timestamp, end: this.parser.pull.timestamp + this.parser.pull.duration}) > 0) {
-			return invulnWindows.filter(window => this.canChargeMoon(window)).length + 1
+		// Edge case where the unableToAct window is larger than the invulWindow.
+		if (this.unableToActWindows.length === 1
+			&& unactableWindows[0].start < invulWindow.start
+			&& unactableWindows[0].end > invulWindow.end) {
```
ghost: This is quite an edge case, is there any way to simplify this?

src/parser/jobs/rpr/modules/HarvestMoon.tsx
```diff
@@ -35,37 +36,60 @@ export class HarvestMoon extends Analyser {
...
+		return this.unableToActWindows(invulWindow)
+			.reduce(this.toActableWindows, [] as Window[])
+			.some(this.longerThanChargeTime.bind(this))
+	}
 
-		// return true if there are ADJUSTED_CAST_TIME milliseconds between the beginning of the window being checked and the beginning of an unableToAct window
-		if (unableToActWindow.start - inputWindow.start >= ADJUSTED_CAST_TIME) {
-			return true
+	// Reduce function for calculating interstitial windows that assumes that there are at least 2 windows.
+	private toActableWindows(acc: Window[], window: Window, idx: number, arr: Window[]): Window[] {
```
ghost: Ack generally prefers that you not abbreviate or shorten your variable / parameter names.

src/parser/jobs/rpr/modules/HarvestMoon.tsx
```diff
@@ -35,37 +36,60 @@ export class HarvestMoon extends Analyser {
...
+			return unactableWindows.concat(this.dummyWindow(unactableWindows[0].end))
 		}
 
-		return invulnWindows.length + 1
+		if (unactableWindows.length === 0 || unactableWindows[0].start > invulWindow.start) {
+			unactableWindows.unshift(this.dummyWindow(invulWindow.start))
+		}
+		if (unactableWindows[unactableWindows.length - 1].end < invulWindow.end) {
+			unactableWindows.push(this.dummyWindow(invulWindow.end))
+		}
```
ghost: This section is pretty unclear, at least to someone who isn't familiar with functional programming in JS. Maybe add some comments for maintainability?

ayyaruq:
Why does this need refactoring? What are you trying to achieve with a refactor? Your PR description didn't make this clear.

src/parser/jobs/rpr/modules/HarvestMoon.tsx
eatmorehufu: Fixed this a little bit but it's kind of the same. When looking at time intervals, they can overlap in 4 ways:
```
1: Overlap at the start of "B" and end of "A"
AAAAA
  BBBBBBBBBBBBB
 2. Overlap at the end of "B" and start of "A"
   AAAAA
 BBBBB
 3. "A" is a subset of "B"
     AAA
BBBBBBBBB
 4. "B" is a subset of "A"
   AAAAAAAAA
      BBB
```

The conditionals here handle all 4 of these situations, and the edge case is 4, where "A" is the unactable window and "B" is the invulWindow. I think this probably won't ever happen in a real fight but it should be handled anyway.

src/parser/jobs/rpr/modules/HarvestMoon.tsx
eatmorehufu: Fixed this with more expressive variable names.

src/parser/jobs/rpr/modules/HarvestMoon.tsx
eatmorehufu: I don't know that I would call this quote a functional programming style. The purpose of this method is to handle the 4 interval overlapping cases I listed above, and I reworded and added more comments to hopefully clear up what it's doing and why it's doing it.

eatmorehufu:
I did it in part because you sort of asked to see this here: https://github.com/xivanalysis/xivanalysis/pull/1631#issuecomment-1039751789

I also did it because there's some duplicate logic/definitions in the existing implementation. See:
- ADJUSTED_CAST_TIME defined twice: [L39](https://github.com/xivanalysis/xivanalysis/blob/endwalker/src/parser/jobs/rpr/modules/HarvestMoon.tsx#L39) [L60](https://github.com/xivanalysis/xivanalysis/blob/endwalker/src/parser/jobs/rpr/modules/HarvestMoon.tsx#L60)
- Duration of window checked against the cast time 3 times: [L47](https://github.com/xivanalysis/xivanalysis/blob/endwalker/src/parser/jobs/rpr/modules/HarvestMoon.tsx#L47) [L51](https://github.com/xivanalysis/xivanalysis/blob/endwalker/src/parser/jobs/rpr/modules/HarvestMoon.tsx#L51) [L61](https://github.com/xivanalysis/xivanalysis/blob/endwalker/src/parser/jobs/rpr/modules/HarvestMoon.tsx#L61)
  - TBF I also check this twice, as a (somewhat unnecessary) mini-optimization.
- Logic of starting with a pre-charged Harvest Moon duplicated twice [L65](https://github.com/xivanalysis/xivanalysis/blob/endwalker/src/parser/jobs/rpr/modules/HarvestMoon.tsx#L65) [L68](https://github.com/xivanalysis/xivanalysis/blob/endwalker/src/parser/jobs/rpr/modules/HarvestMoon.tsx#L68)

Having all of these things defined separately makes maintaining this more bug-prone because you have to change logic in multiple places every time anything needs to change.

PR #1635, opened 2022-02-17
Change Arcane Circle to simulate full duration when overridden

eatmorehufu:
Instead of cutting it short.

Use the new utility functions for SearingLight as well since it 's doing the same checks.


src/parser/jobs/rpr/modules/ArcaneCircle.tsx
```diff
@@ -38,6 +47,13 @@ export class ArcaneCircle extends BuffWindow {
 		}))
 	}
 
+	private maybeReOpenPreviousWindow(event: Events['statusApply']) {
+		// See if another reaper overrode your Arcane Circle
+		if ((this.history.endOfLastEntry() ?? 0) === event.timestamp) {
```
KelosPelmand: I think this should be able to compare without the `??` and will return false if `endOfLastEntry` is undefined.

```suggestion
		if (this.history.endOfLastEntry() === event.timestamp) {
```

src/parser/jobs/smn/modules/SearingLight.tsx
```diff
@@ -165,12 +165,9 @@ export class SearingLight extends Analyser {
 			// If there is no active window, see if the last window just ended
 			// at the exact same timestamp.  If it did, this is likely due to
 			// having multiple summoners, so reopen the last window
-			if (this.history.entries.length > 0) {
-				const last = this.history.entries[this.history.entries.length - 1]
-				if ((last.end ?? 0) === timestamp) {
-					last.end = undefined
-					return
-				}
+			if ((this.history.endOfLastEntry() ?? 0) === timestamp) {
```
KelosPelmand: Same as above

```suggestion
			if (this.history.endOfLastEntry === timestamp) {
```

src/parser/core/modules/ActionWindow/windows/BuffWindow.ts
```diff
@@ -60,14 +60,29 @@ export abstract class BuffWindow extends ActionWindow {
...
+		}
+	}
+
 	private startWindowAndTimeout(event: Events['statusApply']) {
-		this.onWindowStart(event.timestamp)
-		const duration = this.data.getStatus(event.status)?.duration
+		this.startWindow(event.timestamp, this.data.getStatus(event.status)?.duration)
+	}
+
+	private startWindow(timestamp: number, duration?: number) {
```
KelosPelmand: The function naming gets a bit confusing here.  This function should probably be called `startWindowAndTimeout`, since it is doing that, and the other function should probably be renamed to `onSelfBuffApply` or something if it stays a separate function from the party member apply (see below).

src/parser/jobs/rpr/modules/ArcaneCircle.tsx
```diff
@@ -18,6 +20,13 @@ export class ArcaneCircle extends BuffWindow {
 	override initialise() {
 		super.initialise()
 
+		// Multiple reapers can override each other's buffs
+		const dupeFilter = filter<Event>().source(noneOf([this.parser.actor.id]))
+			.target(this.parser.actor.id)
+			.status(this.buffStatus.id)
+
+		this.addEventHook(dupeFilter.type('statusApply'), this.maybeReOpenPreviousWindow)
```
KelosPelmand: The fact that this has to be duplicated into every class that needs it is why I suggested making a `RaidBuffWindow` class or doing what someone else (Hint, I think?) suggested and making an option in `BuffWindow` that says the buff can be overwritten and have the base class set these extra hooks up itself (and include maybeReOpen)

In fact, you probably don't even need an option for the window collapsing.  You should be able to just change the hook to take any source, then check in the hook function if the source was the player or not and handle accordingly.

I'm still torn on if a separate class is better or not.  There's just enough code in BuffWindow that would have to be copied to RaidBuffWindow that duplicating and modifying is annoying, but RaidBuffWindow could also handle the players hit logic.

KelosPelmand (changes requested):
General idea seems sound, but there are probably better ways to organize it for easier reuse by other jobs.

src/parser/core/modules/ActionWindow/windows/BuffWindow.ts
eatmorehufu: Renamed, good note.

src/parser/jobs/rpr/modules/ArcaneCircle.tsx
eatmorehufu: Moved the logic to BuffWindow and put it behind an option that jobs can set for themselves. I'm not sure what you mean by an option for window collapsing, but I think it's more readable to have two hooks (for player-owned IDs, and not) and do flow control there, rather than a single hook that then has conditionals to handle whodunit.

eatmorehufu:
arcane circle

src/parser/core/modules/ActionWindow/windows/RaidBuffWindow.ts
```diff
@@ -0,0 +1,79 @@
+import {Event, Events} from 'event'
+import {BuffWindow, EvaluatedAction, PlayersBuffedEvaluator} from 'parser/core/modules/ActionWindow'
+import {HistoryEntry} from 'parser/core/modules/ActionWindow/History'
+import {ensureArray} from 'utilities'
+import {filter, oneOf, noneOf} from '../../../filter'
+
+const FULL_PARTY = 8
```
ayyaruq: This makes assumptions that aren't guaranteed.

src/parser/core/modules/ActionWindow/windows/RaidBuffWindow.ts
```diff
@@ -0,0 +1,79 @@
...
+	 *
+	 * These might ideally belong in the History, but it would break the
+	 * current paradigm of the ActionWindow.
+	 */
+	protected raidBuffApplications = new Array<Events['statusApply']>()
+
+	/**
+	 * Expected number of people who should get the raid buff every time.
+	 */
+	protected expectedCount = FULL_PARTY
```
ayyaruq: You get the full party in `partyMembers`, could you not use `length` on that to get the actual party size?

src/parser/core/modules/ActionWindow/windows/RaidBuffWindow.ts
```diff
@@ -0,0 +1,79 @@
...
+			.status(oneOf(ensureArray(this.buffStatus).map(s => s.id)))
+
+		this.addEventHook(statusFilter
+			.source(oneOf(playerOwnedIds))
+			.target(oneOf(partyMembers)), this.onRaidBuffApply)
+
+		// Duplicate jobs can override buffs
+		this.addEventHook(statusFilter
+			.source(noneOf(playerOwnedIds))
+			.target(this.parser.actor.id), this.maybeReOpenPreviousWindow)
```
ayyaruq: Our code style is to prefer single line or multiline, but avoid mixing and matching like this. Either each segment gets it's own line, or you just keep it to one. Single line is self explanatory, this suggestion is how multi would look.
```suggestion
		this.addEventHook(
			statusFilter
				.source(oneOf(playerOwnedIds))
				.target(oneOf(partyMembers)),
			this.onRaidBuffApply
		)

		// Duplicate jobs can override buffs
		this.addEventHook(
			statusFilter
				.source(noneOf(playerOwnedIds))
				.target(this.parser.actor.id),
			this.maybeReOpenPreviousWindow
		)
```

src/parser/core/modules/ActionWindow/windows/BuffWindow.ts
```diff
@@ -50,28 +51,42 @@ export abstract class BuffWindow extends ActionWindow {
...
-			.source(this.parser.actor.id)
+			.source(oneOf(playerOwnedIds))
 			.target(oneOf(targets))
 			.status(oneOf(ensureArray(this.buffStatus).map(s => s.id)))
 
-		this.addEventHook(buffFilter.type('statusApply'), this.startWindowAndTimeout)
-		this.addEventHook(buffFilter.type('statusRemove'), this.endWindowByStatus)
+		this.addEventHook(buffFilter.type('statusApply'), this.onStatusApply)
+		this.addEventHook(buffFilter.type('statusRemove'), this.onStatusRemove)
+		this.buffDuration = _.max(ensureArray(this.buffStatus).map(s => s.duration))
```
ayyaruq, line 65: Does this need lodash? Not requesting any change, just wondering if there's a benefit to lodash over normal `Math.max` and spreading the array which seems like it would be more readable.

ayyaruq (commented):
Makes sense to me, although I'd sort out some code style things. The party size bit makes a lot of assumptions that aren't correct for dungeons, alliance raids, or just undersized parties. I'm not seeing the Plentiful Harvest bits here tho, unless that was meant to be a diff PR?

src/parser/core/modules/ActionWindow/windows/BuffWindow.ts
eatmorehufu: I kept the implementation that already existed in the BuffWindow, I don't have an opinion about whether lodash is necessary.

src/parser/core/modules/ActionWindow/windows/RaidBuffWindow.ts
```diff
@@ -0,0 +1,84 @@
...
+	private onRaidBuffApply(event: Events['statusApply']) {
+		this.raidBuffApplications.push(event)
+	}
+
+	private affectedPlayers(buffWindow: HistoryEntry<EvaluatedAction[]>): number {
+		const actualWindowDuration = (buffWindow?.end ?? buffWindow.start) - buffWindow.start
+		// count the number of applications that happened in the window
+		const affected = this.raidBuffApplications.filter(event => {
+			return (buffWindow.start <= event.timestamp &&
+				event.timestamp <= buffWindow.start + actualWindowDuration)
```
KelosPelmand: I think it would make more sense to store `const windowEnd = buffWindow.end ?? buffWindow.start` since the arithmetic is cancelling out anyway.

KelosPelmand (approved):
One minor efficiency point, but otherwise looks good to go.

src/parser/core/modules/ActionWindow/windows/RaidBuffWindow.ts
eatmorehufu: Done.

ayyaruq (approved):
LGTM

PR #1636, opened 2022-02-17
Clean up Esprit generation sim

AkairyuGestalter:
DNC's own self-generation has actually been simplified down to just the 4 combo GCDs and the 4 proc GCDs, so we can simplify the hooks a bit. Since I'm tinkering with things, I'm also adding in the theorycrafter-tested more-specific generation rates that've been determined for specific jobs. Where one isn't known, we'll fall back to the rough-estimate 20% we've been using across the board up until now.

Additional notes:
- Assumed that the now-weaponskill Finishes/Tillana can generate for another DNC if the executing DNC is their partner or has their Technical Finish Esprit buff active, so I removed them from the generation exceptions list
- Assumed that Kaeshi: Namikiri doesn't generate Esprit since it's classed as an Ability like the other Kaeshi Iaijutsu, so it's now included in the exception list
- It'd be awesome if someday the action data could include a flag for Weaponskill vs. Ability vs. Spell so I don't have to keep maintaining this list 🙃

### DIFF
--- a/src/parser/jobs/dnc/changelog.tsx
+++ b/src/parser/jobs/dnc/changelog.tsx
@@ -52,4 +52,9 @@ export const changelog = [
 		Changes: () => <>Update AoE checks for Windmill and mark supported for 6.08.</>,
 		contributors: [CONTRIBUTORS.AKAIRYU],
 	},
+	{
+		date: new Date('2022-02-16'),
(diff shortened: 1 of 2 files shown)


ayyaruq (approved):
Math seems suss but code works so yolo adjust later if/as needed.

PR #1638, opened 2022-02-17
Reaper Cleanup and 6.1 support

ayyaruq:
Buncha changes but it's mostly restructuring for consistency and ease of reading/maintaining, and timeline/navigation stuff. Immortal Sacrifice is removed as Arcane Circle covers it more effectively. This also fixes the rushed window issue in AC. Defensive cooldowns has been temporarily removed until display issues can be sorted out more globally.

Like GNB, Soul Slice/Scythe actions in the timeline are also duplicated due to how they're on the GCD but not the GCD's CDG.

Ignoring the issue of Enshroud severities in this PR so it can be dealt with properly later. There is nothing further to do in this PR, it's ready for review/merge.


ayyaruq:
This closes #1372 as well.

Rjb99:
Defensives don't seem to be displaying properly
image

Rjb99:
Soul Slice is duplicated outside of GCD line, Why?
image

src/parser/jobs/rpr/modules/DISPLAY_ORDER.ts
```diff
@@ -1,6 +1,13 @@
...
+	ARCANE_CIRCLE: 2,
+	ENSHROUD: 3,
+
+	// Super low priority, some of these override default ordering
+	AOE: 50,
+	COMBOS: 51,
+	POSITIONALS: 52,
+	WEAVING: 53,
+	INTERRUPTS: 54,
+	TINCTURES: 55,
```
Rjb99, line 12: Nit; I'd put Tinctures under the rest of Reaper unique logic as we have expected actions which makes it something of note to check.

Arcane Circle > Enshroud > Tinctures > Positional >weaving issues > combo issues > Incorrect AoE Action Usage is the order I'd want to check. but perhaps keeping the "super low priority" ones in the same order as other jobs for consistency would be better.

src/parser/jobs/rpr/modules/DISPLAY_ORDER.ts
```diff
@@ -1,6 +1,13 @@
 export const DISPLAY_ORDER = {
-	IMMORTAL_SACRIFICE: 10,
-	ARCANE_CIRCLE: 15,
-	INTERRUPTS: 20,
-	ENSHROUD: 90,
+	IMMORTAL_SACRIFICE: 1,
```
Rjb99, line 2: Why is a removed file in the Display Order?

ayyaruq:
> Soul Slice is duplicated outside of GCD line, Why?

As noted in the PR description, for the same reason Gnashing is for GNB. GCDs that are in a separate cooldown group to the main GCD group get duplicated for some reason.

ayyaruq:
> Defensives don't seem to be displaying properly

Stat box isn't great for more than 2 lines and RPR has more than 2 defensives. I didn't add this feature and it doesn't scale but not sure if it should be removed, at least until it's more thought out. Open to suggestions.

src/parser/jobs/rpr/modules/Interrupts.tsx
```diff
@@ -8,8 +8,7 @@ export class Interrupts extends CoreInterrupts {
 	static override displayOrder = DISPLAY_ORDER.INTERRUPTS
 
 	override suggestionContent = <Trans id="rpr.interrupts.suggestion.content">
-		Avoid interrupting casts by either prepositioning yourself, utilising slidecasting where possible.
-		If you need to move, consider using <ActionLink action="HELLS_INGRESS"/> or <ActionLink action="HELLS_EGRESS"/>
-		and getting a free <ActionLink action="HARPE"/> out of it.
+		Avoid interrupting casts by either pre-positioning yourself or slidecasting where possible.
+		If you need to move, consider using <ActionLink action="HELLS_INGRESS"/> or <ActionLink action="HELLS_EGRESS"/> and getting a free <ActionLink action="HARPE"/> out of it.
```
Rjb99: Nit; Not sure I like the "free Harpe" in this sentence as even if the harpe is free,  Sentence seems to imply harpe should be used in pair with movement when that is not always the case.

src/parser/jobs/rpr/modules/DefensiveCooldowns.tsx
```diff
@@ -0,0 +1,10 @@
+import {Defensives} from 'parser/core/modules/Defensives'
+
+export class DefensiveCooldowns extends Defensives {
+	protected override trackedDefensives = [
+		this.data.actions.ARCANE_CREST,
+		this.data.actions.FEINT,
+		this.data.actions.SECOND_WIND,
+		this.data.actions.BLOODBATH,
+	]
+}
```
Rjb99: Nit; Why did you remove the Info part of Defensives? It's not a bad idea to explain at a minimum how Arcane Crest and Feint can support the party like it was doing before.

src/parser/jobs/rpr/modules/DefensiveCooldowns.tsx
```diff
@@ -0,0 +1,10 @@
+import {Defensives} from 'parser/core/modules/Defensives'
+
+export class DefensiveCooldowns extends Defensives {
+	protected override trackedDefensives = [
+		this.data.actions.ARCANE_CREST,
+		this.data.actions.FEINT,
+		this.data.actions.SECOND_WIND,
```
Rjb99: Blocking; if stat block isn't good for more than 2, keep the party support options, toss the self-healing for now. it's useful infomation to display.

src/parser/jobs/rpr/modules/Enshroud.tsx
```diff
@@ -13,6 +13,12 @@ import {DISPLAY_ORDER} from './DISPLAY_ORDER'
 
 const REAPINGS_PER_SHROUD = 4
 
+const SEVERITIES = {
+	1: SEVERITY.MINOR,
+	3: SEVERITY.MEDIUM,
```
Rjb99, line 18: Nit; Given that all 4 reapings are greater 2 communios in potency in the same amount of time, I feel like the Severity could be harsher to the player but I'm okay with letting better reaper players than I fine-tune severities

src/data/ACTIONS/root/RPR.ts
```diff
@@ -280,6 +280,8 @@ export const RPR = ensureActions({
 
 	// -----
 	// Soul Gauge oGCDs
+	// Note:
+	//  all of these grant Reaver, but we only keep it on Gluttony since that's the only real use at level cap
```
Rjb99, line 284: Not sure I'm fully understanding why that makes a difference, if it generates a status, it generates a status

Rjb99 (changes requested):
Mostly Nits, only blocking comment is on Defensives, reduce it back to the current Arcane Crest & Feint if the stat block doesn't scale past 2 items well.

src/parser/jobs/rpr/modules/Interrupts.tsx
eatmorehufu: To match the default suggestion of `Avoid interrupting casts by either prepositioning yourself or utilizing slidecasting where possible. If you have to move, try to save an instant cast to keep your GCD rolling.` you can try:

`If you have to move, consider using <ActionLink action="HELLS_INGRESS"/> or <ActionLink action="HELLS_EGRESS"/> and using the instant cast <ActionLink action="HARPE"/> to keep your GCD rolling.`

src/parser/jobs/rpr/modules/ArcaneCircle.tsx
```diff
@@ -20,28 +29,54 @@ export class ArcaneCircle extends RaidBuffWindow {
...
+				{
+					action: this.data.actions.PLENTIFUL_HARVEST,
+					expectedPerWindow: 1,
+				},
 			],
 			suggestionIcon: this.data.actions.ARCANE_CIRCLE.icon,
 			suggestionContent: <Trans id="rpr.arcanecircle.suggestions.content">
-				Each <ActionLink action="ARCANE_CIRCLE"/> window should contain 2 uses of <ActionLink action="COMMUNIO"/> and a use
-				of <ActionLink action="PLENTIFUL_HARVEST"/>.
+				Each <ActionLink action="ARCANE_CIRCLE"/> window should contain 2 uses of <ActionLink action="COMMUNIO"/>
```
eatmorehufu, line 43: I think a more wordy but also more explicit suggestion would be better. Theoretically this language should also be standardized across the site:

"Each {buff} window [after the opener] should contain N uses of {skill} [, M uses of {another_skill}, and O uses of {third_skill}.] The {buff} window in the opener should contain N uses of {skill} [, M uses of {another_skill}, and O uses of {third_skill}.]"

The purpose of this writing is to be informative/instructional, and not to sound necessarily like speech.

src/parser/jobs/rpr/modules/HarvestMoon.tsx
```diff
@@ -30,69 +30,69 @@ export class HarvestMoon extends Analyser {
...
+		const harvestsUsedPercent = this.getUsedPercentage(expectedUses)
 
 		this.checklist.add(new Rule({
 			name: <Trans id="rpr.harvestmoon.checklist.title">
-				Use <DataLink action="SOULSOW" /> during downtime.
+				Use <DataLink action="SOULSOW" /> during downtime
 			</Trans>,
 			description: <Trans id="rpr.harvestmoon.checklist.description">
-				Harvest Moon is one of your highest potency abilities. Aim to get at least one use each time the boss is targetable.
+				<DataLink action="HARVEST_MOON"/> is one of your highest damage abilities. Aim to get at least one use each time the boss is targetable.
```
eatmorehufu: I think this suggestion is a little weird, I took a stab at it but I think it still needs more work:
`<DataLink action="HARVEST_MOON"/> is one of your highest damage abilities. Use it before the end of the fight or before any downtime that is long enough to recharge it.`

Also, this module I think doesn't care if someone just charges soul sow during uptime, and maybe it should? It's incorrect to charge it during uptime.

src/parser/jobs/rpr/modules/DISPLAY_ORDER.ts
ayyaruq: It's immortal. Just a var name tho, not a file in this case.

src/data/ACTIONS/root/RPR.ts
ayyaruq: This is due to a bug with how timeline (and I guess prepull statuses) work when multiple actions define the status applied. Comment explains why it's specifically on Gluttony, beyond that this is in alignment with how other jobs handle a single status coming from multiple actions.

src/parser/jobs/rpr/modules/DefensiveCooldowns.tsx
ayyaruq: I'm removing this module over tuning it for now. End of the day, the way the module works is counter to what it's meant to convey. If this changes, happy to reintroduce it but I don't wanna force that change since other jobs may have different concerns.

src/parser/jobs/rpr/modules/DISPLAY_ORDER.ts
ayyaruq: This probably needs more standardisation in core, module titles as well. Gonna leave it for now even if it's less consistent.

AkairyuGestalter (approved):

PR #1639, opened 2022-02-18
Change Embolden to use RaidBuffWindow.

eatmorehufu:
This should be a no-op to the existing Embolden module functionality. The physical skills are still ignored in the rotation window.


eatmorehufu:
Added Searing Light to this PR as well.

src/parser/core/modules/ActionWindow/evaluators/PlayersBuffedEvaluator.tsx
```diff
@@ -1,26 +1,31 @@
 import {Trans} from '@lingui/react'
 import {EvaluatedAction, EvaluationOutput, WindowEvaluator} from 'parser/core/modules/ActionWindow'
 import {HistoryEntry} from 'parser/core/modules/ActionWindow/History'
+import {Suggestion} from 'parser/core/modules/Suggestions'
 import React from 'react'
 
 interface TrackedBuffOpts {
 	expectedCount: number,
 	affectedPlayers: (window: HistoryEntry<EvaluatedAction[]>) => number
+	suggestion: (windows: Array<HistoryEntry<EvaluatedAction[]>>) => Suggestion | undefined
```
KelosPelmand: This is not the right pattern for suggestion generation.  (You are requiring duplicate code in all the implementors.)  As with other evaluators, TrackedBuffOpts should have

```
suggestionIcon: string
suggestionContent: JSX.Element
suggestionWindowName: JSX.Element
```

and do the suggestion logic itself.  If some jobs with raid buffs strongly object to having a suggestion, add a `suppressSuggestion: boolean` that defaults to false and check that in the suggest method of PlayersBuffedEvaluator.  If classes being moved to RaidBuffWindow just didn't have a suggestion, I would guess it was more that they didn't think about it than that they didn't do it.

src/parser/core/modules/ActionWindow/windows/ActionWindow.tsx
```diff
@@ -137,7 +137,7 @@ export abstract class ActionWindow extends Analyser {
 		this.addEventHook('complete', this.onComplete)
 	}
 
-	private onComplete() {
+	protected onComplete() {
```
KelosPelmand: This should return to private after making PlayersBuffedEvaluator do it's own suggestions.

```suggestion
	private onComplete() {
```

src/parser/core/modules/ActionWindow/windows/ActionWindow.tsx
```diff
@@ -20,7 +20,7 @@ import {History, HistoryEntry} from '../History'
 export abstract class ActionWindow extends Analyser {
 
 	@dependency protected data!: Data
-	@dependency private suggestions!: Suggestions
+	@dependency protected suggestions!: Suggestions
```
KelosPelmand: This should return to private after making PlayersBuffedEvaluator do it's own suggestions.

```suggestion
	@dependency private suggestions!: Suggestions
```

KelosPelmand (changes requested):
Please have PlayersBuffedEvaluator create its own suggestions like the other evaluators do.

I did not check Embolden at all, but it turns out the the Searing Light one doesn't actually work in some cases.  Due to the buff being pet sourced, it is possible for another player to get the buff before you do, which results in them not being counted as buffed in the players buffed count.  (See the 6:18 window in https://xivanalysis.com/fflogs/gAbwHX4BY9rC7n6d/7/7).

I think it would be better to leave the SL file not using BuffWindow rather than try to add a bunch of special hook points to BuffWindow just to make the one existing pet sourced buff work with it.

eatmorehufu:
I reverted the SL changes to keep this PR simple with just RDM changes. I'll add a followup for passing through suggestions later.

Cephid:
If Kelos is happy with the fixes addressing the original concerns I'm fine with this.  The output is identical to what we had, and this removes maintenance further down the road.

ayyaruq (approved):
This should have a code style pass at some point.

Issue #1640, opened 2022-02-18
[SMN] [Feature Request] Egi Order in phases.


Iaotle:
As it is we have to look at the timestamps for each egi. It would be better to have them sorted by time, and put on the small header for each demi phase.

![image](https://user-images.githubusercontent.com/43072246/154688069-4b1d397e-9630-4651-adf7-a71562a5ba82.png)

Alternative:
![image](https://user-images.githubusercontent.com/43072246/154692252-048a0834-6518-4d8a-9067-936a74e8fd4c.png)

Rjb99:
Are they not currently sorted by time? Your pictures change just hides the time stamp which I think can be useful for seeing if there were delays between the egis.

Iaotle:
They are not, in fact, sorted by time:

![image](https://user-images.githubusercontent.com/43072246/154694391-17e4686f-c6ef-4812-be2e-c0c36a066d29.png)


My changes don't hide anything as the text on the header is something I added:
![image](https://user-images.githubusercontent.com/43072246/154694655-a997af6e-2ac9-4ad3-9b62-c59bcca4c267.png)

Iaotle:
This would give a great overview of the kind of things that I want to see in a rotation for a given parse. Egi order would be easy to look at without reading timestamps.

PR #1641, opened 2022-02-19
egi sorting + egi on header

Iaotle:
Sorts egis by time, and adds an easy to see header for each row, like I proposed in https://github.com/xivanalysis/xivanalysis/issues/1640.

image

### DIFF
--- a/src/parser/jobs/smn/modules/Summons.tsx
+++ b/src/parser/jobs/smn/modules/Summons.tsx
@@ -313,7 +313,7 @@ export class Summons extends Analyser {
 				key: summon.start,
 				title: {
 					content: <>
-						{<>{this.parser.formatEpochTimestamp(summon.start)}: <ActionLink {...this.data.getAction(summon.data.demiSummon.action)} /></>}
+						{this.parser.formatEpochTimestamp(summon.start)}: <ActionLink {...this.data.getAction(summon.data.demiSummon.action)} />
 						{this.getEgiIcons(summon)}
 					</>,
 				},
@@ -396,11 +396,11 @@ export class Summons extends Analyser {
 				<Table.Row>
 					<Table.Cell positive={summon.data.garudaSummon != null}>
 						<ActionLink showName={false} action="SUMMON_GARUDA_II" />
-								&nbsp;{this.printUsageTime(summon.data.garudaSummon)}
+						&nbsp;{this.printUsageTime(summon.data.garudaSummon)}
 					</Table.Cell>
 					<Table.Cell positive={summon.data.emeraldGcds === MAX_POSSIBLE_EMERALD_GCDS}>
 						<ActionLink showName={false} action="EMERALD_RITE" />/<ActionLink showName={false} action="EMERALD_CATASTROPHE" />
-								&nbsp;{summon.data.emeraldGcds}
+						&nbsp;{summon.data.emeraldGcds}
 					</Table.Cell>
 					<Table.Cell positive={summon.data.slipstream === MAX_POSSIBLE_SLIPSTREAM}>
 						<ActionLink showName={false} action="SLIPSTREAM" />
--- a/src/parser/jobs/smn/modules/Summons.tsx
+++ b/src/parser/jobs/smn/modules/Summons.tsx
@@ -335,84 +335,11 @@ export class Summons extends Analyser {
 		egis.forEach(egi => {
 			if (egi !== undefined) {
 				if (egi === summon.data.ifritSummon) {
-					retval.push(< > <ActionLink showName={false} action="SUMMON_IFRIT_II" />&nbsp;{this.printUsageTime(summon.data.ifritSummon)} </>)
+					retval.push(< > <ActionLink showName={false} action="SUMMON_IFRIT_II" />&nbsp; </>)
 				} else if (egi === summon.data.garudaSummon) {
-					retval.push(<> <ActionLink showName={false} action="SUMMON_GARUDA_II" />&nbsp;{this.printUsageTime(summon.data.garudaSummon)} </>)
+					retval.push(<> <ActionLink showName={false} action="SUMMON_GARUDA_II" />&nbsp; </>)
 				} else if (egi === summon.data.titanSummon) {
-					retval.push(<> <ActionLink showName={false} action="SUMMON_TITAN_II" />&nbsp;{this.printUsageTime(summon.data.titanSummon)} </>)
-				}
-			}
-		})
-		return retval
-	}
-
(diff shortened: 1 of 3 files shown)


src/parser/jobs/smn/modules/Summons.tsx
```diff
@@ -313,31 +313,57 @@ export class Summons extends Analyser {
 				key: summon.start,
 				title: {
 					content: <>
-						{this.parser.formatEpochTimestamp(summon.start)}: <ActionLink {...this.data.getAction(summon.data.demiSummon.action)} />
+						{<span style={{ width: '200px', display: 'inline-block' }}>{this.parser.formatEpochTimestamp(summon.start)}: <ActionLink {...this.data.getAction(summon.data.demiSummon.action)} /></span>}
```
KelosPelmand: Fixed widths like this are tricky with international sites like xiva.  (Both Japanese and German break in this case.)  I think we will need to either accept the egi icons not lining up or right-aligning them.

Given that this is a relatively fixed list of strings and translations, it's tempting to just size to the largest string, but I'm not sure that will hold up when other browsers, personal settings, and mobile gets involved.

src/parser/jobs/smn/modules/Summons.tsx
```diff
@@ -313,31 +313,57 @@ export class Summons extends Analyser {
...
+					retval.push(<> <ActionLink showName={false} action="SUMMON_GARUDA_II" />&nbsp;{this.printUsageTime(summon.data.garudaSummon)} </>)
+				} else if (egi === summon.data.titanSummon) {
+					retval.push(<> <ActionLink showName={false} action="SUMMON_TITAN_II" />&nbsp;{this.printUsageTime(summon.data.titanSummon)} </>)
+				}
+			}
+		})
+		return retval
+	}
+
+	private getEgiRows(summon: HistoryEntry<SummonWindow>) {
```
KelosPelmand: Please return the table within each row to the original fixed sort order.

I did consider sorting by egi cast time when we first implemented this table but decided against it.  Fixed order was chosen to make it easier to scan through the tables since Ifrit/Titan/Garuda will always be in the same location.

src/parser/jobs/smn/modules/Summons.tsx
```diff
@@ -313,31 +313,57 @@ export class Summons extends Analyser {
...
+			if (a === undefined) { return -1 }
+			if (b === undefined) { return 1 }
+			return (a.timestamp > b.timestamp) ? 1 : -1
+		})
+
+		const retval: JSX.Element[] = []
+		egis.forEach(egi => {
+			if (egi !== undefined) {
+				if (egi === summon.data.ifritSummon) {
+					retval.push(< > <ActionLink showName={false} action="SUMMON_IFRIT_II" />&nbsp;{this.printUsageTime(summon.data.ifritSummon)} </>)
```
KelosPelmand: I think putting the egi timestamps here is not a net positive.  Instead we should just show the icons.

It's subtle, but with the timestamps here, I have to scan the row more to see the egi order, whereas with just the single space, I can see them all at once.  Similarly, I assume that when someone is looking at these, the important bit is the order and the precise timing is rarely what they are looking for.

KelosPelmand (changes requested):
Hey, thanks for the commit! We require a change log entry in the `/parser/<JOB>/index.ts` file for any updates / bug fixes that change job logic, and you'll need to upload an avatar (max 100x100px / 100KB) to /src/data/avatar and add yourself to the /src/data/CONTRIBUTORS.ts file. If you have any questions, please feel free to drop by our Discord and tag yourself with the Developer role.

It also appears that you have some indenting/spacing issues that are causing the linter to fail.  The project comes with a Visual Studio Code configuration that should catch these sorts of things if you don't have another linter you already use.

src/parser/jobs/smn/modules/Summons.tsx
Iaotle: Should I put egi -> egi -> egi or just egi egi egi

src/parser/jobs/smn/modules/Summons.tsx
Iaotle: I'm not sure which size would work best, so if you can find a size that fits all that would be awesome. As long as the size is small enough it should be good for mobile as well I think? Maybe there is an align char that we can use?

src/parser/jobs/smn/modules/Summons.tsx
KelosPelmand: I'd lean towards egi egi egi.  It seems to give enough space between them and minimizes clutter and space needed.

src/parser/jobs/smn/modules/Summons.tsx
KelosPelmand: I don't think there is a size that works best.  I only have basic web layout knowledge, so I'm not sure what kind of tricks are available to force alignments.  It's functional, if not as aesthetically pleasing, to just have them not line up vertically across rows.

Iaotle:
btw linting config wouldn't work for me, hopefully it's minor enough.

Iaotle:
Should be fixed according to your suggestions. Idk if we should have less pixels for the bahamut but without spacing it triggers me really bad.

KelosPelmand:
The fixed px width has to be removed.  It will break for other languages.  Even if you pick a number big enough for the languages we support, I wouldn't trust it to work on all the other browser and font combinations that our users may be using.

image
image

However, some whitespace (either a margin or nbsps) after the demi name is fine if you want to put it in.

KelosPelmand:
Also, it appears you swapped the order of the egis in the table when putting them back.  Did you have a reason for this or was it an oversight?

Iaotle:
image
Alright so they're a bit out of the way now, but this seems like the best solution in order to have it work with all localizations as well as not trigger anyone's OCD for alignment.

Iaotle:
P.S. I don't think that sorting the egi list is a bad idea. You don't really pay attention to the order of the fields anyway if you are skimming because you are only looking at the fields outlined in red, which will be highlighted anyway. It feels distinctly weird to look at the list of egis in the header and having that order not be matched in the table. Let me know if you notice it.

Rjb99:
I'd prefer if we kept output windows uniform. Having checklists/info windows move all over the place can be jarring even if you are only looking for highlighted red field.

Iaotle:
Yeah so this is done now, the way you guys wanted it. Any other suggestions?

KelosPelmand (approved):
Can confirm, this has all the requested changes and is ready to merge.

AkairyuGestalter (approved):

PR #1642, opened 2022-02-20
Fix RDM Swift/Accel double buff extending Swift's covered actions.

eatmorehufu:
Bug spotted here in Swiftcast Actions: https://xivanalysis.com/fflogs/y8mdXkNw6pY29h1t/1/12


Cephid:
Logic here is simple and sound; Good catch putting Impact in the list.

src/parser/jobs/rdm/modules/Swiftcast.tsx
```diff
@@ -12,6 +13,14 @@ export class Swiftcast extends CoreSwiftcast {
...
 
 	@dependency private downtime!: Downtime
+	@dependency private actors!: Actors
+
+	// If Swiftcast and Acceleration are both up, these actions consume Acceleration first.
+	private ACCELERATION_ACTIONS: ActionKey[] = [
+		'VERTHUNDER_III',
+		'VERAERO_III',
+		'IMPACT',
+	]
```
ayyaruq: If you're treating this like a const, make it a const. If it's a class var, you can just use the IDs directly and avoid having to map it later.

src/parser/jobs/rdm/modules/Swiftcast.tsx
```diff
@@ -21,6 +30,11 @@ export class Swiftcast extends CoreSwiftcast {
 			if (!this.downtime.isDowntime() && castTime <= BASE_GCD) {
 				return false
 			}
+			// Ignore acceleration actions since they won't consume Swiftcast when you have Acceleration.
+			if (this.actors.current.hasStatus(this.data.statuses.ACCELERATION.id)
+				&& this.ACCELERATION_ACTIONS.map(actionKey => this.data.actions[actionKey]).includes(action)) {
```
ayyaruq: Prefer to multiline this fully, but I'd lose the map and just keep it on one line.

ayyaruq (commented):
Logically fine, some code style nits. Allowed the shouted var in 1639 but it looks like a bit of a pattern so prefer to sort that out. Not requesting changes so if another maintainer is down for it, cowabunga it is.

src/parser/jobs/rdm/modules/Swiftcast.tsx
eatmorehufu: This has to be a class var in order to use this.data right?

src/parser/jobs/rdm/modules/Swiftcast.tsx
eatmorehufu: Changed the var to be Action[] to avoid the map.

src/parser/jobs/rdm/modules/Swiftcast.tsx
ayyaruq: Correct, so we use the ActionKeys in a const and then map to fill, or just declare outright as a class var with `this.data` for the IDs. If you prefer the former, move the map into the body and set the class var from that, rather than running it in the function.

eatmorehufu: I went with the class var, avoids some mapping.

ayyaruq (approved):

PR #1643, opened 2022-02-20
Well prepared checklist

Malpp:
Top parse log:
image
My garbo 10m50s P3S kill:
image
Rando bad parse log:
image
Log where food buff falls off:
image


src/event.ts
```diff
@@ -219,6 +223,8 @@ interface EventActorUpdate extends FieldsBase {
 	targetable?: boolean
 	/** Current attributes (stats) */
 	attributes?: AttributeValue[]
+	/** Auras on start of pull */
+	auras?: Aura[]
```
ayyaruq: What's an aura meant to be here?

src/parser/core/modules/WellPrepared.tsx
```diff
@@ -0,0 +1,73 @@
...
+import {Event, Events} from 'event'
+import {Analyser} from 'parser/core/Analyser'
+import {filter} from 'parser/core/filter'
+import {dependency} from 'parser/core/Injectable'
+import Checklist, {Requirement, Rule} from 'parser/core/modules/Checklist'
+import {Data} from 'parser/core/modules/Data'
+import React from 'react'
+
+// Give a use time of 5m10s between each potion use
+const TIME_BETWEEN_POTIONS_IN_MS = 310000
```
ayyaruq: What's the rationale for this?

ayyaruq (commented):
Just some questions, haven't done an actual review here yet. We don't get any duration info for medicated or food, so food falling off mid-fight isn't handled, and variance in which pots are used seems to be unhandled as well. Main concern scenario I can think of is if you use an ether, it counts as a pot but isn't really the same. Additionally, the majority of content doesn't need food or pots so this is confusing feedback to users that aren't as experienced. Might be worth having the checklist only warn instead of error (so it's yellow instead of red) but I still think this may be confusing to less experienced players.

src/reportSources/legacyFflogs/eventAdapter/translate.ts
```diff
@@ -99,7 +99,7 @@ export class TranslateAdapterStep extends AdapterStep {
 			return [this.adaptTargetableEvent(baseEvent)]
 
 		case 'combatantinfo':
-			return this.adaptCombatantInfoEvent(baseEvent)
+			return this.adaptCombatantInfoEvent(baseEvent).concat(this.adaptCombatantInfoEventAuras(baseEvent))
```
ackwell: this should all be handled in the main `adaptCombatantInfoEvent` method

if the logic is too clunky to be in one method, feel free to branch out from within there - but it shouldn't be doing non-trivial logic in the primary adapt switch

src/event.ts
```diff
@@ -205,6 +205,10 @@ export interface AttributeValue {
 	estimated: boolean
 }
 
+export interface Aura {
```
ackwell: please use xiv terminology - we have statuses, not auras.

src/event.ts
```diff
@@ -219,6 +223,8 @@ interface EventActorUpdate extends FieldsBase {
 	targetable?: boolean
 	/** Current attributes (stats) */
 	attributes?: AttributeValue[]
+	/** Auras on start of pull */
```
ackwell: introducing a new actor field exclusively for start of pull logic seems like a leaky abstraction.
we already have `statusApply` events, why can't we use those?

src/parser/core/modules/WellPrepared.tsx
```diff
@@ -0,0 +1,73 @@
...
+			.status(this.medicated.id), this.onPotionUse)
+		this.addEventHook('complete', this.onComplete)
+	}
+
+	private onComplete() {
+		this.debug(this.isWellFed)
+		this.debug(this.potionUses)
+		this.debug(this.parser.pull.duration)
+		this.debug(this.parser.pull.duration / TIME_BETWEEN_POTIONS_IN_MS)
+		this.debug(Math.floor(this.parser.pull.duration / TIME_BETWEEN_POTIONS_IN_MS))
```
ackwell: feel free to leave debug calls in, but it'd be nice to format this up as a single line entry with labels so a reader can understand what's going on

ackwell (changes requested):
quick review of adapter scope changes.
xiva event structure is supposed to model xiv logic as closely as possible, we should be avoiding terms from other games here.
but on the same note - the combatantinfo auras field looks to just be a list of status ids - in xiv this would be applied as actual statuses - why not use the events we already have for this?

ackwell:
oh ergh i just remembered that leaving a red blocks the pr until _i_ review it again. please do feel free to ping me on discord to kick the review so i don't forget

src/parser/core/modules/WellPrepared.tsx
Rjb99: To further question this,

Why 5:10? Pots are 4:30 cooldown right now and the ideal time to use them (outside of super fast kills are at the full 2 minute burst windows for most jobs to catch party buffs.

So if anything you would want 4:30 + 1:30 for 6 minutes. There's no point in using a pot on cooldown if you aren't doing your full burst unless you are going to miss a use from it.

Rjb99:
Minor request, I don't see anything accounting for food dropping off. That would also be something I'd like to see.

"Make sure the time left on food buff allows you to have it for the full pull" or something along those lines

AzariahAeroborn:
@Malpp I see this is marked as a draft - is this something you're still interested in pursuing, or does it need updates or should it be abandoned?

AzariahAeroborn:
Closing per abandoned.

PR #1644, opened 2022-02-20
Improve friendly death adjustment logic

ackwell:
Title. This beefs up the logic around fixing fflogs' dumb death bullshit, and moves it into a separate adapter so it's not cluttering up translate.

I've also swapped actor update dedupe onto postprocess instead of adapt because it being on the adapt pass makes it nigh impossible to track how events are changing through the pipeline.

Before:
image

After:
image


ayyaruq (approved):
Looks sane to me but probably worth a second set of eyes.

PR #1645, opened 2022-02-21
Riddle of Wind statistic

Malpp:
Bad log:
image

Perfect play:
image

Info:
image


src/parser/jobs/mnk/modules/DISPLAY_ORDER.ts
```diff
@@ -1,7 +1,8 @@
 export const DISPLAY_ORDER = {
 	RIDDLE_OF_FIRE: 2,
 	PERFECT_BALANCE: 3,
-	DRAGON_KICK: 5,
-	TWIN_SNAKES: 6,
-	DEMOLISH: 7,
+	RIDDLE_OF_WIND: 4,
+	DRAGON_KICK: 6,
+	TWIN_SNAKES: 7,
+	DEMOLISH: 8,
```
ayyaruq: Not sure why this needs to happen, statistic boxes are fixed locations and without any suggestions there's nothing to display so the display order for RoW isn't used. Generally also prefer to not re-order this file.

src/parser/jobs/mnk/modules/RiddleOfWind.tsx
```diff
@@ -0,0 +1,84 @@
...
+	suggest(windows: Array<HistoryEntry<EvaluatedAction[]>>): undefined {
+		this.callback(windows)
+		return undefined
+	}
+}
+
+export class RiddleOfWind extends BuffWindow {
+	static override handle = 'riddleofwind'
+	static override title = t('mnk.row.title')`Riddle of Wind`
+	static override displayOrder = DISPLAY_ORDER.RIDDLE_OF_WIND
```
ayyaruq: Neither of these get used.

src/parser/jobs/mnk/modules/RiddleOfWind.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+	@dependency globalCooldown!: GlobalCooldown
```
ayyaruq: This isn't used anywhere, can kill the import for it as well.

src/parser/jobs/mnk/modules/RiddleOfWind.tsx
```diff
@@ -0,0 +1,84 @@
...
+	@dependency statistics!: Statistics
+
+	private riddleOfWind: ActionRoot['RIDDLE_OF_WIND'] = this.data.actions.RIDDLE_OF_WIND
```
ayyaruq: This doesn't get used anywhere.

src/parser/jobs/mnk/modules/RiddleOfWind.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+		this.statistics.add(new SimpleStatistic({
+			title: <Trans id="mnk.row.statistic.title">
+				Auto attacks hits
+			</Trans>,
+			icon: this.data.actions.RIDDLE_OF_WIND.icon,
+			value: `${actualAttacks}/${maxAttacks}`,
+			info: <Trans id="mnk.row.statistic.info">
+				Missing auto attacks during <ActionLink id={this.riddleOfWind.id}/> means you were not
+				able get full uptime on the boss. This is a DPS loss due to missing auto attacks.
```
ayyaruq:
```suggestion
				Missing auto attacks during <ActionLink action="RIDDLE_OF_WIND"/> means you were not able get full uptime on the boss. This is a DPS loss due to missing auto attacks.
```

src/parser/jobs/mnk/modules/RiddleOfWind.tsx
```diff
@@ -0,0 +1,84 @@
...
+		return super.isRushedEndOfPullWindow(window)
+	}
+
+	private addStatistic(windows: Array<HistoryEntry<EvaluatedAction[]>>) {
+		const maxAttacks = windows.length * MAX_ATTACKS_PER_ROW_WINDOW
+		const actualAttacks = windows.map(history => history.data.length).reduce((previousValue, currentValue) => previousValue + currentValue, 0)
+
+		this.statistics.add(new SimpleStatistic({
+			title: <Trans id="mnk.row.statistic.title">
+				Auto attacks hits
```
ayyaruq:
```suggestion
				Auto Attack Hits
```

src/parser/jobs/mnk/modules/RiddleOfWind.tsx
```diff
@@ -0,0 +1,84 @@
...
+import {DISPLAY_ORDER} from './DISPLAY_ORDER'
+
+/**
+ * While it is possible to get 16 attacks in a Riddle of Wind
+ * window, it is such a niche and time based thing that I'm
+ * not sure if it's worth it. I've looked into a lot of top
+ * parsing monk logs, and it almost feels like it's RNG if
+ * someone has a 16 RoW attack window.
+ */
+const MAX_ATTACKS_PER_ROW_WINDOW = 15
```
ayyaruq: No need to limit width so much for the comment, could rename the const to something like `EXPECTED_AUTOS_PER_WINDOW` since it's expected, not max.

ayyaruq (changes requested):
Mostly wanna tighten up some string formatting, but I feel like BuffWindow is severely overkill here since we use nothing it provides but still have all the overhead. This module would be smaller as just a native analyser since there's really not that much you need to capture.

ghost (commented):
:lgtm:

ayyaruq (approved):

PR #1646, opened 2022-02-21
[SAM] Set Hex Color

Rjb99:
Undoing semantical murders one line at a time


ayyaruq (approved):
I laughed that this is the same colour anyway, what is DRY, yolo LGTM

PR #1647, opened 2022-02-23
fflogs - sort status events dynamically

hintxiv:
This PR removes the total ordering of events which occur on the same timestamp where fflogs is the report source. Instead it trusts the order received from fflogs, with the exception of actions that apply statuses on the same timestamp . In these cases, an adapter is used to dynamically sort status apply events after their corresponding casts.

- **Performance**: the sort adapter takes ~80ms to return on a typical 8 minute log. On the same log, the old sort step took ~30ms. 

- **Testing**: two logs from each job were checked to ensure nothing obvious broke - results [here](https://pastebin.com/KtdVCK05).

    - The Wildfire module needed a small fix to handle WF's unique ordering of `statusRemove > damage`.

- **End-of-status sorting**: you can see an implementation for this in the commit history. This could solve some bugs (e.g. where buff windows are over-reporting their GCD count), but too many other modules currently rely on the ordering of `action > statusRemove` (e.g. consuming a proc). I decided to leave it out.


ghost:
LGTM.
The sortTimestampBucket method could probably be optimized somewhat, but given that you're only reporting an 80ms time for this step, I don't know if this is the place where we should focus optimization efforts.

src/reportSources/legacyFflogs/eventAdapter/sortStatus.ts
```diff
@@ -0,0 +1,70 @@
...
+
+		// This shouldn't be possible, but bail if we don't find a matching status
+		if (statusKey == null) { return false }
+
+		const applyingActions = getDataArrayBy(getActions(this.report), 'statusesApplied', statusKey)
+
+		return applyingActions.some(other => action === other.id)
+	})
+
+	private sortTimestampBucket(events: Event[]): Event[] {
```
ackwell, line 49: yeah there's a few places in this implementation that it would be possible to avoid additional array creation; but as it stands the time burnt isn't high enough to worry about it / bother with reducing readability to do it, i don't think.

src/reportSources/legacyFflogs/eventAdapter/sortStatus.ts
```diff
@@ -0,0 +1,70 @@
...
+			after: [],
+		}))
+
+		for (const [index, node] of nodes.entries()) {
+			const event = node.event
+
+			if (event.type === 'statusApply') {
+				// Sorts the event after its corresponding action event
+				for (const other of nodes.slice(index + 1)) {
+					if (other.event.type === 'action' && this.canActionApplyStatus(other.event.action, event.status)) {
```
ackwell: while absolutely _miniscule_ in likelyhood of ever happening, i _think_ this has a minor logic error;

if two actors use the same action at precisely the same time, then both `statusApply` events will be pushed to both `action`s

to be double sure, you'll probably want to
- check that the action and status events come from the same source
- `break` after pushing it to make sure wild edge cases won't cause double pushes of the same status apply event

ackwell (commented):
looks good, only thing of note is the potential edge case on the adapter

src/reportSources/legacyFflogs/eventAdapter/sortStatus.ts
hintxiv: I'm not too happy with it either. I got rid of one of those extra array creations but I know it's still not optimal. I don't think I quite have the brain wrinkles to improve it further.

src/reportSources/legacyFflogs/eventAdapter/sortStatus.ts
hintxiv: 👍 good catch, I believe these edge cases are handled now.

ackwell (approved):
bueno

PR #1650, opened 2022-03-03
Add CelestialRevolution analyser

skimmedsquare:
Addresses one of the subtasks of #1303 

### Preview

Screen Shot 2022-03-03 at 1 56 44 PM


src/parser/jobs/mnk/modules/CelestialRevolution.tsx
```diff
@@ -0,0 +1,42 @@
...
+	private onComplete(): void {
+		if (this.count > 0) {
+			this.suggestions.add(new Suggestion({
+				icon: this.data.actions.CELESTIAL_REVOLUTION.icon,
+				content: <Trans id="mnk.cr.suggestions.content">
+					Avoid using <ActionLink action="CELESTIAL_REVOLUTION"/> over either <ActionLink action="ELIXIR_FIELD"/> or <ActionLink action="RISING_PHOENIX"/>.
+				</Trans>,
+				severity: SEVERITY.MAJOR,
+				why: <Trans id="mnk.cr.suggestions.why">
+					You cast <ActionLink action="CELESTIAL_REVOLUTION"/> <Plural value={this.count} one="# time" other="# times"/>.
```
Malpp: I would change it to something more along the lines of:
Celestial Revolution was used X time(s)

Malpp (commented):
Don't forget to add a changelog

src/parser/jobs/mnk/modules/CelestialRevolution.tsx
skimmedsquare: Yea, I think that sounds better, will change.

Malpp (approved):

src/parser/jobs/mnk/modules/CelestialRevolution.tsx
```diff
@@ -0,0 +1,42 @@
...
+	private onComplete(): void {
+		if (this.count > 0) {
+			this.suggestions.add(new Suggestion({
+				icon: this.data.actions.CELESTIAL_REVOLUTION.icon,
+				content: <Trans id="mnk.cr.suggestions.content">
+					Avoid using <ActionLink action="CELESTIAL_REVOLUTION"/> over either <ActionLink action="ELIXIR_FIELD"/> or <ActionLink action="RISING_PHOENIX"/>.
+				</Trans>,
+				severity: SEVERITY.MAJOR,
+				why: <Trans id="mnk.cr.suggestions.why">
+					<ActionLink action="CELESTIAL_REVOLUTION"/> was used <Plural value={this.count} one="# time" other="# times"/>.
```
ayyaruq: I'd maaaaaaybe consider lost potency here, but it's iffy since EF and RP aren't the same potency. Could do a range tho, like `{this.count * (600 - 450)} - {this.count * (700 - 450)} potency lost to <Plural value={this.count} blah uses> of <ActionLink action="CELESTIAL_REVOLUTION">` or some such, except using the potency data (can add if needed) or having a const for the string to insert or something since inline calcs are a bit shit.

src/parser/jobs/mnk/modules/CelestialRevolution.tsx
```diff
@@ -0,0 +1,42 @@
...
+import {Event} from 'event'
+import {Analyser} from 'parser/core/Analyser'
+import {filter} from 'parser/core/filter'
+import {dependency} from 'parser/core/Injectable'
+import {Data} from 'parser/core/modules/Data'
+import Suggestions, {Suggestion, SEVERITY} from 'parser/core/modules/Suggestions'
+import React from 'react'
+
+export class CelestialRevolution extends Analyser {
+	static override handle = 'celestialRevolution'
```
ayyaruq: Nit; prefer to keep i18n keys in alignment with the handles as a preference thing. If the i18n key is `cr`, then use `cr` as the handle, although the name of the module itself is kinda verbose, just Revolution is prob good enough and won't make the file name clip the nav panel in vsc.

src/parser/jobs/mnk/modules/CelestialRevolution.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+	@dependency private data!: Data
+	@dependency private suggestions!: Suggestions
+
+	private count: number = 0
```
ayyaruq: Not a fan of generic names, prefer it to be explicit.
```suggestion
	private revolutions: number = 0
```

src/parser/jobs/mnk/modules/CelestialRevolution.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+	override initialise(): void {
+		this.addEventHook(
+			filter<Event>()
+				.source(this.parser.actor.id)
+				.type('action')
+				.action(this.data.actions.CELESTIAL_REVOLUTION.id), () => this.count++)
+		this.addEventHook('complete', this.onComplete)
```
ayyaruq: Style nit, if you're gonna multi-line then multi-line the whole thing. The callback arrow function can be it's own line. Also prefer a newline between multi-line hooks and complete so it's easier to read. Hopefully GH doesn't mangle whitespace on this suggestion.
```suggestion
		this.addEventHook(
			filter<Event>()
				.source(this.parser.actor.id)
				.type('action')
				.action(this.data.actions.CELESTIAL_REVOLUTION.id),
			() => this.count++
		)

		this.addEventHook('complete', this.onComplete)
```

src/parser/jobs/mnk/modules/CelestialRevolution.tsx
```diff
@@ -0,0 +1,42 @@
...
+	}
+
+	private onComplete(): void {
+		if (this.count > 0) {
+			this.suggestions.add(new Suggestion({
+				icon: this.data.actions.CELESTIAL_REVOLUTION.icon,
+				content: <Trans id="mnk.cr.suggestions.content">
+					Avoid using <ActionLink action="CELESTIAL_REVOLUTION"/> over either <ActionLink action="ELIXIR_FIELD"/> or <ActionLink action="RISING_PHOENIX"/>.
```
ayyaruq: Some clarity of wording here, the "either" makes it sound like a choice to replace one of these (to someone who isn't a native speaker) when really it's both. I'd like to add a note about why you'd want to avoid CR tho.
```suggestion
					Avoid using <ActionLink action="CELESTIAL_REVOLUTION"/> over <ActionLink action="ELIXIR_FIELD"/> or <ActionLink action="RISING_PHOENIX"/> as they have higher potency even for a single target.
```

ayyaruq (changes requested):
Some minor code quality changes and a bit of rewording. Nice simple PR tho, thanks!

skimmedsquare:
Screen Shot 2022-03-08 at 3 32 05 PM

src/parser/jobs/mnk/changelog.tsx
```diff
@@ -2,6 +2,11 @@ import CONTRIBUTORS from 'data/CONTRIBUTORS'
 import React from 'react'
 
 export const changelog = [
+	{
+		date: new Date('2022-03-03'),
+		Changes: () => <>Add Celstial Revolution support.</>,
```
ayyaruq:
```suggestion
		Changes: () => <>Add Celestial Revolution support.</>,
```

src/parser/jobs/mnk/modules/Revolution.tsx
```diff
@@ -0,0 +1,48 @@
...
+import {Event} from 'event'
+import {Analyser} from 'parser/core/Analyser'
+import {filter} from 'parser/core/filter'
+import {dependency} from 'parser/core/Injectable'
+import {Data} from 'parser/core/modules/Data'
+import Suggestions, {Suggestion, SEVERITY} from 'parser/core/modules/Suggestions'
+import React from 'react'
+
+const MIN_POTENCY_LOSS = 150
+const MAX_POTENCY_LOSS = 250
```
ayyaruq: I think this can be done more dynamically via data. If you add the potency fields to the Blitzes, you can operate on them directly such as `this.data.actions.ELIXIR_FIELD.potency` so updating the data can be done by layers, allowing the code to work for any version of the expansion even if potencies change. Hardcoding like this basically forces you to have per-patch handling or outright not support certain patches if this changes.

src/parser/jobs/mnk/modules/Revolution.tsx
```diff
@@ -0,0 +1,48 @@
...
+			filter<Event>()
+				.source(this.parser.actor.id)
+				.type('action')
+				.action(this.data.actions.CELESTIAL_REVOLUTION.id),
+			() => this.revolutions++)
+		this.addEventHook('complete', this.onComplete)
+	}
+
+	private onComplete(): void {
+		if (this.revolutions > 0) {
```
ayyaruq, line 30: Personally I'd tiered suggestion to avoid this, but kinda up to you.

ayyaruq (changes requested):
Typo fix aside, I think we should be using the data objects for the potency so it's more future proof.

src/parser/jobs/mnk/modules/Revolution.tsx
skimmedsquare: I think using `TieredSuggestion` is a bit misleading when the resulting `Suggestion` isn't actually tiered, so I think I'll keep this conditional for now. That being said, I wonder if it makes sense to add a `value` field to the current `Suggestion` class that simply acts as a boolean display as given severity or `IGNORED`, though I think that's definitely outside the scope of this particular PR.

ayyaruq (approved):

PR #1652, opened 2022-03-12
AST Optimal Target table for Card usage

otocephaly:
To whomever it may concern,

This is a work in progress; however, I appreciate all feedback given in relation to this.
Please read the description at the top of the relevant section for the related key assumptions before providing feedback because if after the notes it still isn't clear to you, then it won't be clear to the user either and I require that kind of feedback as well.
Please also note this will not be able to be merged in the future without merging PR #1649, as there is a fix in there in relation to DoT damage.

The below code outputs this table which is an accordion within the Arcana Logs table (currently placed at the bottom of the same section.
Collapsed view example:
image

Expanded view example:
image

Also, I updated the regular arcana logs table to include an x or checkmark depending on whether the technically optimal role was chosen

Example:
image

Thank you for taking the time to read this and provide feedback in relation.

Cheers,
Otocephaly


otocephaly:
tentative update (consolidating card played and time columns):
image

suShirow:
Correct me if I'm wrong, but due to crit and RNG wouldn't damage numbers be meaningless? It's not like one can go back in time and pick the winning horse.
What would be useful would be potency (and calculated expected potency if it's a dot). 
The reason I suggest this is because this tool is meant to be 
a) a learning tool 
b) informational in a way that fflogs does not provide.

By showing potency of GCDs + Abilities captured, it would lean more toward a learning tool rather than just analysis for the sake of it.
I also absolutely want to avoid a scenario where it's used just to pick out the player with the best/worst gear/performance.

otocephaly:
Hello!

Additional changes included adding in crit / DH figures to normalize events.
For the crit/DH calculation, it is first based on what fflogs can give us (which is only applicable for DoT events)
If there were no DoTs for the associated actor, then the crit / DH was calculated using an average over the fight and then removing the effective modifications due to other crit/DH adjusting events (such as chain strat).
For all damage events, there is an associated tagged crit/DH modifier which is used to help normalize. This is considered even in DoT cases since fflogs doesn't give the crit/DH at the time of application only the associated crit/DH from the actor.
There are additional considerations based on auto crit/auto DH abilities and statuses; however, for that list, refer to the top of the related section.

For the AST card normalization, the formula is as stated at the top. Crit/DH rates are taken effectively at the point the damage is dealt which is where the crit/DH modifications are utilized.
Assumptions for related calculations are shown typically at source; however, if it is not explained well enough, please let me know and I'll add more notes!

Here is an example of what it looks like now:
image

Here is a console log of the related unmodified crit/DH rates by actor
image

(actor list if needed)
image

Same example without normalizing damage figures
image

src/parser/jobs/ast/modules/ArcanaTracking/ArcanaDoTsGroup.ts
```diff
@@ -0,0 +1,103 @@
...
+	//**PHYS RANGED**
+	//BRD
+	'VENOMOUS_BITE',
+	'WINDBITE',
+	'IRON_JAWS',
+	'CAUSTIC_BITE',
+	'STORMBITE',
+
+	//MCH
+	'FLAMETHROWER',
```
hintxiv: Important to note here that ground DoTs like FT and Doton don't snapshot debuffs like Chain Strat - I think they need a slightly different code path.

src/parser/jobs/ast/modules/ArcanaTracking/CritAndDHRates.tsx
```diff
@@ -0,0 +1,422 @@
...
+		this.data.actions.PRIMAL_REND.id,
+	]
+
+	override initialise() {
+		//6.1 additions
+		if (this.parser.patch.after('6.08')) {
+			this.autoCritAbilities.push(this.data.actions.MIDARE_SETSUGEKKA.id)
+			this.autoCritAbilities.push(this.data.actions.KAESHI_SETSUGEKKA.id)
+			this.autoCritAbilities.push(this.data.actions.OGI_NAMIKIRI.id)
+			this.autoCritAbilities.push(this.data.actions.KAESHI_NAMIKIRI.id)
```
hintxiv: These should be in a const ideally.

src/parser/jobs/ast/modules/ArcanaTracking/CritAndDHRates.tsx
```diff
@@ -0,0 +1,422 @@
...
+		this.data.actions.CHAOTIC_CYCLONE.id,
+		this.data.actions.INNER_CHAOS.id,
+		this.data.actions.PRIMAL_REND.id,
+	]
+	private autoDHAbilities: Array<Action['id']> = [
+		this.data.actions.STARFALL_DANCE.id,
+		this.data.actions.CHAOTIC_CYCLONE.id,
+		this.data.actions.INNER_CHAOS.id,
+		this.data.actions.PRIMAL_REND.id,
+	]
```
hintxiv: Wildfire needs some handling too. It can't crit or DH so the base damage is the only part that counts.

src/parser/jobs/ast/modules/ArcanaTracking/CritAndDHRates.tsx
```diff
@@ -0,0 +1,422 @@
...
+
+	//crit and DH Modifier variables
+	private critPartyModifiers: {[id: Status['id']]: {strength: number}} = {
+		[this.data.statuses.BATTLE_LITANY.id]: {strength: 0.1},
+		[this.data.statuses.DEVILMENT.id]: {strength: 0.2},
+		[this.data.statuses.THE_WANDERERS_MINUET.id]: {strength: 0.02},
+		[this.data.statuses.INNER_RELEASE.id]: {strength: 1},
+		//TO DO: these below moves only increase crit for the next weaponskill, not the oGCDs between
+		[this.data.statuses.REASSEMBLED.id]: {strength: 1},
+		[this.data.statuses.OPO_OPO_FORM.id]: {strength: 1},
```
hintxiv: Only BS and SotD count as auto crits when opo form is up, but it looks like this is going to count all damage (including other GCDs) when the form is up?

src/parser/jobs/ast/modules/ArcanaTracking/CritAndDHRates.tsx
```diff
@@ -0,0 +1,422 @@
...
+	private critEnemyModifierCurrent: {[actor: Actor['id']]: {currentModifier: number}} = {}
+	private DHPartyModifierCurrent: {[actor: Actor['id']]: {currentModifier: number}} = {}
+	private appliedStatuses: Array<{actor: Actor['id'], status: Status['id']}> = [] //reasoning is that bard's statuses are reapplied without removal. this is to catch those
+
+	//crit and DH Modifier variables
+	private critPartyModifiers: {[id: Status['id']]: {strength: number}} = {
+		[this.data.statuses.BATTLE_LITANY.id]: {strength: 0.1},
+		[this.data.statuses.DEVILMENT.id]: {strength: 0.2},
+		[this.data.statuses.THE_WANDERERS_MINUET.id]: {strength: 0.02},
+		[this.data.statuses.INNER_RELEASE.id]: {strength: 1},
```
hintxiv: Inner Release only applies to Fell Cleave and Decimate, but this will buff all damage including autos.

src/parser/jobs/ast/modules/ArcanaTracking/CritAndDHRates.tsx
```diff
@@ -0,0 +1,422 @@
...
+					DHMod: this.DHPartyModifierCurrent[event.source].currentModifier + abilityDHModifier,
+					sourceModifier: target.sourceModifier,
+				}
+			)
+		})
+	}
+
+	private petDamage(event: Events['damage']) {
+		const owner: Actor['id'] | undefined = this.actors.get(event.source)?.owner?.id
+		//don't consider statuses as already considered in Damage and not expecting pets to have ticks
```
hintxiv: I think this is an unsafe assumption since pets have had DoTs in the past.

src/parser/jobs/ast/modules/ArcanaTracking/CritAndDHRates.tsx
otocephaly: Do you happen to know if wildfire is the only ability like this?

hintxiv: I believe so yes

src/parser/jobs/ast/modules/ArcanaTracking/CritAndDHRates.tsx
otocephaly: Fair
Updated to include pet DoTs
See below commit

src/parser/jobs/ast/modules/ArcanaTracking/CritAndDHRates.tsx
otocephaly: Thanks for that catch!
I've updated this to have a separate section along with Opo for MNK
See below commit

src/parser/jobs/ast/modules/ArcanaTracking/CritAndDHRates.tsx
otocephaly: Yeahhhh I was a bit lazy on this one!
I've updated this to have a separate section along with Inner Release for War as you've caught below
See below commit

src/parser/jobs/ast/modules/ArcanaTracking/CritAndDHRates.tsx
otocephaly: Thanks! I added a section to handle this a bit generally in case this comes up again in the future
See below commit

src/parser/jobs/ast/modules/ArcanaTracking/CritAndDHRates.tsx
otocephaly: I've updated this, but let me know if it's the appropriate update
Thank you

src/parser/jobs/ast/modules/ArcanaTracking/ArcanaDoTsGroup.ts
otocephaly: Updated! Thanks for this
I've moved all ground dots to a separate const with separate handling
See below commit

otocephaly:
Edit: there is a bug after reviewing the below. Will update when bug is resolved. Please hold off on reviews for now thx

Updated again!
(I should've put the last log link for retesting purposes, but alas here's another example structured similarly as before
https://www.fflogs.com/reports/a:PrCfLWbH7A2FQjx6)
image


This version is after incorporating Hint's feedback as above
Current version:
image

Console log of unmodified crits/dh by actor:
image

Version without adjusting for crits/DH for comparative purposes:
image

src/event.ts
```diff
@@ -161,6 +161,9 @@ interface EventDamage extends FieldsMultiTargeted {
 		/** Bonus multiplier on the damage from which positionals can be inferred */
 		bonusPercent: number
 	}>>
+	/** direct hit and crit rates provided by fflogs. Note: crit is provided as a stat and not the actual rate */
+	directHitRate?: number
+	expectedCritRate?: number
```
ayyaruq, line 166: If crit is taken as a stat, shouldn't we be using the existing player stats stuff for it? And what about DH then?

src/parser/jobs/ast/modules/ArcanaTracking/CritAndDHRates.tsx
```diff
@@ -0,0 +1,475 @@
...
+import {Analyser} from 'parser/core/Analyser'
+import {filter, oneOf} from 'parser/core/filter'
+import {dependency} from 'parser/core/Injectable'
+import {Actor, Actors} from 'parser/core/modules/Actors'
+import {Data} from 'parser/core/modules/Data'
+import {AUTO_CRIT_ABILITIES, AUTO_DH_ABILITIES, CRIT_ENEMY_MODIFIERS, CRIT_PARTY_MODIFIERS, DH_PARTY_MODIFIERS, GCD_CRIT_MODIFIER, GCD_DH_MODIFIER, INERT_CRIT_DH, SPECIFIC_ACTION_MODIFIER} from './CritAndDHRatesConsts'
+
+const BASE_CRIT_PROBABILITY = 0.05 //5%
+const BASE_CRIT_MULTIPLIER = 0.4
+const ROUNDING_FACTOR = 0.001 //to match rounding used in game for crit rates
```
ayyaruq, line 14: There shouldn't be any rounding, it's just integer flooring.

src/parser/jobs/ast/modules/ArcanaTracking/ArcanaTrackingOptimized.tsx
```diff
@@ -0,0 +1,482 @@
...
+	- damage will be normalized based on relative critical hit and direct hit rates based on the initial calculations done in CritAndDHRates. All of the special cases in CritAndDHRatesConsts need to be considered and applied below since they have varying impacts on the expected value calculations.
+		- considerations: regular modifiers are built into the function .getCritDH and .getCritMulti, so these considerations are
+							inert modifier is taken as base damage with no expected damage formula application
+							auto crits/dh are taken base damage and multiplier without any percentage application. this consideration is also applied for moves that only crit for specific moves or weaponskills
+							Note: there is an assumption for autocrits that the crit multiplier is the same multiplier as a regular crit including all aforeapplied non-autocrit modifiers as part of .getCritMulti
+*/
+
+const OPTIMAL_CARD_PERCENTAGE = 0.06
+const NOT_OPTIMAL_CARD_PERCENTAGE = 0.03
+const MAX_PLAYERS = 8 //won't be processed for alliance raids, hunt groups, etc.
```
ayyaruq, line 40: Why not just get the proper party size?

ayyaruq (commented):
Idle comments while I was reading. I'll note that these modules are massive and needlessly complex tho, this'll need some heavy simplification before it gets merged. Most obvious thing for that is you're defining internal vars to use in hooks but they're not used for anything else, you can just keep the original const and `this.data.matchStatusId(statuses: StatusKey[])` type stuff and ignore the maps.

AzariahAeroborn:
Closing per needing to be redone on Dawntrail branch and updated for new cards.

PR #1656, opened 2022-03-20
Chore: Add CN patch 6.0 timestamp

StarHeartHunt:


suShirow:
Thanks for your PR, however we've merged the earlier duplicate instead here https://github.com/xivanalysis/xivanalysis/pull/1655

PR #1657, opened 2022-03-23
Added support for dark mode

harrjdk:
This adds support for a very simple dark mode based on the existing partial for dark mode. See [Light Mode](https://files.catbox.moe/beg3b3.png) https://files.catbox.moe/beg3b3.png and [Dark Mode](https://files.catbox.moe/bn91t8.png) https://files.catbox.moe/bn91t8.png for reference.


harrjdk:
I'm adding screens to the style changes as found, but wanted to get this up now to start the discussion

harrjdk:
So at this point, without integrating semantic theming like https://react.semantic-ui.com/theming/, remaining elements that might not entirely fit the color schemes would be difficult.  Moving to such a solution would be a much larger process since it would be reasonable in the process of adding less to move from css vars to less files and some entry magic.  If this is the what the primary maintainers of the project would like to move forward with, feel free to close this.

PR #1660, opened 2022-04-03
DRG: Adjust life surge usage in lance charge

ebshimizu:
Title. As per discussion in #fb-drg: https://discord.com/channels/441414116914233364/725970709146239006/959187432874602516

Also finally ported battle lit to RaidBuffWindow.


ayyaruq (approved):
Logically LGTM, I'll note that the Battle Litany module's strings will complicate future changes to expect GCD targets since it'll force retranslation of strings. Better to use a variable for the target. The table is also hella chonky which might be something to address for UX reasons.

PR #1662, opened 2022-04-11
[MCH] 6.1 bump

hintxiv:
I don't think we're using this potency data for anything, but I updated it. Merge after #1661


ayyaruq (changes requested):
Please use patch layers for this.

suShirow (changes requested):
Since this is a major patch, I think we should change these in the data layers to be created in https://github.com/xivanalysis/xivanalysis/pull/1663, or there will be funky differences between our potencies and those that CN and KR get.

hintxiv:
@ayyaruq @suShirow these potency numbers aren't used anywhere and weren't up to date in the first place, there's no point in layering the change. I just updated them for correctness while scrolling through data.

suShirow:
> @ayyaruq @suShirow these potency numbers aren't used anywhere and weren't up to date in the first place, there's no point in layering the change. I just updated them for correctness while scrolling through data.

Oh, okay well alright then.

suShirow (approved):

ayyaruq:
If the potencies aren't used then they should probably be removed, but if they're gonna be there then the values in the root files should reflect 6.0 values and patches should override as needed. Generally we only keep potency where it's needed tho.

ayyaruq:
Might need to re-merge `endwalker` into this.

ayyaruq (approved):
Thanks for re-syncing with `endwalker`, LGTM.

PR #1663, opened 2022-04-11
patch 610 data layers for ACTIONS + STATUSES

suShirow:
I'm surprised we didn't have this yet. 
Statuses layer file includes the duration change for SCH expedience, just to get the ball rolling. (Actually I was lazy to separate it sorry)

### DIFF
--- a/src/parser/core/modules/ActionWindow/windows/ActionWindow.tsx
+++ b/src/parser/core/modules/ActionWindow/windows/ActionWindow.tsx
@@ -24,7 +24,7 @@ export abstract class ActionWindow extends Analyser {
 	@dependency private timeline!: Timeline
 
 	/**
-	* The captured windows.
+	 * The captured windows.
 	 */
 	protected history = new History<Array<Events['action']>>(() => [])
 	/**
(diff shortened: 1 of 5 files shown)


src/parser/core/modules/ActionWindow/windows/ActionWindow.tsx
```diff
@@ -24,7 +24,7 @@ export abstract class ActionWindow extends Analyser {
 	@dependency private timeline!: Timeline
 
 	/**
-	 * The captured windows.
+	* The captured windows.
```
ayyaruq: Bruddah? It was correct before.

ayyaruq (changes requested):
Looks like another file got included by accident, just kill that and gucci sushi.

src/parser/core/modules/ActionWindow/windows/ActionWindow.tsx
suShirow: Oh no crossfire

ayyaruq (approved):

PR #1666, opened 2022-04-12
[GNB] 6.1 Support

Rjb99:
10 -> 3 for all tank stances. Much Change.

### DIFF
--- a/locale/de/messages.json
+++ b/locale/de/messages.json
@@ -853,8 +853,6 @@
   "sge.actiontimeline.addersgall": "",
   "sge.cooldownDowntime.defense-cd-metric": "",
   "sge.cooldownDowntime.ogcd-cd-metric": "",
-  "sge.cooldowns.pepsis.why": "",
-  "sge.cooldowns.suggestions.pepsis.content": "",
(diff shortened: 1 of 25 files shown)


ayyaruq:
Needs rebasing or whatever homie. Layers are merged, so just need the changes to the layer.

Rjb99:
fixed. Should be good to merge now

suShirow (approved):

ayyaruq:
As per ack's instructions in Discord, there's no need for a changelog if nothing changed, we don't need entries just for patch flagging.

suShirow (changes requested):
So sorry, as acc and ack said we gotta remove changelogs

Rjb99:
I added Royal Guard's CD change to the changelog since I did make that minor change. do we still want it dropped/moved elsewhere since it's a change to all 4 tanks? Change really only impacts timeline and minorly at that.

ayyaruq:
Another PR already made that change, and for all tanks, was merged a few days ago. This PR doesn't set that. Also, that change was only in data, not analyser code, so you still don't need a changelog for it since users can't see anything in data.

suShirow (approved):

PR #1673, opened 2022-04-14
DNC 6.1 updates

AkairyuGestalter:
- Added/updated status and action data to handle new proc statuses
- Updated Core procs to allow an action to consume multiple statuses
- Updated DNC procs to separately track new combo-sourced and Flourish-sourced proc statuses

### DIFF
--- a/src/data/ACTIONS/root/DNC.ts
+++ b/src/data/ACTIONS/root/DNC.ts
@@ -277,7 +277,7 @@ export const DNC = ensureActions({
 		onGcd: false,
 		cooldown: 60000,
 		statusesApplied: [
-			'SILKEN_SYMMETRY', // Pre-6.1 proc statuses for base layer
+			'SILKEN_SYMMETRY', // Pre-6.1 proc statuses for base layer (with the post 6.1 key they will have for the future)
 			'SILKEN_FLOW',
 			'THREEFOLD_FAN_DANCE',
 			'FOURFOLD_FAN_DANCE',
--- a/src/data/ACTIONS/root/DNC.ts
+++ b/src/data/ACTIONS/root/DNC.ts
@@ -84,7 +84,7 @@ export const DNC = ensureActions({
 		combo: {
 			start: true,
 		},
-		statusesApplied: ['FLOURISHING_SYMMETRY'],
+		statusesApplied: ['SILKEN_SYMMETRY'],
 	},
 	FOUNTAIN: {
 		id: 15990,
@@ -96,7 +96,7 @@ export const DNC = ensureActions({
 			from: 15989,
 			end: true,
 		},
-		statusesApplied: ['FLOURISHING_FLOW'],
+		statusesApplied: ['SILKEN_FLOW'],
 	},
 	REVERSE_CASCADE: {
 		id: 15991,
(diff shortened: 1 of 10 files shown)


src/data/ACTIONS/root/DNC.ts
```diff
@@ -84,7 +84,7 @@ export const DNC = ensureActions({
 		combo: {
 			start: true,
 		},
-		statusesApplied: ['FLOURISHING_SYMMETRY'],
+		statusesApplied: ['SILKEN_SYMMETRY'],
```
ayyaruq, line 87: I'm unclear why you're changing these in the root instead of a layer, this would break for 6.0 patches. Same goes for the rest of this file. Is this just changing the key to use the newer names instead of changing the IDs in the layer?

ayyaruq (changes requested):
Not so sure about that root data change, but more of a question. Requesting change, you'e updated core procs but not included a core changelog for it.

src/data/ACTIONS/root/DNC.ts
AkairyuGestalter: Yes, I changed the key for the old "formerly combined, now just combo-sourced" status the base layer to match the new name for the status, so the base layer needed to use the new key as well. The layers update the name of the old status to match the key, and fix the statusesApplied for the Flourish action to use the new-id-with-the-old-name proc statuses.

Since there was no noticeable change for end users aside from DNC because of the core change, I didn't put in a core changelog along with it, but if we feel that's useful information for users to have I can add one tonight.

AkairyuGestalter: Checked with Ack, since the core change is transparent to end users other than DNC, we're fine to leave the core changelog alone.

I did switch the base layer definition for the new Flourishing statuses to SHARED.UNKNOWN and moved the real definitions into the 6.1 layer though, now that I know that's a thing!

src/data/ACTIONS/root/DNC.ts
```diff
@@ -277,8 +277,8 @@ export const DNC = ensureActions({
 		onGcd: false,
 		cooldown: 60000,
 		statusesApplied: [
-			'FLOURISHING_SYMMETRY',
-			'FLOURISHING_FLOW',
+			'SILKEN_SYMMETRY', // Pre-6.1 proc statuses for base layer
```
ayyaruq: This comment reads like the opposite to what it seems to be doing, maybe got these mixed up?

ayyaruq (commented):
Still not sure about this data. Not requesting changes as maybe it's ok but I thought the plan was to keep the base layer in 6.0's state and handle the changes to the old statuses, plus the new statuses, in the layers outright. Looks like this is kinda mixing and matching.

src/data/ACTIONS/root/DNC.ts
AkairyuGestalter: Nope, it's correct. The old proc statuses changed their name, so I changed the key in the base layer to reflect the future state, and swapped base Flourish to use the new-key-for-old-status.

In the 6.1 layer I change the name of the old statuses to use the 'Silken' name, officially define the new-status-with-old-names, and update the Flourish action to show the Flourishing-named/keyed statuses again.

Added some more commentary there to clarify.

ayyaruq (approved):

PR #1674, opened 2022-04-17
Pld 6.1

Dynme:
Updates to PLD for 6.1 support.


suShirow (commented):
If those action amendments are from 6.1 they should be added to layers  as `breaksCombo: false` and etc instead of root.
The brand new status should be added to root so no need change on that one.

src/data/STATUSES/root/PLD.ts
```diff
@@ -116,4 +116,10 @@ export const PLD = ensureStatuses({
 		icon: 'https://xivapi.com/i/012000/012518.png',
 		duration: 21000,
 	},
+	BLADE_OF_FAITH_READY: {
+		id: 3019,
+		name: 'Blade of Faith Ready',
+		icon: 'https://xivapi.com/i/012000/012519.png',
+		duration: 30000,
+	},
```
ayyaruq: This doesn't exist in the root for the expac, it should be stubbed in root (so like `BLADE_OF_FAITH_READY: SHARED.UNKNOWN` and import as needed, then patch the details in the layer.

src/parser/jobs/pld/modules/Faith.tsx
```diff
@@ -0,0 +1,72 @@
...
+	@dependency private data!: Data
+	@dependency private suggestions!: Suggestions
+
+	private droppedFaiths: number = 0
+	private usedFaith: boolean = false
+
+	private faithHook?: EventHook<Events['action']>
+
+	override initialise() {
+		const trackedStatuses = TRACKED_STATUSES.map(key => this.data.statuses[key].id)
```
ayyaruq, line 29: I'd probably just assign this directly, since it's a custom analyser anyway, doesn't look like future changes to Faith would add more statuses.

src/parser/jobs/pld/modules/Faith.tsx
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	private onDrop() {
+		if (!this.usedFaith) { this.droppedFaiths++ }
+		if (this.faithHook != null) {
+			this.removeEventHook(this.faithHook)
+			this.faithHook = undefined
+		}
+		this.usedFaith = false
+	}
```
ayyaruq, line 55: Standard approach, but since you only need to track if it's used or not you could likely just count status applies vs actions used as outright numbers directly in the hooks and just check if they're equal. WAR does basically the same thing but it used hooks due to tracking multiple statuses.

So, you could leave this as is (but it's overly complex for little benefit) and either track the raw numbers, or you could build this module on core Procs like I did to Primal Rend in #1558. NBD, just an observation.

ayyaruq (changes requested):
Requesting changes for the data stuff, these should be using layers. I left an explicit comment on statuses, but actions should use layers too.

As a nit, you could simplify the Faith module or use core Procs for it, as noted in the comment there. Up to you tho, just a thought.

ayyaruq (approved):
Might be worth rethinking the Blade of Faith module onto core procs at a later date for simplicity. Works tho so prioritising support over code quality.

PR #1677, opened 2022-04-18
WHM Core gauge lilies

psummers678:
Migrated Lilies to Typescript and used CoreGauge. Changed wording to make clear that it is now ok to overheal with lilies due to misery being DPS neutral and having mana management. Lily timing now 20s and checking patch.

Suggestions: 
BloodLily overcap: 1,2,3 - Minor, Medium, and Major
Lily Overcap: 1, 3 - Medium Minor. This used to always be minor and much higher thresh hold
New: Unspent Blood lilies: Aim to finish with no blood lilies, 1/2 minor, 3 major


Dot-Inni:
Some lily love, very nice. Tiering looks solid to me. Can always be adjusted if the need arises. But: I think you may need a changelog entry since you're changing a job module on a basic level?

psummers678:
Thats the change log added

src/parser/jobs/whm/modules/Lilies.tsx
```diff
@@ -0,0 +1,167 @@
...
+const LILY_TIME_REQUIRED = 20000
+
+const BLOOD_LILY_BLOOM = 3
+const MISERY_COST = 3
+
+const GAUGE_FADE = 0.25
+const TIMER_FADE = 0.75
+const BLOODLILY_FADE = 0.5
+const LILY_COLOR = Color('#4f73b5')
+const BLOODLILY_COLOR = Color('#b52d6c')
```
ayyaruq: I'd probably fold these together, there's no real need for them to be distinct vars when they're mostly one time use.

src/parser/jobs/whm/modules/Lilies.tsx
```diff
@@ -0,0 +1,167 @@
...
+import {filter, oneOf} from 'parser/core/filter'
+import {dependency} from 'parser/core/Injectable'
+import {CounterGauge, Gauge as CoreGauge, TimerGauge} from 'parser/core/modules/Gauge'
+import Suggestions, {SEVERITY, TieredSuggestion} from 'parser/core/modules/Suggestions'
+import React from 'react'
+
+const LILY_INITIAL_STACKS = 0
+const LILY_MAX_STACKS = 3
+const LILY_60_TIME_REQUIRED = 30000
+const LILY_TIME_REQUIRED = 20000
```
ayyaruq: Prefer to be explicit, and slightly more readable:
```suggestion
const LILY_INTERVAL_600 = 30000
const LILY_INTERVAL_610 = 20000
```

src/parser/jobs/whm/index.tsx
```diff
@@ -64,5 +64,10 @@ export const WHITE_MAGE = new Meta({
 			Changes: () => <>Marked WHM as supported for 6.08</>,
 			contributors: [CONTRIBUTORS.INNI],
 		},
+		{
+			date: new Date('2022-04-18'),
+			Changes: () => <>Changed Lily timer and content for 6.1 with misery update</>,
```
ayyaruq: Nit; consider something more like "Migrate Lily gauge to core gauge and update for 6.1 Misery analysis.", it reads a bit easier and is clearer what you actually changed.

src/parser/jobs/whm/modules/Lilies.tsx
```diff
@@ -0,0 +1,167 @@
...
+import Color from 'color'
+import {DataLink} from 'components/ui/DbLink'
+import {Event} from 'event'
+import {filter, oneOf} from 'parser/core/filter'
+import {dependency} from 'parser/core/Injectable'
+import {CounterGauge, Gauge as CoreGauge, TimerGauge} from 'parser/core/modules/Gauge'
+import Suggestions, {SEVERITY, TieredSuggestion} from 'parser/core/modules/Suggestions'
+import React from 'react'
+
+const LILY_INITIAL_STACKS = 0
```
ayyaruq: Initial state of 0 doesn't really need a const.

src/parser/jobs/whm/modules/Lilies.tsx
```diff
@@ -0,0 +1,167 @@
...
+const LILY_MAX_STACKS = 3
+const LILY_60_TIME_REQUIRED = 30000
+const LILY_TIME_REQUIRED = 20000
+
+const BLOOD_LILY_BLOOM = 3
```
ayyaruq: I'd consider reusing `LILY_MAX_STACKS` here since they're the same, but if there's history of SE changing things then it's probably worth the distinction. Consider reordering these vars so they're easier to read tho.

src/parser/jobs/whm/modules/Lilies.tsx
```diff
@@ -0,0 +1,167 @@
...
+	@dependency private suggestions!: Suggestions
+
+	private lilyConsumers = [
+		this.data.actions.AFFLATUS_RAPTURE.id,
+		this.data.actions.AFFLATUS_SOLACE.id,
+	]
+
+	private bloodLilyConsumers = [
+		this.data.actions.AFFLATUS_MISERY.id,
+	]
```
ayyaruq: These are single use, not worth defining inline. Instead, you can have a const with the ActionKeys, and then use those in the hooks like this example:
```ts
// const
const LILY_CONSUMERS: ActionKey[] = [
  'AFFLATUS_RAPTURE',
  'AFFLATUS_SOLUS',
]

// hook
this.addEventHook(playerFilter.type('action').action(this.data.matchActionId(LILY_CONSUMERS)), this.onLilySpend)
```

src/parser/jobs/whm/modules/Lilies.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+	private lilyGauge = this.add(new CounterGauge({
+		maximum: LILY_MAX_STACKS,
+		initialValue: LILY_INITIAL_STACKS,
+		graph: {
+			label: <Trans id="whm.gauge.resource.lily">Lily</Trans>,
```
ayyaruq: Nit; I'd rethink these names, something like `whm.gauge.lily.stacks.label` allows your keys to be more consistent and more easily identified by their purpose in CrowdIn for translators. At the very least, try to be consistent with the keys in the same analyser - none of the suggestion keys match the label keys here, and there's no analyser handle name set so if gauges are split across files you'll run into problems.

src/parser/jobs/whm/modules/Lilies.tsx
```diff
@@ -0,0 +1,167 @@
...
+	private bloodLilyGauge = this.add(new CounterGauge({
+		maximum: BLOOD_LILY_BLOOM,
+		initialValue: 0,
+		graph: {
+			label: <Trans id="whm.gauge.resource.bloodlily">Blood Lily</Trans>,
+			color: BLOODLILY_COLOR.fade(BLOODLILY_FADE),
+		},
+	}))
+
+	override initialise() {
```
ayyaruq, line 88: Hypernit; consider defining your functions in an easier to follow order, so init first, then your gain/spend functions, then your utility functions. Some people like to put formatting functions below `onComplete`, so consider that for `calculateLostLilies`, although not sure it really needs a function at all.

src/parser/jobs/whm/modules/Lilies.tsx
```diff
@@ -0,0 +1,167 @@
...
+		super.initialise()
+
+		const playerFilter = filter<Event>().source(this.parser.actor.id)
+		this.addEventHook(playerFilter.type('action').action(oneOf(this.lilyConsumers)), this.onLilySpend)
+		this.addEventHook(playerFilter.type('action').action(oneOf(this.bloodLilyConsumers)), () => this.bloodLilyGauge.spend(MISERY_COST))
```
ayyaruq: There's only one blood lily consumer so kinda ehhh to use matchers or `oneOf` here, but consider doing the same as I wrote above re the consts and matchers here too.

src/parser/jobs/whm/modules/Lilies.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+	private onLilyGeneration() {
+		this.lilyGauge.generate(1)
+		if (!this.lilyGauge.capped) {
+			this.lilyTimer.start()
+		}
+	}
+
+	private onLilySpend() {
+		this.bloodLilyGauge.capped ? this.bloodLilyGauge.overCap += 1 : this.bloodLilyGauge.modify(1)
```
ayyaruq: This doesn't do anything, overcap is automatically calculated by the counter gauge, so incrementing blindly will manage the overcap for you and is less prone to error. Also for gains, consider using explicit `generate` since it's more consistent with the rest of your code, same for the blood lily hook.

src/parser/jobs/whm/modules/Lilies.tsx
```diff
@@ -0,0 +1,167 @@
...
+		if (this.lilyTimer.expired) {
+			this.lilyTimer.start()
+		}
+		this.lilyGauge.spend(1)
+
+	}
+
+	private calculateLostLilies(): number {
+		const lostLilies = Math.floor(this.lilyTimer.getExpirationTime() / LILY_TIME_REQUIRED)
+		return lostLilies
```
ayyaruq: No point to the const, you can just return directly, but I'm also not convinced this needs a function at all, you could inline the `floor` to a const where it's used directly.

src/parser/jobs/whm/modules/Lilies.tsx
```diff
@@ -0,0 +1,167 @@
...
+	}
+
+	private onComplete() {
+
+		this.suggestions.add(new TieredSuggestion({
+			icon: this.data.actions.AFFLATUS_MISERY.icon,
+			content: <Trans id="whm.lily-blood.suggestion.content">
```
ayyaruq: These keys need a bit more consistency as noted above.

src/parser/jobs/whm/modules/Lilies.tsx
```diff
@@ -0,0 +1,167 @@
...
+				Use <DataLink action="AFFLATUS_MISERY" /> to avoid wasting Blood lily growth.
```
ayyaruq: We usually write suggestions to be more suggestive/passive than demanding.
```suggestion
				Try to use <DataLink action="AFFLATUS_MISERY"/> to avoid wasting Blood Lily growth from overcapping the gauge.
```

src/parser/jobs/whm/modules/Lilies.tsx
```diff
@@ -0,0 +1,167 @@
...
+			} else {
+				content = <Trans id="whm.lily-cap.suggestion.6_1_content">
+					Use <DataLink action="AFFLATUS_RAPTURE" /> or <DataLink action="AFFLATUS_SOLACE" /> before using other GCD heals. It's okay to overheal with your lilies as they are to be used for mana management and movement aswell as healing.
+				</Trans>
+
+				tiers = {
+					1: SEVERITY.MEDIUM,
+					3: SEVERITY.MAJOR,
+				}
+			}
```
ayyaruq: Assign the strings to consts, then just swap out the const used. Tiers are also best done in consts for easier maintenance, for all suggestions in this file.

src/parser/jobs/whm/modules/Lilies.tsx
```diff
@@ -0,0 +1,167 @@
...
+				3: SEVERITY.MAJOR,
+			},
+			value: this.bloodLilyGauge.overCap,
+			why: <Trans id="whm.lily-blood.suggestion.why">
+				<Plural value={this.bloodLilyGauge.overCap} one="# Blood lily" other="# Blood lilies" /> did not bloom due to early lily use.
+			</Trans>,
+		})
+		)
+
+		if (this.calculateLostLilies() > 0) {
```
ayyaruq: This is unnecessary and leads to excess indentation. Since you use TieredSuggestions, they implicitly start at the lowest defined tier, for example by having a tier of `1: SEVERITY.MINOR`, anything *less* than 1 won't show the suggestion at all, so the condition isn't needed.

src/parser/jobs/whm/modules/Lilies.tsx
```diff
@@ -0,0 +1,167 @@
...
+				content: content,
+				tiers: tiers,
+				value: this.calculateLostLilies(),
+				why: <Trans id="whm.lily-cap.suggestion.why">
+					{<Plural value={this.calculateLostLilies()} one="# lily" other="# lilies" />} went unused.
+				</Trans>,
+			}))
+		}
+
+		if (this.bloodLilyGauge.value > 0) {
```
ayyaruq: Same with this condition, it isn't doing anything.

ayyaruq (commented):
Thanks for your PR! Some feedback on code structure and some patterns we use in this project mostly, I didn't cover everything, just enough to set you on the right track. Main thing is some of the i18n keys and wording, and then your use of consts/functions when they're not necessarily required. Overall, these are just nits and not functionally required, but worth addressing to make the code easier to work with IMO.

Leaving this as a comment instead of requesting changes, and feel free to hit me up in Discord if you have any questions.

psummers678:
That should be all done, also lost lilies not changing based on patch

src/parser/jobs/whm/modules/Lilies.tsx
```diff
@@ -0,0 +1,159 @@
...
+	@dependency private suggestions!: Suggestions
+
+	private lilyConsumers: ActionKey[] = [
+		'AFFLATUS_RAPTURE',
+		'AFFLATUS_SOLACE',
+	]
+
+	private bloodLilyConsumers: ActionKey[] = [
+		'AFFLATUS_MISERY',
+	]
```
ayyaruq: These should be consts.

src/parser/jobs/whm/modules/Lilies.tsx
```diff
@@ -0,0 +1,159 @@
...
+	private calculateLostLilies(): number {
+		return Math.floor(this.lilyTimer.getExpirationTime() / this.lilyInterval)
+	}
+
+	private onComplete() {
+		const bloodLilyOvercapTiers = {
+			1: SEVERITY.MINOR,
+			2: SEVERITY.MEDIUM,
+			3: SEVERITY.MAJOR,
+		}
```
ayyaruq: This should also be a const. You could make it a nested object if that's easier too, so using it as `SEVERITIES.LILY_OVERCAP_610` or `SEVERITIES.BLOOD_UNUSED` and keeping it all in one place.

src/parser/jobs/whm/modules/Lilies.tsx
```diff
@@ -0,0 +1,159 @@
...
+		this.bloodLilyGauge.generate(1)
+		if (this.lilyTimer.expired) {
+			this.lilyTimer.start()
+		}
+		this.lilyGauge.spend(1)
+	}
+
+	private calculateLostLilies(): number {
+		return Math.floor(this.lilyTimer.getExpirationTime() / this.lilyInterval)
+	}
```
ayyaruq: I'd probably just const this in `onComplete`, since consts are only evaluated once. More efficient.

src/parser/jobs/whm/modules/Lilies.tsx
```diff
@@ -0,0 +1,159 @@
...
+	private onComplete() {
+		const bloodLilyOvercapTiers = {
+			1: SEVERITY.MINOR,
+			2: SEVERITY.MEDIUM,
+			3: SEVERITY.MAJOR,
+		}
+
+		this.suggestions.add(new TieredSuggestion({
+			icon: this.data.actions.AFFLATUS_MISERY.icon,
+			content: <Trans id="whm.gauge.suggestions.bloodlily.overcap.content">
```
ayyaruq: i18n keys are a bit of a mess still, since we're using `whm.gauge.<name>.<thing>.<subthing>`, it should be `whm.gauge.bloodlily.suggestions.overcap.content` in this case.

src/parser/jobs/whm/modules/Lilies.tsx
```diff
@@ -0,0 +1,159 @@
...
+const GRAPH_DISPLAY_SETTINGS = {
+	LILY_GAUGE_FADE: 0.25,
+	LILY_TIMER_FADE: 0.75,
+	BLOODLILY_GAUGE_FADE: 0.5,
+	LILY_GAUGE_COLOR: Color('#4f73b5'),
+	BLOODLILY_GAUGE_COLOR: Color('#b52d6c'),
+}
+
+export class Lilies extends CoreGauge {
+
```
ayyaruq, line 51: Best to add a handle in here, so that modules don't clash. In this case the handle and i18n keys diverge a bit from the class name but eh. Gauge works better than lily/lilies anyway. There's only one gauge for WHM iirc so it doesn't really matter, could even rename the file/class to just Gauge I guess if you cared.
```suggestion
	static override handle = 'gauge'
```

src/parser/jobs/whm/modules/Lilies.tsx
```diff
@@ -0,0 +1,159 @@
...
+		const lilyOvercapSuggestion_610 = <Trans id="whm.gauge.suggestions.lily.overcap.content.610">
+			Try to use <DataLink action="AFFLATUS_RAPTURE" /> or <DataLink action="AFFLATUS_SOLACE" /> before using other GCD heals. It's okay to overheal with your lilies as they are to be used for mana management and movement aswell as healing.
+		</Trans>
+
+		const lilyOvercapTiers_600 = {1: SEVERITY.MINOR}
+		const lilyOvercapTiers_610 = {
+			1: SEVERITY.MEDIUM,
+			3: SEVERITY.MAJOR,
+		}
```
ayyaruq: Prefer to define consts at the start of the function instead of inline like this, although severities should be file-level consts.

src/parser/jobs/whm/modules/Lilies.tsx
```diff
@@ -0,0 +1,159 @@
...
+			value: this.calculateLostLilies(),
+			why: <Trans id="whm.gauge.suggestions.lily.overcap.why">
+				{<Plural value={this.calculateLostLilies()} one="# lily" other="# lilies" />} went unused.
+			</Trans>,
+		}))
+
+		const bloodlilyLeftoverTier = {
+			1: SEVERITY.MINOR,
+			3: SEVERITY.MAJOR,
+		}
```
ayyaruq: Same deal here.

src/parser/jobs/whm/modules/Lilies.tsx
```diff
@@ -0,0 +1,159 @@
...
+import Suggestions, {SEVERITY, TieredSuggestion} from 'parser/core/modules/Suggestions'
+import React from 'react'
+
+const LILY_MAX_STACKS = 3
+const LILY_INTERVAL_600 = 30000
+const LILY_INTERVAL_610 = 20000
+
+const MISERY_COST = 3
+
+const GRAPH_DISPLAY_SETTINGS = {
```
ayyaruq: Not much value to bundling these, I'd merge them so you define them outright as they're needed with the fade inline, ie `FOO: Color(#696969).fade(0.5)` or whatever, since most of these are single use. That way you can also have a distinct colour for the timer if it's better for UX too.

src/parser/jobs/whm/modules/Lilies.tsx
```diff
@@ -0,0 +1,159 @@
...
+	}))
+
+	override initialise() {
+		super.initialise()
+
+		const playerFilter = filter<Event>().source(this.parser.actor.id)
+		this.addEventHook(playerFilter.type('action').action(this.data.matchActionId(this.lilyConsumers)), this.onLilySpend)
+		this.addEventHook(playerFilter.type('action').action(this.data.matchActionId(this.bloodLilyConsumers)), () => this.bloodLilyGauge.spend(MISERY_COST))
+		this.addEventHook('complete', this.onComplete)
+		this.lilyTimer.start()
```
ayyaruq, line 97: Does this implicitly start at the start of a fight, or do you need to do something to prompt it?

src/parser/jobs/whm/modules/Lilies.tsx
```diff
@@ -0,0 +1,159 @@
...
+			this.lilyTimer.start()
+		}
+	}
+
+	private onLilySpend() {
+		this.bloodLilyGauge.generate(1)
+		if (this.lilyTimer.expired) {
+			this.lilyTimer.start()
+		}
+		this.lilyGauge.spend(1)
```
ayyaruq: Nit; easier to read this way. Looks like indentation was borked (spaces rather than tabs) too.
```suggestion
		this.bloodLilyGauge.generate(1)
		this.lilyGauge.spend(1)
		if (this.lilyTimer.expired) {
			this.lilyTimer.start()
		}
```

ayyaruq (changes requested):
Thanks for the PR, nice work! Requesting changes on the consts, and I think some of the strings can be cleaned up a bit to have more consistent wording with the rest of our strings, and better capitalisation or data links for names referenced in the strings.

The rest of my review is basically nits, but I'd get those i18n key names more standardised before merging so we don't run into issues with the translators.

src/parser/jobs/whm/modules/Lilies.tsx
psummers678: As far as I know and figured out it needs to be explicitly started . If this is removed the timer for lilies doesnt start until the first time a lily is used which calls start.

ayyaruq (approved):
What's up with the extra new line under class def but nice, looks good to me!

PR #1679, opened 2022-04-19
Update SAM for 6.1 support

Rjb99:
After a week of torturing myself in hardstuck plat and making myself want to go back to LoL, I'm channeling my spite and minor depression into the SAM updates

Removed Kaiten Status and Action Ids for 6.1 & beyond
Removed Kaiten hooks and output for 6.1 & beyond
Leaving Kaiten alone in the Kenki.tsx array for spenders. no errors popped up in several tests parses.
Fixed positionals by using potency data instead of hard coding postionals values


src/data/ACTIONS/layers/patch6.1.ts
```diff
@@ -9,5 +10,11 @@ export const patch610: Layer<ActionRoot> = {
 		GRIT: {cooldown: 3000},
 		IRON_WILL: {cooldown: 3000},
 		ROYAL_GUARD: {cooldown: 3000},
+
+		//SAM 6.1 action changes:
+		HISSATSU_KAITEN: {id: SHARED.UNKNOWN.id}, //Kaiten was removed. But is job critical for pre-6.1 analysis.
```
ayyaruq: Not sure why you only wanna kill the ID instead of the whole status here.
```suggestion
		HISSATSU_KAITEN: SHARED.UNKNOWN,
```

src/data/STATUSES/layers/patch6.1.ts
```diff
@@ -12,6 +13,10 @@ export const patch610: Layer<StatusRoot> = {
 			duration: 20000,
 		},
 
+		//SAM 6.1 buff Change
+		THIRD_EYE: {duration: 4000},
+		KAITEN: {id: SHARED.UNKNOWN.id},
```
ayyaruq: Same here, just set it in full:
```suggestion
		KAITEN: SHARED.UNKNOWN,
```

src/parser/jobs/sam/changelog.tsx
```diff
@@ -7,6 +7,11 @@ export const changelog = [
 	// 	Changes: () => <>The changes you made</>,
 	// 	contrubutors: [CONTRIBUTORS.YOU],
 	// },
+	{
+		date: new Date('2022-4-18'),
+		Changes: () => <> Removed Kaiten module for parses from patch 6.1+ as it was removed in patch 6.1</>,
```
ayyaruq: Just killing the space but it could read better.
```suggestion
		date: new Date('2022-04-18'),
		Changes: () => <>Removed Kaiten analysis for the patch 6.1 job update</>,
```

src/parser/jobs/sam/modules/Kaiten.tsx
```diff
@@ -52,37 +52,40 @@ export class Kaiten extends Analyser {
 		super.initialise()
 
 		const playerFilter = filter<Event>().source(this.parser.actor.id)
+		// If before 6.1, we need them hooks for kaiten.
+		if (this.parser.patch.before('6.1')) {
```
ayyaruq, line 55: I'd include the player filter in this, but consider just returning early if it's 6.1 or later instead.

src/parser/jobs/sam/modules/Kaiten.tsx
```diff
@@ -109,40 +112,43 @@ export class Kaiten extends Analyser {
...
-				Avoid using <DataLink action = "HISSATSU_KAITEN"/> when already under the effect of it.
-			</Trans>,
-			tiers: SEVERITY_TIERS,
-			why: <Trans id ="sam.suggestion.doublekaiten.why">
-					You used Kaiten  <Plural value={this.doubleKaitens} one="# time" other="# times"/> when you already had it up.
-			</Trans>,
-			value: this.doubleKaitens,
-		}))
+		// No suggestions if after patch 6.1
+		if (this.parser.patch.before('6.1')) {
```
ayyaruq: You initialise all these vars as 0 and they use TieredSuggestions, so they won't throw anyway. Here too tho, consider just returning early in the function if it's 6.1+ I guess.

src/parser/jobs/sam/modules/Kenki/Kenki.tsx
```diff
@@ -60,7 +60,7 @@ export class Kenki extends CoreGauge {
 		[this.data.actions.HISSATSU_KYUTEN.id, {action: -25}],
 		[this.data.actions.HISSATSU_SHINTEN.id, {action: -25}],
 		[this.data.actions.HISSATSU_SENEI.id, {action: -25}],
-		[this.data.actions.HISSATSU_KAITEN.id, {action: -20}],
+		[this.data.actions.HISSATSU_KAITEN.id, {action: -20}], //Removed. In theory an unknown action will break this if cast by player.
```
ayyaruq: Remove it from the list entirely, then in the initialiser, if it's pre-6.1 add it selectively. Maps are nice like that.

ayyaruq (changes requested):
Minor changes, specifically that map for kenki and the data layers, but I think you could simplify a lot of the kaiten stuff too.

src/data/ACTIONS/layers/patch6.1.ts
```diff
@@ -9,5 +10,11 @@ export const patch610: Layer<ActionRoot> = {
...
+		// Potency buffs, very important, breaks postionals without them.
+		GEKKO: {potency: 330},
+		KASHA: {potency: 330},
```
eatmorehufu: I think more needs to be done to fix the module. The Positional logic relies on the ratio of Combo potency (no positional) to No-combo potency (also no positional) which was also updated.

In both Gekko and Kasha's case, it seems to be:
100 -> 120
320 -> 330.

This fixes only the 330 part, but not the 120 part.

In https://github.com/xivanalysis/xivanalysis/blob/endwalker/src/parser/core/modules/Positionals.tsx I chose to model potencies as a list of potencies with a repeated field of modifiers, which might be necessary if you want to start adding potencies to actions.

src/parser/jobs/sam/modules/Kaiten.tsx
Rjb99: is there a difference besides style for returning early vs checking before moving on?

src/parser/jobs/sam/modules/Kaiten.tsx
Rjb99: removed the redundancy.

src/parser/jobs/sam/modules/Kaiten.tsx
ayyaruq: Return early isn't really a pattern for this use case, I just felt it was better than wrapping a whole module in a giant condition, especially when the onComplete function will never be called if the hooks aren't set. Prob the cleanest way is to roll all of the init function into the condition, and then leave onComplete alone. Might be worth asking ack tho.

src/parser/jobs/sam/modules/Kenki/Kenki.tsx
Rjb99: Change made.

src/data/ACTIONS/layers/patch6.1.ts
Rjb99: I added a "noComboPotency" to the Actions to handle that issue. if we want to fix the actual postionals code to handle it instead, let this be a bandaid.

ayyaruq: Might be better to fix it properly and just wait for that PR, so something like potency being optional still but typed as `number | PotencyConfig` or some such where the latter is an interface describing modifiers like base, combo, positional, proc, etc. Sorta like how combos work except more flexible and backwards compatible.

src/parser/jobs/sam/modules/Kaiten.tsx
```diff
@@ -109,38 +112,40 @@ export class Kaiten extends Analyser {
...
 		this.suggestions.add(new TieredSuggestion({
 			icon: this.data.actions.HISSATSU_KAITEN.icon,
 			content: <Trans id = "sam.kaiten.suggestion.badkaiten.content">
-				Avoid using <DataLink action="HISSATSU_KAITEN"/> on any GCDs besides <DataLink action ="IAIJUTSU"/> moves and <DataLink action = "OGI_NAMIKIRI"/>. These actions are worth it because of the potency gained per kenki spent.
+					Avoid using <DataLink action="HISSATSU_KAITEN"/> on any GCDs besides <DataLink action ="IAIJUTSU"/> moves and <DataLink action = "OGI_NAMIKIRI"/>. These actions are worth it because of the potency gained per kenki spent.
 			</Trans>,
 			tiers: SEVERITY_TIERS,
 			why: <Trans id ="sam.suggestion.badkaiten.why">
-					You used Kaiten  <Plural value={this.badKaitens} one="# time" other="# times"/> on non-optimal GCDs.
+						You used Kaiten  <Plural value={this.badKaitens} one="# time" other="# times"/> on non-optimal GCDs.
```
ayyaruq: These seem unintentional?

src/parser/jobs/sam/modules/Kaiten.tsx
```diff
@@ -109,38 +112,40 @@ export class Kaiten extends Analyser {
...
 		this.suggestions.add(new TieredSuggestion({
 			icon: this.data.actions.HISSATSU_KAITEN.icon,
 			content: <Trans id = "sam.kaiten.suggestion.missedkaiten.content">
-				Always use <DataLink action = "HISSATSU_KAITEN"/> on <DataLink action = "MIDARE_SETSUGEKKA"/>,<DataLink action = "HIGANBANA"/> and <DataLink action="OGI_NAMIKIRI"/>. The gain on these actions from kaiten is too great to miss.
+					Always use <DataLink action = "HISSATSU_KAITEN"/> on <DataLink action = "MIDARE_SETSUGEKKA"/>,<DataLink action = "HIGANBANA"/> and <DataLink action="OGI_NAMIKIRI"/>. The gain on these actions from kaiten is too great to miss.
 			</Trans>,
 			tiers: SEVERITY_TIERS,
 			why: <Trans id ="sam.suggestion.missedkaiten.why">
-					You forgot to use Kaiten  <Plural value={this.missedKaitens} one="# time" other="# times"/> on optimal GCDs.
+						You forgot to use Kaiten  <Plural value={this.missedKaitens} one="# time" other="# times"/> on optimal GCDs.
```
ayyaruq: Same here, prob check the indentation for the lot since even pre-PR the why text here was already overly indented.

src/parser/jobs/sam/modules/Kaiten.tsx
```diff
@@ -109,38 +112,40 @@ export class Kaiten extends Analyser {
...
 		this.suggestions.add(new TieredSuggestion({
 			icon: this.data.actions.HISSATSU_KAITEN.icon,
 			content: <Trans id = "sam.kaiten.suggestion.doublekaiten.content">
-				Avoid using <DataLink action = "HISSATSU_KAITEN"/> when already under the effect of it.
+					Avoid using <DataLink action = "HISSATSU_KAITEN"/> when already under the effect of it.
 			</Trans>,
 			tiers: SEVERITY_TIERS,
 			why: <Trans id ="sam.suggestion.doublekaiten.why">
-					You used Kaiten  <Plural value={this.doubleKaitens} one="# time" other="# times"/> when you already had it up.
+						You used Kaiten  <Plural value={this.doubleKaitens} one="# time" other="# times"/> when you already had it up.
```
ayyaruq: Here too.

src/parser/jobs/sam/modules/Kenki/Kenki.tsx
```diff
@@ -72,6 +71,10 @@ export class Kenki extends CoreGauge {
 	override initialise() {
 		super.initialise()
 
+		if (this.parser.patch.before('6.1')) {
+			this.kenkiGaugeModifiers.set(this.data.actions.HISSATSU_KAITEN.id, {action: -20})
+		}
```
ayyaruq, line 76: Nice.

ayyaruq (changes requested):
Surprised the linter doesn't rage at the borked indentation. Nice work with the 6.1 overrides, this is likely the cleanest approach for it realistically. Requesting changes to fix the indentation but otherwise this looks good to me for the SAM specific bits.

I think the core positionals module needs a major interface rethink tho, in terms of both the data interfaces and what/how things are configured. Maybe worth waiting for that, maybe not, but I think a proper structure in the data system would make the config for jobs much simpler. Prob worth a word from @eatmorehufu on how feasible that is, and up to you if you're down to merge with this bandaid even tho it's kinda ehhh.

src/parser/jobs/sam/modules/Kaiten.tsx
Rjb99: I suspect VSC trying to auto fix indents when I removed the patch check borked it.

ayyaruq (commented):
LGTM pending the positionals stuff now, thanks for addressing the requested changes.

ayyaruq (approved):

PR #1681, opened 2022-04-23
Updated SMN for 6.1

KelosPelmand:


src/data/ACTIONS/layers/patch6.1.ts
```diff
@@ -23,5 +23,17 @@ export const patch610: Layer<ActionRoot> = {
...
+		SEARING_LIGHT: {
+			icon: 'https://xivapi.com/i/002000/002780.png',
+			statusesApplied: ['SEARING_LIGHT'],
+		},
+		RUBY_RITE: {
+			gcdRecast: 3000,
+		},
+		RUBY_CATASTROPHE: {
+			gcdRecast: 3000,
+		},
```
ayyaruq, line 37: Is this missing Ruby Disaster, Ruby Outburst, and Ruby Ruin III? I'm not familiar enough with lv90 SMN to know what's actually available vs replaced by traits, or just not used much since iirc Ruby/Emerald/Topaz Disaster are replaced with Tri-Disaster when you don't have a summon out or something?

src/parser/jobs/smn/modules/SearingLight.tsx
```diff
@@ -65,17 +65,40 @@ export class SearingLight extends Analyser {
 		this.petIds = this.parser.pull.actors
 			.filter(actor => actor.owner === this.parser.actor)
 			.map(actor => actor.id)
+
+		const playerFilter = filter<Event>().source(this.parser.actor.id)
```
ayyaruq, line 69: Not strictly part of the diff, but I'd prob one-line the pets filter too, easier to read.

ayyaruq (commented):
LGTM, one question on the data and otherwise just an observation. If the data is good, happy to merge.

src/data/ACTIONS/layers/patch6.1.ts
KelosPelmand: Those are all the lower level skills that are not currently in the xiva data.  I do plan to put those in at some point to make low level logs not so terrible.  Whenever I put the lower level skills in the base data, the ones you mentioned will get entries here at that point.

ayyaruq: That's fine, I'd not bother if they're lower level. Was hard to tell what's actually available at lv90 from the job guide page on Lodestone.

src/parser/jobs/smn/modules/SearingLight.tsx
KelosPelmand: i figured i would leave the line alone since it didnt need to change, but updated per request

ayyaruq (approved):
LGTM!

PR #1682, opened 2022-04-24
Update RPR texts

ayyaruq:
Fixing a typo in Soul Reaver, and making Soulsow more explicit in Harvest Moon.


src/parser/jobs/rpr/changelog.tsx
```diff
@@ -3,6 +3,11 @@ import CONTRIBUTORS from 'data/CONTRIBUTORS'
 import React from 'react'
 
 export const changelog = [
+	{
+		date: new Date('2022-04-24'),
+		Changes: () => <>Reword Harvest Moon text to make Soulsow's use clearer.</>,
```
eatmorehufu: Use clearer... what? :P

src/parser/jobs/rpr/modules/Reaver.tsx
```diff
@@ -103,7 +103,7 @@ export class Reaver extends Analyser {
 			</Trans>,
 			tiers: SEVERITIES,
 			why: <Trans id ="rpr.reaver.suggestion.dropped.why">
-					You used lost <Plural value={this.droppedReavers} one="# stack" other="# stacks"/> over the course of the fight.
+					You used <Plural value={this.droppedReavers} one="# stack" other="# stacks"/> over the course of the fight.
```
eatmorehufu: this should be "lost" right? or is that not what droppedReavers means?

eatmorehufu (changes requested):

src/parser/jobs/rpr/modules/Reaver.tsx
ayyaruq: Reading this I thought maybe I was high or something when I wrote this but it turns out following the exact request was a bad idea and the user wanted the wrong word dropped, will resolve in a bit.

src/parser/jobs/rpr/changelog.tsx
ayyaruq: Like, to make the usage of Soulsow clearer. Maybe an Australianism, it makes sense to me. Will update.

AzariahAeroborn (approved):

PR #1683, opened 2022-04-24
Mark patch 6.1 support for DRK

ayyaruq:
Additionally, fixed some code legibility issues in Blood gauge and a skill name in data. Main thing was the timeline status changes, both the new status and combining Salted Earth with Salt and Darkness.

Blood gauge change was to prevent the bug where Blood numbers show super high as it previously tracked each stack as a separate application, changing to the action is basically Good Enough™. Blood Weapon itself needs no logical changes as core BuffWindow checks if there's an open window already and safely handles the stacks.

### DIFF
--- a/src/data/STATUSES/layers/patch6.1.ts
+++ b/src/data/STATUSES/layers/patch6.1.ts
@@ -14,7 +14,7 @@ export const patch610: Layer<StatusRoot> = {
 		// Duration varies by the remaining duration on Walking Dead when it was cleansed
 		UNDEAD_REBIRTH: {
 			id: 3255,
-			name: 'Undeath Rebirth',
+			name: 'Undead Rebirth',
 			icon: 'https://xivapi.com/i/013000/013124.png',
 		},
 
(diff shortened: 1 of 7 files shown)


src/data/STATUSES/layers/patch6.1.ts
```diff
@@ -5,6 +5,19 @@ import {SHARED} from '../root/SHARED'
...
+		// DRK 6.1 buff changes
+		BLOOD_WEAPON: {
+			duration: 15000,
+			stacksApplied: 5,
+		},
+
+		// Duration varies by the remaining duration on Walking Dead when it was cleansed
+		UNDEAD_REBIRTH: {
+			id: 3255,
+			name: 'Undeath Rebirth',
```
AzariahAeroborn: Name doesn't match key. Which one matches the game data?

AzariahAeroborn (dismissed):

src/data/STATUSES/layers/patch6.1.ts
ayyaruq: I got bamboozled, the key is correct, updating the rest. Good catch.

AzariahAeroborn (approved):

PR #1684, opened 2022-04-24
Migrate positionals types and definitions from core/jobs to data to better support layering.

eatmorehufu:


eatmorehufu:
These lint failures are from WHM changes, unrelated to this change.

src/data/ACTIONS/root/MNK.ts
```diff
@@ -37,7 +44,6 @@ export const MNK = ensureActions({
 		icon: 'https://xivapi.com/i/000000/000213.png',
 		onGcd: true,
 		speedAttribute: Attribute.SKILL_SPEED,
-		potency: 280,
```
ayyaruq, line 40: Let's not break the Twin Snakes module.

src/data/ACTIONS/root/RPR.ts
```diff
@@ -166,6 +166,24 @@ export const RPR = ensureActions({
...
+		potencies: [{
+			value: 400,
+			bonusModifiers: [],
+			baseModifiers: [],
+		}, {
+			value: 460,
+			bonusModifiers: [BonusModifier.POSITIONAL],
+			baseModifiers: [],
+		}, {
+			// Lance Mastery increases the potency of the 5th hit by 100
```
ayyaruq: Lance Mastery is a DRG skill so...

src/data/ACTIONS/root/RPR.ts
```diff
@@ -175,6 +193,24 @@ export const RPR = ensureActions({
...
```
ayyaruq: Here too, def not a thing for RPR.

src/data/ACTIONS/root/SAM.ts
```diff
@@ -82,7 +82,20 @@ export const SAM = ensureActions({
 		icon: 'https://xivapi.com/i/003000/003158.png',
 		onGcd: true,
 		speedAttribute: Attribute.SKILL_SPEED,
-		potency: 320,
+		// These are 6.0 values updated in the 6.1 layer.
```
ayyaruq: Unnecessary comment, since they'll likely be updated in future layers too and nobody should have to maintain a data comment like this.

src/data/ACTIONS/root/SAM.ts
```diff
@@ -117,7 +130,20 @@ export const SAM = ensureActions({
 		icon: 'https://xivapi.com/i/003000/003164.png',
 		onGcd: true,
 		speedAttribute: Attribute.SKILL_SPEED,
-		potency: 320,
+		// These are 6.0 values updated in the 6.1 layer.
```
ayyaruq: Same deal here too.

src/data/ACTIONS/type.ts
```diff
@@ -9,6 +9,34 @@ export type ActionCombo =
...
+// augments the potency of the 5th hit.
+export enum PotencySpecialCase {
+	DRG_LANCE_MASTERY,
+}
+
+// When calculating the bonusPercent that the game uses to display
+// combo and positional success, BonusModifiers are the modifiers that
+// will increase the percent. Statuses from jobs like RPR and DRG
+// modify the base.
+export type BaseModifier = PotencySpecialCase | keyof StatusRoot
```
ayyaruq: Maybe prefer `StatusKey` over `keyof StatusRoot`.

src/data/ACTIONS/type.ts
```diff
@@ -9,6 +9,34 @@ export type ActionCombo =
...
+
+// Potency is modeled this way because any single potency number
+// can have a combination of states that apply to it, see all
+// of the commented out PotencyModifiers. An empty modifier
+// list means it's the base potency.
+export interface Potency {
+	value: number,
+	baseModifiers?: BaseModifier[]
+	bonusModifiers: BonusModifier[]
+}
```
ayyaruq, line 39: Prefer to not double up usage of a field name and class name like this. Maybe `PotencyConfig` or something instead. That said, I'm not sold on this model yet. A lot of the states overlap so defining discrete states seems weird even if more intuitive than having to calculate the specific modifiers. I'd prob rename `value` here to `potency` and avoid duplicating the existing key name. I'll note that the actual data fields allow any additional stuff so potency itself isn't strictly defined.

ayyaruq (dismissed):
Requested change to not break MNK, we need that `potency` field separately.

Comments could be cleaned up but I'm not sold on this model, the array is difficult to use outside of the module it's for and the naming diverges a bit from what the game does. We've never really modelled traits, but maybe we should. Main thing is, please don't break MNK.

src/data/ACTIONS/root/SAM.ts
eatmorehufu: Fixed

src/data/ACTIONS/root/SAM.ts
eatmorehufu: Fixed

src/data/ACTIONS/root/RPR.ts
eatmorehufu: Fixed

src/data/ACTIONS/root/RPR.ts
eatmorehufu: Fixed

src/data/ACTIONS/type.ts
eatmorehufu: I don't know what you mean by Strictly Defined. I don't _love_ value and I think potency is better but I don't think "config" makes sense as a name for this object. Potency is close to the right name but a more verbose attempt would be ConditionalPotency, PotencyCondition, or ModifiedPotency (even though it's possibly unmodified). Those are just off the top of my head to be more descriptive, but I think they're worse overall.

Even though I think the `value` field should ideally be called potency, the new Action field `potencies` has a list of `Potency` objects with this naming scheme, which works well.

src/data/ACTIONS/type.ts
eatmorehufu: Done.

eatmorehufu:
> 

Migrated MNK potency definitions to data so they can also all be layered, I wrote two potency utility functions and slapped them into action/index but I'm also open to those living in a better place. In terms of expression it would be most convenient if they were defined on the interface itself, but I think it's best to just have external utilities.

src/data/ACTIONS/root/MNK.ts
eatmorehufu: Migrated/fixed the MNK potency definitions to use the new system.

AzariahAeroborn:
I think there may be some value to circle back on this, with respect to ayyaruq's comments on the naming and I'm not 100% sold on why separate modifier types for COMBO & POSITIONAL vs buffs/traits like Lance Mastery or Leaden Fist.  But this does what it says on the tin and provides a good improvement to get positional data into the layers structure, so I'm going to merge it as is and request some additional consideration of the naming for a follow-up PR.

AzariahAeroborn (approved):

PR #1685, opened 2022-04-24
DRG: Update Jump-related Rules

ebshimizu:
our jumps are fast now, so all jumps except for stardiver are generally ok to weave. This also means that we now expect both SSD charges to be held for Dragon Sight, as it's now feasible to weave within those windows.

### DIFF
--- a/src/parser/jobs/drg/modules/Weaving.ts
+++ b/src/parser/jobs/drg/modules/Weaving.ts
@@ -2,11 +2,8 @@ import {ActionKey} from 'data/ACTIONS'
 import {Weaving as CoreWeaving, Weave} from 'parser/core/modules/Weaving'
 import DISPLAY_ORDER from './DISPLAY_ORDER'
 
+// With the reduced animation lock, it's just stardiver that's the bad weave
 const JUMPS: ActionKey[] = [
-	'JUMP',
-	'HIGH_JUMP',
-	'SPINESHATTER_DIVE',
-	'DRAGONFIRE_DIVE',
 	'STARDIVER',
 ]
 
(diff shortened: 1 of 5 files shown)


ayyaruq:
Is DRG going to be bumping the minimum patch to 6.1 then?

ebshimizu:
> Is DRG going to be bumping the minimum patch to 6.1 then?

mmmm yes we should do that.

src/parser/jobs/drg/modules/Weaving.ts
```diff
@@ -2,21 +2,30 @@ import {ActionKey} from 'data/ACTIONS'
...
 ]
 
 export default class Weaving extends CoreWeaving {
 	static override displayOrder = DISPLAY_ORDER.WEAVING
 
-	private jumpIds = JUMPS.map(key => this.data.actions[key].id)
+	private jumpIds = JUMPS_ALL.map(key => this.data.actions[key].id)
+
+	// pre-6.1, all jumps are bad weaves
+	private jump600Ids = [...this.jumpIds, ...JUMPS_600.map(key => this.data.actions[key].id)]
```
ayyaruq: Alternatively, set the var and conditionally append to it, then just use it below as before. Something like this:

```ts
private jumps = JUMPS.map(key => this.data.actions[key].id)

if (this.parser.patch.before('6.1') {
	this.jumps.concat(JUMPS_600.map(key => this.data.actions[key].id))
}

ayyaruq (commented):
More of a comment than requested change on weaving, simpler that way tho I think. Have you tested this on logs from 6.05, 6.08, and 6.1? The cases aren't particularly complex but would be good to validate. Looks good to me, but I'd check these cases. If you're happy with it, let me know and I'll approve.

src/parser/jobs/drg/modules/Weaving.ts
ebshimizu: condensed it a bit, let me know if i've misinterpreted this note

ebshimizu:
> Have you tested this on logs from 6.05, 6.08, and 6.1?

Tested on 6.0-6.1 and looks sane to me.

ayyaruq (approved):
I like how much you simplified the change here, nicely done.

PR #1686, opened 2022-04-25
6.1 warrior stormseye

ghost:
Added updated Surging Tempest early refresh tracking based upon conversations with @ayyaruq


src/parser/jobs/war/modules/SurgingTempest.tsx
```diff
@@ -41,3 +65,62 @@ export class SurgingTempest extends Analyser {
 		return (statusUptime / fightUptime) * 100
 	}
 }
+
+export class StormsEye extends CoreGauge {
```
ayyaruq: Why does this need a separate class?

src/parser/jobs/war/modules/SurgingTempest.tsx
```diff
@@ -1,13 +1,37 @@
...
+	'STORMS_EYE',
+]
+const SURGING_TEMPEST_GENERATION_AMOUNT = 30
+// Surging Tempest Extension
+const SURGING_TEMPEST_EXTENDERS: ActionKey[] = ['INNER_RELEASE']
+const SURGING_TEMPEST_EXTENSION_AMOUNT = 10
+
+const SURGING_TEMPEST_EARLY_REFRESH_GRACE = 7.5
+const STORMS_EYE_LOST_GAUGE = 10
+//TODO: Discuss with acc to select correct tier breakpoints
```
ayyaruq: I'd probably just use a map for these, roll it into a single hook and then check if the action is IR or a member of a list to extend rather than grant.

src/parser/jobs/war/modules/SurgingTempest.tsx
```diff
@@ -41,3 +65,62 @@ export class SurgingTempest extends Analyser {
...
+		this.addEventHook(
+			filter<Event>()
+				.source(this.parser.actor.id)
+				.action(this.data.matchActionId(SURGING_TEMPEST_EXTENDERS)),
+			(_) => this.extendSurgingTempest(SURGING_TEMPEST_EXTENSION_AMOUNT, true)
+		)
+		this.addEventHook('complete', this.onComplete)
+	}
+
+	private extendSurgingTempest(amount: number, onlyIfRunning ?: boolean) {
```
ayyaruq: You can just make the boolean field default to false, or alternatively remove it entirely and have a local var for it - if it's IR, we set it to true, otherwise false.

src/parser/jobs/war/modules/SurgingTempest.tsx
```diff
@@ -41,3 +65,62 @@ export class SurgingTempest extends Analyser {
...
+		onlyIfRunning === undefined ? false : onlyIfRunning
+
+		if (this.surgingTempest.remaining > SURGING_TEMPEST_EARLY_REFRESH_GRACE) {
+			this.earlyRefreshCount++
+		}
+
+		if ((this.surgingTempest.expired || this.surgingTempest.paused) && !onlyIfRunning) {
+			this.surgingTempest.start()
+		}
```
ayyaruq: This isn't necessary, the extend method has the same param for if the gauge is running and will auto start if it's false.

src/parser/jobs/war/modules/SurgingTempest.tsx
```diff
@@ -41,3 +65,62 @@ export class SurgingTempest extends Analyser {
...
+		this.surgingTempest.extend(amount, onlyIfRunning)
+	}
+
+	private onComplete() {
+		if (this.earlyRefreshCount > 0) {
```
ayyaruq: This also isn't necessary, tiered suggestions only show when their minimum tier is hit, so since your lowest tier isn't 0 (which would always show it so kinda derp), this conditional has no effect other than making the code more indented.

src/parser/jobs/war/modules/SurgingTempest.tsx
```diff
@@ -41,3 +65,62 @@ export class SurgingTempest extends Analyser {
...
 	}
 }
+
+export class StormsEye extends CoreGauge {
+	static override handle = 'stormseye'
+
+	@dependency private suggestions!: Suggestions
+
+	private surgingTempest = this.add(new TimerGauge({
+		maximum: 60,
```
ayyaruq: I'd const this, even tho it's only ever changed once.

src/parser/jobs/war/modules/SurgingTempest.tsx
```diff
@@ -41,3 +65,62 @@ export class SurgingTempest extends Analyser {
...
+	}
+
+	private onComplete() {
+		if (this.earlyRefreshCount > 0) {
+			this.suggestions.add(new TieredSuggestion({
+				content: <Trans id="war.stormseye.suggestions.overwrite.content">
+				Avoid refreshing <DataLink status="SURGING_TEMPEST" /> too early
+				</Trans>,
+				why: <Trans id="war.stormseye.suggestions.overwrite.why">
+					<DataLink action="STORMS_EYE" /> Does less damage and generates less Beast Gauge than <DataLink action="STORMS_PATH" />. You lost {this.earlyRefreshCount * STORMS_EYE_LOST_GAUGE} Beast Gauge over the course of the fight due to early refreshes.
```
ayyaruq: Why text doesn't need to explain what they should be doing, only what they did.
```suggestion
					You lost {this.earlyRefreshCount * STORMS_EYE_LOST_GAUGE} Beast Gauge over the course of the fight due to early refreshes.
```

ayyaruq: Why text doesn't need to explain what they should be doing, only what they did.
```suggestion
					You lost {this.earlyRefreshCount * STORMS_EYE_LOST_GAUGE} Beast Gauge over the course of the fight due to early refreshes.
```

src/parser/jobs/war/modules/SurgingTempest.tsx
```diff
@@ -41,3 +65,62 @@ export class SurgingTempest extends Analyser {
...
+			this.surgingTempest.start()
+		}
+		this.surgingTempest.extend(amount, onlyIfRunning)
+	}
+
+	private onComplete() {
+		if (this.earlyRefreshCount > 0) {
+			this.suggestions.add(new TieredSuggestion({
+				content: <Trans id="war.stormseye.suggestions.overwrite.content">
+				Avoid refreshing <DataLink status="SURGING_TEMPEST" /> too early
```
ayyaruq: This could use some detail, specifically what the benefit of avoiding early refreshes is.

src/parser/jobs/war/modules/SurgingTempest.tsx
```diff
@@ -41,3 +65,62 @@ export class SurgingTempest extends Analyser {
...
+			filter<Event>()
+				.source(this.parser.actor.id)
+				.action(this.data.matchActionId(SURGING_TEMPEST_EXTENDERS)),
+			(_) => this.extendSurgingTempest(SURGING_TEMPEST_EXTENSION_AMOUNT, true)
+		)
+		this.addEventHook('complete', this.onComplete)
+	}
+
+	private extendSurgingTempest(amount: number, onlyIfRunning ?: boolean) {
+		onlyIfRunning === undefined ? false : onlyIfRunning
```
ayyaruq: As noted above and below, you can either kill this or remove the param and set this explicitly.

src/parser/jobs/war/modules/SurgingTempest.tsx
```diff
@@ -41,3 +65,62 @@ export class SurgingTempest extends Analyser {
...
+
+	private earlyRefreshCount = 0
+
+	override initialise() {
+		super.initialise()
+		this.addEventHook(
+			filter<Event>()
+				.source(this.parser.actor.id)
+				.action(this.data.matchActionId(SURGING_TEMPEST_GENERATORS)),
+			(_) => this.extendSurgingTempest(SURGING_TEMPEST_GENERATION_AMOUNT)
```
ayyaruq: Prefer not to put the placeholder assignment here, we just use `()` outright if there's no args being used. I'm guessing this is a holdover from either C# or Go.

src/parser/jobs/war/modules/SurgingTempest.tsx
```diff
@@ -41,3 +65,62 @@ export class SurgingTempest extends Analyser {
...
+		)
+
+		this.addEventHook(
+			filter<Event>()
+				.source(this.parser.actor.id)
+				.action(this.data.matchActionId(SURGING_TEMPEST_EXTENDERS)),
+			(_) => this.extendSurgingTempest(SURGING_TEMPEST_EXTENSION_AMOUNT, true)
```
ayyaruq: Don't think a second hook is really justified, but same issue here re the underscore.

ayyaruq (changes requested):
Solid start, some tidying to do tho I think. I'd probably not split these into separate classes, although the timer gauge itself doesn't track overcap or provide total uptime which could be useful. I made a PR for the former, prob gotta think about the latter.

The suggestion content also needs some rewording, it doesn't really cover the benefit. If you're not sure what to write that's fine, we can sort it out with some of the WAR theory peeps as noted in discord.

Worth noting, times are in ms, not seconds, so your early refresh grace is gonna be more like 7500 than 7.5, similarly for the extensions.

src/parser/jobs/war/modules/SurgingTempest.tsx
ghost: C# <\_< >\_>

src/parser/jobs/war/modules/SurgingTempest.tsx
ghost: linter refuses to allow constanting the timergauge

ayyaruq: You don't const the timer gauge, you const the max value and then set `maximum: MAH_CONST,`.

src/parser/jobs/war/changelog.tsx
```diff
@@ -2,6 +2,13 @@ import CONTRIBUTORS from 'data/CONTRIBUTORS'
 import React from 'react'
 
 export const changelog = [
+	{
+		date: new Date('2022-04-25'),
+		Changes: () => <>
+			Reintroduce early Surging Tempest refresh tracking
+		</>,
```
ayyaruq:
```suggestion
		Changes: () => <>Add Surging Tempest early refresh analyser.</>,
```

src/parser/jobs/war/modules/SurgingTempest.tsx
```diff
@@ -1,25 +1,76 @@
...
 
-export class SurgingTempest extends Analyser {
+const SURGING_TEMPEST_GENERATORS: ActionKey[] = [
+	'MYTHRIL_TEMPEST',
+	'STORMS_EYE',
+	'INNER_RELEASE',
+]
+const SURGING_TEMPEST_GENERATION_AMOUNT = 30
+// Surging Tempest Extension
+const SURGING_TEMPEST_EXTENSION_AMOUNT = 10
```
ayyaruq: Thinking about this, tightening up the hooks and using a map for this is probably better, since you want the combo event for Storm's Eye and Mythril Tempest but the action event for Inner Release.

src/parser/jobs/whm/modules/Lilies.tsx
```diff
@@ -15,10 +15,14 @@ const LILY_INTERVAL_610 = 20000
...
+const ONE_HALF = 0.5
+const THREE_QUARTERS = 0.75
+
 const GAUGE_COLOURS = {
-	LILY_GAUGE: Color('#4f73b5').fade(0.25),
-	LILY_TIMER: Color('#4f73b5').fade(0.75),
-	BLOODLILY: Color('#b52d6c').fade(0.5),
+	LILY_GAUGE: Color('#4f73b5').fade(ONE_QUARTER),
+	LILY_TIMER: Color('#4f73b5').fade(THREE_QUARTERS),
+	BLOODLILY: Color('#b52d6c').fade(ONE_HALF),
```
ayyaruq: Out of scope for this PR.

src/parser/jobs/war/modules/SurgingTempest.tsx
```diff
@@ -1,25 +1,76 @@
...
+
+const ONE_MINUTE = 60000
+
+const SURGING_TEMPEST_EARLY_REFRESH_GRACE = 7500
+const STORMS_EYE_LOST_GAUGE = 10
```
ayyaruq: Technically the Storm's Path gauge was lost. Also generally, shorter var names.
```suggestion
const PATH_LOST_GAUGE = 10
```

src/parser/jobs/war/modules/SurgingTempest.tsx
```diff
@@ -1,25 +1,76 @@
...
+	'STORMS_EYE',
+	'INNER_RELEASE',
+]
+const SURGING_TEMPEST_GENERATION_AMOUNT = 30
+// Surging Tempest Extension
+const SURGING_TEMPEST_EXTENSION_AMOUNT = 10
+
+const ONE_MINUTE = 60000
+
+const SURGING_TEMPEST_EARLY_REFRESH_GRACE = 7500
```
ayyaruq: Technically this buffer only applies to Eye, since it's 5000 for Tempest, basically irrelevant for 6.1 IR, and then 5000 for pre-6.1 IR, because you get 3 GCDs consuming IR stacks so need 6 GCDs, but then you can ignore 4 due to the 10s extension from IR itself. Not super sold on that tho, so `SURGE_BUFFER` works too even if it's really more specific than that.
```suggestion
const EYE_BUFFER = 7500
```

src/parser/jobs/war/modules/SurgingTempest.tsx
```diff
@@ -1,25 +1,76 @@
...
+const STORMS_EYE_LOST_GAUGE = 10
+//TODO: Discuss with acc to select correct tier breakpoints
+const SUGGESTION_TIERS = {
+	1: SEVERITY.MINOR,
+	2: SEVERITY.MEDIUM,
+	5: SEVERITY.MAJOR,
+}
```
ayyaruq:
```suggestion

const SUGGESTION_TIERS = {
	1: SEVERITY.MINOR,
	2: SEVERITY.MEDIUM,
	5: SEVERITY.MAJOR,
}
```

src/parser/jobs/war/modules/SurgingTempest.tsx
```diff
@@ -1,25 +1,76 @@
...
+			this.extendSurgingTempest
+		)
 		this.addEventHook('complete', this.onComplete)
 	}
 
+	private surgingTempest = this.add(new TimerGauge({
+		maximum: ONE_MINUTE,
+	}))
+
+	private earlyRefreshCount = 0
```
ayyaruq: Prefer to define vars before the init function. Also, var names:
```suggestion
	private surgeTimer = this.add(new TimerGauge({
		maximum: SURGE_MAX,
	}))

	private earlyEyes = 0
```

src/parser/jobs/war/modules/SurgingTempest.tsx
```diff
@@ -1,25 +1,76 @@
...
+const SURGING_TEMPEST_GENERATORS: ActionKey[] = [
+	'MYTHRIL_TEMPEST',
+	'STORMS_EYE',
+	'INNER_RELEASE',
+]
+const SURGING_TEMPEST_GENERATION_AMOUNT = 30
+// Surging Tempest Extension
+const SURGING_TEMPEST_EXTENSION_AMOUNT = 10
+
+const ONE_MINUTE = 60000
```
ayyaruq:
```suggestion
const SURGE_MAX = 60000
```

src/parser/jobs/war/modules/SurgingTempest.tsx
```diff
@@ -1,25 +1,76 @@
...
 
+	private surgingTempest = this.add(new TimerGauge({
+		maximum: ONE_MINUTE,
+	}))
+
+	private earlyRefreshCount = 0
+
+	private extendSurgingTempest(event: Event) {
+		//if I remove this my editor starts screaming at me and disables code completion, we can remove it in review if it's that malodorous.
+		if (!(event.type === 'action')) { return }
```
ayyaruq: Use the proper event types here, see the WAR gauge module for more details on how to handle the event types, hooks, and how it aligns with what I wrote above re maps.
```suggestion
	private onSurge(event: Events['action' | 'combo']) {
```

src/parser/jobs/war/modules/SurgingTempest.tsx
```diff
@@ -1,25 +1,76 @@
...
+
+		if (this.surgingTempest.remaining > SURGING_TEMPEST_EARLY_REFRESH_GRACE) {
+			this.earlyRefreshCount++
+		}
+		if (event.action === this.data.actions.INNER_RELEASE.id) {
+			this.surgingTempest.extend(SURGING_TEMPEST_EXTENSION_AMOUNT, true)
+		} else {
+			this.surgingTempest.extend(SURGING_TEMPEST_GENERATION_AMOUNT)
+		}
```
ayyaruq: I'd prob rethink this with a switch and some fallthrough. You don't want to count early refreshes that aren't from Eye since it's normal for AoE to overcap - there's no *other* option unlike the single target case. Prob also worth assigning a local const for if it grants or extends and just setting it based on IR, then calling the gauge extend just once.

src/parser/jobs/war/modules/SurgingTempest.tsx
```diff
@@ -32,6 +83,17 @@ export class SurgingTempest extends Analyser {
 				}),
 			],
 		}))
+		this.suggestions.add(new TieredSuggestion({
+			content: <Trans id="war.stormseye.suggestions.overwrite.content">
+			Avoid using <DataLink action="STORMS_EYE" /> too early, as it does less damage and generates less beast gauge than <DataLink action="STORMS_PATH" />
```
ayyaruq: It doesn't do less damage, they're both 400 potency. Use proper sentences and be consistent with case tho.
```suggestion
			Avoid using <DataLink action="STORMS_EYE"/> more than {EYE_BUFFER / 1000} seconds before it expires, as it generates less Beast Gauge than <DataLink action="STORMS_PATH"/> which can cost you uses of your gauge consumers.
```

ghost:
I have zero patience to deal with lint on WHM stuff.
Closing.

ghost:
Depends on #1688

ghost:
I'm out of my depth on that nonsense, gonna leave it as is for now, and thank the stars for us squashing commits

src/parser/jobs/war/modules/SurgingTempest.tsx
```diff
@@ -1,25 +1,76 @@
...
+const SURGING_TEMPEST_GENERATORS: ActionKey[] = [
+	'MYTHRIL_TEMPEST',
+	'STORMS_EYE',
+	'INNER_RELEASE',
+]
+const SURGING_TEMPEST_GENERATION_AMOUNT = 30
+// Surging Tempest Extension
+const SURGING_TEMPEST_EXTENSION_AMOUNT = 10
+
+const TEMPEST_MAX = 60000
```
ayyaruq: I'd like these to be more consistent, and Tempest is easily confused with Mythril Tempest.
```suggestion
const SURGE_MAX = 60000
```

src/parser/jobs/war/modules/SurgingTempest.tsx
```diff
@@ -32,6 +83,17 @@ export class SurgingTempest extends Analyser {
 				}),
 			],
 		}))
+		this.suggestions.add(new TieredSuggestion({
+			content: <Trans id="war.stormseye.suggestions.overwrite.content">
```
ayyaruq: This key doesn't match the module handle or what the string is being used for.
```suggestion
			content: <Trans id="war.surgingtempest.suggestions.early.content">
```

src/parser/jobs/war/modules/SurgingTempest.tsx
```diff
@@ -32,6 +83,17 @@ export class SurgingTempest extends Analyser {
...
+			Avoid using <DataLink action="STORMS_EYE" /> too early, as it does less damage and generates less beast gauge than <DataLink action="STORMS_PATH" />
+			</Trans>,
+			why: <Trans id="war.stormseye.suggestions.overwrite.why">
```
ayyaruq: Same deal here:
```suggestion
			why: <Trans id="war.surgingtempest.suggestions.early.why">
```

src/parser/jobs/war/modules/SurgingTempest.tsx
```diff
@@ -1,25 +1,76 @@
...
+			this.extendSurgingTempest
+		)
 		this.addEventHook('complete', this.onComplete)
 	}
 
+	private surgingTempest = this.add(new TimerGauge({
+		maximum: TEMPEST_MAX,
+	}))
+
+	private earlyRefreshCount = 0
```
ayyaruq: We really only care about early Storm's Eye, so let's be more explicit.
```suggestion
	private earlyEyes = 0
```

ayyaruq (changes requested):
Kind of a lot of feedback, sorry. I'm low key happy to merge this and make the changes myself other than the i18n ones, but I wrote most of the suggestions as actual code changes you can just hit commit on most of if you like them.

The main thing I see is that things are kinda ordered haphazardly, with arbitrary whitespace between them, particularly consts and a bunch of vars being defined *after* the init function. Additionally, I'd really like to see a Map here over an ActionKey array, as it makes the variable gauge changes easier to manage IMO.

src/parser/jobs/war/modules/SurgingTempest.tsx
ghost: made change in 51df176

ghost:
RE: code cleanup, I took cues from both our conversation, and the existing code, especially inside WAR/Gauge.tsx

src/parser/jobs/war/modules/SurgingTempest.tsx
```diff
@@ -1,25 +1,93 @@
...
+			this.onSurge
+		)
+
 		this.addEventHook('complete', this.onComplete)
 	}
 
+	private onSurge(event: Events['action' | 'combo']) {
+		const modifier = this.surgeModifiers.get(event.action)
+		let isGenerator = true
+		let penalizeEarlyRefresh = true
```
ayyaruq: IDK why you need this.

src/parser/jobs/war/modules/SurgingTempest.tsx
```diff
@@ -1,25 +1,93 @@
...
+
+		switch (event.action) {
+		case this.data.actions.INNER_RELEASE.id:
+			isGenerator = false
+
+		// eslint-disable-next-line no-fallthrough
+		case this.data.actions.MYTHRIL_TEMPEST.id:
+			penalizeEarlyRefresh = false
+
```
ayyaruq: This can go entirely.

src/parser/jobs/war/modules/SurgingTempest.tsx
```diff
@@ -1,25 +1,93 @@
...
+		// eslint-disable-next-line no-fallthrough
+		case this.data.actions.STORMS_EYE.id:
+			if (penalizeEarlyRefresh && this.surgingTempest.remaining > EYE_BUFFER) {
```
ayyaruq: Just check the second part of this here.

src/parser/jobs/war/modules/SurgingTempest.tsx
```diff
@@ -1,25 +1,93 @@
...
+				this.earlyEyes++
+			}
+
+			if (modifier != null) {
+				const amount = modifier[event.type] ?? 0
+				this.surgingTempest.extend(amount, isGenerator)
+			}
+			break
```
ayyaruq: Slap this into the default, since it runs every time and allows you to low key ignore Mythril Tempest and meme fallthrough. I'll note that your gauge values are in ms but your changes are in seconds, so you'll either need to `amount *= 1000` or change the values in the map to ms. I'd prob do the latter.

ayyaruq (changes requested):
Gotta fix the scale of those numbers, otherwise looks good. Small note about that switch statement to simplify it too.

PR #1687, opened 2022-04-25
Add overcap tracking to core timer gauge

ayyaruq:
Kinda niche, useful for things like getting wasted time for statuses that are memed by the timer gauge, and possibly DRK if anyone decides we care about Darkside overcap outside of burst windows but there are other better ways to track that.


ayyaruq:
Not sure this is as useful as I thought it was, especially without a cause for the overcap. Gonna kill the PR for now since it's trivial to redo.

PR #1688, opened 2022-04-26
Fix magic number in WHM lily code

ghost:
spinning code fix out from #1686


ayyaruq (commented):
More of a style nit, enable/disable outside the array braces and lose the extra newlines. Let's be consistent with our code style.

ayyaruq (approved):

PR #1690, opened 2022-04-28
fix: MP Gauge calculations

AzariahAeroborn:
The MP gauge needed changes similar to the ones applied for the Blood gauge, so that it handles the stacks on Blood Weapon correctly.  In reviewing that, found that there were also two pre-existing bugs for the MP gauge that kind of canceled each other out: Delirium wasn't being handled correctly so was counting too much MP generation (same issue as Blood Weapon due to changing to a stack based buff), and the generation from successful combos (Syphon Strike & Stalwart Soul) was not ever being called due to a missing hook.

While working on this fix, determined that doing a null-check to prevent creating duplicate hooks on status apply was a preferred approach to hooking based on the action, changed the Blood Gauge approach to match.

### DIFF
--- a/src/parser/jobs/drk/modules/MPUsage.tsx
+++ b/src/parser/jobs/drk/modules/MPUsage.tsx
@@ -100,7 +100,8 @@ export class MPUsage extends Analyser {
 	private onApplyBloodWeapon() {
 		if (this.activeBloodWeaponHook == null) {
 			// Buffs with stacks generate separate apply events for each stack with a single remove at the end.
-			// Make sure we only start hooking for actions effected by Blood Weapon on the first apply event -- no duplicate hooks on the "reapply" events as the stacks go down
+			// Make sure we only start hooking for actions effected by Blood Weapon on the first apply event
+			// ie no duplicate hooks on the "reapply" events as the stacks go down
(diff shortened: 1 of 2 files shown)


src/parser/jobs/drk/modules/MPUsage.tsx
```diff
@@ -108,7 +113,11 @@ export class MPUsage extends Analyser {
 
 	private activeDeliriumHook?: EventHook<Events['damage']>
 	private onApplyDelirium() {
-		this.activeDeliriumHook = this.addEventHook(filter<Event>().source(this.parser.actor.id).type('damage'), this.onGenerator(this.deliriumGenerators))
+		if (this.activeDeliriumHook == null) {
+			// Buffs with stacks generate separate apply events for each stack with a single remove at the end.
+			// Make sure we only start hooking for actions effected by Blood Weapon on the first apply event -- no duplicate hooks on the "reapply" events as the stacks go down
```
ayyaruq: This is Delirium tho, might not need such a verbose comment but.

ayyaruq (changes requested):
Borked comment, otherwise looks good to me.

ayyaruq (approved):
YOLO fixed it myself since just a comment

PR #1693, opened 2022-04-29
Variable Riddle of Wind expected count

ayyaruq:
The 16th hit is largely up to RNG on timing RoW, or otherwise cheating for consistency. We previously ignored it for this reason. The change here is to just increment expected when we *do* get max windows, to avoid situations where the stat box shows `91/90` and similar. The main thing here is that the baseline target doesn't change, it just adjusts if needed.

I don't actually like or agree with this solution, but it's been logged in #fb-monk so here we are. Expected should be based on 15 only, and who cares if you do more. Having variable expectations is unintuitive to users. Up for review and feedback if anyone has any tho.


skimmedsquare:
Ah, I see why this is weird now - maybe it should be a UI change instead of fudging numbers? Is it possible to do something like a table view instead of a fraction format?

AkairyuGestalter (approved):

PR #1697, opened 2022-05-04
Fixed an issue that prevented synthing a pre-cast Searing Light.

KelosPelmand:
Addresses part of https://discord.com/channels/441414116914233364/725971074122121246/970095119472689183

The pre-cast Bahamut is harder to account for since it isn't status based, so that half of this feedback will be addressed separately when I figure out where/how to handle it.

### DIFF
--- a/src/data/ACTIONS/layers/patch6.1.ts
+++ b/src/data/ACTIONS/layers/patch6.1.ts
@@ -56,5 +56,9 @@ export const patch610: Layer<ActionRoot> = {
 			icon: 'https://xivapi.com/i/002000/002649.png',
 			cooldown: 1000,
 		},
+
+		PET_SEARING_LIGHT: {
+			statusesApplied: [],
+		},
 	},
 }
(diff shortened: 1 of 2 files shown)


src/data/ACTIONS/layers/patch6.1.ts
```diff
@@ -56,5 +56,9 @@ export const patch610: Layer<ActionRoot> = {
 			icon: 'https://xivapi.com/i/002000/002649.png',
 			cooldown: 1000,
 		},
+
+		PET_SEARING_LIGHT: {
+			statusesApplied: [],
+		},
```
ayyaruq: Please keep changes for a single job together, this has WHM stuff in between 2 chunks of SMN changes.

ayyaruq (changes requested):

src/data/ACTIONS/layers/patch6.1.ts
KelosPelmand: hmm, not sure why I put it there. moved.

ayyaruq (approved):

PR #1698, opened 2022-05-07
Add some stuff for 6.1 support

Dot-Inni:
Adds Afflatus Heals to the pie chart to show they're not part of the overheal calculations due to them being used for mana management now. Also removed Plenary from defensive cooldowns.
Marks support for 6.1 since these changes bring it up to par while keeping the lily heals seperate enough.


src/parser/jobs/whm/modules/Overheal.tsx
```diff
@@ -16,6 +16,7 @@ export class Overheal extends CoreOverheal {
 				this.data.statuses.REGEN.id,
 				this.data.actions.MEDICA_II.id,
```
psummers678, line 17: Just noticed - Can you add the status for Medica_II aswell?

psummers678 (changes requested):

src/parser/jobs/whm/modules/Overheal.tsx
Dot-Inni: Fixed, guess I missed that oops~

src/parser/jobs/whm/index.tsx
```diff
@@ -74,5 +74,10 @@ export const WHITE_MAGE = new Meta({
 			Changes: () => <>Added Action Timeline structure.</>,
 			contributors: [CONTRIBUTORS.INNI],
 		},
+		{
+			date: new Date('2022-05-07'),
+			Changes: () => <>Added Afflatus Heals to the Overheal pie chart, updated Defensive Cooldowns. 6.1 now supported.</>,
```
ayyaruq: We don't bother with these for general patches and only include a changelog for the actual changes made. If there's no job changes, no need for a changelog entry. I'd prob remove or reword the entries for 6.05/6.08 and remove the note about 6.1 here. The idea is to document what changed and only what changed.

src/parser/jobs/whm/modules/Overheal.tsx
```diff
@@ -29,6 +31,15 @@ export class Overheal extends CoreOverheal {
 			trackedHealIds: [
 				this.data.actions.LITURGY_OF_THE_BELL_ON_DAMAGE.id,
 				this.data.actions.LITURGY_OF_THE_BELL_ON_EXPIRY.id,
+				this.data.actions.LITURGY_OF_THE_BELL_ACTIVATION.id,
+			],
+		},
+		{
+			name: <Trans id="whm.overheal.afflatus.name">Afflatus Healing</Trans>,
+			color: SuggestedColors[4],
```
ayyaruq: Out of scope for this PR but this is not very safe. Core overheal should use an object or map and have actual keys for colours rather than relying on an opaque array index, or this should be entirely up to consuming jobs. IMO set the colour directly for now.

ayyaruq (changes requested):
Requesting to fix up the changelog. The rest is fine, but core overheal desperately needs some refactoring of those colours. Could be a good future PR.

src/parser/jobs/whm/index.tsx
Dot-Inni: Removed the patch update entries.

src/parser/jobs/whm/modules/Overheal.tsx
Dot-Inni: I put the actual hex colors that were obscured by the array prior, not sure if that's better, but at least they're directly visible now?

ayyaruq (approved):

ayyaruq:
@Dot-Inni can you sort out this conflict or set the PR to allow maintainers to edit so we can do it for you? The latter is a good idea for all PRs in general. Since it's Lingui, you can probably just merge `endwalker` and re-run `yarn extract`.

Dot-Inni:
Merge conflict should be resolved now.

PR #1699, opened 2022-05-08
Add talkback

skimmedsquare:
Wanted to speed up local development and not mess around with pulling down the server and running it locally. This adds a dev dependency on [talkback](https://github.com/ijpiantanida/talkback) which stores requests made to the proxy server as local `JSON5` files.

After being saved, a ~10 minute pull (P3S) takes up ~25MB of disk space and takes ~150ms to return from the local proxy as compared to ~6s from the production site.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ yalc.lock
 
 # testing
 /coverage
+/tapes
 
 # production
 /build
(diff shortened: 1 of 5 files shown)


suShirow (commented):
I'd previously saved down fflogs files for dummy data I used to test with, but never bothered to set up a tool. This is really handy and I don't see any downside having this.

PR #1706, opened 2022-06-23
Overheal defects

suShirow:
Add order system for statistics; Fix translation support on coreoverheal


src/parser/core/modules/Overheal.tsx
```diff
@@ -296,12 +296,13 @@ export class Overheal extends Analyser {
...
-					'Type of heal',
-					'% of total overheal',
-					'Overheal % per type',
+					<Trans id="core.overheal.header.type" key="core.overheal.header.type">Type of heal</Trans>,
+					<Trans id="core.overheal.header.percenttotal" key="core.overheal.header.percenttotal">% of total overheal</Trans>,
+					<Trans id="core.overheal.header.percenttype" key="core.overheal.header.percenttype">Overheal % per type</Trans>,
 				],
 				data: data,
 				width: 3, // chart's wide, yo
+				order: 10, // Place it after GCD stat panel and other statistics
```
ayyaruq: I feel like this would be better in `src/parser/core/modules/DISPLAY_ORDER.ts` but that's more for total layout than checklist/stats type stuff so hm. Maybe something to address in the future. Maybe for now allow jobs to override this prop somehow, like how we set display order at the module level.

src/parser/core/modules/Statistics/AbstractStatistic.tsx
```diff
@@ -5,6 +5,7 @@ export interface AbstractStatisticOptions {
 	info?: React.ReactNode
 	width?: ColumnSpan
 	height?: number
+	order?: number
```
ayyaruq: Prefer `displayOrder` for consistency with other modules.

ayyaruq (dismissed):
Not so sure on the interface and how this would be configured by consuming modules, requesting changes just for consistency with other modules tho.

src/parser/core/modules/Statistics/AbstractStatistic.tsx
suShirow: > Prefer `displayOrder` for consistency with other modules.

Can't comply.. these order values are a separate value. Statistic has it's own internal order and that name will conflict with the Statistic's panel's position on the page. I left the module-facing name to be displayOrder though, and just converted it to "statsDisplayOrder" somewhere in the middle and hopefully that won't cause confusion in the future.

suShirow:
image
it werks

src/parser/core/modules/Statistics/AbstractStatistic.tsx
AzariahAeroborn: Would using statsPanelOrder consistently through the stack be a reasonable way to disambiguate this?  I'm not sure I want to block further on a naming debate, but wanted to raise a quick discussion to see if we can get a quick consensus here before I throw another one onto the tech debt pile.

AkairyuGestalter (approved):

PR #1710, opened 2022-07-20
Add RoW to RoF windows

vmaggioli:
Added the expected RoW press during RoF windows. Addresses the RoW checkbox in #1303 

Only thing I can think of that this doesn't cover is maybe instead of showing red when missing RoW just showing yellow since it won't be up for every RoF window.

Proof:
image


Toastdeib:
Not really a MNK guy but I have two general bits of feedback:

1) You'll need to include a changelog entry for MNK (and if this is your first contribution, that means you'll also need to add yourself to the contributors file).
2) Since RoW is 90s and should only be up for every 3rd RoF (or 2nd if you hold it, I don't know the rotation), you could solve the green/red problem by handling it the same way as Blitzes are done in the table, with `1/-` and `0/-` as an informational thing instead of having an expected use count.

vmaggioli:
New proof
RoW no expectation

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -47,6 +48,7 @@ export class RiddleOfFire extends BuffWindow {
 	@dependency globalCooldown!: GlobalCooldown
 
 	private blitzActions = fillActions(BLITZ_ACTIONS, this.data)
+	private riddleActions = fillActions(['RIDDLE_OF_WIND'], this.data)
```
ayyaruq, line 51: If it's a single action and will only ever be a single action, why does this need to fill actions?

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
```diff
@@ -60,6 +62,10 @@ export class RiddleOfFire extends BuffWindow {
 			blitzActions: this.blitzActions,
 		}))
 
+		this.addEvaluator(new RiddleOfWindEvaluator({
+			riddleActions: this.riddleActions,
+		}))
```
ayyaruq, line 67: Could this not just use the expected actions evaluator and have an adjust callback to handle expected=0 for the windows we don't get RoW off cooldown? I'm not following why this needs a custom evaluator, especially since it doesn't handle the windows where we know there can't be a RoW.

ayyaruq (commented):
Thanks for your PR. Good to get this feature in, but I'm not sure about the implementation. Can you explain why it needs a custom evaluator? I'm fine with them where they make sense but it's a bit odd to have one that doesn't handle half the windows when the more generic option in core can do it just fine. Maybe take a look at the core expected actions evaluator.

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
vmaggioli: No reason, I'll get rid of it

src/parser/jobs/mnk/modules/RiddleOfFire.tsx
vmaggioli: Yeah I think that's a much better option. I've made the change to a expected actions evaluator locally, I'm just curious how we can find out whether or not it's on cool down when in the RoF window. I know `adjustCount` method gives us the events in that window, but I'm not sure if that's helpful for checking whether or not it's on cooldown

AkairyuGestalter (approved):
Good enough for a first pass. Default ExpectedActionsEvaluator doesn't allow skipping the suggestion, and this is currently only informational so it's fine as a custom thing for now. We can update core to allow skipping the suggestion at a later date, and deal with calculating whether RoW should have fit into a given window when that happens.

PR #1711, opened 2022-07-21
Allow unavoidable overhealing to be excluded 

psummers678:
## Exclude Ids from overhealing

Allow Core overhealing modules to exclude a list of spells based on a job by job basis and followed method of implementing protected -> private from the tracked heals.

Would like assistance in verification/jest testing 

Discord Server Name: KerrisFang


nonowazu:
I feel like this might be able to be closed - I wasn't looking at this closely when I wrote PR #1734, which should include your "ignoring" feature.

AzariahAeroborn:
Closing this in favor of #1734 - it looks like this might need to be resubmitted to add the ignore flag to Assize to match up with the interface changes in #1734, but otherwise the other changes in this PR are duplicated.  Will flag this in #dev-healer for consideration.

Issue #596, opened 2019-08-25
Add better support for Lily tracking


kkragenbrink:
- [ ] Track when Cure 2 or Medica is used while an Afflatus heal is available and ding it.
- [ ] Track when Cure 3 is used while an Afflatus heal is available and warn (Cure 3 might be okay in some situations).

SaltedyLevity:
First half looks great!

Second half is a general recommendation for players but not an "absolutely must do" so it could just be a wording adjustment? For example, ones that do effective healing only, I would think aren't something to flag as "done wrong" if that makes sense? But more of a "maybe you could've done these things with it, but maybe you couldn't have?"
I know that's super wishy-washy sounding, this one is pretty subjective tbh, i'm just wary of players taking away advice that makes them want to sit on the heals.

Other question, how would you handle what qualifies as effective healing? Just if the skill did like no more than 25% overheal?

kkragenbrink:
I haven't really thought through the idea of effective healing, yet.  Really, though, if used as a weaving window for Asylum when Dia isn't available, you can actually get a free Afflatus Misery every 3m, and it's a DPS gain over clipping with Asylum, or clipping the Dia, so I'm really at a loss for how to track this most effectively.  Trying to figure that out still.  The first half is the most important part, tI think.  Maybe I should break the second half out into a separate task?

SaltedyLevity:
Can you elaborate a bit on what you mean? I'm a bit confused how you're getting that it's a dps gain/free misery. I know it's a different example, but the usual one I use to show if/when afflatus is a gain to weave with over clipping is assize:
In a dummy fight, using one afflatus for the alternating assizes that don't line up with dia, it's a DPS gain to hard clip them UNTIL you get to the 5min15s mark where you can use your first misery to weave assize, then you come out +100 ahead.

So in a more practical fight sense, similar logic applies: you need to have saved yourself more than 3 hard clips with afflatus for it to be a DPS gain - and that assumes you didn't use one to weave and then had to GCD heal sometime later as a result of not having a lily. And if you're only doing one per afflatus, that means you need to weave with Misery. But you might be saving Misery to move, or even for raid buffs that would be more portent than a clip. 

(I sound like i'm regurgitating the stuff i've been writing all weekend, sorry for being so verbose!)

nivthefox:
[...]

PR #1713, opened 2022-07-25
Check for lilies when using GCD heals

vmaggioli:
Addresses #596
Proof below taken from [this fflogs](https://www.fflogs.com/reports/dtFbmkMpLnwgRHxc) for the Ifrit and Garuda kill from the character "Snow Vastal"

image


src/parser/jobs/whm/modules/Lilies.tsx
```diff
@@ -120,6 +133,12 @@ export class Lilies extends CoreGauge {
 
 	}
 
+	private checkAfflatusAvailability() {
+		if (this.lilyGauge.value !== 0) {
```
psummers678: Should add in a check for Blood Lily Gauge so that we dont encourage overcapping

src/parser/jobs/whm/modules/Lilies.tsx
```diff
@@ -32,6 +32,13 @@ const BLOODLILY_CONSUMERS: ActionKey[] = [
 	'AFFLATUS_MISERY',
 ]
 
+// Leaving out CURE3 due to having it only warn instead of error
+const GCD_HEALS: ActionKey[] = [
```
psummers678, line 37: Is the regen on Medica2/Regen enough to not include them? I cant think of many situations where the regen effect would be enough to not call it out?

vmaggioli: I'm on board for Regen but I'm sort of torn on Medica2. I can see heal check phases being started with Medica2 first while lilies are up in order to begin regen first and then using lilies after. However, I'm sure this as a case where if someone's smart enough to order their heals like that then they're smart enough to understand that the tip is selective in its application.

src/parser/jobs/whm/modules/Lilies.tsx
vmaggioli: Good call!

src/parser/jobs/whm/modules/Lilies.tsx
```diff
@@ -120,6 +133,12 @@ export class Lilies extends CoreGauge {
 
 	}
 
+	private checkAfflatusAvailability() {
+		if (this.lilyGauge.value !== 0 && this.bloodLilyGauge.value === 0) {
```
psummers678: We should be checking if the blood lily gauge is not capped rather than if its empty. We are fine to use a lily if we are at 2 blood lily for example.
The gauge is set up to cap at 3 and add 1 every normal lily usage.

src/parser/jobs/whm/modules/Lilies.tsx
psummers678: Had a talk with people in the discord dev channel and its an awkward one. If we keep it to Cure Cure II and Medica for now

src/parser/jobs/whm/modules/Lilies.tsx
vmaggioli: Yeah I'm just dumb and forgot that it takes three charges to create a blood lily 😅 it's been a minute

src/parser/jobs/whm/modules/Lilies.tsx
```diff
@@ -120,6 +133,12 @@ export class Lilies extends CoreGauge {
 
 	}
 
+	private checkAfflatusAvailability() {
+		if (this.lilyGauge.value !== 0 && !this.bloodLilyGauge.capped) {
```
ayyaruq: Not sure about this function name, and it could be inlined with the hook anyway, but there's a getter to check when the gauge is empty instead of a specific value.

```suggestion
		if (!this.lilyGauge.empty && !this.bloodLilyGauge.capped) {
```

src/parser/jobs/whm/modules/Lilies.tsx
```diff
@@ -167,6 +186,17 @@ export class Lilies extends CoreGauge {
 			</Trans>,
 		}))
 
+		this.suggestions.add(new TieredSuggestion({
+			icon: this.data.actions.AFFLATUS_RAPTURE.icon,
+			content: <Trans id="whm.gauge.gcdheals.suggestions.leftover.content">
+        Try to use lilies instead of GCD heals when they're available.
```
ayyaruq:
```suggestion
        Try to use Lilies instead of GCD heals when they're available.
```

src/parser/jobs/whm/modules/Lilies.tsx
```diff
@@ -167,6 +186,17 @@ export class Lilies extends CoreGauge {
 			</Trans>,
 		}))
 
+		this.suggestions.add(new TieredSuggestion({
+			icon: this.data.actions.AFFLATUS_RAPTURE.icon,
+			content: <Trans id="whm.gauge.gcdheals.suggestions.leftover.content">
```
ayyaruq:
```suggestion
			content: <Trans id="whm.gauge.lily.suggestions.leftover.content">
```

src/parser/jobs/whm/modules/Lilies.tsx
```diff
@@ -167,6 +186,17 @@ export class Lilies extends CoreGauge {
...
+        Try to use lilies instead of GCD heals when they're available.
+			</Trans>,
+			tiers: SEVERITIES.UNNECESSARY_GCD_HEALS,
+			value: this.numUnnecessaryGcdHeals,
+			why: <Trans id="whm.gauge.gcdheals.suggestions.leftover.why">
```
ayyaruq:
```suggestion
			why: <Trans id="whm.gauge.lily.suggestions.leftover.why">
```

src/parser/jobs/whm/modules/Lilies.tsx
```diff
@@ -57,6 +67,8 @@ export class Lilies extends CoreGauge {
 
 	@dependency private suggestions!: Suggestions
 
+  private numUnnecessaryGcdHeals = 0;
```
ayyaruq: Prefer to name these by what they do instead of a verbose abbreviation. Something like `wastedGcds` is more than clear enough. Would need to update uses of the var too.

src/parser/jobs/whm/modules/Lilies.tsx
```diff
@@ -49,6 +56,9 @@ const SEVERITIES = {
 		1: SEVERITY.MEDIUM,
 		3: SEVERITY.MAJOR,
 	},
+	UNNECESSARY_GCD_HEALS: {
```
ayyaruq:
```suggestion
	WASTED_GCD_HEALS: {
```

ayyaruq (dismissed):
Thanks for your PR. Logic is mostly good aside from that gauge comparison.

A bit of cleanup to do for consistency with the rest of the WHM code and other jobs in xiva, I've done them as suggestions so you can more easily apply my requested changes, just click the commit button on the comment.

Additionally, could you please resize your avatar to 100x100 as that's the size we use in this project. You'll probably need to rerun `yarn extract` too.

src/parser/jobs/whm/modules/Lilies.tsx
```diff
@@ -120,6 +133,12 @@ export class Lilies extends CoreGauge {
 
 	}
 
+	private checkAfflatusAvailability() {
+		if (!this.lilyGauge.empty && !this.bloodLilyGauge.capped) {
```
AzariahAeroborn: This should also exclude hard casts when the player has the Thin Air status active - if Thin Air is active then there's no MP difference between the hard cast and the lily skill, and we ding elsewhere if someone is overcapping lilies.

See [conversation on Discord](https://discord.com/channels/441414116914233364/470049935077933076/1109949854223966289) for more info.

AzariahAeroborn (dismissed):

AkairyuGestalter (approved):

PR #1714, opened 2022-07-28
Updated sentences 

Jep4:
Revision 1 

- `core`: Unsurprisingly, the core system modules. These provide commonly-used functionality (see the section on dependency below), as well as job-agnostic modules such as "Don't die".
- `jobs/[job]`: Each supported job has its own group of modules, that provide specialised analysis/information for that job.
- `bosses/[boss]`: Like jobs, some bosses have groups of modules, usually used to analyse unique fight mechanics, or provide concrete implementations that fflogs does not currently provide itself.

Changes 1

- `core`:These provide commonly-used functionality (see the section on dependency below), as well as job-agnostic modules such as "Don't die".
- `jobs/[job]`: Each supported job has its own group of modules. These provide specialised analysis/information for that job.
- `bosses/[boss]`: Some bosses have groups of modules. They usually used to analyse unique fight mechanics, or provide concrete implementations that fflogs does not provide.

URL: https://github.com/Jep4/xivanalysis/blob/endwalker/README.md 

Revision 2
## Adding encounter overrides

While we generally avoid adding per-boss handling, as it generates increased maintenance burden for the site as a whole, some modules such as `invulnerability` can't account for all possible variations, and occasionally need to be configured.

Changes 2

Boss handing generates increased maintenance burnen for the site. Some modules such as `invulnerability` can't account for all possible variations, and occasionally need to be configured.

URL: https://github.com/Jep4/xivanalysis/blob/endwalker/docs/patch-checklist.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,9 +41,9 @@ The parser is the meat of xivanalysis. Its primary job is to orchestrate modules
 
 The modules are split into a number of groups:
 
-- `core`: Unsurprisingly, the core system modules. These provide commonly-used functionality (see the section on dependency below), as well as job-agnostic modules such as "Don't die".
-- `jobs/[job]`: Each supported job has its own group of modules, that provide specialised analysis/information for that job.
-- `bosses/[boss]`: Like jobs, some bosses have groups of modules, usually used to analyse unique fight mechanics, or provide concrete implementations that fflogs does not currently provide itself.
+- `core`:These provide commonly-used functionality (see the section on dependency below), as well as job-agnostic modules such as "Don't die".
+- `jobs/[job]`: Each supported job has its own group of modules. These provide specialised analysis/information for that job.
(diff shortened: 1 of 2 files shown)


ayyaruq:
Hi, thanks for your PR. What is the purpose of this update? Most of these changes read better as they were without the change.

As an aside, the description for a PR should be about why the change is proposed, and any other details. We can see the diff so there's no need to repeat it in the description. PR titles should also summarise the change rather than being a comment on it.

Jep4:
Hi.

Sorry for the situation.
It was a mistake while testing how to use a pull request.
ᐧ

ayyaruq:
> Hi. Sorry for the situation. It was a mistake while testing how to use a pull request.

No problem. Could you update the title/description for context on why this PR is needed?

PR #1724, opened 2022-08-29
[AST] 6.2 bump, arcana tracking update, and redraw suggestions

otocephaly:
Hiiii reviewer~~~

Some notes for this PR
- Crown play still works as intended even though the button itself has been combined in the UI. No changes were necessary
- Redraw still functions within AST and no changes were necessary here either
- Arcana tracking now includes technically optimal targets (targets who would receive 6% would be technically optimal and are shown with a green checkmark). See attachment 1
- Arcana tracking now includes redraw effectiveness both suggestions and a display on the table. Redraw effectiveness includes:
   - Redraws performed when not necessary (already has a unique seal)
   - Redraws not performed when AST would benefit (held non-unique seal and didn't redraw when available)
   - See attachment 2 for redraw effectiveness on arcana tracking examples
   - See attachment 4 for examples on redraw suggestions
   - Note: the redraw effectiveness can be applied to 6.0 onwards since the only change to redraw is that a different seal is guaranteed from the one you have so no case management has been done in that respect.

Attachment 1: technically optimal target
image

Attachment 2: Redraw effectiveness arcana tracking examples
image
image

Attachment 3: Redraw effectiveness suggestions examples
image
image

Source logs used above:
https://www.fflogs.com/reports/a:9fDvcZdQYTbtm7hN#fight=4&source=80
https://www.fflogs.com/reports/a:V2LGRvdbcZ3tynrW#fight=10&source=8


otocephaly:
Splitting out instead. Closing PR for ease

PR #1729, opened 2022-09-09
Bugfix: DNC missed steps

AkairyuGestalter:
Missed a 'not' when checking whether a dance should be marked dirty based on whether the boss was invulnerable when the dance _should_ have finished if they'd done all the expected steps.


nonowazu (approved):
Not that my review counts for anything, but if I'm stepping back to maintain stuff I need to review PRs.

That being said, it looks good to me – simple negation so not a lot to think about.

suShirow (approved):
Thanks @nonowazu ! lgtm2

PR #1730, opened 2022-09-11
[GNB] 6.2 Support

Rjb99:
This applies the change to  GNB's Bloodfest in 6.2 90s -> 120s
The Opener offsets were changed to support 1NM23, which is like 123NM, except it's safe for getting all skills in raid buffs and overcaps a minute early compared to 123NM.

I've removed out of date guides and updating the comments on Bloodfest evaluator, BF = 3 Burst Strikes now after the opener.
I will re-add info graphs if/when GNB graphic artists remake them for 6.2.

~~this job is a nightmare trying to find non-video resources for. resources were updated for 6.1 a day before 6.2 hit~~


nonowazu (approved):
The logic changes all look sound - and the comments help greatly for understanding the changes.

AkairyuGestalter (approved):

PR #1731, opened 2022-09-14
Fixed cooldowns for three role actions

Toastdeib:
Small fixes for some incorrect data that got brought up in #fb-general. Didn't bother to include a changelog on this since it should only impact the timeline display and not actual analysis, but I can if needed.


nonowazu (approved):
Assuming these were all correct at the start of the expac, this is a simple approval. I know rescue is correct, at least.

AkairyuGestalter (approved):

PR #1733, opened 2022-09-21
[DRG] Fix issue with solo Dragon Sight suggestion appearing all the time

ebshimizu:
The suggestion to use Dragon Sight on a partner was showing up constantly due to SE changing the IDs used for Right Eye, presumably after ~~patch 6.08 when they took the visual away from us~~ they fixed the bug where dragon sight goes away when your partner dies in the 6.21 hotfix. This PR fixes it so we're not checking for an ID, but instead are checking to see if we used DS on ourselves and displaying the suggestion accordingly. This works back through patch 6.0.

### DIFF
--- a/src/parser/jobs/drg/modules/Buffs.tsx
+++ b/src/parser/jobs/drg/modules/Buffs.tsx
@@ -1,6 +1,6 @@
 import {t} from '@lingui/macro'
 import {Trans, Plural} from '@lingui/react'
-import {ActionLink, StatusLink} from 'components/ui/DbLink'
+import {DataLink} from 'components/ui/DbLink'
 import {ActionKey} from 'data/ACTIONS'
 import {Cause, Event, Events} from 'event'
 import {Analyser} from 'parser/core/Analyser'
@@ -12,7 +12,7 @@ import {Data} from 'parser/core/modules/Data'
 import {Invulnerability} from 'parser/core/modules/Invulnerability'
 import {PieChartStatistic, Statistics} from 'parser/core/modules/Statistics'
 import {Statuses} from 'parser/core/modules/Statuses'
-import Suggestions, {Suggestion, TieredSuggestion, SEVERITY} from 'parser/core/modules/Suggestions'
+import Suggestions, {TieredSuggestion, SEVERITY} from 'parser/core/modules/Suggestions'
 import React from 'react'
 import DISPLAY_ORDER from './DISPLAY_ORDER'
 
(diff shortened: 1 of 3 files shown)


src/parser/jobs/drg/modules/Buffs.tsx
```diff
@@ -112,6 +114,13 @@ export default class Buffs extends Analyser {
 		}
 	}
 
+	private onDs(event: Events['action']) {
```
nonowazu: Minor nit, might as well use the full `onDragonSight` name for the method since the shorthand just makes it confusing for reviewers. I can't block on it, but something to consider.

nonowazu (approved):
Logic looks fine, only feedback is the function name for people that don't read the drg modules often (like me!)

src/parser/jobs/drg/modules/Buffs.tsx
ebshimizu: makes sense to me, easy fix :)

src/parser/jobs/drg/modules/Buffs.tsx
```diff
@@ -77,7 +77,9 @@ export default class Buffs extends Analyser {
 	override initialise() {
 		const playerFilter = filter<Event>().source(this.parser.actor.id)
 		this.addEventHook(playerFilter.type('action'), this.onCast)
-		this.addEventHook(playerFilter.type('statusApply').status(this.data.statuses.RIGHT_EYE_SOLO.id), () => this.soloDragonSight = true)
+
+		this.addEventHook(filter<Event>().source(this.parser.actor.id).action(this.data.actions.DRAGON_SIGHT.id), this.onDragonSight)
```
Toastdeib: Minor nit, but is there any particular reason not to reuse `playerFilter` here? It looks to be the same base filter as the other hooks.

src/parser/jobs/drg/modules/Buffs.tsx
```diff
@@ -162,15 +171,15 @@ export default class Buffs extends Analyser {
 			</Trans>,
 		}))
 
-		if (this.soloDragonSight) {
+		if (this.soloDragonSightCount > 0) {
```
ayyaruq: Wouldn't this be easier with tiered suggestions? No need for the conditional or indenting that way, and you get more flexibility if there are conditions where solo eyes aren't just minor, just make sure to clean up unused imports.

ayyaruq (commented):
Minor nit for easier to read code. If you think it's not necessary, I'll just merge, but imo it's cleaner to not need as much indentation and more flexible for future maintenance with my suggestion.

ayyaruq (approved):

PR #1122, opened 2021-04-04
As-yet unused overheal addition to manually bucket heals

redshadowhero:
As always – open to suggestions.

This adds the ability to manually bucket heals by an arbitrary numerical id. I created a test with WAR to bucket nascent heals into equilibrium, direct (really just storm's path), glint (heals on someone else), and self heals, and results were promising:

image

As a very brief example:

```js
export default class Overheal extends CoreOverheal {
        // etc
    	trackedHealCategories = [
		{
			id: SELF_HEALS,
			name: <Trans id="war.overheal.self">Nascent heals (self)</Trans>,
			color: SuggestedColors[3],
			trackedHealIds: [],
		},
		{
			name: <Trans id="war.overheal.equilibirum">Nascent heals (other)</Trans>,
			color: SuggestedColors[1],
			trackedHealIds: [
				STATUSES.NASCENT_GLINT.id,
			],
		},
		{
			name: <Trans id="war.overheal.christianbale">That movie with Christian Bale</Trans>,
			color: SuggestedColors[2],
			trackedHealIds: [
				ACTIONS.EQUILIBRIUM.id,
			],
		},
	]

        // etc
        overrideHealCategory(event, petHeal) {
		if (this._flashOn && !IGNORED_HEALS.includes(event.ability.guid)) {
			return SELF_HEALS
		}
		return -1
	}
}
```

PR #1734, opened 2022-09-24
Add bucketing of heals and a nice "ignore" feature

nonowazu:
Attempt 2 of #1122, now with a feature to ignore heals in a bucket. I think a combination of this with manual bucketing, ignoring buckets, and considerHeal will allow flexibility in ignoring heals altogether.

Though, I should be convinced to make `debug = true` show ignored heals somewhere.

### DIFF
--- a/src/parser/core/modules/Overheal.tsx
+++ b/src/parser/core/modules/Overheal.tsx
@@ -15,9 +15,11 @@ interface SeverityTiers {
 }
 
 interface TrackedOverhealOpts {
+	bucketId?: number
 	name: JSX.Element,
 	color?: string
 	trackedHealIds?: number[],
+	ignore?: boolean
 }
 
 const REGENERATION_ID: number = 1302
@@ -44,6 +46,8 @@ export const SuggestedColors: string[] = [
 ]
 
 export class TrackedOverheal {
+	bucketId: number = -1
+	ignore: boolean
 	name: JSX.Element
 	color: string = '#fff'
 	protected trackedHealIds: number[]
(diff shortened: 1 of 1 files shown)


src/parser/core/modules/Overheal.tsx
```diff
@@ -235,6 +251,17 @@ export class Overheal extends Analyser {
 
 		const guid = event.cause.type === 'action' ? event.cause.action : event.cause.status
 		const name = event.cause.type === 'action' ? this.data.getAction(guid)?.name : this.data.getStatus(guid)?.name
+
+		const bucketId = this.overrideHealBucket(event, petHeal)
+		if (bucketId >= 0) {
+			for (const trackedHeal of this.trackedOverheals) {
+				if (trackedHeal.bucketId === bucketId) {
+					this.debug(`Heal ${name} (${guid}) at ${event.timestamp} MANUALLY shoved into bucket ${trackedHeal.name.props.default}`)
```
AkairyuGestalter: small nit: it should be ${trackedHeal.name.props.defaults} (see the normal 'matched into category' debug string)

AkairyuGestalter (changes requested):
one minor nit, otherwise it does what it says on the tin. I don't think it's necessary to have the `debug = true` flag override the `ignore` flag and show the bucket, since we have the console debug logging to refer to.

src/parser/core/modules/Overheal.tsx
nonowazu: The worst part about this? I'm the one who wrote the other debug string, so I should know this.

AkairyuGestalter (approved):

PR #1735, opened 2022-09-24
Small bugfixes for NIN

Toastdeib:
Fixes include:

- An actual bug in the Huton module that was flagging _any_ Huton extender as Armor Crush if used when Huton wasn't active
- Bumping the severity of dropped Kassatsu from Medium to Major
- Inaccurate text on the Rabbit Medium suggestion that suggested Trick was still a raid buff
- Weighting the calculation in the Positionals module so every action is treated the same, rather than averaging the percents for each individual action


nonowazu (approved):
While it's a lot of little changes, I didn't see anything glaring.

AkairyuGestalter (approved):

PR #1736, opened 2022-09-26
[DRG] Fix power surge checklist description

ebshimizu:
Checklist description for Power Surge now correctly identifies Sonic Thrust as the aoe action that grants the buff.

### DIFF
--- a/src/parser/jobs/drg/modules/Buffs.tsx
+++ b/src/parser/jobs/drg/modules/Buffs.tsx
@@ -135,7 +135,7 @@ export default class Buffs extends Analyser {
 		this.checklist.add(new Rule({
 			name: <Trans id="drg.buffs.checklist.name">Keep {this.data.statuses.POWER_SURGE.name} up</Trans>,
 			description: <Trans id="drg.buffs.checklist.description">
-				<ActionLink {...this.data.actions.DISEMBOWEL}/> and <ActionLink action="SONIC_THRUST" /> grant <StatusLink status="POWER_SURGE" /> which provides a 10% boost to your personal damage and should always be kept up.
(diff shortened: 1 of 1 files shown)


src/parser/jobs/drg/modules/Buffs.tsx
```diff
@@ -135,7 +135,7 @@ export default class Buffs extends Analyser {
 		this.checklist.add(new Rule({
 			name: <Trans id="drg.buffs.checklist.name">Keep {this.data.statuses.POWER_SURGE.name} up</Trans>,
 			description: <Trans id="drg.buffs.checklist.description">
-				<ActionLink {...this.data.actions.DISEMBOWEL}/> and <ActionLink action="COERTHAN_TORMENT" /> grant <StatusLink status="POWER_SURGE" /> which provides a 10% boost to your personal damage and should always be kept up.
+				<ActionLink {...this.data.actions.DISEMBOWEL}/> and <ActionLink action="SONIC_THRUST" /> grant <StatusLink status="POWER_SURGE" /> which provides a 10% boost to your personal damage and should always be kept up.
```
Toastdeib: For consistency's sake, can you change `<ActionLink {...this.data.actions.DISEMBOWEL}/>` to `<ActionLink action="DISEMBOWEL" />`? Seems kind of silly to have both flavors of `ActionLink` keying in a single string.

ebshimizu: hmm who could've possibly gotten lazy when the new action link format came out (it was me). fixed!

ayyaruq: If you're gonna be using the newer link formats may as well use DataLink since this module uses more than one link type anyway. Softer imports, but nbd really.

ebshimizu:
whoops this touches the same file as #1733 so i'll fix it in that PR instead

PR #1738, opened 2022-09-27
[PLD] Updated pld notes and supported patches

Styrfire:
Add 6.2 as a supported patch for PLD. 6.2 contained no changes outside of dmg for the magic combo.


.vscode/extensions.json
```diff
@@ -7,7 +7,7 @@
 		"dbaeumer.vscode-eslint",
 		"editorconfig.editorconfig",
 		"mikestead.dotenv",
-		"msjsdiag.debugger-for-chrome"
+		"ms-vscode.js-debug"
```
nonowazu: I wouldn't update the extensions without discussing it in the channel first, and I would suggest not packaging it as part of this PR - it may delay the PR being merged for various reasons.

nonowazu (changes requested):
I would also suggest changing/shrinking/compressing the avatar image. It's 1MB in size, and the previous winner goes to Furst who's at a whopping 1/18th that size

```
hosakaii:avatar king$ du -s * | sort -n | tail -n 3
88	Arkevorkhat.png
88	aria.png
112	Furst.png
```

src/parser/jobs/pld/changelog.tsx
```diff
@@ -2,6 +2,13 @@ import CONTRIBUTORS from 'data/CONTRIBUTORS'
 import React from 'react'
 
 export const changelog = [
+	{
+		date: new Date('2022-10-01'),
+		Changes: () => <>
+			Updated supported patches to include 6.2.
+		</>,
+		contributors: [CONTRIBUTORS.STYRFIRE],
+	},
```
nonowazu: Since nothing functional has changed, I don't think we do changelogs on just patch bumps anymore – the fact that it's not greyed out in the selection menu should be notification enough.

nonowazu (changes requested):

nonowazu (approved):
Looks good to me

ayyaruq (changes requested):
Thanks for your PR. Sorry to nitpick your avatar again, this should be better documented but we expect it to be 100x100 in resolution on top of a reasonable file size.

Some Q's tho:
- is the notes file actually used for anything? Most jobs have shifted this stuff into GH issues
- why are the changes for magic combos not updated? I don't see any code using the potencies so they can likely be removed, but if they're present they should be correct and updated for the layer
- are you taking over support of 6.2? If not, it would be worth getting an approval from Aria so we know someone is maintaining it going forward

I'm requesting a fix for the avatar, the rest isn't a big deal but should be sorted out at some point.

nonowazu:
Small nitpick from me - if we're going to remove the notes.md (and I have no problem with that) – can we ensure we have github issues for them?

Styrfire:
As mentioned in the discord, there is only generic class information in the notes. No issues to track in Github.

ayyaruq:
Generic class information maybe, but it details exactly what we'd expect in an issue. Easily converted, unless nothing is pending and it's out of date anyway.

Styrfire:
> Generic class information maybe, but it details exactly what we'd expect in an issue. Easily converted, unless nothing is pending and it's out of date anyway.

Nothing is pending in the notes. I brought it up to date but it was just generic class information that you could find on The Balance. (like 11 gcd's and 2 goring per fof, etc...) It hadn't been modified in 3 years... I doubt anyone is using it, and again, I double checked that we don't need the information.

ayyaruq (approved):

PR #1741, opened 2022-09-29
Potential bugfix for weaves that go on after death

nonowazu:
...zombie jokes aside, [this](https://www.fflogs.com/reports/a:q64tW7PkHy2VXmDN#fight=4) parse has a death where:

* The MCH did a GCD, then two oGCDs (which is correct)
* Died
* Revived and used two *more* oGCDs (which, in theory, is still correct for whatever reason)
* Used a GCD, which triggers the weaving module to ding them for 24 seconds of weaves

Now, I'm no expert on core weaving, but I think this fixes the issue (or at least opens the door to you judging me harshly until I fix the issue).

### DIFF
--- a/src/parser/core/modules/Weaving.tsx
+++ b/src/parser/core/modules/Weaving.tsx
@@ -71,6 +71,7 @@ export class Weaving extends Analyser {
 		this.addEventHook(playerFilter.type('prepare'), this.onBeginCast)
 		this.addEventHook(playerFilter.type('action'), this.onCast)
 		this.addEventHook(filter<Event>().type('complete'), this.onComplete)
+		this.addEventHook(filter<Event>().type('death'), this.clearWeave)
 	}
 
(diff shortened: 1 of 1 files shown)


Toastdeib:
I haven't checked any parses but this looks fine to me. Having death save a possible bad weave and then reset to the pre-pull state makes sense as a fix for the issue, so consider this a tentative 👍 until I get a chance to actually take a peek at a log (although I'm assuming you already did that yourself).

nonowazu:
There was some discussion [here](https://discord.com/channels/441414116914233364/441424599310270464/1025066144316522529), but I think the consensus worked out to "this is okay-ish", but I had a question [later](https://discord.com/channels/441414116914233364/441424599310270464/1025511642575937577) that made me wonder if this is doing the Right Thing™.

AzariahAeroborn (approved):

PR #1743, opened 2022-09-30
SGE Overheal Update

AkairyuGestalter:
- Added action data for Pneuma's heal component, thanks again SE
- Added ignore prop to Addersgall Overheal bucket, since they're not just used for healing
- Updated Overheal buckets now that the Addersgall bucket doesn't display

### DIFF
--- a/src/parser/core/modules/Overheal.tsx
+++ b/src/parser/core/modules/Overheal.tsx
@@ -88,8 +88,8 @@ export class TrackedOverheal {
 	}
 
 	/**
-     * Gets a printable name for the category
-     */
+	 * Gets a printable name for the category
+	 */
 	get debugName(): string {
 		if (typeof this.name === 'string') {
 			return this.name
--- a/src/parser/core/modules/Overheal.tsx
+++ b/src/parser/core/modules/Overheal.tsx
@@ -20,6 +20,7 @@ interface TrackedOverhealOpts {
 	color?: string
 	trackedHealIds?: number[],
 	ignore?: boolean
+	debugName?: string
 }
 
 const REGENERATION_ID: number = 1302
@@ -53,6 +54,7 @@ export class TrackedOverheal {
 	protected trackedHealIds: number[]
 	heal: number = 0
 	overheal: number = 0
+	internalDebugName: string | undefined
 
 	constructor(opts: TrackedOverhealOpts) {
 		this.name = opts.name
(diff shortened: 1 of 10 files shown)


src/parser/jobs/sge/modules/Overheal.tsx
```diff
@@ -9,23 +10,30 @@ export class Overheal extends CoreOverheal {
 	override displaySuggestion = true
 
 	override suggestionIcon = this.data.actions.DIAGNOSIS.icon
+	override overhealName = <Trans id="sge.overheal.direct.name">GCD Heals</Trans>
 
 	override trackedHealCategories = [
 		{
 			name: <Trans id="sge.overheal.addersgall.name">Addersgall</Trans>,
```
nonowazu: Not worth blocking over, but since this doesn't appear anywhere anymore, it might be worth replacing this with plain `<></>` tags instead of Trans tags, but just my 2¢.

nonowazu (approved):

src/parser/jobs/sge/modules/Overheal.tsx
ayyaruq: If it doesn't appear anywhere, why does it still exist?

ayyaruq (approved):
Approved, but I'd like to see a resolution to that translation string @nonowazu raised before merging.

src/parser/jobs/sge/modules/Overheal.tsx
AkairyuGestalter: As we discussed on Discord, I'd left it as-is due to how the core Overheal module's debug output worked, but we figured out a plan to revise that so I could change the bucket name to just be a string.

src/parser/core/modules/Overheal.tsx
```diff
@@ -87,6 +87,37 @@ export class TrackedOverheal {
 		return false
 	}
 
+	/**
+     * Gets a printable name for the category
+     */
```
nonowazu: This looks a little mis-indented.

nonowazu (changes requested):
I'm not sure if there's a better way to do the debugName, to be honest. However, since it's *only* called when `debug = true`, I'd say it's probably not worth nitpicking too much.

src/parser/core/modules/Overheal.tsx
AkairyuGestalter: Sure enough. I feex.

nonowazu (approved):

suShirow (approved):
Nono's given their approval, LGTM

PR #1744, opened 2022-09-30
Bugfix - Improve accuracy of 'expected finish' check

AkairyuGestalter:
Bug was reported that a Quadruple Technical Finish was being flagged as having the wrong number of steps:
https://xivanalysis.com/fflogs/qzJ9Wvrny14xAXmH/1/9

In this case the log was started late, but this could technically (har) apply to any pre-pull Technical Finish. The check that figures out what finish was expected went off the action that opened the window, which is expected to be the Step, but in this case was also Quadruple Technical Finish since it was the first action in the log.

Fixing the getter to just return the actual finisher if it's the Double Standard or Quad Technical, and to determine the proper finisher based on the actual finisher, instead of the assumed starter.

### DIFF
--- a/src/parser/jobs/dnc/changelog.tsx
+++ b/src/parser/jobs/dnc/changelog.tsx
@@ -54,7 +54,7 @@ export const changelog = [
 	},
 	{
 		date: new Date('2022-02-16'),
-		Changes: () => <>Improved accuracy of Esprit gauge generation simulation</>,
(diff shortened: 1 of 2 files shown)


hintxiv:
Confirming that the bugged log works after this change, lgtm 👍

Toastdeib:
Looks good to me as well - didn't personally check a log but the code seems fine, so I'm good to merge it.

Toastdeib (approved):

PR #1747, opened 2022-10-12
[SAM] Enpi Bug fix

Rjb99:
Fix Enpi consuming Meikyo when it don't


nonowazu (approved):
1 line functional change, makes sense.

Toastdeib (approved):
Looks good to me.

PR #1748, opened 2022-10-20
Standardize DoT suggestion wording

hintxiv:
Core DoTs specifically advises against a generic suggestion, so I just updated each module individually even though the wording is currently the same for all of them.


suShirow (approved):
Much clearer thank you

PR #1751, opened 2022-11-01
[AST] divination update, checklist reorder, random cleanup

otocephaly:
Hiiiii

I updated display order on some core modules in order to be able to rearrange things in the checklist
Divination was updated based on what's listed in the balance

This is what both look like after running it using example log below
image

image

https://www.fflogs.com/reports/a:KtJ3NaWZm48kDLXF#fight=4&source=6


otocephaly:
Hiii

Subsequent to the many changes in the meantime, I've updated the relevant sections which included removing the defensives from the checklist. In spirit of this and since it'll be handy for planning purposes to have them on this list, I've added collective unconsciousness (as discussed), macrocosmos since it's relatively under utilized without planning, and earthly star since we don't have a specific section dealing with it. Here are the updated screenshots as a result

Please note the following however,
Bump to 6.4 will come at a later PR to separate that out and keep that PR simple
Undraw seems to come up in the arcana logs whenever there is a redraw. Will work on fixing that (or someone can help if they wish)

cc: @suShirow 

Divination:
image

Defensives:
image

Checklist after reordering:
image

src/parser/jobs/ast/modules/Lightspeed.tsx
```diff
@@ -1,199 +0,0 @@
...
-				deathTruncated: false,
-
-			}
-			this.castHook = this.addEventHook(
-				filter<Event>()
-					.source(this.parser.actor.id)
-					.type('action'),
-				this.onCast,
-			)
-			this.castTimeIndex = this.castTime.setTimeAdjustment('all', LIGHTSPEED_REDUCTION)
```
AzariahAeroborn, line 76: This code (line 76, and the matching code on lines 94 and 95 for tryCloseWindow) is critical for marking spells as instant casts during a Lightspeed window, and needs to be ported to _somewhere_.  There's a visibility/maintainability argument for keeping a separate Lightspeed module to do this castTime modification, or I think having the Lightspeed handling part of the Divination module handle it would be fine as well (though my preference is the more discoverable separate Lightspeed module, even though it wouldn't be suggesting / recommending anything).

Without this code, we're dinging multi-weaves in lightspeed (see http://localhost:3000/fflogs/A6TWmwzX9xkQbtva/117/84)

AzariahAeroborn (changes requested):
The MAJOR suggestion for a single usage of Lightspeed that was decoupled from Divination seems a little stringent to me, but I did see you and Sushi had discussed this previously, so if you're comfortable with that level then we can roll with it.  See comment on the castTime adjustment from Lightspeed below.

otocephaly:
sushi didn't state a level at which to put it, but I think major makes sense based on her statement
I'll leave it at major for now and see if there are any user or balance feedback

(here's the conversation for my own reference in case I forget in the future)
image

src/parser/jobs/ast/modules/Lightspeed.tsx
otocephaly: very good point! I see what you mean
I'll add in a lightspeed module and reference it in divination as suggested

otocephaly: added lightspeed section
lemme know if it's sufficient
thank you ♥

src/parser/jobs/ast/modules/Lightspeed.tsx
```diff
@@ -0,0 +1,44 @@
...
+const LIGHTSPEED_REDUCTION = -2500
+
+// in this module we only want to track Lightspeed windows for castTime/weaving purposes
+export class Lightspeed extends Analyser {
+	static override handle = 'Lightspeed'
+
+	@dependency private data!: Data
+	@dependency private castTime!: CastTime
+
+	private castedTime: number | undefined = undefined //when lightspeed was casted
```
AzariahAeroborn, line 17: @otocephaly 
Might be worth cleaning this up in a follow-up, since this is really going unused.  I realize it was probably lifted from the previous incarnation of the code and so I'm not going to block on it, but might be worth removing the vestigal code so future maintainers don't have to wonder what's going on here.

AzariahAeroborn (approved):

src/parser/jobs/ast/modules/Lightspeed.tsx
otocephaly: oop I thought I cleaned it up
I'll keep in mind for a clean up in the near future then
thanks for the advice

suShirow:
I'm late to the party - apologies that my availability this patch has been tight. 
It appears to be the general consensus that 2min burst with lightspeed feels awful and defeats the purpose, since you'll end up drifting cards out of burst. 

There's some specific situations where it might be useful to decouple it 
This expac, off the top of my head, I can only think of holding it for prog situations for raise and maybe P3S fountain of fire, but even that you can get away with not using LS. 

So I concur with leaving it as major
Appreciate ya'll handling all of this

suShirow:
I realized Aza's point was "a single usage", 
Maybe we can have it as 
1 decouple: medium
2 decouples: major

But we can re-evaluate that

PR #1752, opened 2022-11-07
Updated Dark Knight to be supported with 6.2

Styrfire:
The only change for 6.2 to Dark Knight was lowering Living Dead's effect duration from 24 to 20 seconds and potency changes. (https://na.finalfantasyxiv.com/lodestone/topics/detail/6eee1ca8a733856669d901d95d2fa9db46a466e6/)


ayyaruq (approved):
I assume you meant Living Shadow instead than Living Dead. Seems fine since we don't even use the duration for it anywhere. LGTM.

PR #1753, opened 2022-11-16
[MNK] Overwritten LeadenBuff Check

Rjb99:
This PR adds a suggestion for when a monk dragon kicks when they already have the leaden fist buff.

(Dragon Kick rotation is a meme, don't do it kids!)


src/parser/jobs/mnk/modules/DragonKick.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+	private LeadBuffUp = false
+	private overKicks = 0 // Amount of times they kicked over a buff
+
+	override initialise(): void {
+		this.addEventHook(
+			filter<Event>()
+				.source(this.parser.actor.id)
+				.type('damage')
+				.cause(filter<Cause>().action(this.data.actions.DRAGON_KICK.id)),
```
ayyaruq: Not sure how I feel about this. It's better this way if you're manually tracking their status state, but if going for the more basic "do I have this status" you need to check on use time instead of damage time, unless it misses. Kinda annoying now I think about it more. An observation tho.

ayyaruq (commented):
Not a proper review, just 2 Q's for now:
1. any reason not to roll this into the existing Bootshine analyser?
2. is it actually worth hooking on status apply when you can query the statuses active at a point in time? So in `onKick`, just check if `this.combatants.selected.hasStatus()` or whatever. Might potentially get out of sync or something so not sure if it's actually better, but it's definitely simpler since you only need a single hook.

Rjb99:
> Not a proper review, just 2 Q's for now:
> 
> 1. any reason not to roll this into the existing Bootshine analyser?
> 2. is it actually worth hooking on status apply when you can query the statuses active at a point in time? So in `onKick`, just check if `this.combatants.selected.hasStatus()` or whatever. Might potentially get out of sync or something so not sure if it's actually better, but it's definitely simpler since you only need a single hook.

I've merged it into BootShine and I've moved to just checking status on kick.

PR #1754, opened 2022-11-30
MCH - Looser heat overcap guidelines, clean up unnecessary output

hintxiv:


src/parser/jobs/mch/modules/YassQueen.tsx
```diff
@@ -152,7 +149,7 @@ export class YassQueen extends Analyser {
 					content: <Fragment>
 						{this.parser.formatEpochTimestamp(queen.start)}
 						<span> - </span>
-						{queen.battery} Battery spent ({this.parser.formatDuration(queen.battery * BATTERY_TO_MILLIS_FACTOR)}), {totalDamage} total damage
+						{queen.battery} Battery spent, {totalDamage} total damage
```
Toastdeib, line 152: Do we need to run a `yarn extract` here or is this not translated?

hintxiv: I ran yarn extract, looks like that text was never wrapped in \<Trans\>. I feex

Toastdeib (approved):

PR #1766, opened 2023-01-14
MCH - 6.3 Updates

hintxiv:
- Updated the Hypercharge module to allow non-HC GCDs as long as there are 5 total HC GCDs in the window (+ updated some wording and added timestamp buttons).
- Added Dismantle to data and to the timeline.
- Added a simple statistic for Dismantle and Tactician usage.
- Rewrote the "About" section.


AkairyuGestalter (approved):
Would be nice if there was a way to skip tracking Dismantle in the Defensives config for patches before 6.3, but I don't think there's a good way to do that. Since it's informational only, I'm not gonna block this.

PR #1768, opened 2023-01-14
Updated SMN for 6.3.

KelosPelmand:
Also removed outdated Swiftcast message (that has actually been outdated for a few patches when they made Ruby Rite cast < recast), since it was confusing users.

### DIFF
--- a/src/parser/jobs/smn/modules/Swiftcast.tsx
+++ b/src/parser/jobs/smn/modules/Swiftcast.tsx
@@ -1,23 +1,11 @@
 import {Trans} from '@lingui/macro'
 import {ActionLink} from 'components/ui/DbLink'
 import {Swiftcast as CoreSwiftcast} from 'parser/core/modules/Swiftcast'
-import React, {Fragment} from 'react'
-import {Message} from 'semantic-ui-react'
+import React from 'react'
 import {DISPLAY_ORDER} from './DISPLAY_ORDER'
 
 export class Swiftcast extends CoreSwiftcast {
 	static override displayOrder = DISPLAY_ORDER.SWIFTCAST
 
 	override suggestionContent = <Trans id="smn.swiftcast.missed.suggestion.content">Use a spell with <ActionLink action="SWIFTCAST"/> before it expires.</Trans>
-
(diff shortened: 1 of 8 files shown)


src/parser/jobs/smn/modules/Swiftcast.tsx
```diff
@@ -1,23 +1,11 @@
...
 
 export class Swiftcast extends CoreSwiftcast {
 	static override displayOrder = DISPLAY_ORDER.SWIFTCAST
 
 	override suggestionContent = <Trans id="smn.swiftcast.missed.suggestion.content">Use a spell with <ActionLink action="SWIFTCAST"/> before it expires.</Trans>
-
-	override output() {
-		return <Fragment>
-			<Message>
-				<Trans id="smn.swiftcast.note">Swiftcast should only be used on <ActionLink action="RESURRECTION" />, <ActionLink action="RUBY_RITE" />
```
AkairyuGestalter, line 16: Isn't it still the case that raising and snapshotting slipstream are still the two cases you'd want to hold Swift for? Unless I'm missing something it seems overkill to remove the entire message when those two cases might still apply...

AkairyuGestalter (changes requested):
One thing that might just be me being out of date on SMN...

src/parser/jobs/smn/modules/Swiftcast.tsx
KelosPelmand: It is still the case where those are generally the "best case uses".  However, there are some reasons I still think it is worth removing:

1) It is more guide-like than we normally put, and in that vein, should not have been there in 6.0 in the first place.
2) We don't put similar text for BLM, which would ideally use all triples/swifts on F4/Despair.
3) Although rare, you could need to hold it for movement (I recall doing this in DSU at one point.)

If we think it needs to stay in some form, let me know.

AkairyuGestalter: Reasoning seems fine to me!

AkairyuGestalter (approved):

PR #1769, opened 2023-01-15
Fix ex5's dodgy pull actor

ackwell:
pain.

docs in the module comment


ackwell:
all the other fights are fine, so core is good to go once this is in

AkairyuGestalter (approved):
Top ten janky fixes 2023

PR #1772, opened 2023-01-15
SGE - 6.3 updates

AkairyuGestalter:
Changes to SGE for 6.3:
- Reduced Phlegma cooldowns to 40s
- Eukrasian Prognosis shields cast by self on self are now eligible for Addersting generation

### DIFF
--- a/src/parser/jobs/sge/modules/Gauge.tsx
+++ b/src/parser/jobs/sge/modules/Gauge.tsx
@@ -26,7 +26,8 @@ const ADDERSGALL_CONSUMERS: ActionKey[] = [
 	'TAUROCHOLE',
 ]
 
-const OVERWRITES_DIAGNOSIS: StatusKey[] = [
+const OVERWRITES_SHIELDS: StatusKey[] = [
+	'EUKRASIAN_DIAGNOSIS',
 	'EUKRASIAN_PROGNOSIS',
 	'GALVANIZE',
 ]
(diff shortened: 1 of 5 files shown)


src/parser/jobs/sge/modules/Gauge.tsx
```diff
@@ -95,22 +96,35 @@ export class Gauge extends CoreGauge {
...
+
+		// Hook Addersgall consumer and generator actions
 		this.addEventHook(playerFilter.type('action').action(this.data.matchActionId(ADDERSGALL_CONSUMERS)), this.onConsumeAddersgall)
 		this.addEventHook(playerFilter.action(this.data.actions.RHIZOMATA.id), this.onRhizomata)
 
 		// Hook shield applications/removals that could generate Addersting
 		this.addEventHook(playerDiagnosisPartyFilter.type('statusApply'), this.onShieldApply)
 		this.addEventHook(playerDiagnosisPartyFilter.type('statusRemove'), this.onShieldRemove)
+		// For 6.3 and later patches, self-targeted Prognosis can generate Addersting
+		if (!this.parser.patch.before('6.3')) {
```
ackwell, line 114: `!before` feels like it should be `after`, but i realise that `after('6.3')` isn't actually going to do what we want here.

looking at the patch utility, i'm... tempted to say it needs a bit of a re-think, or at least extension - comparing `patch <-> patch` doesn't really make sense - in common vernacular, we reference patches as the point in time of their introduction, and it would be accurate to say "a fight logged right now is logged after 6.3)

maybe the comparisons can accept either a patch or a timestamp, and can then act as-current on patches (which isn't _incorrect_, semantically), and act as point-in-time for timestamps. idk. thoughts.

ackwell (approved):

PR #1773, opened 2023-01-16
Feature/pld update 6.3

Styrfire:


pattyjogal (approved):
Looks great! I don't think there was anything I had to comment on, but it was easy to follow along with the changes from the patch

src/parser/jobs/pld/modules/Requiescat.tsx
```diff
@@ -92,34 +94,42 @@ export class Requiescat extends BuffWindow {
...
 			],
 			downtime: this.downtime,
 			globalCooldown: this.globalCooldown,
-			suggestionIcon: this.data.actions.HOLY_SPIRIT.icon,
+			suggestionIcon: this.data.actions.REQUIESCAT.icon,
 			suggestionContent: <Trans id="pld.requiescat.suggestions.wrong-gcd.content">
-				GCDs used during <DataLink action="REQUIESCAT" /> should consist of 4 uses of <DataLink action="HOLY_SPIRIT" /> (or
-				multi-hit <DataLink action="HOLY_CIRCLE" />) and 1 use of <DataLink action="CONFITEOR" /> for optimal damage.
+				GCDs used during <DataLink action="REQUIESCAT" /> should consist of <DataLink action="CONFITEOR" />
+				, <DataLink action="BLADE_OF_FAITH" />, <DataLink action="BLADE_OF_TRUTH" />, and <DataLink action="BLADE_OF_VALOR" /> for optimal damage.
```
ackwell: i feel this is a bit heavy on action links (well, the old one was too) - but not sure how better to word it. maybe something like "the [confetior] combo" or similar - but i'm not sure. to think about for later.

src/parser/jobs/pld/modules/Requiescat.tsx
```diff
@@ -92,34 +94,42 @@ export class Requiescat extends BuffWindow {
...
 			</Trans>,
 			suggestionWindowName: <DataLink action="REQUIESCAT" showIcon={false} />,
 			severityTiers: SEVERITIES.MISSED_CASTS,
 		}))
 		this.addEvaluator(new ExpectedActionsEvaluator({
```
ackwell, line 48: i get the sense that these two evaluators are now gonna start stepping on each other's toes:

image

log: http://localhost:3000/fflogs/a:17QXyFbaMDWpmZC8/3/8

if anything, i'd probably err towards removing the custom requi window and just using expected actions, given it's now a much simpler expected behavior during that rotation phase

src/parser/jobs/pld/modules/Requiescat.tsx
```diff
@@ -92,34 +94,42 @@ export class Requiescat extends BuffWindow {
...
 				{action: this.data.actions.CONFITEOR, expectedPerWindow: 1},
+				{action: this.data.actions.BLADE_OF_FAITH, expectedPerWindow: 1},
+				{action: this.data.actions.BLADE_OF_TRUTH, expectedPerWindow: 1},
+				{action: this.data.actions.BLADE_OF_VALOR, expectedPerWindow: 1},
 			],
 			suggestionIcon: this.data.actions.CONFITEOR.icon,
 			suggestionContent: <Trans id="pld.requiescat.suggestions.missed-confiteor.content">
-				Be sure to end each <DataLink status="REQUIESCAT" /> window with <DataLink action="CONFITEOR" /> for optimal damage.
+				Be sure to use <DataLink action="CONFITEOR" />, <DataLink action="BLADE_OF_FAITH" />, <DataLink action="BLADE_OF_TRUTH" />
+				, and <DataLink action="BLADE_OF_VALOR" /> for each each <DataLink status="REQUIESCAT" /> window for optimal damage.
```
ackwell:
```suggestion
				, and <DataLink action="BLADE_OF_VALOR" /> in each <DataLink status="REQUIESCAT" /> window for optimal damage.
```

src/parser/jobs/pld/modules/RoyalAuthority.tsx
```diff
@@ -0,0 +1,193 @@
...
+			type: 'statusRemove',
+			target: this.parser.actor.id,
+			status: this.data.statuses.SWORD_OATH.id,
+		}, this.onRemoveSwordOath)
+
+		this.addEventHook({
+			type: 'statusApply',
+			target: this.parser.actor.id,
+			status: this.data.statuses.DIVINE_MIGHT.id,
+		}, this.onApplyDivineMight)
```
ackwell, line 77: general heads up; you _might_ be able to lean on the `data` fields of `statusApply` to help track stack count (it'll be the stack count for this status). ymmv, no need to change.

src/parser/jobs/pld/modules/RoyalAuthority.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+const SWORD_OATH_SEVERITY = {
+	89: TARGET.WARN,
+	94: TARGET.SUCCESS,
+}
+
+const DIVINE_MIGHT_SEVERITY = {
+	89: TARGET.WARN,
+	94: TARGET.SUCCESS,
+}
```
ackwell: please don't use `WARN`. i've been trying to get rid of it for years. i know, i know.

idea behind checklist is that it's either fine, or very much not fine. key rotational stuff that _must_ be done or you're dropping serious balls - i.e. not rolling your gcd. anything else should be a suggestion.

src/parser/jobs/pld/modules/RoyalAuthority.tsx
```diff
@@ -0,0 +1,193 @@
...
+					name: <Trans id="pld.sword-oath.checklist.requirement.name">
+						Uses of <DataLink status="SWORD_OATH" /> out of possible uses
+					</Trans>,
+					overrideDisplay: `${swordOathStacksUsed} / ${swordOathPotentialStacks} (${this.getPercent(swordOathStacksUsed, swordOathPotentialStacks).toFixed(2)}%)`,
+					percent: this.getPercent(swordOathStacksUsed, swordOathPotentialStacks),
+				}),
+			],
+		}))
+
+		this.checklist.add(new TieredRule({
```
ackwell, line 168: these two should probably be combined into a single metric - if keeping as a checklist, probably some form of "use all resources granted by royal authority" - unless someone's got a strong call for keeping seperate?

src/parser/jobs/pld/modules/RoyalAuthority.tsx
```diff
@@ -0,0 +1,193 @@
...
+		this.onRemoveSwordOath()
+		this.onRemoveDivineMight()
+
+		const swordOathStacks = this.data.statuses.SWORD_OATH.stacksApplied
+		const swordOathStacksUsed = this.swordOathHistory.reduce((used, swordOath) => used + swordOath.used, 0)
+		const swordOathPotentialStacks = this.swordOathHistory.reduce((potential, swordOath) => potential + swordOath.initial, 0)
+		// Leaving the following commented out since it's not actually being used right now, but here's the calculation for checking dropped stacks if that data becomes useful
+		// const droppedStacks = this.swordOathHistory.reduce((dropped, swordOath) => dropped + Math.max(swordOath.initial - swordOath.used, 0), 0)
+
+		this.checklist.add(new TieredRule({
```
ackwell, line 150: i'm not super on top of the current recommendations for pld; i've seen some rotations with additional hardcasts of hs, and/or dropped atonements, for alignment purposes.

are they recommended enough to care about?
and if they are, is this checklist going to start dinging people who drop atonements for that purpose?
and if it _is_ - is it niche enough to not worry about anyway?
genuine questions, idk

src/parser/jobs/pld/modules/Tincture.tsx
```diff
@@ -33,18 +33,21 @@ export class Tincture extends CoreTincture {
...
+				{action: this.data.actions.CONFITEOR, expectedPerWindow: 1},
+				{action: this.data.actions.BLADE_OF_FAITH, expectedPerWindow: 1},
+				{action: this.data.actions.BLADE_OF_TRUTH, expectedPerWindow: 1},
+				{action: this.data.actions.BLADE_OF_VALOR, expectedPerWindow: 1},
+				{action: this.data.actions.HOLY_SPIRIT, expectedPerWindow: 1},
 			],
 			suggestionIcon,
 			suggestionContent: <Trans id="pld.tincture.suggestions.trackedactions.content">
-				One use of <ActionLink {...this.data.actions.EXPIACION}/> and at least one use of <ActionLink {...this.data.actions.CIRCLE_OF_SCORN}/> should occur during every <ActionLink {...this.data.actions.INFUSION_STR}/> window.
+				Try to land at least one cast of <DataLink action="EXPIACION" />, <DataLink action="CIRCLE_OF_SCORN" />
```
ackwell, line 47: this really doesn't read well i don't think - would be good to cut out the bulk of the big-list-of-actions, as it just becomes a wall of text and people glaze over.

this is just crap i'm thinking up off the top of my head, but something to the tune of

"try to land the [confetior] combo and other key damage abilities within [tincture] windows. check the [link to tinctures module] for details"

would help - i think the table in the module is much more clear around expectations than written text.

src/parser/jobs/pld/modules/Confiteor.tsx
```diff
@@ -0,0 +1,73 @@
...
+		if (!this.usedConfiteor) { this.droppedConfiteors++ }
+		if (this.confiteorHook != null) {
+			this.removeEventHook(this.confiteorHook)
+			this.confiteorHook = undefined
+		}
+		this.usedConfiteor = false
+	}
+
+	private onComplete() {
+		this.suggestions.add(new TieredSuggestion({
```
ackwell, line 59: does this suggestion add anything that the requi required actions one does not? same link i logged on the other one trips this one as well, and it really feels like it's repeating the same thing.

src/parser/jobs/pld/modules/FightOrFlight.tsx
```diff
@@ -109,34 +46,50 @@ export class FightOrFlight extends BuffWindow {
...
 		}))
 
+		this.addEvaluator(new ExpectedActionsEvaluator({
+			expectedActions: [
+				{action: this.data.actions.GORING_BLADE, expectedPerWindow: 1},
+				{action: this.data.actions.CONFITEOR, expectedPerWindow: 1},
+				{action: this.data.actions.BLADE_OF_FAITH, expectedPerWindow: 1},
+				{action: this.data.actions.BLADE_OF_TRUTH, expectedPerWindow: 1},
+				{action: this.data.actions.BLADE_OF_VALOR, expectedPerWindow: 1},
+				{action: this.data.actions.HOLY_SPIRIT, expectedPerWindow: 1},
```
ackwell, line 67: afaik this isn't a safe assumption - there will be times in the rotation where you'll end up with i.e. 3 atonements as filler unless you're doing spreadsheet optimisations. afaik current "general" recommended rotation burst window just locks in goring and confetti combo, and leaves the remaining 3 as priority based filler

src/parser/jobs/pld/modules/FightOrFlight.tsx
```diff
@@ -109,34 +46,50 @@ export class FightOrFlight extends BuffWindow {
...
+			],
+			suggestionIcon: this.data.actions.EXPIACION.icon,
+			suggestionContent: <Trans id="pld.fightorflight.suggestions.gcd_actions.content">
+				Try to land at least one cast of <DataLink action="GORING_BLADE" />
```
ackwell, line 71: gunna ditto myself about action lists in suggestions

ackwell (commented):
mostly looking good!

most comments are just text wording or ideas/questions

main actionable thing to look into is the duplication in/between reqi module and confetior module, and the checklist in royal authority.

thanks for the work, and sorry it took so long for my flakey ass to check over it! i'll try to keep iteration time lower i swear 😔

ackwell:
oh and i forgot to push, just a quick fix for the status keys - notes in commit long description

src/parser/jobs/pld/modules/Requiescat.tsx
Styrfire: Why would I not want to list out each ability? That was done previously. I just followed the convention.

src/parser/jobs/pld/modules/RoyalAuthority.tsx
Styrfire: There are no data fields for the divine might status because there are no stacks. I tried that initially

src/parser/jobs/pld/modules/RoyalAuthority.tsx
Styrfire: I'm fine with getting rid of it, but is there something that can be said for that, so that it doesn't block commits like this?

src/parser/jobs/pld/modules/RoyalAuthority.tsx
Styrfire: No, hardcasts of hs are not recommended because they are damage losses.

Styrfire: image

src/parser/jobs/pld/modules/FightOrFlight.tsx
Styrfire: I don't agree with this because you should move around the Holy Spirit to fit in the window. It doesn't break combos and you can hold onto it and start a new combo if needed. Buffed Holy Spirits can and should always be worked into a Fight or Flight window

src/parser/jobs/pld/modules/Requiescat.tsx
Styrfire: I agree and will push up the change.

src/parser/jobs/pld/modules/RoyalAuthority.tsx
Styrfire: "is there something that can be said for that" What I mean by this is can this be a public notice and documented somewhere so people can start phasing it out of their modules? I don't feel like this should be a pr blocker because I'm not aware of any documentation that exists regarding phasing this out.

src/parser/jobs/pld/modules/Tincture.tsx
Styrfire: Do you have any "link to modules" examples? I would love to use that more! That being said, I do think this is subjective and it's hard for me to justify not being clear in the notes before graphing it out in the module. (I think linking to the module is a fantastic suggestion though)

src/parser/jobs/pld/modules/Confiteor.tsx
Styrfire: I personally think this is specific and major enough to deserve it's own call out. Letting the buff time out is HUGE imo.

src/parser/jobs/pld/modules/FightOrFlight.tsx
Styrfire: I could have been a bit nicer with my previous comment. The Balance thought that was the case early on but it was proven to be a bad assumption because hard casting Holy Spirit as filler turned out to be a damage lost (because of the damage loss on the auto) or damage neutral in the long run. Paladin works with a priority system now (and the math behind it backs it up).

src/parser/jobs/pld/modules/FightOrFlight.tsx
Styrfire: I ditto my comment above.

src/parser/jobs/pld/modules/FightOrFlight.tsx
ackwell: Ah I just double checked now and they've added info w/r/t prioritising DM HS. Cool, easy.

src/parser/jobs/pld/modules/Tincture.tsx
ackwell: `this.parser.scrollTo(handle)`

there's an example in `core/modules/Weaving` (for much the same purpose, linking to the module for more detail).

src/parser/jobs/pld/modules/RoyalAuthority.tsx
ackwell: Yeah, had read that. The thing of note is the "almost exactly the same" bit - _iif_ the looping rotations are common _enough_ to ding people regularly for dropped atonements, then we'll probably need to look at i.e. reducing the breakpoint. But for now, it should be fine.

src/parser/jobs/pld/modules/RoyalAuthority.tsx
ackwell: uh, no idea why i put the comment here, it was intended to be attached to the sword oath stacks for atonement.

src/parser/jobs/pld/modules/RoyalAuthority.tsx
ackwell: It's... _sort_ of covered here: https://docs.xivanalysis.com/composing-job-analysis#checklist, but not explicitly. I should probably actually document it, but it's one of those things where the active developer base has more-or-less been around the entire time and nobody thought to do it, hah.

src/parser/jobs/pld/modules/Confiteor.tsx
ackwell: oh, huh, confetti ready is a 30s buff, so it's entirely possible to hold it and use outside reqi. whack. _that said_, it's only actually a particularly strong skill with reqi, each skill loses 500 (fucking hell that's a lot) potency without it. so it's sorta a two-way relation - confetti combo is basically shit outside reqi, and reqi should be wholly consumed on confetti.

so... yeah, i _think_ it's still dupey, but perhaps a combination will require rethinking at least the wording to be very clear that _"reqi is for confetti combo"_, rather than "you want to get these in i guess". perhaps we can, like, consider the 4-hit combo as a single "thing" to do in reqi, and _any_ drop of those as a full-failure, rather than (i.e.) HS->combo just considering the last hit dropped.

just thoughts.

src/parser/jobs/pld/modules/Requiescat.tsx
ackwell: eh, mostly just looks messy, slash reads poorly. it's not make or break, just a general thing to consider.

ackwell (approved):
thanks for getting those changes in

i think there's still some open points for discussion but that can be done async

src/parser/jobs/pld/modules/Confiteor.tsx
Styrfire: I definitely agree that requi is for confetti only, and if you do let it time out, there's no way that the requi module will be happy (because it won't have the confetti spells). I can definitely see the possibility of the duplication making a bigger deal about it than it should be. I'm just struggling with the idea of there being a use case that isn't getting addressed directly (although it is indirectly in the requi module).

src/parser/jobs/pld/modules/RoyalAuthority.tsx
Styrfire: This may be a personal preference thing, but I like the idea for keeping them separate because they are two separate resources. Missing a divine might stack has a bigger dps loss than missing sword oath, but there are also fewer divine might stacks so the missing one is a bigger hit to the severity whereas missing a sword oath is harder to trigger because there are more instances.

I ultimately think it works out and would struggle to want to group them because they represent two different issues. Maybe someone is good at using all their divine might stacks but not sword oath, or the other way around?

PR #1774, opened 2023-01-16
Fix ex5 but less broken this time

ackwell:
This converts the hacky parser-land fix into a somewhat less hacky but still pretty hacky adapter-land fix. At least the adapter is a fairly homogenous state of "hacky but works i guess" in the first place so hey it's not out of place!

It also deletes the inverse dependency injector because nothing else uses it, and as shown by the last day or so, actually using it is just a footgun we don't need.


ackwell:
gunna merge to get the errors to fuck off. if another maint can just eyes over the new one so they know how it works that'd be great

PR #1782, opened 2023-01-24
Silently catch and report cache errors without nuking the page

ackwell:
Title.


ackwell (commented):
lgtm great job akk thanks for getting this in this is going to save so many headaches

no worries akk, hopefully it does

akk, why am i reviewing my own pr i think i'm going insane

i dunno akk, i think you're well past the "going" stage honestly

PR #1783, opened 2023-01-28
Use custom filter predicate for requi to avoid dm holy

ackwell:
Title. Filters are just functions - I _think_ this is the first instance of a custom (non-`filter`) filter function, but I thought on it a bit and I'm happy to support this usage pattern. Pretty sure it was intended to be able to be used like this in the first place, hah.

|Play|Before|After|
|--|--|--|
|[Correct](http://localhost:3000/fflogs/a:mR4WwJjFPgAG2Nk9/11/9)|image|image|
|[Incorrect](http://localhost:3000/fflogs/a:qv7GV2WJHjCQbMAT/2/1)|image|image|


Styrfire (approved):
The changes look fantastic! I don't have anything to add!

Styrfire (approved):

Styrfire (approved):

PR #1786, opened 2023-02-05
Feature - Overhaul usage analysis for defensive cooldowns

AkairyuGestalter:
Large-ish overhaul of the display output for defensive cooldowns. Moving away from a checklist-oriented use-maximization focus, towards acknowledging existing usage, and helping players find more opportunities without messing with their current mitigation plans. To that end:

- Defensives is now included in core analysis by default, and role actions are displayed without the need for jobs to specify those individually.
- All jobs (aside from DRG, which _has_ no job-specific defensive cooldown) now include an override of core.
- CooldownDowntime no longer supports defensive cooldown tracking (yay, one less TieredSuggestion)

Note: A couple jobs have some pretty short cooldowns, some of which also consume a secondary resource that I'm on the fence about including:
- PLD: Holy Sheltron and Intervention
- DRK: The Blackest Night
- SAM: Third Eye

I'm defaulting to including them for now, but willing to be talked out of it by job maintainers.

Since this applies to all jobs, I'm just including a core changelog note for it. Seems superfluous to include both core and job ones...


src/parser/jobs/sge/modules/Defensives.tsx
```diff
@@ -0,0 +1,24 @@
...
+		this.data.actions.PHYSIS_II,
+		this.data.actions.ZOE,
+		this.data.actions.SOTERIA,
+		this.data.actions.RHIZOMATA,
+		this.data.actions.KRASIS,
+	]
+
+	// Retaining old Trans ID to maintain i18n
+	protected override headerContent = <Trans id="sge.cooldownDowntime.defense-cd-metric">
+		Using your mitigation and healing cooldowns allows you to help keep the party healthy while continuing to deal damage and healing to your <DataLink showIcon={false} action="KARDIA" /> target.
```
AzariahAeroborn: Not sure it's worth blocking on, but this wording is clunky, sounds like I'm slapping my tank around when on SGE
(while continuing to deal damage ... to your Kardia target)

src/parser/jobs/whm/modules/oGCDs.tsx
```diff
@@ -14,7 +14,7 @@ export class OGCDs extends CooldownDowntime {
 		{cooldowns: [this.data.actions.ASSIZE], weight: 1},
 		{cooldowns: [this.data.actions.PRESENCE_OF_MIND], weight: 1},
 	]
-	override defensiveCooldowns = [
+	override suggestionOnlyCooldowns = [
```
AzariahAeroborn, line 17: Should these be removed instead and consolidated with the defensives module?  Seems like this is going to result in duplication.

src/parser/jobs/sge/modules/Defensives.tsx
```diff
@@ -0,0 +1,24 @@
...
+import React from 'react'
+
+export class Defensives extends CoreDefensives {
+	protected override trackedDefensives = [
+		this.data.actions.PNEUMA,
+		this.data.actions.HOLOS,
+		this.data.actions.PANHAIMA,
+		this.data.actions.HAIMA,
+		this.data.actions.PHYSIS_II,
+		this.data.actions.ZOE,
```
AzariahAeroborn: Might be worth some further discussion as to whether Zoe should be excluded since it's often bound with Pneuma.

src/parser/core/modules/Defensives.tsx
```diff
@@ -1,89 +1,163 @@
...
-		this.addEventHook('complete', this.onComplete)
+	private getUses(defensive: Action): number {
+		return this.cooldowns.cooldownHistory(defensive).length
 	}
 
-	private formatUsages(defensive: Action): React.ReactNode {
-		const uses = this.uses.get(defensive.id) ?? 0
-		const maxUses = this.cooldownDowntime.calculateMaxUsages({cooldowns: [defensive]})
+	private getMaxUses(defensive: Action): number {
+		return this.cooldownDowntime.calculateMaxUsages({cooldowns: [defensive]})
```
AzariahAeroborn: This isn't accounting for drift required to match the existing usages, and results in showing a Max Uses that's higher than what we're recommending in the details.  See Physis II on this log - 3/8 uses shown, but the breakdown only shows 3 possible extra uses in the downtimes.
http://localhost:3000/fflogs/A948DrvXpx1MzRaL/5/6

src/parser/core/modules/Defensives.tsx
```diff
@@ -1,89 +1,163 @@
...
+		roleDefensives.forEach(roleAction => {
+			if (!this.trackedDefensives.find(action => roleAction.id === action.id)) {
+				this.trackedDefensives.push(roleAction)
+			}
 		})
+	}
 
-		this.addEventHook('complete', this.onComplete)
+	private getUses(defensive: Action): number {
+		return this.cooldowns.cooldownHistory(defensive).length
```
AzariahAeroborn: This is failing to exclude interrupted casts and is instead counting them as uses.  See Pneuma on this log
http://localhost:3000/fflogs/A948DrvXpx1MzRaL/5/6

AzariahAeroborn (changes requested):
I didn't dig into additional specific classes beyond WHM and SGE (that were in the log I checked and referenced for the comments on usage & max usage calculations).  High level, seems reasonable, but concerned about the WHM duplication and the usage calcs.

src/parser/jobs/sge/modules/Defensives.tsx
AkairyuGestalter: Updated the wording to say "...deal damage to enemies and healing to your Kardia target." Happy to tweak the wording farther if we need to but this should at least clarify it a bit.

src/parser/jobs/sge/modules/Defensives.tsx
AkairyuGestalter: Yknow, that's a fair point. I can't remember the last time I used Zoe for something other than Pneuma, aside from mis-pressing the button. I'll remove it for now.

src/parser/core/modules/Defensives.tsx
AkairyuGestalter: That is intentional. I want to both show where the user can get additional uses given their existing ones, and also show them whether or not they're truly maximizing the potential usage of the skill. Think of something linke Pneuma that's got a long CD. If someone were to consistently use it at an interval of 235s for example, we wouldn't be showing any extra opportunities since there's no way to fit an additional 120s CD in between, but they'd only be getting about half of the possible uses in a fight. If there's a better way to communicate that then I'm all ears, but the numbers are technically correct to what they should be.

src/parser/jobs/whm/modules/oGCDs.tsx
AkairyuGestalter: I was leaning towards leaving this in at first, and deciding based on user feedback later whether the suggestions were still helpful. Yes it'll duplicate information about them, but since the new Defensives doesn't suggestions

src/parser/core/modules/Defensives.tsx
AkairyuGestalter: Yeah, that's a thing. I updated the code to filter out any cooldown history entries where the cooldown ended due to interruption and that seems to have fixed the problem.

src/parser/core/modules/Defensives.tsx
AkairyuGestalter: After our additional discussion on Discord, revising this to show the maximum as the sum of actual usage plus the extra uses shown in the table. I toyed with having the accordion labels be [ActionName] - Used: [count] - Available: [count] or something like that and it didn't feel better than the simple [used]/[total] version so I left it that way with the calculation fix in place

AzariahAeroborn (approved):

PR #1798, opened 2023-05-20
Account for weave delay for buff windows

Maliziacat:
**Issue:** GNB's No Mercy window asks for 9 GCDs for an estimated 2.49 GCD.

**Fix:** Introduce a very conservative weave delay estimate for buff windows to the ExpectedGcdCountEvaluator module.

Because buffs with stacks (Meikyo Shisui, Blood Weapon, Swiftcast, Zoe) end immediately after the stack is consumed, this weave delay calculation would change the GCD incorrectly. Buffs with stacks are excluded from this change.

**Test Plan:**
The following buffs use this module: DRG (Dragon Sight, Battle Litany, Lance Charge), DRK (Blood Weapon), PLD (Fight or Flight, Tincture), SAM (Meikyo Shisui), GNB (No Mercy), BRD (Raging Strikes), MNK (Riddle of Fire), NIN (Trick Attack), SGE (Zoe), AST/SGE/RDM/SMN/WHM (Swiftcast).

I used the following logs to compare buff modules:

https://www.fflogs.com/reports/zHFwdv9bkBPgnYVX - test logs for DRG, DRK, PLD 2.44, 2.50, SAM, GNB 2.44, 2.45, 2.46, 2.49, 2.50, BRD, MNK, NIN, SGE. Each job includes both successful and failed buff windows. PLD/GNB include some early deaths during buff window.
- fight 2 GNB 2.44 expects 7 GCDs instead of 8. It's an early death so probably OK.
- fight 14 GNB 2.45 expects 6 GCDs instead of 7. Another early death.
- fight 18, 19, 20 at GNB 2.49 expects 8 GCDs instead of 9. Intended change.

https://xivanalysis.com/fflogs/NHyJtja7kGMDLFCY - GNB at 2.49 (fight 4) vs 2.50 (fight 14)
- fight 4 GNB expects 8 GCDs. Intended change.

https://xivanalysis.com/fflogs/myFVR4WGCZ6PBhgH/ - tests logs for PLD at 2.42, 2.43, 2.50, MNK at 1.94, 2.00, and SSS 4.0, RDM swiftcast, PLD tincture
- no changes

https://xivanalysis.com/fflogs/72pCAcYwRXyjqhDF#fight=12&type=damage-done - high end MNK, DRK, GNB
- no changes

https://xivanalysis.com/fflogs/9k4hq3BLF8RDnaMQ#fight=23&type=damage-done - another high end MNK, DRK, GNB with GNB at 2.49 estimated GCD
- GNB now shows 8/8 GCDs on No Mercy

https://xivanalysis.com/fflogs/6WY24f7gPNThBnxA#fight=13&type=damage-done - PLD with 2.49 estimated GCD on fight 9
- no changes

### DIFF
--- a/src/parser/core/modules/ActionWindow/evaluators/ExpectedGcdCountEvaluator.tsx
+++ b/src/parser/core/modules/ActionWindow/evaluators/ExpectedGcdCountEvaluator.tsx
@@ -12,8 +12,7 @@ const weaveDelay = 250
 
 // exported for use in AllowedGcdsOnlyEvaluator
 export function calculateExpectedGcdsForTime(defaultExpected: number, gcdEstimate: number, start: number, end?: number) {
-	const duration = Math.max((end ?? start) - start, 1)
-	return Math.min(defaultExpected, Math.ceil(duration / gcdEstimate))
+	return Math.min(defaultExpected, Math.ceil(((end ?? start) - start) / gcdEstimate))
 }
 
 interface ExpectedGcdCountOptions {
--- a/src/parser/core/modules/ActionWindow/evaluators/ExpectedGcdCountEvaluator.tsx
+++ b/src/parser/core/modules/ActionWindow/evaluators/ExpectedGcdCountEvaluator.tsx
@@ -112,7 +112,7 @@ export class ExpectedGcdCountEvaluator implements WindowEvaluator {
 		if (!this.hasStacks) {
 			adjustedStart += weaveDelay
 		}
-		
+
 		return calculateExpectedGcdsForTime(this.expectedGcds, this.globalCooldown.getDuration(), adjustedStart, window.end) + this.adjustCount(window)
 	}
 
--- a/src/parser/core/modules/ActionWindow/evaluators/ExpectedGcdCountEvaluator.tsx
+++ b/src/parser/core/modules/ActionWindow/evaluators/ExpectedGcdCountEvaluator.tsx
@@ -6,9 +6,12 @@ import {EvaluatedAction} from '../EvaluatedAction'
 import {HistoryEntry} from '../History'
 import {EvaluationOutput, WindowEvaluator} from './WindowEvaluator'
 
+// 700ms weave delay with "ok" ping
+const weaveDelay = 700
+
 // exported for use in AllowedGcdsOnlyEvaluator
 export function calculateExpectedGcdsForTime(defaultExpected: number, gcdEstimate: number, start: number, end?: number) {
-	return Math.min(defaultExpected, Math.ceil(((end ?? start) - start) / gcdEstimate))
+	return Math.min(defaultExpected, Math.ceil(Math.max((end ?? start) - start - weaveDelay, 0) / gcdEstimate))
 }
 
 interface ExpectedGcdCountOptions {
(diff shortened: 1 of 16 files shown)


Rjb99:
I haven't fully read this yet. But I'm not sure I like the idea of windows adjusting themselves based on death. I feel like that might be confusing to users unless we add a note to explain they died mid buff window.

Maliziacat:
> I haven't fully read this yet. But I'm not sure I like the idea of windows adjusting themselves based on death. I feel like that might be confusing to users unless we add a note to explain they died mid buff window.

I actually did not introduce that behavior! This was already baked into the analysis tool. Your buff expires on death, so we look at the shortened buff duration.

src/parser/core/modules/ActionWindow/evaluators/ExpectedGcdCountEvaluator.tsx
```diff
@@ -6,9 +6,23 @@ import {EvaluatedAction} from '../EvaluatedAction'
...
+
+	// Buffs with stacks have durations tightly coupled to the GCD
+	// and do not benefit from accounting for weave delay
+	if (!hasStacks) {
+		usableWindow -= weaveDelay
+	}
+
+	usableWindow = Math.max(usableWindow, 1)
+
+	return Math.min(defaultExpected, Math.ceil(usableWindow / gcdEstimate))
```
AzariahAeroborn: I would prefer to see the window adjustment logic in the `calculateExpectedGcdsForWindow` function instead, and this function should just naively take start and end values of a time range, and not care about whether the underlying buff has stacks.  As is, this change is going to cause a change to the way Dragon Sight calculates "forgiven" GCDs when the tether partner dies and may result in forgiving one GCD too few (or having a really counterintuitive need to pass "true" hasStacks on that forgiveness calculation).

src/parser/jobs/gnb/modules/NoMercy.tsx
```diff
@@ -77,10 +78,11 @@ export class NoMercy extends BuffWindow {
 		this.addEvaluator(new ExpectedGcdCountEvaluator({
 			expectedGcds: EXPECTED_USES.GCD,
 			globalCooldown: this.globalCooldown,
+			hasStacks: false,
 			suggestionIcon,
 			suggestionContent: <Trans id="gnb.nomercy.suggestions.gcds.content">
-				Try to land 9 GCDs during every <ActionLink action="NO_MERCY" /> window. A 20 second duration is sufficient
-					to comfortably fit 9 GCDs with full uptime if you wait until the last one-third of your GCD timer to activate it.
+				Try to land {EXPECTED_USES.GCD} GCDs ({EXPECTED_USES.GCD_SLOW} without gearing skill speed) during every <ActionLink action="NO_MERCY" /> window.
```
AzariahAeroborn: Should this copy specify the GCD threshold that's too slow to fit 9 GCDs?  Seems ambiguous as written.

AzariahAeroborn (changes requested):

src/parser/jobs/gnb/modules/NoMercy.tsx
Maliziacat: Something like this?
"Try to land {EXPECTED_USES.GCD} weaponskills ({EXPECTED_USES.GCD_SLOW} weaponskills at {GCD_SLOW} GCD or slower) during every <ActionLink action="NO_MERCY" /> window."

Or without the variable names,
"Try to land 9 weaponskills (8 weaponskills at 2.47 GCD or slower) during every No Mercy window."

AzariahAeroborn: That seems like an improvement to me.

AzariahAeroborn (approved):

PR #1803, opened 2023-05-23
DRG: 6.4 support

ebshimizu:
DRG is the same*

*but with added positional buffs. Since Dragon Sight is still primarily used for damage, it'll typically be used on CD and not for optimizing positionals (and if you're doing that we're going to assume you know what you're doing). This is a lot of words to say there are no plans for additional analysis of DS.

### DIFF
--- a/src/parser/jobs/drg/index.tsx
+++ b/src/parser/jobs/drg/index.tsx
@@ -12,7 +12,7 @@ export const DRAGOON = new Meta({
 	</>,
 	supportedPatches: {
 		from: '6.0',
-		to: '6.3',
+		to: '6.4',
 	},
 	contributors: [
 		{user: CONTRIBUTORS.FALINDRITH, role: ROLES.MAINTAINER},


nonowazu (approved):
~~Untenable amount of chan~~ Simple one-line change, doesn't impact anything.

AkairyuGestalter:
Question: have we verified that the missed positional checks won't bitch about a nominally missed positional while Dragon Sight is active? We probably ought do that before we say no changes needed

ebshimizu:
> Question: have we verified that the missed positional checks won't bitch about a nominally missed positional while Dragon Sight is active? We probably ought do that before we say no changes needed

good question, my understanding from reading the code and vaguely recalling the discussion from when it was implemented is that the `Positionals` module in core looks at the potency of the relevant actions to determine if a positional would hit, so assuming new DS handles that as it does for other skills (e.g. MNK's riddle of earth) we should be ok.

ebshimizu:
quick striking dummy test shows positionals registering ok: https://xivanalysis.com/fflogs/1xr4C8mwdAVLg3bc/1/1
stood in the front, hit some buttons. DS was active for one FC and one WT.

AkairyuGestalter:
Good enough for me, thanks for checking!

AkairyuGestalter (approved):

PR #1812, opened 2023-05-26
[SMN] Add level 70/80 skills for better Ultimate analysis

Maliziacat:
**Issue:**
A level 70/80 smn log like https://xivanalysis.com/fflogs/kaPYJ28j61NHL9Ky/7/4 will show large gaps in the timeline and various modules because Titan/Garuda/Ifrit summon actions are not present.

**Fix:**
Add Summon Titan, Summon Garuda, Summon Ifrit, Topaz Ruin III, Emerald Ruin III, Ruby Ruin III, Topaz Outburst, Emerald Outburst, and Ruby Outburst. This is enough to fix the timeline module.

**Extra Notes:**
I'm specifically avoiding additional changes even though we could support things like the summons module, because as discussed that could cause complications long term. We're not really trying to support 70/80 content here, just trying to make the timeline useable.

I used these logs for testing:
https://www.fflogs.com/reports/v8WVxQzgKYrcB7J1#fight=4 - level 90 smn parse
https://www.fflogs.com/reports/Hzf2bRYdCgKFphQ1#fight=153 - level 90 smn parse
https://www.fflogs.com/reports/kaPYJ28j61NHL9Ky#fight=7 - level 70 smn parse on UWU
https://www.fflogs.com/reports/hTkHrMvzqVagpnNc#fight=5 - level 80 smn parse on TEA

**Before** ABC / summons module / timeline:
image

**After** ABC / summons module / timeline:
image

### DIFF
--- a/src/data/ACTIONS/root/SMN.ts
+++ b/src/data/ACTIONS/root/SMN.ts
@@ -157,10 +157,28 @@ export const SMN = ensureActions({
 		cooldown: 20000,
 	},
 
+	RUBY_RUIN_III: {
+		id: 25817,
+		name: 'Ruby Ruin III',
+		icon: 'https://xivapi.com/i/002000/002682.png',
+		onGcd: true,
+		speedAttribute: Attribute.SPELL_SPEED,
+		castTime: 2800,
+	},
+
 	RUBY_RITE: {
 		id: 25823,
 		name: 'Ruby Rite',
 		icon: 'https://xivapi.com/i/002000/002760.png',
+		onGcd: true,
+		speedAttribute: Attribute.SPELL_SPEED,
+		castTime: 2800,
+	},
+
+	TOPAZ_RUIN_III: {
+		id: 25818,
+		name: 'Topaz Ruin III',
+		icon: 'https://xivapi.com/i/002000/002682.png',
 		onGcd: true,
 		speedAttribute: Attribute.SPELL_SPEED,
 		castTime: 2800,
@@ -172,6 +190,15 @@ export const SMN = ensureActions({
 		icon: 'https://xivapi.com/i/002000/002761.png',
 		onGcd: true,
 		speedAttribute: Attribute.SPELL_SPEED,
+	},
+
+	EMERALD_RUIN_III: {
+		id: 25819,
+		name: 'Emerald Ruin III',
+		icon: 'https://xivapi.com/i/002000/002682.png',
+		onGcd: true,
+		speedAttribute: Attribute.SPELL_SPEED,
+		castTime: 2800,
 	},
 
 	EMERALD_RITE: {
@@ -293,6 +320,14 @@ export const SMN = ensureActions({
 		statusesApplied: ['SLIPSTREAM'],
 	},
 
+	SUMMON_IFRIT: {
+		id: 25805,
+		name: 'Summon Ifrit',
+		icon: 'https://xivapi.com/i/002000/002680.png',
+		onGcd: true,
+		speedAttribute: Attribute.SPELL_SPEED,
+	},
+
 	SUMMON_IFRIT_II: {
 		id: 25838,
 		name: 'Summon Ifrit II',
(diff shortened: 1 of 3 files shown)


nonowazu:
Not trying to be a downer on this PR, but the goal of xiva is only to support current max-level analysis. When content from previous expansions becomes 'old', xiva branches and sets up a secondary domain for that expansion (for [example](https://stormblood.xivanalysis.com/)). However, that analysis is only good for parses made during that expansion.

There is not as-yet a mechanism for supporting analysis at lower level breakpoints.

Maliziacat:
> Not trying to be a downer on this PR, but the goal of xiva is only to support current max-level analysis. When content from previous expansions becomes 'old', xiva branches and sets up a secondary domain for that expansion (for [example](https://stormblood.xivanalysis.com/)). However, that analysis is only good for parses made during that expansion.
> 
> There is not as-yet a mechanism for supporting analysis at lower level breakpoints.

I want to at least make legacy ultimate analysis viable, but I wouldn't call this change "official support", nor do I plan on making further changes to make 70/80 fully supported. This is a minimum viable bar. Without this change, XIVA is completely useless. Neither the Stormblood nor the Shadowbringers domains are capable of handling Endwalker summoner.

Think of it from a new raider's perspective, exactly the kind of user we want to attract. A new raider does some extremes, then the current savage tier. Along the way they discover XIVA and find it to be quite useful. After current savage, I'd expect this user to try things like BLU or legacy ultimates, as a gateway to current ultimates. Wouldn't it be nice if all of those progression steps were at least functional in XIVA?

In the professional world this kind of change wouldn't be made because it costs time that could be spent on a higher value feature, while adding risk to the product. Since we're all volunteers here, the time cost is only to the code reviewer, and I'm hoping this adds zero risk to 90 SMN.

nonowazu:
This adds module logic that the maintainer needs to account for; even further on this, this becomes a slippery slope for things like [bosses](https://github.com/xivanalysis/xivanalysis/tree/endwalker/src/parser/bosses), which suddenly might need custom invuln config (compare with [endwalkers](https://github.com/xivanalysis/xivanalysis/tree/shadowbringers/src/parser/bosses) branch) - otherwise the analysis becomes useless again anyway.

I think in a professional world where funding is available, this change probably would be made because the core product would likely be funded to support level breakpoints. However, given our volunteer situation is light at best (and not all jobs are even supported for this expansion), adding the expectation from the end user of "oh hey, this now works on old ultimates at some bare minimum, when can we expect more support?" feels like a bridge too far.

Fully agree with the action changes, since it just makes the timeline work at a bare level and adds exactly 0 overhead once added.

src/data/ACTIONS/root/SMN.ts
```diff
@@ -157,6 +157,15 @@ export const SMN = ensureActions({
 		cooldown: 20000,
 	},
 
+	RUBY_RUIN_III: {
```
KelosPelmand, line 160: This will only give Ruby Ruin III a 2.8s GCD, not the full 3.0s it has in the job guide.  I didn't check the patch notes to be sure, but I assume this changed in patch 6.1, like it did for Ruby Rite, so you will need to add something to the 6.1 patch file (src\data\ACTIONS\layers\patch6.1.ts) for that.

src/data/ACTIONS/root/SMN.ts
```diff
@@ -234,6 +260,15 @@ export const SMN = ensureActions({
 		statusesApplied: ['REKINDLE', 'UNDYING_FLAME'],
 	},
 
+	RUBY_OUTBURST: {
```
KelosPelmand, line 263: Same as for Ruby Ruin III.

KelosPelmand (changes requested):
I looked at 5e595d6b74f8 (the one with the module changes) and it looked fine.  Up to you if you want to roll back the last commit.

Either way, the patch data fixes will need to be added.

src/data/ACTIONS/root/SMN.ts
Maliziacat: I checked the patch notes - it started with 2.8s cast, 2.5s recast in 6.0, then became 2.8 cast, 3.0s recast in 6.1. I'll update it :)

Maliziacat:
> I looked at [5e595d6](https://github.com/xivanalysis/xivanalysis/commit/5e595d6b74f80f9c580ad77bced7a55846a49b47) (the one with the module changes) and it looked fine. Up to you if you want to roll back the last commit.
> 
> Either way, the patch data fixes will need to be added.

I'll defer to you over Nono as you're the SMN maintainer. Summon module changes are back in :)

KelosPelmand (approved):
I am now happy with this PR and think it is ready to merge.

AkairyuGestalter (approved):
Didn't see anything obviously wrong, and Kelos is happy with it. Approved!

PR #1813, opened 2023-05-27
fflogs event adapter: handle prepull auras as a statusApply

Hugmeir:
Auras seem like a bit of a holdover from WoW, but for us they seem to cover pre-pull statuses, like stances (Grit, Mighty Guard, the BLU mimickries), as well as Well Fed and the Squadron Manual.

This makes those auras show up for us as a statusApply event at pull start.

### DIFF
--- a/src/reportSources/legacyFflogs/eventAdapter/translate.ts
+++ b/src/reportSources/legacyFflogs/eventAdapter/translate.ts
@@ -376,7 +376,7 @@ export class TranslateAdapterStep extends AdapterStep {
 		}
 	}
 
-	private adaptCombatantInfoEvent(event: CombatantInfoEvent): Array<Events['actorUpdate']> {
+	private adaptCombatantInfoEvent(event: CombatantInfoEvent): Array<Events['actorUpdate'] | Events['statusApply']> {
 		// TODO: Use more info in here. We're currently extracting the speed attribute values for the logging player, but there's also player level, prepull statuses, and more in there.
 
 		const attributeMapping: Array<[number | undefined, Attribute]> = [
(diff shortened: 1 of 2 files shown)


src/reportSources/legacyFflogs/eventAdapter/translate.ts
```diff
@@ -391,20 +391,34 @@ export class TranslateAdapterStep extends AdapterStep {
...
-			return []
+		const ourEvents = new Array<Events['actorUpdate'] | Events['statusApply']>()
+		// "Auras" seems to be a holdover from WoW, but over here they mainly cover Stances;
+		// Tank Stance (Grit, etc), as well as the BLU mimickries.
+		// The minimum thing we can do here is just report these as a statusApply at the
+		// start of each pull.
+		event.auras.forEach(aura => {
+			const statusApplyEvent: Events['statusApply'] = {
+				...this.adaptTargetedFields(event),
+				type: 'statusApply',
```
AkairyuGestalter, line 402: If you take a look at these events in the Events View module while debugging, they all end up saying that the target of the status apply is "Unknown" since there isn't actually a targetID property on the CombatantInfoEvent for resolveTargetId to use (by way of the adaptTargetedFields call). Can we add this after the adaptTargetedFields call to set the target to the source player (since no FFXIV 'aura' applies to anyone other than the source):
```
target: resolveActorId({
	id: event.sourceID,
	instance: event.sourceInstance,
	actor: event.source,
}),
```
This is basically the same thing that the actorUpdate synthing is doing into the actor property.

AkairyuGestalter (dismissed):
One comment about the target of these adapted statusApply events...

src/reportSources/legacyFflogs/eventAdapter/translate.ts
Hugmeir: I had tried this originally, but it runs into a potentially annoying issue:  Ignore BLU, and say that we have a log with a SCH, and they do a pre-pull Succor; everyone gets the `Galvanize` shield.

The logs we get from fflogs will have that `Galvanize` as a pre-pull aura, with no target; If we make the target default to the source, then you end up in the bizarre situation of everyone in the party being able to apply `Galvanize` to themselves.

I don't know what the side effects of that would be; for BLU it caused the Overheal report to start triggering for non-healers.

AkairyuGestalter: Ah, ok. I didn't realize other statuses like that would fall into this category... never mind on this then. I'll need to do some more testing to see if the 'unknown' target ends up doing anything weird, but if not then this change seems fine

AkairyuGestalter (approved):
Not seeing any unintended side effects of the 'unknown' targets on the logs I've checked through so this seems to be fine.

PR #1816, opened 2023-05-27
Fix Devilment timing check for 6.4+

AkairyuGestalter:
A knock-on effect of the buff radius changes made in 6.4 was that Technical Finish no longer immediately applies to the source Dancer upon action execution. This opened up the possibility for Devilment to be executed between the execution of the Technical Finish, and the application of its status, which was causing the "Was Devilment used on time?" check to fail incorrectly.

Updated to hook the player dancer's Technical Finish actions as well as Technical Finish status applications for the purpose of opening a new window, and adjusted the check to determine whether Devilment was used on time accordingly.

### DIFF
--- a/src/parser/jobs/dnc/changelog.tsx
+++ b/src/parser/jobs/dnc/changelog.tsx
@@ -77,4 +77,9 @@ export const changelog = [
 		Changes: () => <>Fix a bug where fights with unavoidable downtime could report Standard Step uptime as a negative percentage.</>,
 		contributors: [CONTRIBUTORS.AKAIRYU],
 	},
+	{
+		date: new Date('2023-05-27'),
+		Changes: () => <>Adjust Devilment timing check to account for changed Technical Finish status application timing after 6.4.</>,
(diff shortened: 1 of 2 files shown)


AzariahAeroborn (commented):
If we're in a 2 DNC party and we're the second tech step, it looks like tryOpenWindow is going to throw a new window onto the stack before we get the remove event from the previous DNC and cause problems - can you test for this case before we merge this?

AkairyuGestalter:
Made up a log on a dummy and no, it's doing what I'd expect it to do. The status application from the other DNC initiates the window, and my action keeps it open since we haven't had the end timestamp set yet. Also double-checked that my status application doesn't inappropriately close the other DNC's window.

AkairyuGestalter:
Neglected to make sure the "Don't use Devilment outside of Technical" suggestion was performing correctly wrt to hooking the action in addition to the status applications... Hurf. It be fixed now too.

AzariahAeroborn (approved):

PR #1817, opened 2023-05-28
PLD - Skip tracking defensives with shared gauge resources

AkairyuGestalter:
Intervention, Holy Sheltron, and Cover all use Oath Gauge, so their usage is more complicated than just the "is the cooldown up?" question. Consequently, they shouldn't be tracked by the Defensives analysis


Toastdeib (approved):
Looks good to me.

PR #1823, opened 2023-05-30
WHM Update Assize cooldown from 6.3 patch

psummers678:
Noticed Assize cooldown was incorrect in the data so updated to match 6.3 notes

### DIFF
--- a/src/data/ACTIONS/root/WHM.ts
+++ b/src/data/ACTIONS/root/WHM.ts
@@ -133,7 +133,7 @@ export const WHM = ensureActions({
 		id: 3571,
 		name: 'Assize',
 		icon: 'https://xivapi.com/i/002000/002634.png',
-		cooldown: 45000,
+		cooldown: 40000,
 	},
 
 	ASYLUM: {


src/data/ACTIONS/root/WHM.ts
```diff
@@ -133,7 +133,7 @@ export const WHM = ensureActions({
 		id: 3571,
 		name: 'Assize',
 		icon: 'https://xivapi.com/i/002000/002634.png',
-		cooldown: 45000,
+		cooldown: 40000,
```
nonowazu, line 136: This will have to be made in the [6.3 patch file](https://github.com/xivanalysis/xivanalysis/blob/endwalker/src/data/ACTIONS/layers/patch6.3.ts) instead of in the root - this allows calculations for things like assize to be consistent for a 6.1 vs a 6.3 log (for example)

nonowazu (changes requested):

src/data/ACTIONS/root/WHM.ts
psummers678: Ah ok mb closed

Issue #1707, opened 2022-06-29
[MCH] [Minor] Incorrect calculation of Gauss Round cooldown


LonelyCat124:
In this analysis:
https://xivanalysis.com/fflogs/yFv3zHLqWY6dPNTD/1/6

I apparently used 55/53 gauss round, but 55/56 ricochet.
Ignoring the 1s gap between the first ricochet and first gauss round, I should be otherwise able to use the same number of both as far as I can tell. I also definitely shouldn't be able to use more than possible.

The same bug doesn't seem to have occurred for this log though:
https://xivanalysis.com/fflogs/7bgPmYN26BX4QKWj/1/4

Maliziacat:
calculateMaxUsages in parser\core\modules\CooldownDowntime.tsx
gResets = 70, meaning we cast Heat Blast 70 times, for 70 * 15 = 1050 seconds of extra time.
gUsages = 55, meaning we cast Gauss Round 55 times.

expectedFirstUseTime: 3.000s
actualFirstUseTime: 1.513s
pullEndTimestamp: 542.244s, or 9 minutes 2 seconds.

with a cooldown of 30.000s, 542.244 / 30 = 18 extra casts.
Plus 3 initial charges = 21 casts.
70 Heat Blasts * 15 seconds / Heat Blast = 1050s extra time. 1050 / 30 = 35 extra casts from Heat Blasts.
21 + 35 = 56, which is our expected number.

Running with debug logging, however, shows we only looked at 65 Heat Blasts. It's not seeing the last set of 5 Heat Blasts starting at 8:53. This is because it thinks the next charge should be at 9:17, which is after the end of the log, so it doesn't check for more potential cooldown resets (Heat Blasts).

I've published a fix for this, https://github.com/xivanalysis/xivanalysis/pull/1825.

PR #1825, opened 2023-06-02
[Core] Ensure calculation of max cooldown usage always applies cooldown resets

Maliziacat:
**Issue:** https://github.com/xivanalysis/xivanalysis/issues/1707
The calculation of the max usages of a cooldown may fail if cooldown-reset abilities are used shortly before the parse ends and shortly after a period of downtime.

`calculateMaxUsages` in core\modules\CooldownDowntime.tsx has a core loop:
`while currentTime < pullEndTimestamp`
`  1. currentTime += step`
`  2. spend any saved charges, increase current time`
`  3. apply any cooldown resets, decrease current time`
`  4. if charge is available but in downtime, move to end of downtime then spend charge, increase current time`
`  5. if in downtime, increase current time to end of downtime`

This fails for logs like https://xivanalysis.com/fflogs/yFv3zHLqWY6dPNTD/1/6. At 8:47 in the log, we have a charge available to spend, but we're in downtime. So we go to step 4, spend our charge at 8:48, and move our current time to 9:18. 9:18 is past the end of the log and so we think we're done. However, we haven't dealt with the 5 Heat Blasts from 8:53-8:58, which we should do in step 3.

**Fix:** Move step 3 to the end. Apply any cooldown resets after dealing with all possibilities in which we could increase the current time.

**Test Notes:** This affects MCH's Gauss Round / Ricochet (via Heat Blast), DRK's Plunge (via Unmend), and WAR's Infuriate (via Fell Cleave, Decimate, Chaotic Cyclone, Inner Chaos). The bug is most likely to show up in clears where there is downtime shortly before the clear, AND one of these cooldown reset abilities is used between that downtime and the clear.

Test logs:
https://xivanalysis.com/fflogs/yFv3zHLqWY6dPNTD/1/6 - MCH bug report log. Gauss Round now shows 56 potential uses instead of 53.
https://xivanalysis.com/fflogs/qcpAQBKY1Dzdmfkb/6 - two WAR logs, but no interesting downtime. No change.
https://xivanalysis.com/fflogs/APM4TFqyGhNXdcCv/71 - high end DRK/WAR log on P9S, no interesting downtime. No change.
https://xivanalysis.com/fflogs/ALfJhPnCHK4mRT7Y/33 - high end MCH/WAR log on P9S, no interesting downtime. No change.

*To find interesting downtime, I looked at P8S phase 1. If Gorgon was first, there's a big period of downtime for Hippo-2 mechanics shortly before enrage. So we're looking for logs with a clear time between 7:27 and 7:53, with a WAR / DRK / MCH, and a noticeable drop in DPS around 7:00 indicating Hippo-2.*
https://xivanalysis.com/fflogs/6LhHG3WF8gdX4Zxy/4 - DRK and MCH. No change to DRK, because who uses Unmend anyways. MCH Gauss Round / Ricochet potential uses change to 48 from the original 45 because of the post-downtime Hypercharge.
https://xivanalysis.com/fflogs/gQnCyp6FjYXM9Pdz/12/45 and https://xivanalysis.com/fflogs/zFn3t1Z4GLTfKVhP/7/3 - WAR. No change, even with the Inner Release after downtime.
https://xivanalysis.com/fflogs/yJmf84DxzjtF271R/7 - WAR and MCH. No change to either. MCH does enter Hypercharge in the relevant window but only gets a single Heat Blast off.

### DIFF
--- a/src/parser/core/changelog.tsx
+++ b/src/parser/core/changelog.tsx
@@ -10,6 +10,13 @@ export const changelog: ChangelogEntry[] = [
 	// 	contributors: [CONTRIBUTORS.YOU],
 	// },
 	{
+		date: new Date('2023-07-11'),
+		Changes: () => <>
+			Better handling for defensives with a shared cooldown, and fix a bug with the extra use available calculations.
+		</>,
+		contributors: [CONTRIBUTORS.AKAIRYU],
+	},
+	{
 		date: new Date('2023-07-09'),
 		Changes: () => <>
 			Fixed prepull actions for defensive cooldowns module.
--- a/src/parser/core/modules/Defensives.tsx
+++ b/src/parser/core/modules/Defensives.tsx
@@ -9,7 +9,7 @@ import {CooldownDowntime} from 'parser/core/modules/CooldownDowntime'
 import {Data} from 'parser/core/modules/Data'
 import React, {Fragment, ReactNode} from 'react'
 import {Accordion, Button, Icon, Message, Table} from 'semantic-ui-react'
-import {CooldownEndReason, CooldownHistoryEntry, Cooldowns} from './Cooldowns'
+import {ChargeHistoryEntry, CooldownEndReason, CooldownHistoryEntry, Cooldowns} from './Cooldowns'
 import DISPLAY_ORDER from './DISPLAY_ORDER'
 import {Timeline} from './Timeline'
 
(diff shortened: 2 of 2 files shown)


KelosPelmand (changes requested):
While there is definitely an issue that needs fixed in the original algorithm, this fix introduces a different problem.  Consider the following hypothetical:

4 min encounter with downtime from 1:57-2:15
tracked skill with 1 min cd, no charges, first used at 0:00
5s reset skill used at 1:30

Current algo usages at:
0:00
1:00
2:00 -> 1:55 (reduced by reset, used here)
2:55
3:55
5 usages total

Modified algo usages at:
0:00
1:00
2:00 -> 2:15 (moved to end of dt) -> 2:10 (reduced by reset, use here)
3:15
4 usages total

This fix only trades one problem for another, but I don't have a better fix yet, so if you come up with something, feel free to update the PR.

PR #1826, opened 2023-06-02
Update MNK.ts to have the correct cooldown for RoE

cptroot:
Noticed that xivanalysis was saying I could use RoE 27 times in the new extreme.


Dot-Inni:
Depending on when this was changed, it should be in the corresponding patch layer (src/data/ACTIONS/layers), but I'm pretty sure Monk is *generally* still kind of scuffed even at 6.0.

AkairyuGestalter:
Superseded by #1835. Closing.

PR #1828, opened 2023-06-04
Feature/pld 6.4 fixes

czyba:
Prepares PLD for 6.4 by:

* Unflagging Atonement as combo-breaker
* Improve FoF Analysis due to Atonement no longer breaking Combo
* Marking PLD supported Patches range to `[6.3, 6.4]`

Here's an example of the FoF Analysis:
image


czyba:
Due to discussions about scope I've decided to split the PR into seperate parts, the atonement fix and the rest.

PR #1831, opened 2023-06-23
bugfix: Fix report loading for Firefox in Private Browsing mode

czyba:
Firefox rejects Cache access in private browsing mode with security errors. This causes reports to not load. The cache is used in two places:

* `legacyStore::fetchReport`
* `fflogsApi::getFflogsEvents`

The latter already has a try-catch block for `getCache` throwing an exception, the former does not. Essentially replicates the try-catch block to the former code location.

Alternatively this can be moved into `fflogsApi::getCache` method so that this fix is does not need to be replicated.

For more information about the SecurityError see [MDN Documentation](https://developer.mozilla.org/en-US/docs/Web/API/CacheStorage?retiredLocale=de).

### DIFF
--- a/src/reportSources/legacyFflogs/legacyStore.ts
+++ b/src/reportSources/legacyFflogs/legacyStore.ts
@@ -1,3 +1,4 @@
+import * as Sentry from '@sentry/browser'
 import * as Errors from 'errors'
 import ky from 'ky'
 import _ from 'lodash'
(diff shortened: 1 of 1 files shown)


ackwell (approved):
probably could go in `getCache` but it's neither here nor there. a fix is a fix.

me? reviewing prs? it's less likely than you think

PR #1832, opened 2023-06-29
Convert Technicalities to core RaidBuffWindow

AkairyuGestalter:
Finally had the impetus (and the time) to convert DNC's bespoke implementation to use the core RaidBuffWindow logic so we're more consistent with other jobs. Notable end-user visible changes:
1. The output table now includes a column showing the number of GCDs executed within the window, with the expectation of hitting 9 GCDs (8 normal GCDs or 6 GCDs + Standard Step and Finish, plus Tillana's shorter GCD).
2. The rotation column of the output table no longer shows the Technical Finish that began the window. This was an unintended side effect of the 6.4 Devilment timing change that this port allowed me to work around.
3. The rotation column no longer shows the dance steps performed for Standard Step under the window, as it simplifies the "did you finish it under the window" check, and the steps aren't particularly relevant otherwise.
4. Added a new suggestion when we detect that Standard Step was initiated under the window, but was not finished under the window.
5. The output table no longer displays windows that were initiated by other DNCs.
6. Cloned DRG's window interference evaluator to provide better information about whether the player overwrote or was overwritten by another DNC's buff.
7. The Devilment "on time" check will no longer treat non-use of Devilment in the window as a "late" use, since we're no longer showing the windows initiated by other players. Use outside of Technical Finish will still be tracked and suggested against as before.


AkairyuGestalter:
Based on discussion in Discord:
- Converted the multi-job buff overwriting information to a core evaluator and added to RaidBuffWindows by default
- Disabled said evaluator for RPR since Arcane Circle indirectly functions as a gauge generator
- Created a core RulePassedEvaluator to encapsulate all the X vs checkmark evaluators we have scattered around, and converted the Feather pooling and Devilment timing evaluators created herein to extend that type

src/parser/jobs/rpr/modules/ArcaneCircle.tsx
```diff
@@ -24,6 +24,9 @@ export class ArcaneCircle extends RaidBuffWindow {
 
 	override buffStatus = this.data.statuses.ARCANE_CIRCLE
 
+	// RPR should not evaluate Arcance Circle overwrites since it's also a gauge generator
+	override evaluateOverwrites = false
```
hintxiv: I think this is a mistake. Napkin math follows:

```
PH = 1000p
50 shroud gauge ~= 313*5 = 1565p
Odds of losing a use of AC by holding it for 20s = (20/120) = 1/6
Avg. potency gain from overwriting another RPR's AC = (1000 + 1565)(1/6) = 427.5p
```

For a gain of 427.5p you're losing 3% on 8 players for 20s. That works out to around 1780 RPR-equivalent potency (healer/caster/ranged potency is worth more, tank potency is worth less) per player in 20s, which is easily surpassed by all jobs - even Glare spam for 20s is significantly more than this. Therefore on average, overwriting AC seems like a bad practice to recommend.

Thoughts?

AkairyuGestalter: The counterargument to the effective potency numbers is that delaying AC completely screws the delaying player's rotation since now their own burst is going to drift out of alignment with everyone else unless something drastic happens. That "something drastic" basically falls into the category of advanced non-standard RPR play, which isn't really what we're about here. In content where you're even likely to see multiple reapers (24mans mainly) it gets even messier so leaving this off for RPR seemed the right move.

It's been a hot minute since I looked back at this code holistically, but I could be convinced to change this setting to only skip the suggestion to not overwrite other player's buffs, but keep the informational column around, if we think that's still useful information to have.

AkairyuGestalter: Since I'm an idiot and forgot that the whole thing was informational only, and didn't have suggestions attached to it, I'm now fine with this. Done!

hintxiv (approved):
Seems good to me 👍

PR #1836, opened 2023-07-02
[BLM] Display timestamps for paradox and polyglot proc overwrites

Maliziacat:
**Issue:** you may be notified that you overwrote Paradox or Polyglot, like in https://xivanalysis.com/fflogs/A3pPfcTgFarKHLnb/22/7. Unfortunately, you have no easy way to determine when you overwrote your proc.

**Fix:** Display a timestamp and a button to the timeline for the proc overwrites.
image

**Testing:**
https://xivanalysis.com/fflogs/A3pPfcTgFarKHLnb/22/7 - real BLM parse with a single paradox overwrite.
https://xivanalysis.com/fflogs/A3pPfcTgFarKHLnb/16/7 - BLM parse with no overwrites. No additional UI is displayed.
https://xivanalysis.com/fflogs/bHLWDM8wqknyCBPK/1/1 - BLM parse with multiple paradox and polyglot overwrites. Note the Rotation Outlier module mentions when Umbral Ice paradox was missed, but the Proc Overwrites module shows when the overwrite actually occurred.


src/parser/jobs/blm/modules/Gauge.tsx
```diff
@@ -73,14 +75,17 @@ const POLYGLOT_COLOR = Color(JOBS.BLACK_MAGE.colour)
 
 export class Gauge extends CoreGauge {
 	static override handle = 'gauge'
-	static override title = t('blm.gauge.title')`Gauge`
+	static override title = t('blm.gauge.title')`Proc Overwrites`
```
AkairyuGestalter: Please leave the title alone. Paradox and Polyglot aren't proc statuses so they should not be labeled as such.

src/parser/jobs/blm/modules/Gauge.tsx
```diff
@@ -575,5 +587,48 @@ export class Gauge extends CoreGauge {
 			}))
 		}
 	}
+
+	private overwriteMessage(timestamp: number) {
+		if (this.overwritePolyglotTimestamps.indexOf(timestamp) > -1) {
```
AkairyuGestalter: This feels really awkward... Instead of storing two arrays of timestamps and inferring which thing was overwritten based on the passed-in timestamp, we should keep a single array of an interface that holds both the timestamp and some identifier for which thing was overwritten so this method can just check that property.

src/parser/jobs/blm/modules/Gauge.tsx
```diff
@@ -575,5 +587,48 @@ export class Gauge extends CoreGauge {
...
+
+		return <Trans id="blm.gauge.overwrites.unknown">Unknown proc overwritten</Trans>
+	}
+
+	override output() {
+		const allOverwrites = this.overwritePolyglotTimestamps.concat(this.overwriteParadoxTimestamps).sort()
+		if (allOverwrites.length === 0) { return false }
+
+		return (
+			<Table collapsing unstackable compact="very">
```
AkairyuGestalter: Especially after putting the title property back the way it was, we should probably have some table header message explaining what we're showing to the user here.

AkairyuGestalter:
One more food for thought, if we're keeping a list of all the timestamps where polyglot is overwritten, we can probably dispense with the separate counter variable and just do a .length check on that list for the suggestion.

Maliziacat:
> One more food for thought, if we're keeping a list of all the timestamps where polyglot is overwritten, we can probably dispense with the separate counter variable and just do a .length check on that list for the suggestion.

Fixed! Thanks for the note.

src/parser/jobs/blm/modules/Gauge.tsx
```diff
@@ -575,5 +584,48 @@ export class Gauge extends CoreGauge {
...
+				<Table.Header>
+					<Table.Row>
+						<Table.HeaderCell><Trans id="blm.gauge.header.time">Time</Trans></Table.HeaderCell>
+						<Table.HeaderCell><Trans id="blm.gauge.header.event">Event</Trans></Table.HeaderCell>
```
AkairyuGestalter: I think this column might better be labeled as something like "Overwrote which?", given the other request to leave the title alone. It would also be valid to have the content just be the datalink for the skill in question, either Paradox or Xenoglossy (since we don't really handle AoE as much as single-target).

src/parser/jobs/blm/modules/Gauge.tsx
```diff
@@ -549,15 +558,15 @@ export class Gauge extends CoreGauge {
 			}))
 		}
 
-		if (this.overwrittenPolyglot > 0) {
+		if (this.overwritePolyglotTimestamps.length > 0) {
```
AkairyuGestalter: wrt the other comment about holding a single array containing all of the data, it'll be better to do something like
const overwrittenPolyglot = this.overwriteData.filter(event => etc.).length
and use that const in the condition and the Plural rather that calculating the count twice

AkairyuGestalter (changes requested):
Apparently the rest of my comments didn't actually post, go figure. A couple things I'd like to see changed/polished up here.

src/parser/jobs/blm/modules/Gauge.tsx
Maliziacat: I see a future opportunity to include dropped Enochian timestamps as well - but definitely in a separate changelist. That would mean a more generic header title like "Event" and perhaps more descriptive cell entries.

But for now I'll go with the "Overwrote" header title and keep the notes small.

AkairyuGestalter: Dropped Enochian can be gotten from the Rotation Outliers table since that's the most serious error the player can make (aside from dying). Adding it elsewhere would be needlessly duplicative, so yeah, let's just keep this table limited to the paradox and Polyglot overwrites.

AkairyuGestalter (approved):

PR #1837, opened 2023-07-06
BLM - Convert RotationWatchdog to use RestartWindow

AkairyuGestalter:
If we actually had buff statuses still, I probably would've done this a while ago... Moved RotationWatchdog from being a bespoke Analyser to extending the RestartWindow class, so we're more consistent with other UI and take better advantage of shared code.

All rotation analysis and suggestion generation is moved into evaluator classes, and made a few minor adjustments to the core ActionWindow and ExpectedActionsEvaluator modules to preserve some of the display behavior already present in RotationWatchdog (ie. filtering out of the table the rotations without errors, and showing which actions were procs). While I was testing, I also found and fixed bug with the expected despair count where it would claim a missing despair even if the entire fire phase of a rotation took place during extended downtime (such as if the player generates 3 Hearts via Umbral Soul, then Transposes to Fire and back to Ice to generate a Paradox marker for when the boss is targetable again). That's really the only notable change for end users so that's all I mentioned in the changelog...


AkairyuGestalter:
Changed to draft because I want to update some of these evaluators based on changes I was making in #1832

AkairyuGestalter:
If I can figure out how to squash the revision history on this monstrosity, I will, but it's ready for review now that the PR it was built on top of has been merged in.

src/parser/jobs/blm/modules/RotationWatchdog.tsx
```diff
@@ -1,66 +1,79 @@
...
+import {ExpectedFireSpellsEvaluator} from './RotationWatchdog/ExpectedFireSpellsEvaluator'
+import {ExtraF1Evaluator} from './RotationWatchdog/ExtraF1Evaluator'
+import {ExtraHardT3Evaluator} from './RotationWatchdog/ExtraHardT3Evaluator'
+import {IceMageEvaluator} from './RotationWatchdog/IceMageEvaluator'
+import {ManafontTimingEvaluator} from './RotationWatchdog/ManafontTimingEvaluator'
+import {MissedIceParadoxEvaluator} from './RotationWatchdog/MissedIceParadoxEvaluator'
+import {RotationErrorNotesEvaluator} from './RotationWatchdog/RotationErrorNotesEvaluator'
+import {SkipB4Evaluator} from './RotationWatchdog/SkipB4Evaluator'
+import {SkipT3Evaluator} from './RotationWatchdog/SkipT3Evaluator'
+import {UptimeSoulsEvaluator} from './RotationWatchdog/UptimeSoulsEvaluator'
```
AzariahAeroborn, line 32: It makes sense that the main module needs to pull evaluators from all of the sub components, but it feels like a footgun in waiting that then many of these sub components have dependencies on the main module for constants like severity tiers and other constants that are specific to just a few evaluators.  Can we push down the constants from RotationWatchdog.tsx into either the appropriate evaluator or the EvaluatorUtilities sub-class?

src/parser/jobs/blm/modules/RotationWatchdog.tsx
```diff
@@ -1,66 +1,79 @@
...
+
+const DEBUG_SHOW_ALL = false && process.env.NODE_ENV !== 'production'
 
 const MAX_POSSIBLE_FIRE4 = 6
-const NO_UH_EXPECTED_FIRE4 = 4
-const FIRE4_FROM_MANAFONT = 1
+export const NO_UH_EXPECTED_FIRE4 = 4
```
AzariahAeroborn: Only used in SkipB4Evaluator

src/parser/jobs/blm/modules/RotationWatchdog.tsx
```diff
@@ -1,66 +1,79 @@
...
 	3: SEVERITY.MEDIUM,
 	5: SEVERITY.MAJOR,
 }
 
-const ENHANCED_SEVERITY_TIERS = {
+export const ENHANCED_SEVERITY_TIERS = {
 	1: SEVERITY.MINOR,
 	2: SEVERITY.MEDIUM,
 	3: SEVERITY.MAJOR,
 }
```
AzariahAeroborn: Evaluators seem to use only one or the other of these - would it be more clear to just have each evaluator define its own severity tiers and "eat" the duplication?  Or at least can these be moved to Utilities?

src/parser/jobs/blm/modules/RotationWatchdog.tsx
```diff
@@ -1,66 +1,79 @@
...
+const ROTATION_ENDPOINTS: ActionKey[] = [
 	'BLIZZARD_III',
 	'TRANSPOSE',
 	'BLIZZARD_II',
 	'HIGH_BLIZZARD_II',
 ]
 
 // This is feelycraft at the moment. Rotations shorter than this won't be processed for errors.
-const MIN_ROTATION_LENGTH = 3
+export const MIN_ROTATION_LENGTH = 3
```
AzariahAeroborn: Not consumed outside of this module.

src/parser/jobs/blm/modules/RotationWatchdog.tsx
```diff
@@ -78,666 +91,414 @@ const CYCLE_ERRORS = ensureRecord<CycleErrorCode>()({
...
-		}
-
-		// Make sure we don't go wild and return a larger expected count than is actually possible, in case the above logic misbehaves...
-		return Math.min(expectedCount, MAX_POSSIBLE_FIRE4)
-	}
-
-	public get actualFire4s(): number {
-		return this.events.filter(event => event.action === this.data.actions.FIRE_IV.id).length
-	}
+export const EMPTY_GAUGE_STATE: BLMGaugeState = {
```
AzariahAeroborn: Not referenced outside of this file

src/parser/jobs/blm/modules/RotationWatchdog/ExpectedFireSpellsEvaluator.tsx
```diff
@@ -0,0 +1,104 @@
...
+	override determineExpected(window: HistoryEntry<EvaluatedAction[]>, action: TrackedAction) {
+		const windowMetadata = getMetadataForWindow(window, this.metadataHistory)
+
+		if (action.action.id === this.fire4Action.id) {
+			if (windowMetadata.finalOrDowntime) { return undefined }
+			if (windowMetadata.expectedFire4s >= 0) { return windowMetadata.expectedFire4s }
+		}
+		if (action.action.id === this.despairAction.id && windowMetadata.expectedDespairs >= 0) {
+			return windowMetadata.expectedDespairs
+		}
```
AzariahAeroborn, line 77: I'm not super on board with exposing determineExpected for this logic - it feels like this should be part of the adjustExpectedActionsCount function instead, since really all this is doing is using the logic from that function to return early and never call through to super.determineExpected.  Doubly so since we're defining the base expected per window as 0 for both actions anyway - the adjustment value being returned by adjustExpectedActionsCount is going to wind up being the final value returned anyway except for the case of marking F4 undefined in the last window before downtime.  I think we'd be better off to look at the interface of the base class and change it as necessary so that any consumer can mark an expected action as undefined in specified circumstances, rather than add an override point that's going to result in copy/pasting code around when other jobs want to do this.

AzariahAeroborn (changes requested):
I'm kind of on the fence as far as the way the "business logic" of the window metadata is being scattered all over the place with this carve up, especially given some of the evaluators wind up checking if a higher priority suggestion overrides them (e.g. don't ding for ice mage if they died).   

Also, judging by this log: https://www.fflogs.com/reports/8bCKJNWFqTY2H9gc#fight=359&type=damage-done 
the changes seem more substantial than indicated for supporting no-F1 rotations.  Is the change in the rotation watchdog output table for this log expected?  Is the change large enough to merit expanding on the changelog message?

src/parser/jobs/blm/modules/RotationWatchdog/ExpectedFireSpellsEvaluator.tsx
AkairyuGestalter: Trying to do something like this I think would end up meaning deeper changes to the TrackAction interface and might brush against all the other jobs that use it. At this point I'd rather leave the one-off as a one-off and deal with the interface implications if and when someone else wants to do this, or at least as a separate PR from the rest of this.

src/parser/jobs/blm/modules/RotationWatchdog.tsx
AkairyuGestalter: It was also used locally for the Fire 4 count calculation. I dropped the export and passed it through to SkipB4 via constructor options instead.

src/parser/jobs/blm/modules/RotationWatchdog.tsx
AkairyuGestalter: Rather that mixing them in with the utilities, I moved the exported constants to a WatchdogConstants file that everything including the main module could import from.

AzariahAeroborn (approved):

PR #1839, opened 2023-07-08
BLU: Final Sting handling

Hugmeir:
`Final Sting` is one of BLU's "signature" moves --  At the cost of our lives, we do a significant chunk of damage.

The changes in this MR accomplish two things:

1. Deaths caused by `Final Sting` should not be counted as Deaths at all
2. `Final Sting` is normally buffed by using Moon Flute first; thus buff windows that end in a Final Sting are excempt from most requirements.

Also, note that all of this also applies to `Self-Destruct`.

### DIFF
--- a/src/parser/jobs/blu/modules/Death.tsx
+++ b/src/parser/jobs/blu/modules/Death.tsx
@@ -1,8 +1,12 @@
+import {Plural, Trans} from '@lingui/react'
+import {DataLink} from 'components/ui/DbLink'
 import {Event, Events} from 'event'
 import {filter, oneOf} from 'parser/core/filter'
 import {dependency} from 'parser/core/Injectable'
+import {Actor} from 'parser/core/modules/Actors'
 import {Data} from 'parser/core/modules/Data'
-import {Death} from 'parser/core/modules/Death'
+import {ActorDeathInfo, Death} from 'parser/core/modules/Death'
+import React from 'react'
 
 // Deaths from Final Sting and Self-Destruct should not count for
 // the report.
@@ -24,7 +28,10 @@ export class BLUDeath extends Death {
 			])), this.onFinalSting)
 	}
 
-	private onFinalSting() {
+	private actorStings: Map<Actor['id'], number> = new Map<Actor['id'], number>()
+	private onFinalSting(event: Events['action']) {
+		const stingerId = event.source
+		this.actorStings.set(stingerId, (this.actorStings.get(stingerId) ?? 0) + 1)
 		this.isFinalSting = true
 	}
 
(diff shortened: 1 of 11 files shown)


Hugmeir:
A sample log: https://www.fflogs.com/reports/yGdD6tr842KYMxVF#fight=2&source=8

src/parser/jobs/blu/modules/Death.tsx
```diff
@@ -0,0 +1,68 @@
...
+		if (actorStings === 0) {
+			super.deathSuggestionWhy(actorId, playerInfo)
+		}
+
+		return <Trans id="blu.deaths.why">
+			<Plural id="blu.deaths.why.death_count"
+				value={playerInfo.count}
+				_1="# death"
+				other="# deaths"
+			/>.  That number does not include the <Plural id="blu.deaths.why.final_stings"
```
AkairyuGestalter: This is really wordy. Can we do something like '<Plural [number of deaths] />, excluding those from Final Sting'? That should be enough to get the point across.

src/parser/jobs/blu/modules/MoonFlute.tsx
```diff
@@ -132,7 +138,27 @@ export class MoonFlute extends BuffWindow {
...
 		}))
 	}
+
+	private adjustExpectedGcdCount(window: HistoryEntry<EvaluatedAction[]>) {
+		const finalStingUsed = window.data.filter(event => (event.action.id === this.data.actions.FINAL_STING.id || event.action.id === this.data.actions.SELF_DESTRUCT.id)).length
+		return finalStingUsed >= 1 ? (-window.data.length+1) : 0
+	}
+
+	private adjustExpectedActionOutcome(window: HistoryEntry<EvaluatedAction[]>, _action: TrackedActionGroup) {
+		const finalStingUsed = window.data.filter(event => (event.action.id === this.data.actions.FINAL_STING.id || event.action.id === this.data.actions.SELF_DESTRUCT.id)).length
```
AkairyuGestalter: Minor nit: Since you're running the same logic in both adjust functions, you might consider extracting that to a helper function

AkairyuGestalter (changes requested):
Overall this looks fine, just two smaller requests. Also looks to be worth a pull from master and a re-run of yarn extract to get rid of the merge conflicts on the i18n files.

src/parser/jobs/blu/modules/Death.tsx
Hugmeir: Yeah, that wording sucks. Hm... I was trying to avoid "excluding X" because it makes it a bit ambigious for the reader -- is that telling me that the deaths are not being counted? Or that I should not count one of the deaths there because it came from Final Sting?

Latest commit has different wording, going from this:

    2 deaths. That number does not include the 1 death from using Final Sting

To this:

    2 deaths, not counting 1 death from using Final Sting.

src/parser/jobs/blu/modules/MoonFlute.tsx
Hugmeir: Done!

AkairyuGestalter (approved):

PR #1840, opened 2023-07-08
BLU: Simplify the White Death cast counter

Hugmeir:
The previous version was trying to exclusively track casts that happened when the Touch of Frost effect was active, but the way we were adding and removing the event hook ended up mis-counting actions in some edge cases.

So this just simplifies the entire thing, adding a hook during initialization for the White Death action.


Hugmeir:
Sample log: https://www.fflogs.com/reports/QNnLX27TH3d16YGj#fight=4&source=7

AkairyuGestalter (approved):

PR #1842, opened 2023-07-08
BLU: Overheal report should not consider Devour a heal

Hugmeir:
Devour is a 250 potency GCD that works roughly like WAR's Thrill of Battle -- it gives a buff that temporarily increases your max HP, and heals for the amount increased.

BLU tanks essentially pop this on cooldown.  BLU DPSes don't normally carry it, but if they do, they either use it on cooldown for a minor potency gain (30 or 40 potency vs our filler GCD) or hold it for specific mechanics (A8S Gavel).

BLU's Overheal report is special in that it only appears if the person used any non-White Wind heals, and the heal from Devour was causing a false positive.

### DIFF
--- a/src/parser/jobs/blu/changelog.tsx
+++ b/src/parser/jobs/blu/changelog.tsx
@@ -3,6 +3,11 @@ import React from 'react'
 
 export const changelog = [
 	{
+		date: new Date('2023-05-14'),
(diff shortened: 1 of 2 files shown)


Hugmeir:
Sample log: https://www.fflogs.com/reports/yGdD6tr842KYMxVF#fight=2&source=3

AkairyuGestalter (approved):

PR #1843, opened 2023-07-08
BLU ABC is special, let people know!

Hugmeir:
BLU's ABC report considers the following as uptime:
1. Time in Diamondback
2. Time in Waning Nocturne, the second half of Moon Flute
3. Time spent doing the Phantom Flurry channel
4. Time spent spamming Surpanakha

This MR includes an explanationf or this on the ABC report


AkairyuGestalter (changes requested):
Can you pull from master and re-run yarn extract to resolve the i18n merge issues? Might be worth only doing so for this and #1839 one at a time in case they decide to conflict with each other though...

AkairyuGestalter (approved):

PR #1844, opened 2023-07-09
Stacked Gauge timeline display and mark MNK as supported

AkairyuGestalter:
Update to ResourceGraphs to allow sets of gauges to show up as stacked rows instead of 0-based. The two gauges updated to use this (since they're both logically shared-maximums) are:
- Reaper's Death Gauge (Lemure and Void Shroud). Now much more closely mimics the in-game display, a definite plus 
image
- Monk's Beast Gauge (Opo, Raptor, and Coeurl) 
image
This also adds gauge tracking and display for Monk in general, since it wasn't previously in use


AkairyuGestalter:
With all the outstanding PR cleanup I was doing this evening, the gauge display is now the final required piece to fulfill issue 1303. There's still an outstanding question mark about requiring Rising Phoenix when Brotherhood is detected under Riddle of Fire, but since that's not a sure thing, I'm opting to additionally mark MNK as supported for Endwalker as of this PR.

src/parser/jobs/rpr/modules/DeathGauge.tsx
```diff
@@ -33,30 +32,31 @@ const VOID_ACTIONS: ActionKey[] = [
 	'LEMURES_SCYTHE',
 ]
 
+const LEMURE_SHROUD_COLOR = Color('#4ec8dc')
+const VOID_SHROUD_COLOR = Color('#b614c4')
```
ayyaruq, line 36: What was wrong with the existing colours?

src/parser/jobs/mnk/modules/PerfectBalance.tsx
```diff
@@ -27,36 +32,114 @@ const PB_BAD_ACTIONS: ActionKey[] = [
...
+	}))
+	private solarNadiGauge = this.add(new CounterGauge({
+		maximum: 1,
+		initialValue: 0,
+		graph: {
+			handle: NADI_GAUGE_HANDLE,
+			label: <Trans id="mnk.gauge.resource.nadi.solar">Solar Nadi</Trans>,
+			color: SOLAR_NADI_COLOR.fade(GAUGE_FADE),
+		},
+	}))
```
ayyaruq, line 112: These should be a boolean gauge, or history of bool states, not counters.

src/parser/jobs/mnk/modules/PerfectBalance.tsx
```diff
@@ -65,10 +148,50 @@ export class PerfectBalance extends Analyser {
...
+				this.solarNadiGauge.generate(1)
+			}
+		}
+		if (action.id === this.data.actions.RISING_PHOENIX.id) {
+			this.solarNadiGauge.generate(1)
+		}
+		if (action.id === this.data.actions.PHANTOM_RUSH.id) {
+			this.lunarNadiGauge.reset()
+			this.solarNadiGauge.reset()
 		}
```
ayyaruq: This would probably make more sense as a switch.

ayyaruq (dismissed):
Thanks for looking into this. The RPR changes fix a pretty annoying UI bug, I'm not sold on the MNK gauge implementation though. It's only used for the resource graphs but they don't really offer much info to MNK gameplay, just clutter in the timeline. Might be better to split this up into the RPR fixes to get that in and handle the MNK gauge separately.

Also before support is flagged, the PRs that got merged today will need to be fixed.

src/parser/jobs/rpr/modules/DeathGauge.tsx
AkairyuGestalter: It felt kludgy/fragile to be basing the colors off, in part, other jobs' defined color. They also felt... dull? I guess is the best word to describe it. These were color-picked from a screenshot of the actual gauge UI, similar to how I set up the Astral Fire/Umbral Ice/Feather/Addersting gauge colors.

src/parser/jobs/mnk/modules/PerfectBalance.tsx
AkairyuGestalter: BooleanGauge doesn't exist today, and it's not *wrong* per se, so they'll stay as they are. I have thoughts about how to polish both the nadis and the Beast gauge even further but that will be a separate future PR.

hintxiv: The only other "boolean" gauge in the game I can think of is BLM's paradox [which we also currently model with a CounterGauge](https://github.com/xivanalysis/xivanalysis/blob/626d6406f5ca35a8f824d0909e56e424280202a6/src/parser/jobs/blm/modules/Gauge.tsx#L148). Since it's a relatively large undertaking to create a new gauge type and it has such limited applicability, I don't think it's worth blocking on.

AkairyuGestalter: This is handled in #1868

src/parser/core/modules/ResourceGraphs/Graph.tsx
```diff
@@ -44,8 +44,8 @@ export function Graph({resource, scaleX}: GraphProps) {
 				data={data}
 				curve={(resource.linear ?? false) ? curveLinear : curveStepAfter}
 				x={datum => scaleX(datum.time)}
-				y0={scaleY(0)}
-				y1={datum => scaleY(datum.current)}
+				y0={datum => scaleY(datum.base || 0)}
+				y1={datum => scaleY((datum.base || 0) + datum.current)}
```
ackwell: given these are `T | undefined`, can write as `datum.base ?? 0`

i recognise that the only case that `||` would trip on here is `0` which is also the fall-through, but _semantics_

src/parser/core/modules/ResourceGraphs/ResourceGraphs.tsx
```diff
@@ -49,6 +51,11 @@ const DEFAULT_ROW_HEIGHT: number = 64
 export interface ResourceGroupOptions extends ResourceGraphOptions {
 	/** The handle for this group of data */
 	handle: string,
+	/** Should the group display it's gauges stacked on top of each other, instead of all starting from the 0 baseline (ie. RPR's Death Gauge to help it look like it does in-game)
+	 * 		NOTE: If using this option, all gauges within the resource group should have a constant and identical maximum or it's gonna look hella weird.
+	 * 			Also recommended to use the forceCollapsed option along with this (unless debugging) since that'll also look weird.
+	  */
```
ackwell: hmmmmm

this is effectively simulating a stacked graph, but with all of the timeline infrastructure getting in the way while you're doing it.

the library being used for graph rendering, visx, has [a utility for generating data pre-stacked](https://airbnb.io/visx/stacked-areas) - it'll probably take a bit of massaging to make it work (primarily around the clip path handling), but it'd be good to lean on.

bonus, if you put the stacked graph on the group timeline row, then in debug you can render the individual gauges to child rows and it won't have wonky baselines

ackwell (commented):
only partially reviewed, just wanted to flag the graph stuff before i pass out

src/parser/core/modules/ResourceGraphs/ResourceGraphs.tsx
AkairyuGestalter: I spent a... frustratingly long time trying to make the StackedArea work before falling back to this as a "dumb but it works" option. Willing to keep tinkering with things to see if I can figure out what it was I was too dumb to figure out at first tho.

src/parser/core/modules/ResourceGraphs/Graph.tsx
AkairyuGestalter: Always have trouble remembering which to do and why. Fixed.

Toastdeib (approved):
Gave this a look over with Akairyu, and given that it's been looked over by ack as well I'm gonna go ahead and approve it so we can finally get MNK stuff moved forward.

PR #1847, opened 2023-07-09
[Core] Weaving interrupt issue and updates

otocephaly:
Hiii
Thanks for reviewing ♥

The updates include:
- to only refresh weaving window in specific actor death and not any death in the party
- - this is in reference to conversation https://discord.com/channels/441414116914233364/725970639910993931/1122895772703273102
- an update to the look. this was done partially to match the other tables in the analysis but also to allow clicking on the timestamp to see what's up. Also I added the next action to allow for context (e.g. resurrection)
- added pullstart and pullend to allow suggestions for weaving at the start and end of the fight in the cases where the actor does not do an action close to either

Sample logs used: https://www.fflogs.com/reports/a:f6JKaMA4rjH9Y1n7#fight=17&source=3

Here is a preview of the new look:
image

(this is what it looked like before on the same logs):
image


src/parser/core/modules/Weaving.tsx
```diff
@@ -217,38 +247,57 @@ export class Weaving extends Analyser {
...
+								<Plural
+									id="core.weaving.panel-count"
+									value={item.weaves.length}
+									_1="# weave"
+									other="# weaves"
+								/>
+								&nbsp; - &nbsp;
+								{this.parser.formatDuration(item.gcdTimeDiff)}
+								&nbsp;
+								<Trans id="core.weaving.between-gcds">between GCDs</Trans>
```
AkairyuGestalter, line 288: Should we split up the weave count and the time between GCDs into two separate table columns? Food for thought...

AzariahAeroborn: I think I would rather remove the time after, and maybe change the Weave Info column to be actual/expected.  Even in the current accordion style, there's been a fair amount of confusion caused by the "time between GCDs" value, and I wonder if it's a net value add or not.

otocephaly: @AkairyuGestalter
originally I was thinking two columns, but I didn't wholly like the look
but I am also not wholly liking the one I posted above either since it lumps things together
I'm open to suggestions if you think there would be a nicer way to present the info
this is what it looks like as two columns (below screenshot)

@AzariahAeroborn
would it look better if we were descriptive in the actions instead of just saying gcds like this?

image

src/parser/core/modules/Weaving.tsx
```diff
@@ -59,10 +64,26 @@ export class Weaving extends Analyser {
 
 	protected severity = WEAVING_SEVERITY
 
+	//used to synth a start and end in case there were no actions before the pull or close to when the pull ended. it's also used to ensure that leading and trailing gcd events are not nullable
+	private pullStart: Events['action'] = {
+		action: 0,
```
otocephaly, line 69: fyi, this parameter is fixed to be a number
0 was chosen, but I'm sure any other number could work if we don't want specifically zero

otocephaly:
@ayyaruq I looked at a sub 1.96 gcd log and it seemed to be ok, but I am not sure if this example meets what you'd also be looking for
https://www.fflogs.com/reports/a:Dh3KJBnkzRL2wqGc#fight=last&source=4

src/parser/core/modules/Weaving.tsx
KelosPelmand: I like the columns this way, but not a fan of the gcd names in the third column personally, since it ends up showing the same info (which GCDs) twice.  Maybe that's just me knowing that "big skills = gcds, little skills = ogcds", which may or may not be obvious to the target audience tho.

AkairyuGestalter (approved):

PR #1850, opened 2023-07-12
[BugFix] Defensives - Handle shared CDs and fix extra use calculations

AkairyuGestalter:
Reported in https://xivanalysis.com/fflogs/a:MDpmzT4wAcYx1Bqr/20/3 - Shake It Off was displaying as having two additional uses at the end of the fight, when it should only have been reporting one. This was an unintended downstream effect of the cooldowns module adjusting the timestamps of the usage history, combined with this modules method for looking up the next time a charge of the cooldown was available.

Also noted in the same log was Bloodwhetting and Nascent Flash displaying in each other's output window, making it difficult to tell when each was actually used. Added some code to handle ensuring that the defensive used was exactly the one we're looking for in both the actual usage calculation, as well as the output display, while retaining the shared additional use calculations between them.

Additionally, we're now storing the cooldown and charge history of each defensive after retrieving it from the cooldowns module, instead of constantly calling back to it since that module has to recalculate it on the fly with each call. Huzzah performance.


src/parser/core/modules/Defensives.tsx
```diff
@@ -35,6 +35,11 @@ export class Defensives extends Analyser {
 	 * Implementing modules should provide a list of job-specific defensive actions to track
 	 */
 	protected trackedDefensives: Action[] = []
+
+	// Private lists of the cooldown usage and charge histories so we don't keep recalculating it via calls to the cooldowns module
+	private cooldownHistories: {[key: number]: CooldownHistoryEntry[]} = {}
+	private chargeHistories: {[key: number]: ChargeHistoryEntry[]} = {}
```
ackwell, line 41: caching these is technically fine, because this module doesn't generate anything until the rest of the parse is complete - but worth documenting as such.

if performance is a concern, i'd be inclined to move caching down to the source (i.e. the cooldowns module), as it allows cache _breaking_ to be colocated.

src/parser/core/modules/Defensives.tsx
```diff
@@ -53,11 +58,22 @@ export class Defensives extends Analyser {
 	}
 
 	private getUsageCount(defensive: Action): number {
-		return this.getUses(defensive).length
+		// The cooldowns module actually returns events based on cooldown *group* so make sure we're actually getting the uses for the cooldown we asked for
+		return this.getUses(defensive).filter(event => event.action.id === defensive.id).length
```
ackwell: is it safe for `getUses().length != getUsageCount()`? semantics seem a bit askew that "usage" means two different things here

src/parser/core/modules/Defensives.tsx
```diff
@@ -115,17 +131,22 @@ export class Defensives extends Analyser {
...
-						compact
-						size="mini"
-						icon="time"onClick={() => this.timeline.show(entry.start - this.parser.pull.timestamp, entry.end - this.parser.pull.timestamp)}>
-					</Button> {this.parser.formatEpochTimestamp(entry.start)}
-					</Trans>
-				</Table.Cell>
-			</Table.Row>
+			{
+				// Only create the usage rown if this history entry was for this cooldown, not another in the same cooldown group
+				entry.action.id === defensive.id ?
```
ackwell: nit; possibly worth a function for the jsx on this just so it doesn't all end up in a big ol' ternary

ackwell (approved):
lgtm with comments, feel free to merge as-is if you're happy with it

src/parser/core/modules/Defensives.tsx
AkairyuGestalter: Eh, probably not. BLU's override is using getUses().length since getUsageCount didn't exist yet so I should probably make sure those match now. Made a getGroupUses that functions like the original getUses, put the filter to the specific action in getUses, and changes this module's calls to getUses over to getGroupUses where they formerly expected to work off the full group.

src/parser/core/modules/Defensives.tsx
AkairyuGestalter: Cleaner, agreed.

src/parser/core/modules/Defensives.tsx
AkairyuGestalter: Since we're doing some filtering on the endReason prop, I think we can leave this here for now but it might not hurt to update the cooldowns module to give those options in the future.

PR #1852, opened 2023-07-12
BLU: Show a statistic for time spent under Diamondback

Hugmeir:
BLU's Diamondback is pretty busted: it gives you a 90% damage mitigation for 10 seconds, at the cost of being unable to do anything while the effect is active.

We use Diamondback to resolve certain mechanics, but ideally time spent under Diamondback should be zero, or as close to zero as we can get it, since every Diamondback means ~12 seconds of zero damaging GCDs, and it hard-clips the GCD after Diamondback expires.

EDIT: The MR is now adding an statistic, not a suggestion, see comments below for an updated screenshot.

So this PR adds a new MINOR Suggestion:
Screenshot 2023-07-12 at 12 51 53


src/parser/jobs/blu/modules/AlwaysBeCasting.tsx
```diff
@@ -143,6 +156,23 @@ export class AlwaysBeCasting extends CoreAlwaysBeCasting {
...
+		this.suggestions.add(new TieredSuggestion({
+			icon: this.data.actions.DIAMONDBACK.icon,
+			content: <Trans id="blu.diamondback.time.content">
+				The ABC report will count time spent under <DataLink action="DIAMONDBACK" /> as GCD uptime, but you should still aim to minimize this, since it means dropping damaging GCDs.
+			</Trans>,
+			why: <Trans id="blu.diamondback.time.why">
+				<Plural value={diamondBackSeconds ?? 0} one="# second" other="# seconds" /> spent in Diamondback.
+			</Trans>,
+			tiers: {1: SEVERITY.MINOR},
+			value: diamondBackSeconds,
```
AkairyuGestalter: I feel like this would make more sense as just a count of how many times Diamondback was used. You're always going to spend ten seconds unable to act per cast (unless you manage to die anyways I guess), so I'm not sure what value we're gaining by displaying it as time here.

src/parser/jobs/blu/modules/AlwaysBeCasting.tsx
```diff
@@ -143,6 +156,23 @@ export class AlwaysBeCasting extends CoreAlwaysBeCasting {
 	override onComplete() {
 		super.onComplete()
 
+		this.diamondBackHistory.closeCurrent(this.parser.pull.timestamp + this.parser.pull.duration)
+		const diamondBackTimeMs = this.diamondBackHistory.entries.reduce((acc, e) => {
+			return acc + ((e.end ?? e.start) - e.start)
+		}, 0)
+		const diamondBackSeconds = Math.ceil(diamondBackTimeMs / 1000)
+		this.suggestions.add(new TieredSuggestion({
```
AkairyuGestalter: Should this perhaps be a statistic instead of a suggestion? You can still give the context about wanting to minimize the number of uses in the statistic's Info property, and it might fare better as an at-a-glance "Did I overuse Diamondback this pull?" display that way.

AkairyuGestalter (commented):
Some food for thought that I'm ok with leaving the way it is if you'd prefer, but I wanted us to at least consider before merging this in.

src/parser/jobs/blu/modules/AlwaysBeCasting.tsx
xiashtra: There are boss actions that will knock you out of Diamondback early, so assuming 10s for every cast is not always true (e.g. O4S phase 2).

Hugmeir: Aye, A8S also has a similar "you get knocked out of Diamondback".  It's also fine to spend time in Diamondback during some downtime mechanics, though this patch isn't handling that

src/parser/jobs/blu/modules/AlwaysBeCasting.tsx
Hugmeir: I like that -- sometimes Diamondback is unavoidable, and it sucks to have an unactionable suggestion.  Haven't poked at the statistics code before, but how hard could it be?

Hugmeir: Screenshot 2023-07-14 at 14 07 11

Screenshot 2023-07-14 at 14 07 57


The recent push now has this as a statistic.

src/parser/jobs/blu/modules/AlwaysBeCasting.tsx
AkairyuGestalter: Aha, that wasn't something I was aware of. Fair enough then, we'll leave it as a time!

AkairyuGestalter (approved):

PR #1853, opened 2023-07-12
Core ActionTimeline: Option to resolve the label name late

Hugmeir:
This is really just groundwork for a QoL change in BLU; we have several abilities that share cooldowns, like Quasar and J Kick -- because of spell slot limitations, people will only take one or the other, so it's much nicer if the timeline shows the name of the ability they used.

Internally, this introduces a `lateResolveLabel` attribute to the action row config.  Normally, timeline labels are resolved immediately (during `initialise()`), so we cannot check which cooldowns have been used yet.
If `lateResolveLabel` is there and true, then when passing multiple alternative abilities for one row, it will wait until the last possible moment, then look at the cooldowns and find which one of the provided abilities was used.

You can see a sample of the end result in these two:
Screenshot 2023-07-12 at 13 12 34
Screenshot 2023-07-12 at 13 12 49

Note that J Kick / Quasar, Eruption / Feather Rain, Angel's Snack / Matra Magic each share cooldowns, but the timeline just shows whichever they first used.

### DIFF
--- a/src/parser/core/modules/ActionTimeline/ActionTimeline.tsx
+++ b/src/parser/core/modules/ActionTimeline/ActionTimeline.tsx
@@ -30,6 +30,11 @@ export interface ActionRowConfig {
 	label?: ReactNode
 	/** Order of the row within the timeline. */
 	order?: number
+	/** For multi-action rows, should we resolve the name of the action late?
+	*  This allows setting the name of the row to whatever actions was first used
+	*  of the set of possibilities.
+	* */
+	lateResolveLabel?: boolean
 }
 
 /** Configuration for a single row. */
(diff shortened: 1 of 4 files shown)


AkairyuGestalter:
Can you go ahead and include the BLU ActionTimeline change that this is meant to support? It's hard to tell how useful this actually is/whether it's implemented in a good way without getting to see the full intended end result in action.

Hugmeir:
Sure, the PR now includes the changes to BLU's ActionTimeline.

A sample log:
DPS (shows Matra Magic, J Kick, Feather Rain)
https://www.fflogs.com/reports/yGdD6tr842KYMxVF#fight=20&source=1

Healer (shows Angel's Snack, Quasar, Eruption)
https://www.fflogs.com/reports/yGdD6tr842KYMxVF#fight=20&source=2

Tank (shows Chelonian Gate, Dragon Force)
https://www.fflogs.com/reports/yGdD6tr842KYMxVF#fight=20&source=3

src/parser/core/modules/ActionTimeline/ActionTimeline.tsx
```diff
@@ -82,13 +87,36 @@ export class ActionTimeline extends Analyser {
...
+		for (const maybeAction of config.content) {
+			let fullAction = undefined
+			if (typeof maybeAction === 'object') {
+				fullAction = maybeAction
+			} else if (typeof maybeAction === 'number') {
+				fullAction = this.data.getAction(maybeAction)
+			} else if (typeof maybeAction === 'string') {
+				fullAction = this.data.actions[maybeAction as ActionKey]
+			}
+			if (fullAction === undefined || fullAction === null) { continue }
```
AkairyuGestalter: Minor nit on this one: this is the one case where the == sign is ok to use. ```fullAction == null``` will do a check for null or undefined so you don't need to explicitly look for both

AkairyuGestalter (changes requested):
One minor nitpick. This makes a lot more sense now that I can see the whole thing :)

src/parser/core/modules/ActionTimeline/ActionTimeline.tsx
Hugmeir: TIL, that makes sense! Patched.

AkairyuGestalter (approved):

PR #1856, opened 2023-07-14
BLU: Suggest people turn Mighty Guard off

Hugmeir:
Well, sort of.  This adds a MINOR severity suggestion that shows how many damaging GCDs non-tanks did while under MG.

Ideally that should be zero, but occasionally you need MG to survive mechanics.

NOTE: BLU tank detection requires handling prepull auras, which is separately done here: https://github.com/xivanalysis/xivanalysis/pull/1813
This can be merged without those changes, it just means that BLU tanks will get a false-positive suggestion.


src/parser/jobs/blu/modules/MightyGuardGCDing.tsx
```diff
@@ -0,0 +1,90 @@
...
+		const gcds = this.damagingActionsUnderMG
+		this.suggestions.add(new TieredSuggestion({
+			icon: this.data.actions.MIGHTY_GUARD.icon,
+			content: <Trans id="blu.mighty_guard.gcds.content">
+				<DataLink action="MIGHTY_GUARD"/> comes with a significant damage penalty. If you don't need the damage mitigation to survive a mechanic, turn it off!
+			</Trans>,
+			tiers: {1: SEVERITY.MINOR},
+			value: gcds,
+			why: <Trans id="blu.mighty_guard.gcds.why">
+				<Plural value={gcds} one="# damaging GCD" other="# damaging GCDs" /> were cast under <DataLink action="MIGHTY_GUARD" showIcon={false} />
```
AkairyuGestalter: minor nit: 'were' should be included in the Plural's one/other properties (and be 'was' in the one prop) for full grammatical correctness

AkairyuGestalter (approved):

PR #1858, opened 2023-07-14
BLU: Triple Trident and SpS need special handling

Hugmeir:
The standard burst for BLU has us using Triple Trident under Moon Flute, but the story is a bit more complicated than that.

For Spell Speed builds, using Triple Trident on cooldown is a (small!) DPS gain over holding it for the Moon Flute window.

This commit adds handling for this case -- If TT was off cooldown during the burst window, either because this is the opener or because it re-aligned itself, then the report requires it to be used. If it was on cooldown, we don't dock them points for not being able to use it.

Sample log:
https://www.fflogs.com/reports/QNnLX27TH3d16YGj#fight=4&source=7
Particularly, the burst at 2:08 cannot use TT, since it was used on cooldown earlier, at around ~1:27

Screenshot 2023-07-14 at 20 00 24


AkairyuGestalter (approved):
I approve of spell speed builds :)

PR #1859, opened 2023-07-15
BLU: Weaving report fixes

Hugmeir:
Three closely-related changes here:

1. Several BLU actions were defined wrong, with `castTime: 2500` instead of `castTime: 2000, gcdRecast: 2500`, which was mistakenly triggering the Weaving report.
2. The core Weaving module allows a single weave after a 2s cast time GCD, but that always results in a hard clip -- Unsure if this is a bug in the core module, I just now brought it up on the Discord, but in the meanwhile, patching BLU to disallow weaves on any spell with a cast time higher than 1.5s
3. ...except we do allow & expect people to clip 2s cast time spells during Moon Flute windows, so this adds an exception for that case


Hugmeir:
Forgot to include an example log:
https://www.fflogs.com/reports/QNnLX27TH3d16YGj#fight=4&source=7&view=timeline&type=casts&start=845000&end=856723

Here, at ~02:55, we have the sequence Triple Trident (Feather Rain) => The Rose of Destruction, which is clipping the second GCD since Triple Trident has a 2s cast time; you can also see a similar example at ~01:25, with Tingle (Swifcast) => Triple Trident, since Tingle has a 2s cast time too.

AkairyuGestalter (approved):

PR #1860, opened 2023-07-18
Patch definition for 6.45

AkairyuGestalter:
Since we actually have BLU analysis now (yay!), we're going to need to be able to split between pre-6.45 level 70 cap spell expectations and 6.45+ level 80 cap spells. PR adds 6.45 to the patch list and creates empty action and status layer files

### DIFF
--- a/src/parser/jobs/sch/index.js
+++ b/src/parser/jobs/sch/index.js
@@ -18,7 +18,7 @@ export const SCHOLAR = new Meta({
 	</>,
 	supportedPatches: {
 		from: '6.0',
-		to: '6.4',
+		to: '6.45',
 	},
 	contributors: [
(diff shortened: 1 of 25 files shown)


AkairyuGestalter:
One new UnableToAct status added by yarn generate. Went ahead and saved other job maintainers some busywork by updating job support patch levels, since only BLU has job changes this patch.

src/parser/jobs/sch/index.js
```diff
@@ -18,7 +18,7 @@ export const SCHOLAR = new Meta({
 	</>,
 	supportedPatches: {
 		from: '6.0',
-		to: '6.4',
+		to: '6.45',
```
nonowazu, line 21: I know you don't need my explicit approval, but I :+1: this

nonowazu (approved):

ackwell (approved):

PR #1863, opened 2023-07-19
BLU: New spells added in 6.45

Hugmeir:
The only change to existing spells:

    * Magic Hammer shares a cooldown with the newly-added Candy Cane
    * The Rose of Destruction & Chelonian Gate, which shared a
      cooldown, now also share it with the newly added Ruby Dynamics

Since those changes would only matter for logs generated after 6.45, I think it's safe to just add these without messing with the layers.

NOTE: Winged Reprobation is *weird* and it's actual cooldown is not going to be properly tracked yet.  That will be handled in a future commit.


src/data/ACTIONS/root/BLU.ts
```diff
@@ -14,6 +14,8 @@ export const BLU_COOLDOWN_GROUP = {
 	NIGHTBLOOM: 23290,
 	THE_ROSE_OF_DESTRUCTION: 23275,
 	MATRA_MAGIC: 23285,
+	MAGIC_HAMMER: 18305,
```
AkairyuGestalter, line 17: Minor nitpick I didn't pick up on before. Can we call this const BLU_COOLDOWN_GROUPS to better reflect what it's defining?

src/data/ACTIONS/root/BLU.ts
```diff
@@ -1008,4 +1011,196 @@ export const BLU = ensureActions({
 		statusesApplied: ['BLEEDING'],
 		damageType: MAGICAL,
 	},
+	GOBLIN_PUNCH: {
+		id: 34563,
+		name: 'Goblin Punch',
+		icon: 'https://xivapi.com/i/003000/003358.png',
+		onGcd: true,
+		gcdRecast: 2500,
```
AkairyuGestalter, line 1116: Should we go ahead and tag the GCDs (or otherwise? BLU's weird) with the speedAttribute: Attributes.SPELL_SPEED property to make sure the timeline display respects the player's speed?

AkairyuGestalter (commented):
Two comments not going to block on either if you'd rather not change 'em though. It looks like they might've double-posted though... I'll try to clean that up... [edit] Nope, I guess they actually only showed once after I refreshed the page .-.

src/data/ACTIONS/root/BLU.ts
Hugmeir: I think this is a good idea, though I'd rather do it in a separate PR since I'll need to go and check every spell, since SOME of our oGCDs are also affected by SpS.

(also, to be honest... I play with zero SpS both as BLU and RDM so I didn't even realize this was an option)

src/data/ACTIONS/root/BLU.ts
Hugmeir: Done!

src/data/ACTIONS/root/BLU.ts
Hugmeir: Well, that actually only took a couple of minutes to verify in-game, I didn't remember you could just see the tooltip of spells in the spell book without slotting them into your actions.  PR now includes `speedAttribute: Attributes.SPELL_SPEED` for everything that's affected, including three weird edge cases (White Death and Divine Cataract are affected, Phantom Flurry's kick is not)

AkairyuGestalter (approved):
Looks good to me!

PR #1864, opened 2023-07-20
BLU: Winged Reprobation cooldown fun!

Hugmeir:
Winged Reprobation (spell #118) works weird.  The description says it is a 1s cast time, 90s cooldown GCD.

But when you cast it, you get a stack of a buff. When you get that buff, it resets the cooldown of Winged Reprobation, up to 4 stacks -- when you get 4 stacks, you lose the buff and the cooldown actually happens.

So if we're getting the Winged Reprobation buff, and it's not the 4-stack version, then reset the skill's cooldown.

NOTE: This also includes the changes in https://github.com/xivanalysis/xivanalysis/pull/1863 since that's what actually introduces the spell.

### DIFF
--- a/src/parser/jobs/blu/modules/GeneralCDDowntime.ts
+++ b/src/parser/jobs/blu/modules/GeneralCDDowntime.ts
@@ -1,8 +1,26 @@
+import {Events} from 'event'
+import {dependency} from 'parser/core/Injectable'
+import {Actors} from 'parser/core/modules/Actors'
 import {CooldownDowntime} from 'parser/core/modules/CooldownDowntime'
 
 const DEFAULT_ALLOWED_DOWNTIME = 1000
+const MAX_WINGED_STACKS_BEFORE_CD = 3
 export class GeneralCDDowntime extends CooldownDowntime {
 	override defaultAllowedAverageDowntime = DEFAULT_ALLOWED_DOWNTIME
+
+	@dependency private actors!: Actors
+
+	override countUsage(e: Events['action']): boolean {
+		const actionId = e.action
+		if (actionId !== this.data.actions.WINGED_REPROBATION.id) {
+			return true
+		}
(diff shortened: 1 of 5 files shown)


src/parser/jobs/blu/modules/GeneralCDDowntime.ts
```diff
@@ -1,9 +1,27 @@
...
 	override defaultAllowedAverageDowntime = DEFAULT_ALLOWED_DOWNTIME
 
+	@dependency private actors!: Actors
+
+	override countUsage(e: Events['action']): boolean {
+		const actionId = e.action
+		if (actionId !== this.data.actions.WINGED_REPROBATION.id) {
+			return true
+		}
+		const wingedReprobationStacks = this.actors.current.getStatusData(this.data.statuses.WINGED_REPROBATION.id) ?? 0
```
AkairyuGestalter, line 18: Do we need this countUsage override if we're not tracking Winged Reprobation? Alternatively, should we be tracking it? If the latter, that's definitely going to need patch level support (and can be a later PR once we get someone to merge the one adding the patch information...)

AkairyuGestalter (changes requested):
One comment that we can take or leave, and we ought to start having changelogs for BLU stuff, now that we're nicely caught up and working on post-level cap increase changes.

src/parser/jobs/blu/modules/GeneralCDDowntime.ts
Hugmeir: It should be tracked, but Winged is also optional for some new types of openers so I need to do more work before we start telling people about missing uses.

AkairyuGestalter: Fair enough. Since we're planning to track it eventually, I'm fine with leaving the code in.

Hugmeir: Rebased the branch & included a changelog entry!

AkairyuGestalter (approved):

PR #1865, opened 2023-07-22
[BugFix] RaidBuffWindow window treating idle Carbuncle as an enemy

AkairyuGestalter:
If a SMN does not use Radiant Aegis at all during a fight, fflogs will not treat their Carbuncle as a pet since no actions originate from it. It will adapt into Xiva as an 'unknown' actor, which has an unfortunate interaction with the RaidBuffWindow extension of the base BuffWindow.

BuffWindow tracks status applications by any actor that the player controls (the player themselves or a pet of the player, to handle early Endwalker Carbuncle-sourced Searing Light), targeting the player (for buffs to themselves) or any enemy actor (for things like Trick Attack/Mug).  The 'enemy' target list was being built by looking for actors that didn't match the player's 'team' meaning that it does include 'unknown' team actors like the idle Carbuncle in this situation. The net result was that RaidBuffWindow was detecting the status application of the tracked raid buff to the Carbuncle by way of mirroring from the owning Summoner, which caused it to incorrectly close/open windows in unexpected ways.

Since no actor a player could reasonably apply intentionally-enemy-targeted buffs like Trick Attack or Mug to should wind up with the 'Unknown' actor team, we can safely ignore them when building the potential targets list for all BuffWindows, thus preventing idle Carbuncles from breaking RaidBuffWindow's analysis and display.

See: https://xivanalysis.com/fflogs/VtnRyYMJ9Gr4PXDq/5/1


src/parser/core/modules/ActionWindow/windows/BuffWindow.ts
```diff
@@ -47,7 +47,7 @@ export abstract class BuffWindow extends ActionWindow {
 		// or on an enemy actor (trick attack)
 		// enemies are not on the same team as the parser actor
 		const enemyTargets = this.parser.pull.actors
-			.filter(actor => actor.team !== this.parser.actor.team)
+			.filter(actor => actor.team > this.parser.actor.team) // Ignore actors with an 'unknown' team (team 0, where player's are team 1: Friend)
```
ackwell: i know it's not as much of a one-line-er, but i'm not a huge fan of treating enums as numbers - would be more self-descriptive to just check `&&!=unknown`

AkairyuGestalter: I can live with that. Done!

src/parser/core/changelog.tsx
```diff
@@ -23,6 +23,13 @@ export const changelog: ChangelogEntry[] = [
 		</>,
 		contributors: [CONTRIBUTORS.HINT, CONTRIBUTORS.AYA],
 	},
+	{
+		date: new Date('2023-07-21'),
+		Changes: () => <>
+			Fixed a bug that was breaking raid buff window detection for players in a party with a Summoner that did not use <DataLink action="RADIANT_AEGIS" showIcon={false} />.
```
hintxiv, line 29: I know we're not wildly consistent on this but I think all the other Core changelogs are written in present tense (with one exception directly below this particular change).

hintxiv (approved):
Looks good to me. You can adjust the changelog wording or not, up to you.

src/parser/core/changelog.tsx
AkairyuGestalter: We're kind of all over the place in Core too, some of the older ones have past tense as well. We'll leave it for now.

PR #1866, opened 2023-07-22
[BugFix] Correct Masterful Blitz count evaluation in Riddle of Fire

AkairyuGestalter:
The BlitzEvaluator's calculation for how many Masterful Blitz actions were expected to be used in a Riddle of Fire window was not properly taking into account whether the player was already in a Perfect Balance window when Riddle of Fire was activated. Created a function in the Perfect Balance module to calculate that based on either having the status or having contents in the beast Chakra gauge. Also cleaned up the available charge calculations by using the cooldowns module's charge history instead of trying to replicate it.

Note that this builds on top of the changes added in PR #1844 so approving this one will also pull through the changes from that one. Approve/merge that one first if you want to see the minimal set of changes this PR adds.


src/parser/jobs/mnk/modules/evaluators/BlitzEvaluator.tsx
```diff
@@ -119,16 +117,14 @@ export class BlitzEvaluator implements WindowEvaluator {
...
 
-		const castsAvailable = this.pbCharges - Math.min(this.pbCharges, priorCasts.length - priorRegens.length)
+		// If we entered the window with a PB already active and have enough time to use it before the window ends, expect that to get used too
+		const castsAvailable = chargesBeforeWindow + chargesInWindow + (this.perfectBalance.inBalance(window.start) && window.start + END_OF_WINDOW_TOLERANCE < (window.end ?? this.pullEnd) ? 1 : 0)
 
-		// Even though it is possible to have a triple blitz window
-		// we do not recommend it as a target since it is an advanced
-		// optimization that requires specific types of downtime
-		return Math.min(this.pbCharges, castsAvailable)
+		// If we're going into RoF with a blitz prepped, we can fit 3 blitzes in 11 GCDs (or at least one, if we're a short window at the end of the fight)
```
Toastdeib, line 136: Nit - can you confirm whether this comment is accurate? It says 3 but the `MAX_EXPECTED_BLITZES` const is 2.

src/parser/jobs/mnk/modules/evaluators/BlitzEvaluator.tsx
```diff
@@ -100,17 +107,8 @@ export class BlitzEvaluator implements WindowEvaluator {
...
-				isOnCooldown = this.pbCasts[i] < this.rechargeTimes[i-1]
-			}
-
-			if (!isOnCooldown) {
-				this.rechargeTimes[i] = this.pbCasts[i] + this.pbCooldown
-			} else {
-				this.rechargeTimes[i] = this.rechargeTimes[i-1] + this.pbCooldown
-			}
+		if (this.pbChargeHistory.length === 0) {
+			this.pbChargeHistory = this.cooldowns.chargeHistory('PERFECT_BALANCE')
```
Toastdeib: I *think* this should work with `this.cooldowns.chargeHistory(this.data.actions.PERFECT_BALANCE)` if I'm reading the TS correctly in the Cooldowns module, which (at least in my book) would be cleaner and safer than using the string key. That said, I do see some precedent elsewhere in these files for using a string instead of the action, so you're welcome to take this with a grain of salt if it's a standard practice.

Toastdeib (commented):
I won't pretend to understand all of the math, but overall this looks fine to my untrained eye. I left a couple of small comments but I'm happy to approve and merge this once they're addressed as long as no one else objects.

src/parser/jobs/mnk/modules/evaluators/BlitzEvaluator.tsx
AkairyuGestalter: Three is possible in the case where a blitz is prepared ahead of time, but the logic to try to reflect that was bad. Added a specific const for that case and polished the returns to make it more readable and accurate.

src/parser/jobs/mnk/modules/evaluators/BlitzEvaluator.tsx
AkairyuGestalter: Yeah, I like that better. Have to pass it through from the RoF analyser since evaluators don't have direct access to `data` but still.

src/parser/jobs/mnk/modules/PerfectBalance.tsx
```diff
@@ -148,7 +148,7 @@ export class PerfectBalance extends Gauge {
 	// Any of the chakra gauges have values in them, or
 	// There is a history window that contains the specified timestamp
 	public inBalance(timestamp: number): boolean {
-		const latestHistory = this.history.find(entry => entry.start <= timestamp && entry.start + this.data.statuses.PERFECT_BALANCE.duration < timestamp)
+		const latestHistory = this.history.find(entry => entry.start <= timestamp && entry.start + this.data.statuses.PERFECT_BALANCE.duration > timestamp)
```
Toastdeib, line 151: Oop, good catch - I didn't spot this one in my initial review.

src/parser/jobs/mnk/modules/evaluators/BlitzEvaluator.tsx
```diff
@@ -108,7 +114,7 @@ export class BlitzEvaluator implements WindowEvaluator {
 
 	private calculateRechargeTimes(): void {
 		if (this.pbChargeHistory.length === 0) {
-			this.pbChargeHistory = this.cooldowns.chargeHistory('PERFECT_BALANCE')
+			this.pbChargeHistory = this.cooldowns.chargeHistory(this.pbAction)
```
Toastdeib, line 117: 👍

Toastdeib (approved):
I think this is good to go now; I'll hold off on merging until this evening just in case anyone else wants to weigh in but the fixes all seem solid (and I even put in the effort to understand the math a bit).

PR #1867, opened 2023-07-22
Feature - Enum Gauge

AkairyuGestalter:
This PF builds off of the naive gauge stacking functionality introduced by #1844 to provide a better internal API for dealing with gauges that have a set number of slots that can be filled with any number of a limited set of values. Includes some timeline and tooltip display enhancements for empty values. Gauges displaying timeline content using this new type are:
- Monk: Beast Chakra
- Astrologian: Astrosigns
- Reaper: Death Gauge
- Black Mage: Astral Fire and Umbral Ice

I also made a few timeline ordering tweaks to BLM to coincide with the gauge display update while I was at it.

~~Branch contains the changes also made in PRs #1844 and #1866. Merging this one will also merge those changes in. If you want to see the minimal set of changes introduced by this PR, merge those first.~~

PR #1868, opened 2023-07-22
Feature - Set Gauge

AkairyuGestalter:
This PF builds off of the naive gauge stacking functionality introduced by #1844 to provide a better internal API for dealing with gauges that have a set number of slots that can each only be filled by a specific value. Includes some timeline and tooltip display enhancements for empty values. Gauges displaying timeline content using this new type are:
- Monk: Lunar and Solar Nadis
- Samurai: Sen
- Bard: Codas

~~Branch contains the changes also made in PRs #1844 and #1866. Merging this one will also merge those changes in. If you want to see the minimal set of changes introduced by this PR, merge those first.~~


AkairyuGestalter:
Note that this PR will have merge conflicts with #1867. Take care when merging or get a hold of me to clean the other up after merging one of the two.

src/parser/jobs/mnk/modules/PerfectBalance.tsx
```diff
@@ -153,5 +288,22 @@ export class PerfectBalance extends Analyser {
...
+		this.suggestions.add(new TieredSuggestion({
+			icon: this.data.actions.PHANTOM_RUSH.icon,
+			content: <Trans id="mnk.pb.suggestions.nadi-overcap.content">
+				Generating a Lunar or Solar Nadi while already in possession of that Nadi means fewer uses of <DataLink action="PHANTOM_RUSH" /> over the course of the fight. Try not to overcap either of your Nadis.
+			</Trans>,
+			tiers: {
+				1: SEVERITY.MINOR,
+				2: SEVERITY.MEDIUM,
+				3: SEVERITY.MAJOR,
+			},
```
hintxiv: Overcapping 1 Nadi is a fairly standard rotation adjustment (e.g. "double solar opener", "double lunar opener") to shift Phantom Rush into 2 min windows in cases where you won't lose a use of it. The nuanced solution would be to actually calculate whether overcapping lost you a use, but that might be beyond the scope of this PR.

For now I think that either the suggestion should start at 2+ overcaps, or the wording should be adjusted to reflect that you aren't guaranteed to lose a use of Phantom Rush by overcapping once (it's only a 1 in 3 chance that you'd lose a use after all).

src/parser/jobs/mnk/modules/evaluators/BlitzEvaluator.tsx
```diff
@@ -119,16 +114,16 @@ export class BlitzEvaluator implements WindowEvaluator {
 	}
 
 	private expectedBlitzes(window: HistoryEntry<EvaluatedAction[]>): number {
-		const priorCasts = this.pbCasts.filter(time => time <= window.start)
-		const priorRegens = this.rechargeTimes.filter(time =>
-			time <= (window.end ?? window.start) - END_OF_WINDOW_TOLERANCE
-		)
+		// We expect the player to burn off PB charges that were available when the window began, and any that became available with enough time before the window ended
+		const chargesAvailable = (_.last(this.pbChargeHistory.filter(entry => entry.timestamp < window.start))?.current ?? this.maxCharges) +
+			(this.pbChargeHistory.some(entry => entry.timestamp > window.start && entry.timestamp <= (window.end ?? window.start) - END_OF_WINDOW_TOLERANCE) ? 1 : 0)
```
hintxiv: This one-liner is doing a LOT of work and it's a little hard to read, could it be extracted into its own function?

src/parser/jobs/mnk/modules/evaluators/BlitzEvaluator.tsx
```diff
@@ -119,16 +114,16 @@ export class BlitzEvaluator implements WindowEvaluator {
...
 
-		const castsAvailable = this.pbCharges - Math.min(this.pbCharges, priorCasts.length - priorRegens.length)
+		// If we entered the window with a PB already active, expect that to get used too
+		const castsAvailable = chargesAvailable + (this.perfectBalance.inBalance(window.start) ? 1 : 0)
 
 		// Even though it is possible to have a triple blitz window
 		// we do not recommend it as a target since it is an advanced
 		// optimization that requires specific types of downtime
-		return Math.min(this.pbCharges, castsAvailable)
+		return Math.min(castsAvailable, MAX_EXPECTED_BLITZES)
```
hintxiv: I think we can be a bit more precise to accommodate triple blitz windows -- even though it's a niche optimization, the logic is pretty straightforward and the player would have to be doing some very weird things with their rotation in order for this to return 3.

```typescript
const castsAvailable = chargesAvailable + (this.perfectBalance.inBalance(window.start) ? 1 : 0)
const blitzReady = this.perfectBalance.functionThatReturnsTrueIfYouHaveThreeBeastChakra

// If we're going into RoF with a blitz prepped, we can fit 3 blitzes in 11 GCDs
return blitzReady ? castsAvailable : Math.min(castsAvailable, MAX_EXPECTED_BLITZES)
```

hintxiv (commented):
Amazing! Hopefully I'm putting these comments on the right PR.

src/parser/jobs/mnk/modules/evaluators/BlitzEvaluator.tsx
hintxiv: There seems to be a bug with end of fight tolerance here. In this situation my PB charge came up ~1 GCD before the fight ended, but it expected a blitz:

image

Live version:

image

https://xivanalysis.com/fflogs/FwZtxbynRKVTP94g/1/8

src/parser/jobs/mnk/modules/PerfectBalance.tsx
```diff
@@ -27,48 +36,164 @@ const PB_BAD_ACTIONS: ActionKey[] = [
...
+
+	private opoBeastGauge = this.add(new CounterGauge({
+		maximum: 3,
+		initialValue: 0,
+		graph: {
+			handle: BEAST_GAUGE_HANDLE,
+			label: <Trans id="mnk.gauge.resource.beast.opo">Opo-opo</Trans>,
+			color: OPO_GAUGE_COLOR.fade(GAUGE_FADE),
+		},
+	}))
```
hintxiv: Not sure exactly where or why this happens but the Beast Gauge resource group seems to default to "Opo-opo" as its title (because it's the first gauge added maybe?):

image

Ideally this would be called "Beast Chakra" or something instead.

AkairyuGestalter: Hm, it should have been labeled that way, I probably goofed something in the un-stashing/detangling from the Enum gauge PR (which definitely will fix this) but I'll get that corrected in case this one gets merged first

src/parser/jobs/mnk/modules/PerfectBalance.tsx
AkairyuGestalter: bumped up the severity tiers and left a note that we should try to evaluate whether the overcap mattered

src/parser/jobs/mnk/modules/evaluators/BlitzEvaluator.tsx
AkairyuGestalter: I didn't pull it out to a function, but I did turn it into two const assignments with useful names as part of figuring out the end of fight leniency adjustment

src/parser/jobs/mnk/modules/evaluators/BlitzEvaluator.tsx
AkairyuGestalter: Seems sane to me, done.

src/parser/jobs/mnk/modules/evaluators/BlitzEvaluator.tsx
AkairyuGestalter: Should be fixed now!

src/parser/jobs/mnk/modules/PerfectBalance.tsx
AkairyuGestalter: Yep, lost the resource group definition somewhere along the way. Fixed now.

src/parser/jobs/mnk/modules/PerfectBalance.tsx
```diff
@@ -303,5 +302,23 @@ export class PerfectBalance extends Gauge {
...
+		this.suggestions.add(new TieredSuggestion({
+			icon: this.data.actions.PHANTOM_RUSH.icon,
+			content: <Trans id="mnk.pb.suggestions.nadi-overcap.content">
+				Generating a Lunar or Solar Nadi while already in possession of that Nadi means fewer uses of <DataLink action="PHANTOM_RUSH" /> over the course of the fight. Try not to overcap either of your Nadis.
+			</Trans>,
+			tiers: {
+				2: SEVERITY.MINOR, // Start at 2 since 1 might be expected depending on opener
+				3: SEVERITY.MEDIUM,
+				4: SEVERITY.MAJOR,
+			},
```
AzariahAeroborn: This severity seems too low.  Losing a Phantom Rush would be more potency loss than overwriting a Polyglot (and more potency loss than using 5 Celestial Revolutions to avoid doubling up), but we score both of those other mistakes as always MAJOR.  If we need to forgive one in the opener, can we implement a start of fight forgiveness for overcap and then otherwise flag any overcap as bad?

See this log for one where the overcap definitely lost a usage - looks like went for an opener with double Solar but then did two more PBs before getting to Phantom Rush http://localhost:3000/fflogs/GjQ1RAVaD4cWk6Xv/1/49

AzariahAeroborn (changes requested):

src/parser/jobs/mnk/modules/PerfectBalance.tsx
AkairyuGestalter: I'm ok with changing 2 to Medium and 3 to Major for this PR. Trying to add a fight timing forgivenness factor and/or the noted "figure out if the overcap actually lost them a Phantom Rush usage" changes I think can wait for another PR and/or a 7.0 change.

AkairyuGestalter: Changed to MAJOR at overcap >=2 based on Discord discussion

AzariahAeroborn (approved):

PR #1870, opened 2023-07-22
BLU 6.45: Apokalypsis

Hugmeir:
BLU got a new channelled ability in 6.45, `Apokalypsis` -- this behaves pretty much exactly as the `Phantom Flurry` ability we already had, but it doesn't have the optional ending kick GCD, so it's just a 10 second channel with no extra fuss.

This PR first simplifies the tracking of `Phantom Flurry` a fair bit, making it generic enough that adding Apokalypsis support just meant modifying three lines.

Like with `Phantom Flurry`, it also adds a suggestion in case people drop ticks of the `Apokalypsis` channel by moving too early.

And since it was too minor to make it into a PR of it own, this also patches an issue in the ColdFog module.  The issue was not affecting anything, it just meant that the module was tracking the end time of the Cold Fog buff wrong, but that data is never used.


src/parser/jobs/blu/modules/AlwaysBeCasting.tsx
```diff
@@ -23,18 +24,19 @@ const SURPANAKHA_ANIMATION_LOCK_MS = 1000
 // This also removes the time under Waning Nocturne (the second half of Moon Flute)
 // from the ABC report. Can't cast, am waning.
 
-interface PhantomFlurryWindow {
-	start: number
-	end: number
+interface channelWindow {
```
AkairyuGestalter: Style nit: should be ChannelWindow

src/parser/jobs/blu/modules/AlwaysBeCasting.tsx
```diff
@@ -59,26 +61,26 @@ export class AlwaysBeCasting extends CoreAlwaysBeCasting {
 		super.initialise()
 
 		const playerFilter = filter<Event>().source(this.parser.actor.id)
-		const phantomFlurryCastFilter = playerFilter
-			.action(this.data.actions.PHANTOM_FLURRY.id)
+
+		const channelCastFilter = playerFilter
+			.action(oneOf([this.data.actions.PHANTOM_FLURRY.id, this.data.actions.APOKALYPSIS.id]))
```
AkairyuGestalter, line 66: Since were' building this list multiple times, can we make a `private channelCastIds` or something similar?

src/parser/jobs/blu/modules/AlwaysBeCasting.tsx
```diff
@@ -105,74 +107,60 @@ export class AlwaysBeCasting extends CoreAlwaysBeCasting {
...
+		this.channelHistory.closeCurrent(event.timestamp)
 	}
 
 	override considerCast(action: Action, castStart: number): boolean {
 		if (action === this.data.actions.PHANTOM_FLURRY) {
 			this.debug(`Phantom Flurry began channeling at ${this.parser.formatEpochTimestamp(castStart)}`)
 			return false
 		}
+		if (action === this.data.actions.APOKALYPSIS) {
+			this.debug(`Apokalypsis began channeling at ${this.parser.formatEpochTimestamp(castStart)}`)
```
AkairyuGestalter: If you do the `channelCastIds` suggestion, you can merge this condistion with the other as `this.channelCastIds.includes(action.id)` then just do `action.name + ' began channeling at`... for your message, rather than needing to duplicate the code for each channel we eventually have.

AkairyuGestalter (changes requested):
A couple code style cleanup comments, but otherwise this looks good.

src/parser/jobs/blu/modules/AlwaysBeCasting.tsx
Hugmeir: One of them is an action list, the other is a status list -- And yeah, annoyingly Phantom Flurry gives you a status called Phantom Flurry, and Apokalypsis gives you a status called Apokalypsis. It's not confusing at all, squeenix >:/

src/parser/jobs/blu/modules/AlwaysBeCasting.tsx
Hugmeir: Patched!

src/parser/jobs/blu/modules/AlwaysBeCasting.tsx
Hugmeir: Didn't go with the list idea per above, but did some changes so this won't be duplicated code anymore.

src/parser/jobs/blu/modules/AlwaysBeCasting.tsx
AkairyuGestalter: Ah, I missed that one was a status list and the other an action list. Thanks SE!

src/parser/jobs/blu/modules/AlwaysBeCasting.tsx
AkairyuGestalter: Works for me for now. If/when we get a third we can revisit the list definition idea.

AkairyuGestalter (approved):

PR #1871, opened 2023-07-22
BLU 6.45: New DoTs!

Hugmeir:
BLU got two extremely strong DoTs this time around, `Mortal Flame` (an eternal 40 potency DoT that only needs to be applied once) and `Breath of Magic`(60 seconds, 120 potency).

Both have the funny catch that they can only be applied once per target, so generally only one person will carry one of each.  Due to that restriction, we only show the uptime percent if they ever applied it on any target.

Because of how strong these are, they both must always be applied with `Bristle` (*1.5) and `Moon Flute` (*1.5), so this PR adds suggestions for that too.


src/parser/jobs/blu/modules/DoTs.tsx
```diff
@@ -72,36 +100,92 @@ export class DoTs extends CoreDoTs {
...
+				<DataLink action="BREATH_OF_MAGIC"/> must always be buffed with <DataLink action="BRISTLE"/> and <DataLink action="MOON_FLUTE"/> for a colossal damage gain.
+			</Trans>,
+			tiers: {1: SEVERITY.MAJOR},
+			why: <Trans id ="blu.breath_of_magic.suggestion.unbristled.why">
+				<Plural value={this.unbuffedBoM} one="# Breath of Magic was" other="# Breaths of Magic were"/> cast without both buffs.
+			</Trans>,
+			value: this.unbuffedBoM,
+		}))
+
+		this.suggestions.add(new TieredSuggestion({
```
AkairyuGestalter: Since the content of these two suggestions is basically the same, I feel like they should be combined. It'd look awkward if someone drew the short straw and was supposed to be managing both and they get two different suggestions that look identical other than the action name specified.

src/parser/jobs/blu/modules/DoTs.tsx
```diff
@@ -72,36 +100,92 @@ export class DoTs extends CoreDoTs {
 			</Trans>,
 			value: this.unbuffedSongsOfTorment,
 		}))
+
+		this.suggestions.add(new TieredSuggestion({
+			icon: this.data.actions.BREATH_OF_MAGIC.icon,
+			content: <Trans id = "blu.breath_of_magic.suggestion.unbristled.content">
+				<DataLink action="BREATH_OF_MAGIC"/> must always be buffed with <DataLink action="BRISTLE"/> and <DataLink action="MOON_FLUTE"/> for a colossal damage gain.
+			</Trans>,
+			tiers: {1: SEVERITY.MAJOR},
```
AkairyuGestalter, line 130: Technically, if you're only including one tier then a regular Suggesiton is probably more correct, but then you do have to add the check to see whether it needs to be created or not. If you don't feel like changing this, I'll let it pass :)

AkairyuGestalter (changes requested):
Two comments about the suggestions.

src/parser/jobs/blu/modules/DoTs.tsx
Hugmeir: It's generally strongly recommended NOT to do that -- It's a fairly harsh DPS loss to have one person carry both of the new DoTs, and with our spell slot economy being what it is, it's an even harder sell.

src/parser/jobs/blu/modules/DoTs.tsx
Hugmeir: Ah yeah, I did that to avoid the extra layer of indentation. Don't really mind either way tho.

Hugmeir:
I can confirm that `Mortal Flame` does in fact reapply itself every tick. On the bright side, this does not affect the uptime report, so that just works as usual.

It does affect the clipping report -- by entirely skipping it, because Mortal Flame has no duration, and the clipping report only handles DoTs with a duration. So  while not ideal, this is fine -- It's supposed to be an ability you apply once, maybe twice per fight; there's not going to be any real "clipping".

So I've removed the WIP from the title & pushed a rebase.

src/parser/jobs/blu/modules/DoTs.tsx
AkairyuGestalter: I can live with it for the sake of reduced indentation :)

src/parser/jobs/blu/modules/DoTs.tsx
AkairyuGestalter: Recommended or not, it's still a thing that *can* be done, and I don't want the suggestion list to look wierd if they choose to do so. Adding the `unbuffed<thing>` values together and providing a single suggestion that says along the lins of `BOM and MF must always be...` etc. still gets the meaning across.

src/parser/jobs/blu/modules/DoTs.tsx
```diff
@@ -72,36 +100,92 @@ export class DoTs extends CoreDoTs {
...
+				percent: flameUptimePercent,
+			}))
+		}
+
 		this.checklist.add(new TieredRule({
 			name: <Trans id="blu.dots.rule.name">Keep your DoTs up </Trans>,
 			description: <Trans id="blu.dots.rule.description">
-				The <DataLink status="BLEEDING" showIcon={false} showTooltip={false} /> effect from Nightbloom and SoT is a solid 15% of your total damage.
+				The <DataLink status="BLEEDING" showIcon={false} showTooltip={false} /> effect from <DataLink action="NIGHTBLOOM" showIcon={false} showTooltip={false} /> and <DataLink action="SONG_OF_TORMENT" showIcon={false} showTooltip={false} /> is a solid 15% of your total damage. <br />
+			If you are one of the DPSes applying <DataLink action="BREATH_OF_MAGIC" showIcon={false} showTooltip={false} /> or <DataLink action="MORTAL_FLAME" showIcon={false} showTooltip={false} />, it should be a top priority to keep these DoTs rolling.
```
AkairyuGestalter, line 165: Something I missed in earlier review passes: We're trying to discourage the use of `TieredRule` since it muddies the intended meaning of the checklist as a simple 'did you or did you not do `thing`' check. Can you change this to a regular `Rule` and skip the 90% warning target? The 95% success target is sufficient.

Hugmeir:
Newest changes removed the `TieredRule` and merged the BoM/Mortal Flame buff suggestions into one

src/parser/jobs/blu/modules/DoTs.tsx
```diff
@@ -72,36 +100,90 @@ export class DoTs extends CoreDoTs {
...
 		}))
+
+		const unbuffedWhyBoM = <Trans id ="blu.breath_of_magic.suggestion.unbuffed.why">
+			<Plural value={this.unbuffedBoM} one="# Breath of Magic was" other="# Breaths of Magic were"/> cast without both buffs.
+		</Trans>
+		const unbuffedWhyFlame = <Trans id ="blu.mortal_flame.suggestion.unbuffed.why">
+			<Plural value={this.unbuffedMortal} one="# Mortal Flame was" other="# Mortal Flames were"/> cast without both buffs.
+		</Trans>
+		const unbuffedWhyBoth = <Trans id ="blu.big-dots.suggestion.unbuffed.why">
+			<Plural value={this.unbuffedBoM} one="# Breath of Magic" other="# Breaths of Magic"/>,
```
AkairyuGestalter: grammar nit: stick an 'and' between the two plurals

src/parser/jobs/blu/modules/DoTs.tsx
```diff
@@ -72,36 +100,90 @@ export class DoTs extends CoreDoTs {
...
+			<Plural value={this.unbuffedMortal} one="# Mortal Flame" other="# Mortal Flames"/> were cast without both buffs.
+		</Trans>
+		const unbuffedWhy = (this.unbuffedBoM && this.unbuffedMortal) ? unbuffedWhyBoth
```
AkairyuGestalter: Rather that putting this into a big set of const assignmens and a ternary selection, can we just make a function that takes in the two `unbuffed<thing>` values and spits back out the correct `why`?

AkairyuGestalter (approved):

PR #1872, opened 2023-07-24
BLU Cooldowns: Only track used spells, & track three new 6.45 cooldowns

Hugmeir:
This first half of this is similar in vein to a change BLU has for `Defensives`, where we filter out defensives that were completely unused, assuming they just didn't slot it in.

This does the same but for our general cooldown downtime report, since even more so unlike level 70 BLU, the extra new spells mean that we now have a fair chunk of spells we may not be able to take due to spell slot limitations (Glass Dance, Sea Shanty, Rose of Destruction, even Cold Fog can be on the cut list!).

And speaking of new spells, this adds tracking for `Being Mortal` / `Apokalypsis`, `Sea Shanty`, and `Winged Reprobation`.

### DIFF
--- a/src/parser/jobs/blu/modules/GeneralCDDowntime.ts
+++ b/src/parser/jobs/blu/modules/GeneralCDDowntime.ts
@@ -74,4 +74,9 @@ export class GeneralCDDowntime extends CooldownDowntime {
 		{cooldowns: [this.data.actions.DRAGON_FORCE]},
 		{cooldowns: [this.data.actions.ANGELS_SNACK]},
 	]
+
+	override onComplete() {
+		this.trackedCds = this.trackedCds.filter(cdGroup => this.calculateUsageCount(cdGroup) !== 0)
+		return super.onComplete()
+	}
 }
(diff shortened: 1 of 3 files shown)


src/parser/jobs/blu/modules/GeneralCDDowntime.ts
```diff
@@ -74,4 +87,8 @@ export class GeneralCDDowntime extends CooldownDowntime {
 		{cooldowns: [this.data.actions.DRAGON_FORCE]},
 		{cooldowns: [this.data.actions.ANGELS_SNACK]},
 	]
+
+	override onCompletePrepwork() {
+		this.trackedCds = this.trackedCds.filter(cdGroup => this.calculateUsageCount(cdGroup) !== 0)
```
AkairyuGestalter, line 92: I feel like a cleaner solution to this need to filter the tracked cooldowns before creating the output would be to make the core file's `output` method `protected`, then override that here, run this filtering, then call `super.output()` to generate the output off the now-filtered list.

Hugmeir: `CooldownDowntime` doesn't use `output()` at all, since -- like with the ABC and DoT report -- they just use their `onComplete` to directly add to the checklist array.

I could make `onComplete` protected and override that? That's actually how BLU & MCH's ABC report does this already, so there's prior art for it.

AkairyuGestalter: Yeah, that's still the gist of what I was getting at, even though I got the exact method name wrong. Needs more coffee before I go code reviewing!

Hugmeir: Alright, should be patched now!

AkairyuGestalter (approved):

PR #1873, opened 2023-07-24
BLU: 6.45 Moon Flute update

Hugmeir:
With the new spells in 6.45, BLU's burst window got even more
busy.  It is, technically, still a gain to cast Rose of Destruction,
to use Bristle on Matra Magic, and to use Glass Dance, but they are
all roughly on the category of "you may drop these, either entirely
or just out of the opener".

Adding 4 or 6 new potential actions to the report would've rendered it
mostly useless, so instead, this PR actually relaxes the restrictions a bit.

We now expect a Moon Flute to include the following oGCDs:

    J Kick
    Shock Strike
    Feather Rain
    Nightbloom
    Being Mortal
    Sea Shanty
    Surpankaha x4
    Phantom Flurry

And the following two GCDs:

    Matra Magic
    Triple Trident (only if it was off cooldown, for SpS builds)

While there's optimal filler GCDs for the opener, they aren't
useful generic recomendations (Swiftcast => Wild Rage), and they
aren't standard between all openers (Winged Reprobation and Rose of
Destruction are in practice mutually exclusive, and some openers
drop Bristle for a third Winged Reprobation, for roughly the same
potency on every burst but the first one), so we're just
not enforcing them anymore.

It's also optimal to do an extra weave for Glass Dance somewhere in
the opener, but that's now so rarely taken due to spell slot
limitations that giving it its own column in the report will just
add noise.

To complicate matters even further, one person in the party may
also be doing an extra odd-minute burst to reapply `Breath
of Magic`:

`Breath of Magic` is a new DoT BLU got this patch.  It lasts 60
seconds and does a silly amount of potency, with the caveat that
only one person can apply it on the target.

Due to BoM's silly potency, ideally it should be reapplied under
Moon Flute every minute, meaning that in addition to the normal
even-minute burst, the BoM applier should be doing an odd-minute
burst too.

Since "odd-minute" and "even-minute" are going to be entirely
encounter-dependent, this commit is using a heuristic:
If they used both `Breath of Magic` and `Song of Torment` under
a given Moon Flute, then this is an odd-minute burst, so few
of the usual requirements apply.


xiashtra:
> We now expect a Moon Flute to include the following oGCDs:
> 
> ```
> J Kick
> Shock Strike
> Feather Rain
> Nightbloom
> Being Mortal
> Sea Shanty
> Surpankaha x4
> Phantom Flurry
> ```

You may want to relax the requirement on J Kick.  The new opener is more difficult to execute than before, and the most common suggestion for slow GCD/high ping/slow reaction time players is to move J Kick outside of Moon Flute to help fit everything else in.  The potency loss is minimal compared to missing Phantom Flurry at the end.

src/parser/jobs/blu/modules/MoonFlute.tsx
```diff
@@ -72,19 +73,115 @@ export class MoonFlute extends BuffWindow {
...
+		const extraFilter = playerFilter.type('statusApply')
 			.status(this.data.statuses.WANING_NOCTURNE.id)
 		this.addEventHook(extraFilter, this.onWaningNocturneApply)
 
+		const bomFilter = playerFilter.type('action')
+			.action(this.data.actions.BREATH_OF_MAGIC.id)
+		this.addEventHook(bomFilter, this.onCastBreathOfMagic)
+
+		const level80BLU = this.parser.patch.after('6.4')
+		const bluLevel70MoonFlute = [
```
AkairyuGestalter, line 92: This would be easier to maintain/read if you create a single `const` holding the expected actions common to both pre and post-patch BLU, then based on the patch level push additional entries into the list depending on what was expected at the time. That way we're not holding both full lists in memory regardless of patch level.

src/parser/jobs/blu/modules/MoonFlute.tsx
```diff
@@ -186,31 +242,50 @@ export class MoonFlute extends BuffWindow {
...
+			// Final Sting used, so don't dock any points for missed actions:
+			return neutralOrPositiveOutcome
+		}
+
+		// Final Sting not used here
+		if (this.breathOfMagicApplier) {
+			// It's the Breath of Magic applier!  They may be doing an off-minute window to
+			// reapply BoM, and we should not dock them any points for that.
+			// ...but how can we tell if it's an off-minute window?
+			// Yeah, that's a good question.  Hopefully someone will have a better
```
AkairyuGestalter: This isn't the silliest thing I've ever seen done for guesswork! You should see the nonsense that is DNC's Gauge code :)

AkairyuGestalter (changes requested):
One adjustment I'd recommend to how you're building the list of expected actions in the moon flute window, but otherwise I think this looks fine!

src/parser/jobs/blu/modules/MoonFlute.tsx
xiashtra: What about doing a check against last Nightbloom use < 120s and Breath of Magic.  That way you don't get a false positive against someone not bringing Nightbloom at all.

src/parser/jobs/blu/modules/MoonFlute.tsx
Hugmeir: Hm, I have the opposite take here. I think simply having the lists in plain text is much easier for a human to maintain/read, compared to having a partial list and then figuring out what gets injected into it.

Hugmeir:
Regarding dropping the requirement on J Kick (or at least making it an optional one), I like the idea -- to parrot what I said on the discord, due to ping I always drop J Kick out of the opener myself, so thumbs up -- but I think I'll add it in its own PR later.

src/parser/jobs/blu/modules/MoonFlute.tsx
Hugmeir: Latest patch has a reimplementation of @xiashtra's heuristic using the Nightbloom cooldown as the heuristic -- if they didn't take Nightbloom then everything is strict.  If they took Nightbloom, and a moon flute happens while Nightbloom was still on cooldown, AND they used BoM, we assume it's an odd-minute and relax the check.

As a TODO for later, the odd-minute check should still disallow Surpanakha unless it is exactly 0 or 4. Anything in-between in lost potency.

I also reworded the suggestions from the module slightly, both for clarity and to point out that the odd-minute MF is excempt from some checks.

Hugmeir: ...also, oh gosh, I just looked at DNC's Gauge calculations.  BLU has it good.

AkairyuGestalter: And it used to be worse! PrObAbIlItY!

AkairyuGestalter (approved):

PR #1874, opened 2023-07-27
Improve GCD estimates from FFLogs reports

hintxiv:
Rewrites our speedStat fflogs adapter to use parts of the process that https://gcdcalc.fly.dev/ uses to estimate GCDs, which seems to result in a significant improvement in accuracy. Jobs with prickly speed stats like BLM and BRD should be validated before merging. Below are some random logs I checked -- the new method was more accurate in all cases:

**1.94 Monk**
https://www.fflogs.com/reports/FwZtxbynRKVTP94g#fight=1&type=damage-done
* Old estimate: 1.92
* New estimate: 1.94

https://www.fflogs.com/reports/xJkvZHcr3bTFK8PN#fight=5&type=damage-done
* Old estimate: 1.96
* New estimate 1.94

**2.48 RDM**
https://www.fflogs.com/reports/YMGbt8CfHa4gnZ1Q#fight=107&type=damage-done
* Old estimate: 2.50
* New estimate: 2.48

https://www.fflogs.com/reports/6jHhrzFd7mAZq8PG#fight=16&type=damage-done
* Old estimate: 2.49
* New estimate: 2.48


AkairyuGestalter (approved):
Logic seems sound, and a selection of logs for my static shows our BRD's and my GCDs the same pre and post fix, and off by at most .01s (I think mine is 2.13 with food, and it's showing as 2.14 on both so eh). It's certainly no worse for the logs I've checked so I'm good with this!

PR #1877, opened 2023-07-30
Leylines to buffwindow

AkairyuGestalter:
Swapped LeyLines' implementation from standalone Analyser to buffwindow to cut down on how much bespoke window handling code was around, and to change its UI to be more consistent with other windows.

Pulled over the new RulePassedEvaluator added on #1832 and added a 'ruleContext' option to the output in addition to the checkmark/x icons, to allow this to retain the per-window uptime % display alongside the indicator of whether that window met the uptime expectation of 90%.

### DIFF
--- a/src/parser/core/modules/ActionWindow/evaluators/NotesEvaluator.ts
+++ b/src/parser/core/modules/ActionWindow/evaluators/NotesEvaluator.ts
@@ -6,12 +6,13 @@ import {EvaluationOutput, WindowEvaluator} from './WindowEvaluator'
 
 export abstract class NotesEvaluator implements WindowEvaluator {
 
-	protected abstract header : RotationTarget
+	protected abstract header : RotationTarget | undefined
 	protected abstract generateNotes(window: HistoryEntry<EvaluatedAction[]>): JSX.Element
 
 	suggest(_windows: Array<HistoryEntry<EvaluatedAction[]>>): Suggestion | undefined { return undefined }
 
-	public output(windows: Array<HistoryEntry<EvaluatedAction[]>>): EvaluationOutput  {
+	public output(windows: Array<HistoryEntry<EvaluatedAction[]>>): EvaluationOutput | undefined {
+		if (this.header == null) { return }
 		return {
 			format: 'notes',
 			header: this.header,
--- a/src/parser/core/modules/ActionWindow/evaluators/RulePassedEvaluator.tsx
+++ b/src/parser/core/modules/ActionWindow/evaluators/RulePassedEvaluator.tsx
@@ -1,0 +1,44 @@
+import {RotationTarget} from 'components/ui/RotationTable'
+import React, {ReactNode} from 'react'
+import {Icon} from 'semantic-ui-react'
+import {Suggestion} from '../../Suggestions'
+import {EvaluatedAction} from '../EvaluatedAction'
+import {HistoryEntry} from '../History'
+import {NotesEvaluator} from './NotesEvaluator'
+
+export interface RulePassedEvaluatorOpts {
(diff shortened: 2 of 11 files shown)


src/parser/core/modules/ActionWindow/evaluators/NotesEvaluator.ts
```diff
@@ -6,12 +6,13 @@ import {EvaluationOutput, WindowEvaluator} from './WindowEvaluator'
 
 export abstract class NotesEvaluator implements WindowEvaluator {
 
-	protected abstract header : RotationTarget
+	protected abstract header : RotationTarget | undefined
```
KelosPelmand, line 9: Not sure I see the benefit of this change.  If someone wants to make a suggestion only evaluator, they can already inherit from WindowEvaluator directly and override suggest with the suggestion and override output with `return undefined;`.

You don't even seem to be using this new feature in this PR anyway.  If you had some other purpose in mind, I'd be interested to know.

src/parser/core/modules/ActionWindow/evaluators/RulePassedEvaluator.tsx
```diff
@@ -0,0 +1,44 @@
+import {RotationTarget} from 'components/ui/RotationTable'
+import React, {ReactNode} from 'react'
+import {Icon} from 'semantic-ui-react'
+import {Suggestion} from '../../Suggestions'
+import {EvaluatedAction} from '../EvaluatedAction'
+import {HistoryEntry} from '../History'
+import {NotesEvaluator} from './NotesEvaluator'
+
+export interface RulePassedEvaluatorOpts {
```
KelosPelmand: Style point: Not personally a huge fan of how this essentially is "derive from RulePassedEvaluator by passing in a bunch of function objects or not" vs the more standard "derive from RulePassedEvaluator that has virtual methods with default implementations".

(I realize other evaluators take some function objects, but I think this struck me differently than those due to these functions feeling more core logic of the class as compared to handle edge cases in the other evaluators.)

I see why you did it that way, since you need data from the Leylines window object.  The alternative to have RulesPassedEvaluator use regular virtual methods instead would mean that the derived StayedInEvaluator would have to take an `owner: Leylines` and call the methods that way.  Personally, I'm used to doing things that way from other projects I've done, but I'm not sure what the general preference is for this project.

AkairyuGestalter: I wrote RulesPassed as an abstract Evaluator at first but that ended up meaning a lot additional boilerplate for each extension (5 in BLM, 3 in DNC, so far, plus others I haven't PRed yet) and yes, frequently having to pass quite a lot of data in through constructor properties to let it do the checks, which is why I eventually gravitated to this model.

This comment made me take a closer look at other places we'd eventually want to use it to replace/polish up all the places we're showing a column with the checkmark for success and X for failure, and I think you're right that having it be an abstract other things can inherit from is probably the better long-term model. I'll swap back to that.

src/parser/core/modules/ActionWindow/evaluators/NotesEvaluator.ts
AkairyuGestalter: This was an artifact of how I originally wrote the RulePassedEvaluator on #1832 since not all of the 'rules' there have column output. It might go away with the re-implementation of RulesPassed as an abstract based on your other comment. will see.

AkairyuGestalter: still need it for DNC's implementation after all (see current state of #1832)

PR #1879, opened 2023-08-02
[Bugfix] Countergauge death bugfix

AkairyuGestalter:
bloody null coalesce/ternary order of operations.

See: https://xivanalysis.com/fflogs/bLVmRzXJq4hyKpFT/11/1
Expected medidate gauge generation: 41
Actual prior to fix: 42 across 43 events marked as 'generate'... the real ones and the death and raise events at ~5:30

### DIFF
--- a/src/parser/core/modules/Gauge/CounterGauge.ts
+++ b/src/parser/core/modules/Gauge/CounterGauge.ts
@@ -57,7 +57,7 @@ export interface CounterChartOptions {
 
 export class CounterGauge extends AbstractGauge {
 	private initialValue: number
-	private _value: number
+	private currentValue: number
 	private minimum: number
 	private maximum: number
 	overCap: number = 0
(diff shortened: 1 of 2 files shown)


src/parser/core/modules/Gauge/CounterGauge.ts
```diff
@@ -148,7 +148,7 @@ export class CounterGauge extends AbstractGauge {
 	set(value: number, reason?: GaugeEventReason) {
 		const delta = value - this.value
 
-		const _reason = reason ?? delta > 0 ? 'generate' : 'spend'
+		const _reason = reason ?? (delta > 0 ? 'generate' : 'spend')
```
hintxiv: I think the convention for an underscore prefix in a variable name is "we don't intend to use this variable for anything", but we are using it here. I also think the inline null + ternary operators together are a little hard to parse (which is probably why this error got through in the first place). Could we change it to something like this instead?

```ts
    const defaultReason = (delta > 0) ? 'generate' : 'spend'
    
    // ...
    
    this.pushHistory(reason ?? defaultReason, delta)
```

hintxiv (changes requested):

src/parser/core/modules/Gauge/CounterGauge.ts
AkairyuGestalter: Honestly can't remember why I wrote it that way to begin with, so yes, happy to fix.

hintxiv (approved):

PR #1880, opened 2023-08-05
Feature - Home page support summary

AkairyuGestalter:
Based on feedback on Discord, this change adds a summary of what is currently supported by xivanalysis to help keep users informed of the state of the site, especially when notable patches with changes to job mechanics roll around.

image

Example before core catches up to current patch:
image

Examples of start-of-expansion appearance (before and after core, but w/o job support):
image
image


src/components/Home/SupportSummary.module.css
```diff
@@ -0,0 +1,94 @@
...
+	overflow: hidden;
+}
+
+@media sm {
+	.summary {
+		font-size: unset;
+	}
+}
+
+@media md {
```
hintxiv: I think this breakpoint should be wider. Here is a "medium" screen with unreadable support text:

image

If we just swap md to lg (and sm to md), the transition is much cleaner imo.

src/components/Home/SupportSummary.module.css
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	.supportGrid .jobs {
+		flex-direction: column;
+	}
+
+	.supportGrid .job {
+		width: 100%;
+	}
+}
```
hintxiv, line 103: Not super important, but a small touch that I think would be nice is to make the job icons unselectable. E.g.:

```css
.job > *:first-child  {
    -webkit-touch-callout: none;
    -webkit-user-select: none;
    -khtml-user-select: none;
    -moz-user-select: none;
    -ms-user-select: none;
    user-select: none;
}
```

src/components/Home/SupportSummary.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+interface RoleData {
+	roleKey: RoleKey
+	jobKeys: JobKey[]
+}
+
+class SupportSummary extends Component {
+	private roleJobs: RoleData[] = []
+	override render() {
+		const maxPatch = [...Object.entries(PATCHES)].map(patch => patch[0]).pop() ?? FALLBACK_KEY
```
hintxiv: I know that the PATCHES object is chronologically ordered for the sake of convenience, but this fact isn't necessarily guaranteed by the data type. Maybe something like:

```suggestion
		const maxPatch = Object.entries(PATCHES)
			.map(patch => ({name: patch[0], date: patch[1].date[GameEdition.GLOBAL]}))
			.sort((a, b) => b.date - a.date)[0].name
```

src/components/Home/Home.tsx
```diff
@@ -43,12 +44,14 @@ class Home extends Component {
...
 					<Modal.Header><Trans id="core.home.about.title">
-						About xivanalysis
+							About xivanalysis
 					</Trans></Modal.Header>
 					<Modal.Content>
 						<TransMarkdown source={about}/>
 					</Modal.Content>
 				</Modal>
+
+				<SupportSummary />
```
hintxiv: Adding the summary here pushes the URL input away from the horizontal centre of the page, which might be a little confusing given how thin it is relative to the summary. This is just an idea, but maybe some css tweaks to keep the URL input centred and slightly separated from the summary? Let me know what you think.

image

hintxiv (changes requested):
Very cool feature!

src/components/Home/Home.tsx
AkairyuGestalter: Didn't have a ton of time to try to fix this before raid started today. Will see if I can cook something up tomorrow, cause I think getting some spacing in between the two elements is a good idea. Probably won't try to keep the input centered still since that causes other problems, but not having them smashed together the way they are now will be nice.

hintxiv (approved):

PR #1889, opened 2023-08-14
WIP: Core Weaving module: Tighten the check

Hugmeir:
Previously this was looking at the cast time for an action and going by a simple table to decide how many weaves you had.

The problem is that this lead to several false-negatives.

The initial implementation of this PR made the check slightly more strict, first by assuming that the animation lock (for both GCDs and weaves) is 800ms[*], then by running the equivalent of this:

    floor( (recastTime - castTime - animationLock) / animationLock )

However, that actually made it *too* strict, causing it to display some real-but-unactionable issues, like this from the current top RDM report:
Screenshot 2023-08-14 at 14 58 08

So to reduce the noise, this also adds a 50ms grace period for clips -- Essentially allowing GCD clips so long as they are 50ms or less.

Weaving issues on current HEAD:
Screenshot 2023-08-14 at 15 12 56

Weaving issues with this PR applied (note the 3:17 and 5:11 clips):
Screenshot 2023-08-14 at 15 13 15

[*] I'm told that for people with low latency this goes as low as 600ms, but this seems a bit more forgiving for anyone not hugging the DCs.

### DIFF
--- a/src/parser/core/changelog.tsx
+++ b/src/parser/core/changelog.tsx
@@ -10,6 +10,13 @@ export const changelog: ChangelogEntry[] = [
 	// 	contributors: [CONTRIBUTORS.YOU],
 	// },
 	{
+		date: new Date('2023-08-14'),
+		Changes: () => <>
+			Tightened the detection of Weaving issues.
+		</>,
+		contributors: [CONTRIBUTORS.HUGMEIR],
+	},
+	{
 		date: new Date('2023-07-28'),
 		Changes: () => <>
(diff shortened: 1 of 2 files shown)


Hugmeir:
Thanks to some feedback on the discord, this now accounts for the different animation lock for instant GCDs and casted GCDs.

Hugmeir:
Looks like what I wanted to achieve (pointing out *slow* weaves) and what the Weaving module is trying to point out don't exactly match, so I'm going to close this PR.

PR #1890, opened 2023-08-19
Feature - Adapt gaugeUpdate events for Dancer

AkairyuGestalter:
Thanks to some chatter on both ours and the ranged theorycraft Discord, I was finally able to make a viable adaptation of DNC's gaugeUpdate events to use in place of/in addition to the existing probabilistic simulations for Esprit and Feathers.

Not sure if this is the best path forwards for adapting these events but it does work, so I'm sticking it here for review.

Misc Notes:
- I tried to have the adaptation step store to the actor whether they have gaugeUpdate events, and then use that in the gauge module initialiser to tinker with what was getting hooked. Apparently all the initialise functions get called _before_ event adaptation takes place, so I was getting weird results. Instead, the event hooks themselves keep track of whether gaugeUpdate events were detected and unhook/skip processing as needed.
- I updated the Esprit overcap suggestion to handle the fact that for the logging player we can distinguish between definitive overcap (ie, logged gauge set at 100 and they used one of their own generators) and potential overcap (existing probabilistic gauge generation from partnered/technical finished party members). There's a three way split now depending on what case we're in (definitely missed at least one saber dance, maybe missed at least one due to probabilistic generation, or definitely missed at least one _and_ potentially missed at least an additional one)

Sample logs (My own, a couple 24 man runs as DNC):
- https://www.fflogs.com/reports/xg6fWHwtjkGVF17m
- https://www.fflogs.com/reports/C8JFDdRQGWXgtvwL


src/parser/jobs/dnc/changelog.tsx
```diff
@@ -8,43 +8,48 @@ export const changelog = [
...
+		date: new Date('2023-09-04'),
+		Changes: () => <>Fix a bug related to the Technical Finish status application changes from 6.4 that occasionally caused incorrect 'missed' Finish reports.</>,
 		contributors: [CONTRIBUTORS.AKAIRYU],
 	},
 	{
-		date: new Date('2021-11-27'),
-		Changes: () => <>Removed combo drop leniency around back-to-back Standard/Technical Steps due to lengthened combo timeout.</>,
+		date: new Date('2023-08-19'),
+		Changes: () => <>Esprit and Feather Gauges now show the actual values for logs the Dancer uploaded themselves! Overcap is still partly simulated though.</>,
 		contributors: [CONTRIBUTORS.AKAIRYU],
```
AzariahAeroborn: Update changelog entry.  Make sure we don't have any other unintended changes happening here and it's just GH being weird with the change displayed.

src/parser/jobs/dnc/modules/Gauge.tsx
```diff
@@ -135,12 +140,38 @@ export class Gauge extends CoreGauge {
 		this.addEventHook(statusApplyFilter.status(espritStatusMatcher), this.addEspritGenerationHook)
 		this.addEventHook(statusRemoveFilter.status(espritStatusMatcher), this.removeEspritGenerationHook)
```
AzariahAeroborn, line 141: Can we also unhook these once we know we have gauge events, and clear the esprit partner damage hook if it exists?  Should also mean that we wouldn't need to do checks in the probabilistic esprit generation (onDamage) to bail if this log has gauge events.

AzariahAeroborn (changes requested):

src/parser/jobs/dnc/modules/Gauge.tsx
AkairyuGestalter: The remaining generation event stuff is used to calculate parts of the overcap suggestion for when the log events show the gauge is capped but we're getting another event that should have generated Esprit or Feathers. Gauge changed events can't get that information for us, so we still need the generation events to infer it.

src/parser/jobs/dnc/changelog.tsx
AkairyuGestalter: Yeah, fixing the changelogs on all these stale PRs is definitely on the list to do before we approve them. Should be good here with the latest commit.

AzariahAeroborn (approved):

PR #1891, opened 2023-08-23
[BugFix] Sharpcast Thundercloud fix

AkairyuGestalter:
Typical event order for a sharpcasted thunder reapplying the thundercloud proc buff looks something like this (all at the same timestamp though):
Thunder damage event
Thundercloud status apply event
Sharpcast status remove event

On this log (https://xivanalysis.com/fflogs/BV3MDXg12bv6AwFz/10/11), just after the 6:16 mark, the thundercloud status apply preceded the other events by half a second for some reason. This resulted in an event order that the BLM Procs code interpreted as a dropped Thundercloud because the next use of Thunder 3 after this didn't appear to have an active Thundercloud to consume and the status reapplication didn't sync up with a Sharpcast usage.

Added a safety backstop to create a new Thundercloud window if one is not currently open when a Sharpcast consumption by a Thunder spell is detected.

Also did some minor housekeeping to reorder the changelog to make it consistent with other jobs and remove an icon that we don't use anymore.


src/parser/jobs/blm/modules/Procs.tsx
```diff
@@ -48,10 +34,14 @@ export default class Procs extends CoreProcs {
...
 
 		// Thunder procs used while sharpcast is up re-grant the proc status without technically removing it, so we need to forcibly add the 'removal' here to keep the 'dropped' counting correct
-		if ((event.action === this.data.actions.THUNDER_III.id || event.action === this.data.actions.THUNDER_IV.id) && this.hasSharpcast) {
+		if ((event.action === this.data.actions.THUNDER_III.id || event.action === this.data.actions.THUNDER_IV.id) && this.actors.current.hasStatus(this.data.statuses.SHARPCAST.id)) {
 			this.tryAddEventToRemovals(procGroup, event)
+
+			// Make sure a Sharpcasted thunder always results in an open window, in case the log had weird status timing
+			if (!this.currentWindows.has(procGroup)) {
+				this.currentWindows.set(procGroup, {start: event.timestamp})
+			}
```
hintxiv, line 43: This seems to produce two procs, which I don't believe is intended:

image

AkairyuGestalter: It's a side effect of the "oops, the status went away earlier than it should have" event ordering, but in practice it's not going to matter much for BLM since it'll just end up looking to users (if they pay that close attention in the first place) like a DoT tic that procced right before the Sharpcast got used. It's better than the alternative of telling players they dropped a proc/Sharpcast buff when they actually did use them.

hintxiv (approved):

PR #1894, opened 2023-09-05
[Bugfix] - Incorrect 'missed' Technical Finishes

AkairyuGestalter:
Report link: https://xivanalysis.com/fflogs/14DVaFHBgwzNPRyn/14/140

Apparently the 6.4 buff range changes not only messed with devilment timing, it also occasionally (always?) produces a 'damage' event without a sequence number that targets the source player and 'misses'. If this event comes before the real damage events, it was closing the window early and marking it as a missed finish.

We probably ought to handle this down in adapter space, but I can't debug into that to try to see how it's winding up making these events, so I'm just dealing with it here.


hintxiv (approved):
LGTM

PR #1895, opened 2023-09-13
Port Divination to core RaidBuffWindow

AkairyuGestalter:
Follow-on from porting Technical Finish over. PR moves Divination tracking into the core RaidBuffWindow, makes use of new core RulePassedEvaluators to handle the "checkmark" table columns, and will include multi-AST Divination overwrite tracking by default.


suShirow (approved):
LGTM

PR #1896, opened 2023-09-13
Port Searing Light to core RaidBuffWindow

AkairyuGestalter:
Fairly straightforward port to allow SMN to benefit from core RaidBuffWindow functionality. Minor updates to handle the pre-6.1 Carbuncle-sourced buff status and ghosting logic, but otherwise it's mostly stock.


AkairyuGestalter:
After further discussion with the SMN maintainers, we're going to wait until Dawntrail to port this to the core Analyser so we don't have to deal with the pet support nonsense. It'll be an even more straightforward port then anyways.

PR #1897, opened 2023-09-14
Standardize Brotherhood RaidBuffWindow report

AkairyuGestalter:
Monk was doing a really weird version of the RaidBuffWindow for Brotherhood, in part due to the "Brotherhood should always be used inside Riddle of Fire" usage rule. However, the way this ended up being displayed was hard to read on the Riddle of Fire table, would hide any raidbuff overwrite information unless Brotherhood was used fully outside RoF, and was generally just not consistent with the way other jobs show raid buffs information alongside other buff windows.

Changes:
- Removed the atypical MissedBrotherhood version of RaidBuffWindow
- Created a standard RaidBuffWindow instance for Brotherhood
- Moved the Brotherhood drift evaluator from Riddle of Fire to Brotherhood, using the RulePassedEvaluator with the type of drift as context
- Added the MissedBrotherhoodEvaluator to the Brotherhood RaidBuffWindow to keep the distinct Suggestion test for Brotherhoods used fully outside Riddle of Fire.


src/parser/jobs/mnk/changelog.tsx
```diff
@@ -12,6 +12,11 @@ export const changelog = [
 		Changes: () => <>Add gauge visualisation for Perfect Balance, and mark as supported for Endwalker.</>,
 		contributors: [CONTRIBUTORS.AKAIRYU],
 	},
+	{
+		date: new Date('2023-09-13'),
```
AzariahAeroborn: Should we update the changelog date on this?

src/parser/jobs/mnk/modules/Brotherhood.tsx
```diff
@@ -137,59 +109,4 @@ export class Brotherhood extends Analyser {
...
-			icon: this.data.actions.BROTHERHOOD.icon,
-			content: <Trans id="mnk.brotherhood.suggestions.missed-players.content">
-				Try to make sure your <DataLink action="BROTHERHOOD"/> casts buff your full party with each use. Failing to do so is a raid damage loss.
-			</Trans>,
-			matcher: matchClosestHigher,
-			tiers: BROTHERHOOD_SEVERITY_TIERS,
-			value: (expectedBuffs - actualBuffs) / expectedBuffs,
-			why: <Trans id="mnk.brotherhood.suggestions.missed-players.why">
-				<Plural value={windowsWithMissedBuffs} one="# cast" other="# casts"/> of <DataLink action="BROTHERHOOD"/> missed hitting the entire party for a total of <Plural value={expectedBuffs - actualBuffs} one="# players" other="# players"/>.
-			</Trans>,
```
AzariahAeroborn, line 192: This suggestion appears to be not working in the new approach, is that expected?  
image

https://xivanalysis.com/fflogs/wkGCrNh2p3BWVtzx/2/63 generates this suggestion in production, but not in the PR.

src/parser/jobs/mnk/modules/evaluators/BrotherhoodEvaluator.tsx
```diff
@@ -1,98 +1,105 @@
...
 
-		switch (matchedWindow.status) {
+		return new Suggestion({
+			icon: this.suggestionIcon,
+			content: <Trans id="mnk.brotherhood.drifted-window.content.">
+				Try and make sure your <DataLink action="BROTHERHOOD"/> casts are fully included within a <DataLink action="RIDDLE_OF_FIRE"/> window.
+			</Trans>,
+			severity: SEVERITY.MINOR,
+			why: <Trans id="mnk.brotherhood.drifted-window.why">
+				<Plural value={driftedWindows.length} one= "# cast" other="% casts"/> of <DataLink action="BROTHERHOOD"/> was either early or late for a <DataLink action="RIDDLE_OF_FIRE"/> window.
```
AzariahAeroborn:
```suggestion
				<Plural value={driftedWindows.length} one="# cast" other="# casts"/> of <DataLink action="BROTHERHOOD"/> <Plural value={driftedWindows.length} one="was" other="were"/> either early or late for a <DataLink action="RIDDLE_OF_FIRE"/> window.
```

Template error for the plural value, holdover from production.  Might be worth fixing the verb while we're mucking with it?

AzariahAeroborn (changes requested):
It looks like converting to the Buff Window approach is removing the suggestion for players buffed by Brotherhood... which looks like would put it in line with other RBW modules, but wasn't called out as an intended change so I figured I'd double check if that was intended.  Otherwise LGTM, one minor copy bug that already exists in production but would be worth fixing while we're here.

src/parser/jobs/mnk/modules/Brotherhood.tsx
AkairyuGestalter: We made the decision for basically every other raid buff window to not ding the player for missing someone since that's not fully under their control (party member in Narnia or dead for example). This just brings MNK in line with other jobs, and is intended.

AzariahAeroborn (approved):

PR #1918, opened 2023-10-12
BLM - show tincture window

### LINKED ISSUE #1917: [BLM][Improvement] Show spells in tincture window
XIVAnalysis for BLM currently does not display the tincture window module. While it is not the highest priority thing for the job, it's still helpful to know what spells get covered by pots.

noloerino:
Resolves #1917


AkairyuGestalter (changes requested):
I don't know why it hadn't occurred to me to add a Tinctures display to BLM before now...

Can you add a changelog entry for this? We want to have those for any end-user visible change, and you should take credit for your contribution!

noloerino:
Done, thanks for the reminder!

AkairyuGestalter (changes requested):
One more nitpick: Can you drop the image resolution for your icon to something like 100-150px? These don't show particularly large anyways, so it's nice to have the files a bit smaller to save on memory/bandwidth

AkairyuGestalter (approved):

Issue #1926, opened 2023-12-16
The Show/Hide i18n overlay button causes bunshin on homepage


Odaimoko:
<img width="1248" alt="image" src="https://github.com/xivanalysis/xivanalysis/assets/7827700/b8de2eb8-bc9a-4eeb-becb-753561194e22">

PR #1927, opened 2023-12-22
Remove bunshin from support summary

AkairyuGestalter:
Observe issue #1926
Observe issue #1926 go away
Something something react routing, idk

### DIFF
--- a/src/components/Home/SupportSummary.tsx
+++ b/src/components/Home/SupportSummary.tsx
@@ -5,6 +5,7 @@ import {GameEdition} from 'data/EDITIONS'
 import {JobKey, JOBS, RoleKey, ROLES} from 'data/JOBS'
 import {patchSupported} from 'data/PATCHES'
 import {FALLBACK_KEY, PATCHES} from 'data/PATCHES/patches'
+import {observer} from 'mobx-react'
 import {AVAILABLE_MODULES} from 'parser/AVAILABLE_MODULES'
 import React, {Component, Fragment} from 'react'
 import {Message} from 'semantic-ui-react'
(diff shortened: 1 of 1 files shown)


ackwell:
ah i see what's going on. adding `@observer` actually just hides the issue, not fix the problem.

`@observer` implies `extends PureComponent` (effectively), and `PureComponent` means it will avoid re-rendering if none of its props change. `Component` will always just re-render. in essence, adding `@observer` is papering over the issue by just preventing it re-rendering at all (because it has no props that will change).

this... is getting a bit into react-weeds, but the underlying issue is that regardless of the `Pure`-ness of the component declaration, the `render` method (or the function body in non-class instances) _must_ be pure - i.e. if it runs `render` with the same props and state twice, it should get the same output each time.

this isn't held by the current impl of this UI, as it has a class property (`roleJobs`) that is mutated during render - i.e. if `render` is called twice, you end up with two sets of values in `roleJobs`.

there's a whooooooooooole bunch of ways to approach this and so on and so forth, but in this case the most dead-simple approach would be to delete the `roleJobs` class property, and then use an array in the render function a la

```diff
 // Build the list of roles and their associated jobs
 let jobKey: JobKey
+const roleJobs = [] 
 for (jobKey in JOBS) {
 	const roleKey = JOBS[jobKey].role
 	if (roleKey === 'UNSUPPORTED') { continue }
 	const roleData = this.roleJobs.find(rd => rd.roleKey === roleKey) ?? {roleKey, jobKeys: []}
 	roleData.jobKeys.push(jobKey)
 	if (!this.roleJobs.some(rd => rd.roleKey === roleKey)) {
-		this.roleJobs.push(roleData)
+ 		roleJobs.push(roleData)
 	}
 }

 return <>
 	{supportMessage}
 	<div className={styles.summary}>
-		<SupportSummaryGrid roles={this.roleJobs} />
+		<SupportSummaryGrid roles={roleJobs} />
 	</div>
 </>
```

ackwell (approved):

PR #1935, opened 2024-02-27
[BugFix] Bring Searing Light's buff count tracking in-line with RaidBuffWindow

AkairyuGestalter:
Searing Light was using the outdated assumption that the number of players to be buffed was always 8. While we don't target this at dungeons, it still looks bad to have most jobs (which are mostly/all ported over to core RaidBuffWindow) recognizing when the party size is smaller, and SMN isn't.

We'll handle this in Dawntrail with the expected port of Searing Light to core, but in the meantime, I replicated the expected buff denominator logic from there into Searing Light. Also taking the opportunity to remove the suggestion to hit all players to make that consistent with other jobs, since that is not fully in the player's control.

### DIFF
--- a/locale/de/messages.json
+++ b/locale/de/messages.json
@@ -1021,8 +1021,6 @@
   "smn.searinglight.players-count": "Spieler gebufft",
   "smn.searinglight.suggestiongs.ghosted.why": "",
   "smn.searinglight.suggestions.ghosted.content": "",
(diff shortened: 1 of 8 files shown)


KelosPelmand:
LGTM 👍 

Suggestion was always minor because it wasn't entirely under your control, but I'm not sad to see it go either.

AzariahAeroborn (approved):

PR #1938, opened 2024-03-11
feat: add documentation for example logs

AzariahAeroborn:
Add a documentation file for how (and why) to add example logs.  I can try to get screen snips of the FFLogs parts of the process if we think that would be helpful... my initial thought is that the process is relatively simple and the description should be reasonably good, so I didn't spend time trying to figure out how to do sample pictures in markdown.


AzariahAeroborn:
Closing to rebase on dawntrail branch.

PR #1939, opened 2024-03-20
WHM: Update Lilies Module for better Blood Lily calculations

samehhamadeh:
This PR is based on the issue mentioned here: https://discord.com/channels/441414116914233364/1166220286308388885
Changes I made were discussed in #dev-healer channel in discord.

Updated 2 of the suggestions for Blood Lily usages:
-Overcapping suggestion is now based on how many Misery usages were lost from overcapping. It also checks how many stacks you had at the end of the fight and whether overcapping led you to end the fight with leftovers and lose a usage. Due to this, I updated the severities to be medium for losing 1 full Blood Lily, and major for losing 2.
-End of fight leftovers suggestion now only penalizes you for ending the fight with a fully bloomed Blood Lily, since you cant use leftover stacks unless they are fully bloomed. The leftover stacks did affect anything, the previous change will catch it and it will be reflected there.

These are logs that I used for verification:
- https://www.fflogs.com/reports/mQkbrpCXwahTVxgL#fight=16&source=3
This was the report initially mentioned in the thread about the issue. On the current version of xiva, it mentioned 8 missed Misery usages which was incorrect. On my updated version it will mention 2 which is accurate, since 11 Lilies were used, which should bloom 3 Blood Lilies, which means 2 were missed due to overcaps.
- https://www.fflogs.com/reports/rHk92hTn7KFvLpBg#fight=6&source=2
This report shows the modified suggestion to finishing the fight with leftover Blood Lilies. In the updated version it will say 1 Blood Lily was missed (as in, 1 Misery usage), instead of 3 Blood Lilies for each individual growth stack.
- https://www.fflogs.com/reports/Lp7Bw12kgAQTqZ6Y#fight=2
This report shows the overcapping suggestion working correctly with the change to include end of fight leftover stacks to the calculation. In the log, I overcapped 2 blood lily growth stacks, then ended the fight with 1 leftover stack, and the report shows 1 missed usage due to overcapping, this wouldnt be the case with the old version.


src/parser/jobs/whm/modules/Lilies.tsx
```diff
@@ -183,7 +184,7 @@ export class Lilies extends CoreGauge {
 			tiers: SEVERITIES.BLOODLILY_LEFTOVER,
 			value: this.bloodLilyGauge.value,
 			why: <Trans id="whm.gauge.bloodlily.suggestions.leftover.why">
-				{<Plural value={this.bloodLilyGauge.value} one="# Blood Lily" other="# Blood Lilies" />} went unused.
+				{<Plural value={this.bloodLilyGauge.value / MISERY_COST} one="# Blood Lily" other="# Blood Lilies" />} went unused.
```
AkairyuGestalter: Setting this plural value here feels a bit awkward with respect to the tiers/value props of the suggestion overall. We're sending in the full value of the blood lily gauge, with a tiers array that is effectively doing this 'divide by misery cost', but not actually using the MISERY_COST constant, which could lead to weird behavior in the future if the cost of spending a Misery changes. Since we're always showing a severity of MAJOR if we hit this anyways, I think we'd be better off storing the result of this division into a const, and creating a normal Suggestion with a defined MAJOR Severity if that value is > 0, rather than using TieredSuggestion.

AkairyuGestalter (changes requested):
One note about the leftover lilies suggestion, but otherwise this looks good!

src/parser/jobs/whm/modules/Lilies.tsx
samehhamadeh: Done, I changed the TieredSuggestion to a Suggestion that displays if >1 blood lily was used. I kept the plural value in as future proofing in case they change whm to have more than 1 blood lily at a time.

AkairyuGestalter: I'm getting a "0.333 Blood Lilies went unused" suggestion now on your third log. Can you floor the `const unusedLilies` assignment to make sure it's always an integer value?

samehhamadeh: Fixed

AkairyuGestalter (approved):

PR #1940, opened 2024-03-21
Add forgiveness for defensives

nonowazu:
This adds a forgiveness window for defensive usages; as our team has started to use xiva for understanding some defensive mappings, I've seen some situations where the availability of any one given defensive is listed as 1s (but upon looking at the code is < 400ms).

I fully expect someone to catch my miss at the top of the section where it says something like 3/5 instead of a now "forgiven" 3/4, so I'm open to feedback about restructuring this little change to accommodate that.

### DIFF
--- a/src/parser/core/modules/Defensives.tsx
+++ b/src/parser/core/modules/Defensives.tsx
@@ -21,6 +21,8 @@ const DEFENSIVE_ROLE_ACTIONS: Map<RoleKey, ActionKey[]> = new Map<RoleKey, Actio
 	['HEALER', []],
 ])
 
+const DEFAULT_FORGIVENESS_MS: number = 800
+
 export class Defensives extends Analyser {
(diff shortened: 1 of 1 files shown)


AkairyuGestalter:
Since all of this is informational anyways, I don't personally think we should obfuscate where those admittedly miniscule openings are. As it is, someone can look at that and say "yes technically it's available, but not really" and move on, but with this change someone might see a header indicating one additional potential use, but the details don't show them where that was possible, leading them to think it's bugged.

If we're going to pursue this idea, plugging the forgiving code into getAdditionalUsageData would be preferable, so all references to how many uses were reasonably possible respect that, not just the table display.

AkairyuGestalter (changes requested):
Should've left that comment here, but the point stands.

AkairyuGestalter (approved):

PR #1941, opened 2024-03-22
MNK: Add April Fools dragon kick rotation module

obi-hardy-dev:
This is just an April fools goof module for monk. 

The logic of the rotation is that any filler should be dragon kick, until it's time to reapply twin fists buff and demolish, then, if applicable, consume leaden fist buff with a bootshine and go back to dragon kick spam until either a buff window or a twin application. 

Ends up looking like this 
image

Adds a checklist Item and table of missed dragon kicks opportunities. 

Check List:
image

Table:
image

A DK log: https://www.fflogs.com/reports/nwqbMNWQt4hBR9d8#type=damage-done&fight=last&source=2
A not as cool log: https://www.fflogs.com/reports/3qVWx4tDBwK7v8Mz#fight=67&type=damage-done&source=1674


src/parser/jobs/mnk/modules/DKOptimalGoof.tsx
```diff
@@ -0,0 +1,202 @@
...
+				this.dragonKicks.push(gcd)
+			} else {
+				gcd.reason = 'Not a dragon kick :('
+				this.missedKicks.push(gcd)
+			}
+		}
+	}
+
+	onComplete() {
+		this.checklist.add(new Rule({
```
AkairyuGestalter: My one concern with putting this in the checklist is that we might accidentally get a newer player thinking they really should be doing this! I'd be more comfortable putting this as a SEVERITY.MEMES Suggestion, since that way it's pretty clear that we're not serious about it. Otherwise, hilarious!

AkairyuGestalter (changes requested):
One note wrt the checklist. We'll see if this survives the cutover to Dawntrail, since the meme might not still be applicable after those job changes, but it'll still be funny for this year at least!

src/parser/jobs/mnk/modules/DKOptimalGoof.tsx
obi-hardy-dev: Understood, that makes sense, and it will give me a chance to play with suggestions! I'll rework to use suggestions instead. Do you think the output table is fine?  Is there something I should add there to make it clear its a meme?

AkairyuGestalter: I think the 'not a dragon kick (frown)' should be silly enough to be a hint that it's not real!

obi-hardy-dev: Removed the checklist item and added a suggestion 
image

obi-hardy-dev:
I'll close this for now but keep the code on my end in case the DK rotation is still a semi-optimal goofy option next expansion

PR #1942, opened 2024-03-24
Gauge Graph Cleanup

AkairyuGestalter:
This PR finishes the process of cleaning up the interface for working with the newer graphing options I added for the Enum/Set Gauge types, as well as accomplishing a few miscellaneous graph cleanup tasks I encountered along the way:
- A bunch of graphs were titled with raw text, instead of Trans tags
- Finally fixed the bug where MP recovery due to raising from death caused an MP graph to show for jobs that don't use the resource 
- - (See https://www.fflogs.com/reports/j3BhrDy6wkWKAf4G#fight=14 for an example, the BRD died on this one)
- Consolidated WAR and DRK gauge graphs into the timeline, since they were the final outlier jobs not doing so
- - This change should allow us to remove the chartjs dependency, finally
- Fixed a few minor graph row collapse state glitches for jobs that have only one gauge type

Aside the MP graph change, no specific logs to use for testing this. Any log with an affected job should suffice.


src/parser/core/modules/Gauge/CounterGauge.ts
```diff
@@ -48,11 +44,11 @@ export interface CounterGaugeOptions extends AbstractGaugeOptions {
 	deterministic?: boolean
 }
 
-export interface CounterChartOptions {
-	/** Label to display on the data set. */
-	label: string
-	/** Color to draw the data set in. Defaults to grey. */
-	color?: string | Color
+type CounterGraphOptions =
+	&Omit<ResourceGraphOptions, ''> // Not currently omitting any options, but making easier to do so in the future
```
hintxiv: Any reason why there's an & here?

It might be cleaner to have a base `GraphOptions` interface which `ResourceGraphOptions` and `CounterGraphOptions` (and others) extend.

src/parser/core/modules/Gauge/EnumGauge.ts
```diff
@@ -23,19 +22,22 @@ export interface EnumGaugeOptions extends AbstractGaugeOptions {
...
 	maximum: number,
 	/** Graph options. Omit to disable graphing in the timeline for this gauge. */
-	graph?: GaugeGraphOptions
+	graph?: EnumGraphOptions
 }
 
-interface EnumResourceData extends ResourceData {
-	base?: number
+type EnumGraphOptions =
+	& Omit<ResourceGraphOptions, ''> // Not currently omitting any options, but making easier to do so in the future
```
hintxiv: Same comment as above

src/parser/core/modules/Gauge/SetGauge.ts
```diff
@@ -21,18 +20,22 @@ export interface SetEntryOption {
 export interface SetGaugeOptions extends AbstractGaugeOptions {
 	options: SetEntryOption[],
 	/** Graph options. Omit to disable graphing in the timeline for this gauge. */
-	graph?: GaugeGraphOptions
+	graph?: SetGraphOptions
 }
 
-interface SetResourceData extends ResourceData {
+type SetGraphOptions =
+	&Omit<ResourceGraphOptions, 'tooltipHideMaximum'>
```
hintxiv: Same comment as above

src/parser/jobs/rpr/modules/OtherGauges.tsx
```diff
@@ -41,11 +41,18 @@ export class OtherGauges extends CoreGauge {
 
 	// Initialise our gauges - default max is 100 so yolo it is
 	private soulGauge = this.add(new CounterGauge({
-		graph: {label: 'Soul Gauge', color: Color(JOBS.WARRIOR.colour).fade(GAUGE_FADE), collapse: true},
+		graph: {
+			label: <Trans id="rpr.gauge.soulgauge.label">Soul Gauge</Trans>,
+			color: Color(JOBS.WARRIOR.colour).fade(GAUGE_FADE),
+			collapse: false,
```
hintxiv, line 47: Is `collapse: false` here intentional? I don't have a preference but currently the two MCH gauges are collapsed by default, and if we're changing it here then I'd prefer for it to be changed for MCH as well for consistency's sake.

src/parser/core/modules/Gauge/CounterGauge.ts
AkairyuGestalter: The & I think is leftover from a previous iteration. Removed.

The trouble with starting from a base GraphOptions and adding to it elsewhere is that the graphing code needs to have access to all of the options, regardless of which kind of graph it is to know how to display it, but we don't want some of the gauge types making use of some of the options since they either don't make sense or don't interact well with the others. Ack suggested enforcing that by narrowing the options list, so that's what I did here.

src/parser/jobs/rpr/modules/OtherGauges.tsx
AkairyuGestalter: I don't remember, but I like the way it looks better, so sure, we can do it for MCH (and DNC actually) too. There's a few other jobs that are still defaulting collapsed, but I'd want their maintainers to make the decision before mucking with them so I'll leave them be for now.

hintxiv (approved):

PR #1945, opened 2024-05-19
Dawntrail branch initial setup

AkairyuGestalter:
This PR contains a bunch of start-of-expac housekeeping and setup to prepare for Dawntrail.

Summary of changes:
- Add github config for the dawntrail branch
- Update the home page background image to one for the Arcadion raid series
- Add VPR and PCT to the job data and stub out their parser code
- Squash and remove the Endwalker action and status layer files
- Remove Endwalker patch-specific code from analysis code
- Purge the contributor and changelog files to give us a clean slate for Dawntrail contributions
- Set core and all jobs to an unsupported patch range
- Remove Endwalker patches from the patch list, and add 7.0 patch date
- Minor bugfix for the home page job support summary to keep jobs that aren't done from showing as supported once core gets marked clean for 7.0
- Other miscellanous cleanup and i18n re-extract


src/data/ACTIONS/root/PLD.ts
Toastdeib, line 1: Leaving this as a file comment because I apparently can't comment on lines that weren't modified - it looks like we de-comboed the Confiteor combo in `patch6.3.ts`, but the steps in the flattened file still have their combo properties. Is that correct, or should they be removed? I'm assuming correct since the actual tooltips still list the combo actions, but I wanted to confirm.

src/parser/jobs/dnc/changelog.tsx
```diff
@@ -1,105 +1,10 @@
...
-import React from 'react'
+// import CONTRIBUTORS from 'data/CONTRIBUTORS'
+// import React from 'react'
 
 export const changelog = [
 	// {
 	// 	date: new Date('2021-11-19'),
 	// 	Changes: () => <>The changes you made</>,
-	// 	contrubutors: [CONTRIBUTORS.YOU],
+	// 	contributors: [CONTRIBUTORS.YOU],
```
Toastdeib, line 8: lel

src/parser/jobs/mch/modules/Tincture.tsx
Toastdeib, line 1: These seem like functional changes rather than patch squishing; any reason for including them here? I'm assuming it doesn't actually change the logic but I'm curious.

src/reportSources/legacyFflogs/__tests__/__snapshots__/eventAdapter.ts.snap
```diff
@@ -39,7 +39,7 @@ Array [
 exports[`Event adapter individual events adapts applydebuff 1`] = `
 Array [
   Object {
-    "action": 2258,
+    "action": 2248,
```
Toastdeib, line 42: No clue what this file is; do you know what the change here means?

Toastdeib (commented):
Couple of small comments, but I think this looks good to go for the most part. I'm mainly curious about the one I left on the MCH Tincture file. I'm happy to approve and merge it once you've had a chance to respond.

src/parser/jobs/mch/modules/Tincture.tsx
AkairyuGestalter: It's pulling in the changes from #1946 so we don't lose them. That PR merged in after I branched this, and figured I'd grab it since this was still under review

src/parser/jobs/dnc/changelog.tsx
AkairyuGestalter: It bothered me since last expac and just didn't do anything about it then 🙃

src/reportSources/legacyFflogs/__tests__/__snapshots__/eventAdapter.ts.snap
AkairyuGestalter: It's some unit test file that broke now that the old Trick Attack raid buff status was fully gone from the data files. I had to switch it to the Mug raid buff status to get it to pass.

src/data/ACTIONS/root/PLD.ts
AkairyuGestalter: Yeah, I'm not sure what happened there, but they're definitely still a combo in-game. I re-checked it today and every step of Confetti -> Swordsx3 gets combo broken if you do your normal combo in between

src/parser/jobs/mch/modules/Tincture.tsx
Toastdeib: Are the `endwalker` and `dawntrail` branches fully diverged now? I don't actually know our branching structure for the major ones like these two, but I assumed we would want to merge the changes in that PR up from `endwalker` into `dawntrail` rather than duplicate them.

Toastdeib: After a discussion with ack on Discord, sounds like we can keep these without it being much of a problem.

Toastdeib (approved):
Marking this as approved. I'll let it sit a little longer before merging in case anyone else wants to chime in, but if not, I'll push the button this afternoon.

PR #1946, opened 2024-05-21
[MCH] Improve Reassemble counts under Tincture windows

hintxiv:


AkairyuGestalter (commented):
This doesn't contain any provision for updating the expectedPerWindow denominator count for Reassemble right now. Should it? Showing 2/1 Reassembles in a window seems odd.

hintxiv:
@AkairyuGestalter the expected count is still 1, since it's not always possible to get 2 in. However, if someone does get 2 in their pot window, it's confusing that the graph only shows 1/1.

(can I not reply to PR comments anymore? weird)

AkairyuGestalter (approved):
Good enough explanation for me!

PR #1947, opened 2024-05-27
Limit Breaks: mark as GCDs and add their cast & "recast" time

Hugmeir:
Marking limit breaks as GCDs makes them show up in the timeline; adding the cast time is pretty self-explanatory, but the "recast" time is actually cast time + animation lock, since that's how long you have to wait until you can pop another GCD.

Cast & animation lock times taken from https://www.akhmorning.com/allagan-studies/limit-break/actions-in-detail/

### DIFF
--- a/src/data/ACTIONS/root/LIMIT_BREAK.ts
+++ b/src/data/ACTIONS/root/LIMIT_BREAK.ts
@@ -7,6 +7,8 @@ export const LIMIT_BREAK = ensureActions({
 	SHIELD_WALL: {
 		id: 197,
 		name: 'Shield Wall',
+		onGcd: true,
+		gcdRecast: 1930,
 		icon,
 	},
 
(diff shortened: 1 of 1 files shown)


Hugmeir:
I know that _technically_ LBs aren't GCDs, you can weave them, but they have a cast time, and the animation lock effectively works as a long GCD recast.  The only other non-GCD with a cast time is BLU's Off guard & Peculiar Light, but those don't have a recast.

This seemed easiest than adding a special case just for LBs.

ackwell (approved):
thanks for getting all those recasts!

mmmmmmmh, i'm not _super_ keen on marking lbs as gcds semantically, despite it being easier - it's a bit muddy, and does mean we're technically getting a double whammy of the "short" gcd prior, and the abnormally "long" lb "gcd". speed calculations should plaster over this, but it doesn't sit right.

that said; ABC currently is calculated based on manual calculations of gcd (re)casts - so anything we do here _other_ than mark lbs as gcds is gonna cause jank.

---

i think the best _long term_ solution here would be to refactor ABC to utilise the cooldown tracking system - we could then mark LBs seperately from GCDs, but have a module set the GCD recast (or other similar effect).

for now, this should be Good Enough

PR #1949, opened 2024-06-09
[DRG] 7.0 Initial Support

ebshimizu:
Getting back into the code after a year or so away.

Drgtrail pre-launch to-do list:
- [x] scaffold skill data
- [x] scaffold procs
- [x] delete dragon sight
- [x] delete mirage dive eye gauge
- [x] update life surge consumers
- [x] update standard buff windows
- [x] rewrite the BotD module (likely going to change to a standard buff window with checks for overlapping LC+BL in the window)
- [x] Remove drift tables.
  - Reasoning: iirc the original intent of those tables was to help avoid LotD drift due to unintended delay of HJ and GSK. With LotD tied to GSK 'drifting' life isn't as much of an issue. The checklists should cover the 'number of times jump was used' cases, though seeing where specifically you were losing uses was rather useful. These can be put back if necessary.
- [x] update contributors

post-launch to-do list:
- [x] retrieve ids and icons
- [x] populate combo data
- [x] sanity check with log when available (checked with dummy log https://www.fflogs.com/reports/mYJAQdghzCq9PWGM#fight=last, can add to xiva guild)


ebshimizu:
i think that's a good enough start for 7.0 DRG. most of the changed simplify the analysis, will revisit this PR a bit after DT launch to add actual skill and status info and tweak things according to early guides.

Lemme know what needs changed with this PR. There's a few small core changes to note. One adds to add the `isRushedEndOfPullWindow` from `BuffWindow` to `TimedWindow`, and the other removes Dragon Sight from the raid buffs list (RIP)

AkairyuGestalter (approved):
Changes seem sane, and I got reasonable results from the logs I checked. I do see we haven't marked the job as supported yet, which is probably fine since we expect changes next week with Normal, but feel free to stick in another PR marking the job supported if you're happy with things.

PR #1951, opened 2024-06-10
Retrieve logged level for content and display in about

AzariahAeroborn:
Draft PR for initial interface and criticism.  At a minimum, I want to clean up the About.tsx file further before this merges this way (and it looks like I may need to touch up the unit tests for the ActorUpdate synths as well)

### DIFF
--- a/src/parser/core/modules/About.tsx
+++ b/src/parser/core/modules/About.tsx
@@ -8,14 +8,18 @@ import {AVAILABLE_MODULES} from 'parser/AVAILABLE_MODULES'
 import {Analyser, DisplayMode} from 'parser/core/Analyser'
 import React from 'react'
 import {Header} from 'semantic-ui-react'
+import {dependency} from '../Injectable'
 import styles from './About.module.css'
+import {Actors} from './Actors'
 import DISPLAY_ORDER from './DISPLAY_ORDER'
 
 export default class About extends Analyser {
-	static handle = 'about'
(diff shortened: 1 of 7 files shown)


src/parser/core/modules/About.tsx
```diff
@@ -25,6 +29,7 @@ export default class About extends Analyser {
 	//		from: ...,
 	//		to: ...,
 	// }
+	supportedLevels = null
```
ackwell: hm. i'm not sure how to feel about explicitly mentioning "support". iiuc, we're not really trying to support synced content, but rather making sure synced content isn't missing data completely in the timeline.

ackwell (commented):
general setup makes sense. i'm not sold on calling out support ranges explicitly, happy to discuss.

src/parser/core/modules/About.tsx
AzariahAeroborn: Removed the supportedLevels support range.

AzariahAeroborn:
I updated the way to access level to make it a getter on ActorResources -- while I was digging into this, found that my initial implementation besides being a pain to use also wound up crawling the actor updates a bunch (every time the ActorResources constructor got called) to retrieve the level property.  Getter seems to work as expected since the actorUpdate with the actor's level is going to be the first packet we receive for any actor in a given pull.

ackwell (approved):

PR #1952, opened 2024-06-11
Initial NIN changes for Dawntrail

Toastdeib:
Bit of a big PR (sorry about that), but it encompasses most of the changes we'll want prior to having the finalized tooltips/rotational guidance and the actual IDs/icons available. Changes in this PR include:
* Updates for action and status data (some new, some changed, some removed)
* Text copy changes to account for new skill names (Mug -> Dokumori, Trick Attack -> Kunai's Bane) as well as some minor cleanup
* Removal of the Huton module
* Conversion of the Huton attack speed modifier module into a `SpeedAdjustments` subclass (based on MNK's Greased Lightning implementation)
* Adjustments to the Ninki module to account for new spenders (Zesho Meppo and Deathfrog Medium), as well as Mug's upgrade to Dokumori
* Adjustments to the Ninjutsu and Kassatsu modules to account for the new version of Huton
* A new (untested) module for tracking and providing feedback on Kazematoi stack usage
* Updates to the (now refactored) Kunai's Bane window module to check for Tenri Jindo (if TCJ was used) and allow one Armor Crush in the opener


src/parser/jobs/nin/index.tsx
```diff
@@ -16,12 +16,12 @@ export const NINJA = new Meta({
 	</>,
 
 	supportedPatches: {
-		from: '✖',
-		to: '✖',
+		from: '7.0',
+		to: '7.0',
```
ackwell: If you want; can revert this back to ✖-support, get it merged in, then have a smaller pr for 7.0 support once we've got the rest of the details?

src/parser/jobs/nin/modules/KunaisBaneUsage.tsx
ackwell, line 1: gotta love the "we swear we changed something" renames

src/parser/jobs/nin/modules/KunaisBaneWindow.tsx
```diff
@@ -93,17 +93,21 @@ export class TrickAttackWindow extends BuffWindow {
...
+				{
+					action: this.data.actions.TENRI_JINDO,
+					expectedPerWindow: 1,
+				},
 			],
 			suggestionIcon,
-			suggestionContent: <Trans id="nin.taWindow.suggestions.trackedactions.content">
-				Every <ActionLink action="TRICK_ATTACK"/> window should contain <ActionLink action="HYOSHO_RANRYU"/>, 2 <ActionLink action="RAITON"/> casts, and <ActionLink action="DREAM_WITHIN_A_DREAM"/> in order to maximize damage.
+			suggestionContent: <Trans id="nin.kb-window.suggestions.trackedactions.content">
+				Every <ActionLink action="KUNAIS_BANE"/> window should contain <ActionLink action="HYOSHO_RANRYU"/>, 2 <ActionLink action="RAITON"/> casts, and <ActionLink action="DREAM_WITHIN_A_DREAM"/> in order to maximize damage.
```
ackwell, line 103: should this also mention `TENRI_JINDO`? if so, wondering if the text should be rewritten - the expectations are already pretty clear due to the rest of the info, worried it's getting a little dense

src/parser/jobs/nin/modules/Ninki.tsx
```diff
@@ -69,19 +72,23 @@ export class Ninki extends CoreGauge {
...
 		// Spenders
 		[this.data.actions.BHAVACAKRA.id, {action: -50}],
+		[this.data.actions.ZESHO_MEPPO.id, {action: -50}],
 		[this.data.actions.HELLFROG_MEDIUM.id, {action: -50}],
+		[this.data.actions.DEATHFROG_MEDIUM.id, {action: -50}],
 		[this.data.actions.BUNSHIN.id, {action: -50}],
 	])
 
-	private erroneousFrogs: number = 0 // This is my new band name
+	// Honestly not sure which of these is the better band name
```
ackwell, line 92: depends on the genre, really.

src/parser/jobs/nin/index.tsx
Toastdeib: Makes sense; I can flip that back, especially since most of the module changes are me making assumptions and not actually concrete info from the math wizards.

src/parser/jobs/nin/modules/KunaisBaneWindow.tsx
Toastdeib: It doesn't mention Tenri Jindo for the same reason it doesn't mention Ten Chi Jin; it's only expected in every other window. That said, I could rewrite it to something that doesn't mention the actions explicitly and just refers users to the table instead. Maybe something like:
```ts
Every <ActionLink action="KUNAIS_BANE"/> window should contain several important high-damage abilities detailed in the Kunai's Bane Window table.
```
Let me know what you think, I'm good with either option.

src/parser/jobs/nin/modules/KunaisBaneWindow.tsx
```diff
@@ -130,16 +135,30 @@ export class TrickAttackWindow extends BuffWindow {
...
-			// We push one Raiton outside the Trick window in the opener to fit an extra Bhava
+			// We push one Raiton outside the Kunai's Bane window in the opener to fit an extra Bhava
 			if (bufferedWindowStart <= this.parser.pull.timestamp) {
 				return -1
 			}
 		}
 
+		if (action.action.id === this.data.actions.TENRI_JINDO.id) {
+			// Tenri Jindo should only be expected when the window contains TCJ
+			if (window.data.find(value => value.action.id === this.data.actions.TEN_CHI_JIN.id) !== undefined) {
```
AkairyuGestalter: I don't think this check is behaving correctly, this log has 1/0 expected on the TCJ windows, and 0/1 on the non-TCJ windows, instead of 1/1 and 0/0:
http://127.0.0.1:3000/fflogs/Mbd9P6V4fvCN8tDG/4/293

AkairyuGestalter (changes requested):
Found what I think is a goof in the Kunai's Bane Window tracking for Tenri Jindo. Otherwise the stuff in there seems fine, so if we get that and the merge conflicts fixed this should be good to go.

src/parser/jobs/nin/modules/KunaisBaneWindow.tsx
Toastdeib: I feex.

AkairyuGestalter (approved):
Changes lgtm, I press button

PR #1957, opened 2024-06-25
DO NOT MERGE: Emergency revert for icon update.

ackwell:
This will update the icon utility to redirect icons back to the old xivapi static endpoint.

Please do _not_ merge this unless required. cc @supamiu, drop a line here, on discord, or anywhere if server isn't coping and this needs to be rolled back.


ackwell:
seems stable. closing - can reapply if something goes boom.

PR #1974, opened 2024-07-06
Fixed the Kazematoi gauge

Toastdeib:
Small PR this time; the Kazematoi gauge wasn't appearing due to a shared handle with the Ninki gauge, so those both have their own explicitly unique ones now. I also updated the colors so they wouldn't be identical, for a clearer visual.
image


AkairyuGestalter (approved):
Silly me missing those in the previous PR review, the Jindo bug blinded me 🙃

PR #1980, opened 2024-07-10
Dawntrail - PCT gauge state tracking

AkairyuGestalter:
## Pull request type

- [X] This is a patch bump

## Pull request details

This PR adds gauge state tracking for Pictomancer. No suggestions or checklist stuff yet. A few notes for reviewers to look for:
- Technically speaking, under the hood black and white paint are the same thing, and having Monochrome Tones just forces the next paint consumer to be Comet in Black, while showing one paint as Black. I'm simulating it here by just saying "if Subtractive is used with no paint, generate a black next time you would have generated white"
- In order to compress the display better, I intend to make a SetEnumGauge (EnumSetGauge?) that functions like a set, but allows for Enums to be in one of the set slots. This will be used for the Canvas gauge, where the weapon and landscape slots are set, but the creature slot should realistically be an enum to properly display _which_ motif is ready. This will be done on a separate future PR, so I've done a "good enough" implementation as a simple set gauge for now.
- SetGauge and EnumGauge don't have history correction the way CounterGauge do, so I'm doing some switch case fallthrough shenanigans to get close to the behavior. Please don't kill me :(
- There was also a bug with SetGauge's spend function, that nobody had used yet anyways. It's fixed now lol

## Testing / Validation

- [X] I used the log(s) listed below to develop and test this bugfix: http://localhost:3000/fflogs/c4tGYTC1LzmDrAwy/24/79

Realistically, any PCT log will be fine

## Job Maintenance


src/parser/jobs/pct/modules/Gauge.tsx
```diff
@@ -0,0 +1,367 @@
...
+		case this.data.actions.WINGED_MUSE.id:
+			if (!this.depictionGauge.getStateAt(WING_MOTIF)) {
+				this.depictionGauge.generate(WING_MOTIF)
+			}
+		case this.data.actions.POM_MUSE.id:
+			if (!this.depictionGauge.getStateAt(POM_MOTIF)) {
+				this.depictionGauge.generate(POM_MOTIF)
+			}
+			/* eslint-enable no-fallthrough */
+		}
```
AzariahAeroborn, line 346: Shouldn't we be able to generalize this history correction so that we can make it available to all SetGauge consumers?  SAM stickers are going to have a similar problem when the user uses Midare after only one sticker is generated in a pull.

src/parser/jobs/pct/modules/Gauge.tsx
```diff
@@ -0,0 +1,367 @@
...
+			break
+		case this.data.actions.FANGED_MUSE.id:
+			this.portraitGauge.generate(MADEEN_PORTRAIT)
+			break
+		}
+	}
+
+	private onComplete() {
+		// Suggestions and stuff
+	}
```
AzariahAeroborn: Remove unused code or implement suggestions.

AzariahAeroborn (changes requested):

src/parser/jobs/pct/modules/Gauge.tsx
AkairyuGestalter: Theoretically yes, but I'd like to do a separate PR to add core history correction to SetGauge, rather than try to stuff it onto this one.

AzariahAeroborn (approved):

PR #1999, opened 2024-07-13
RDM Baseline DawnTrail

Cephid:
## Pull request type

- [X] This is new functionality or an addition to existing functionality
- [X] This is a patch bump

## Pull request details

- [ ] This is in response to a discussion or thread on Discord: *[Link to thread or start of discussion]*

		-Added New Actions and Statuses for DawnTrail
		-Modified existing Modules to account for DawnTrail modifications
		-Added new Modules for DawnTrail Additions
		-All Potencies have been adjusted, any functionality such as combo breaking has been adjusted
		-This does not fix the issue with Manafication not breaking combos, or address the issue of the faster GCD for RDM Melee combo breaking the estimated GCD count for BuffWindow for tracking stacked buffs consumed on use.  These issues are the only reason we are not marking the job as supported with this submission.

## Testing / Validation

- [] The changes being implemented with this bugfix include comments that contain example log(s) to test with

https://www.fflogs.com/reports/7BHtJ6kDYKp412a9 - Good Manafication, Embolden, and Grand Impact usage
https://www.fflogs.com/reports/RAqxQcVp23WgnfKM - Failure to use Grand Impact
https://www.fflogs.com/reports/NJyZkQAtVwKYMXfC - Failure to use Prefulgence and Vice of Thorns
https://www.fflogs.com/reports/nKabzNchfwA7RrD4 - Failure to finish Manafication Combo, proper usage of Vice of Thorns
https://www.fflogs.com/reports/y9bc6qpf4KgVnvTX - Complete Failure to use Manaficaiton and Magicked Swordplay stacks.

## Job Maintenance


src/data/ACTIONS/root/RDM.ts
```diff
@@ -69,7 +69,33 @@ export const RDM = ensureActions({
...
+		id: 37007,
+		name: 'Prefulgence',
+		icon: iconUrl(3244),
+		potency: 900,
+	},
+	MAGICK_BARRIER: {
+		id: 25857,
+		name: 'Magick Barrier',
+		icon: iconUrl(3237),
+		speedAttribute: Attribute.SPELL_SPEED,
```
AzariahAeroborn: I'm fairly sure this cooldown doesn't scale with speed, it's just the instant OGCD defensive, correct?

I guess you moved an existing value, but probably worth correcting it while we're here.

src/parser/jobs/rdm/modules/ManaGauge.tsx
```diff
@@ -179,6 +158,34 @@ export class ManaGauge extends CoreGauge {
...
+	}
+
+	//Honestly I don't really need to do anything here; but I needed something to hook for onApplyMagickedSwordPlay
+	//This code was pretty much taken as is from the BloodGauge handling for Blood Weapon
+	private onHitUnderMagickedSwordPlay(event: Events['damage']) {
+		if (event.cause.type === 'status') {
+			return
+		}
+	}
+
```
AzariahAeroborn: Since we're not doing anything with the hook for attacks under Magicked Swordplay, we shouldn't include any of this.

However, we do need to do _something_ so that onGaugeSpender doesn't decrement gauge while Magicked Swordplay is in effect (easiest thing to do would be to check for hasStatus(MAGICKED_SWORDPLAY) at the top of the gauge spender method and bail early, would just need to make sure we don't have event ordering hijinx between when the action (cast) event happens and when the buff falls off on consuming the final stack.

src/parser/jobs/rdm/modules/Procs.tsx
```diff
@@ -38,6 +42,20 @@ export class Procs extends CoreProcs {
...
+		//We are uncertain if we want this here or on their own, for now they'll be on their own.
+		//Leaving this here in case we change our minds
+		// {
+		// 	procStatus: this.data.statuses.PREFULGENCE_READY,
+		// 	consumeActions: [this.data.actions.PREFULGENCE],
+		// },
+		// {
+		// 	procStatus: this.data.statuses.THORNED_FLOURISH,
+		// 	consumeActions: [this.data.actions.VICE_OF_THORNS],
+		// },
```
AzariahAeroborn, line 58: Is the only reason to separate these out to have separate timeline rows for them?  If so, would we be better off to add a property to CoreProcs to allow consumers to choose to show the Procs rows expanded by default (set up line 260 in CoreProcs to use collapse: this.collapseProcsRowsInTimeline and then let consumers override that to false if they have multiple distinct procs they want to show separated by default)?

I think the UI would be much better with putting in a way to show the Procs group as uncollapsed by default, rather than having 3 rows that all just are titled "Procs" from the different modules...

Roughly after (aside from all of the "Procs" headers being one header for all 4 rows):
image

As is with separated modules:
image

If we do keep separated modules, remove this commented code (we can always re-add it later).

src/parser/jobs/rdm/modules/Procs.tsx
```diff
@@ -136,6 +154,42 @@ export class Procs extends CoreProcs {
...
+			return <Trans id="rdm.procs.suggestions.prefulgence.content">
+				Try to consume <DataLink status="PREFULGENCE_READY"/> before it expires as <DataLink action="PREFULGENCE"/> is your strongest skill.
+			</Trans>
+		}
+		if (id === this.data.statuses.THORNED_FLOURISH.id) {
+			return <Trans id="rdm.procs.suggestions.viceofthorns.content">
+				Try to consume <DataLink status="THORNED_FLOURISH"/> before it expires as <DataLink action="VICE_OF_THORNS"/> Gives 3 White & Black Mana and is one of your strongest GCDs outside of your Finisher Combo.
+			</Trans>
+		}
+	}
```
AzariahAeroborn: There's no real value in this separation here, since it's just hard coded Trans tags.  It'd be clearer to just put the respective Trans tags directly in their suggestion.

src/parser/jobs/rdm/modules/Procs.tsx
```diff
@@ -136,6 +154,42 @@ export class Procs extends CoreProcs {
...
+			return <Trans id="rdm.prefulgence.suggestions.dropped.why">
+				<DataLink status="PREFULGENCE_READY"/> timed out <Plural value={this.getDropCountForStatus(this.data.statuses.PREFULGENCE_READY.id)} one="# time" other="# times"/>
+			</Trans>
+		}
+		if (id === this.data.statuses.THORNED_FLOURISH.id) {
+			return <Trans id="rdm.viceofthorns.suggestions.dropped.why">
+				<DataLink status="THORNED_FLOURISH"/> timed out <Plural value={this.getDropCountForStatus(this.data.statuses.THORNED_FLOURISH.id)} one="# time" other="# times"/>
+			</Trans>
+		}
+	}
```
AzariahAeroborn: Same comment as getCriticalProcContent.

src/parser/jobs/rdm/modules/Procs.tsx
```diff
@@ -154,6 +208,31 @@ export class Procs extends CoreProcs {
...
+		// }))
+
+		// this.suggestions.add(new TieredSuggestion({
+		// 	icon: this.data.actions.VICE_OF_THORNS.icon,
+		// 	content: this.getCriticalProcContent(this.data.statuses.THORNED_FLOURISH.id),
+		// 	tiers: SEVERITY_CRITICAL_PROCS,
+		// 	value: this.getDropCountForStatus(this.data.statuses.THORNED_FLOURISH.id),
+		// 	why: this.getCriticalProcContent(this.data.statuses.THORNED_FLOURISH.id),
+		// }))
+
```
AzariahAeroborn: See comment above for commented code when tracking these procs.  If we do keep separated modules, remove this commented code.

src/parser/jobs/rdm/modules/index.tsx
```diff
@@ -7,12 +7,19 @@ import {DualCast} from './Dualcast'
...
 import {ManaStackGauge} from './ManaStackGauge'
 import {MeleeCombos} from './MeleeCombos'
+import {Prefulgence} from './Prefulgence'
 import {Procs} from './Procs'
 import {Swiftcast} from './Swiftcast'
+import {ViceOfThorns} from './ViceOfThorns'
 import {Weaving} from './Weaving'
+//For now commenting this out, it's uncertain if we want GrandImpact to be within Procs or on its own.
+//For now it'll be contained within Procs
+//import { GrandImpact } from './GrandImpact'
```
AzariahAeroborn: See comments on Procs module, remove commented code and unused modules as necessary.

src/parser/jobs/rdm/modules/MeleeCombos.tsx
```diff
@@ -179,17 +180,20 @@ export class MeleeCombos extends Analyser {
...
 			how they were when they were one large combo.  As such for now this fix is no longer a bandaid but a permanent fixture.
 			*/
-			if (action.id === this.data.actions.MANAFICATION.id &&
-				current && current.data.lastAction.action === this.data.actions.ENCHANTED_REDOUBLEMENT.id) {
-				return
-			}
+			// if (action.id === this.data.actions.MANAFICATION.id &&
+			// 	current && current.data.lastAction.action === this.data.actions.ENCHANTED_REDOUBLEMENT.id) {
+			// 	return
+			// }
```
AzariahAeroborn: Did they change the behavior of Manafication to drop Red Magic stacks if you cast it between Redoublement (or Moulinet Trois) and VerFinisher?  If not, we need to keep this and make it check the AOE melee finisher too to keep the "unified combo" into the VerFinisher combo in place.

AzariahAeroborn (changes requested):
Main concern is the MagickedSwordplay and MeleeCombos behaviors.  I won't block on the Procs comments if you want to keep it as is, but I think it would be better to have things consolidated for future-us's maintenance reasons.

src/data/ACTIONS/root/RDM.ts
Cephid: I'll remove, I didn't even look at it; I just noticed it was down with the casted GCDs instead of with all the ability oGCDs.

src/parser/jobs/rdm/modules/ManaGauge.tsx
Cephid: It appears as though the action happens before the status modification.  I've removed all of this, added a dependency on actors so I can check hasStatus and added a check to onGaugeSpender.  Below is an example of the ordering of events.

image

src/parser/jobs/rdm/modules/Procs.tsx
Cephid: Correct, Thorned Flourish and Prefulgence are 2 of the highest potency skills, and the feedback I got was that those might be better served as having their own timeline rows.  So I commented out the code that had them in the base Procs module in case they changed their minds after having it for a while.  The intention is, whatever they decide I'll cleanup the method that wasn't wanted.  Since I already had code for both ways, I decided to just comment it out so it'd be a quick change if they decide to change it to all in one, or moving Grand Impact out into it's own row and keeping VerStone/VerFire under 1 proc.

As for the UI change for core; my thought would be as part of the proc declaration have title, and ownRow or some other bool named to denote it'll get it's own row.  Then on the if (this.showProcTimelineRow check for that ownRow bool and if so, add a new simple row with no collapse using the provided label; otherwise do what it does today.  The idea being it'd appear separate, not have the collapse bit and just look like any other tracked timeline skill.

src/parser/jobs/rdm/modules/Procs.tsx
Cephid: Moved and Removed.

src/parser/jobs/rdm/modules/Procs.tsx
Cephid: Moved and Removed

src/parser/jobs/rdm/modules/Procs.tsx
Cephid: That is the plan, this code is only being kept in case they want it all under the one proc window.  However if I don't get much feed back after the release, then I'll just remove these and keep it with the setup I have implemented currently

src/parser/jobs/rdm/modules/index.tsx
Cephid: I went ahead and removed it here, it'll be simple enough to re-add the 2 lines.

src/parser/jobs/rdm/modules/MeleeCombos.tsx
Cephid: This fix was put in place to handle for the old method of

Melee Combo --> Manafication --> Finisher.

This was because Manafication didn't clear your mana stacks, but it did give you 50|50 + the 6x potency buff stack; so if you were over 50|50 you'd spend first then manafic to rush the buff onto the finisher.  In application the melee and finisher combo are 2 separate combos, but for visualization and modeling strategies we treat them as one this is why the special handling was here.  

However, now that we don't have to worry about overcapping when using manafication; and the risk of losing out on Prefulgence if you are forced to disengage before you get to the melee end of the combo if you pop it before a finisher series, I removed it for the moment.  I still have a pending question out to the mentors on the balance, but haven't gotten a response.  As such I left it here but commented out until I do.

AzariahAeroborn: But since we still have Manafication marked as breaks combo (and it **does** need to be, since it does break combos), commenting this out means that if someone does use MF in between the sword combo and the finisher combo we're now going to record that as a break.  Testing the melee combos on a dummy, it still does not break the longer combo if you use MF in between sword combo and finisher combo and the same general behavior with both the ST and AOE melee combos, so we need to not remove this and also need to add in the "don't break combo if MF used after AOE sword combo" check.

Cephid: Manafication should break combos (within the melee or within the finisher combo) but it doesn't because of a bug in the core combos module that only considers damage actions which is another issue entirely.  I'll add this back in and modify the final check on the action to be either Redoublement or Moulinet Trois.  I also am fixing the extract error I introduced in my previous check in.

AzariahAeroborn (approved):

PR #2007, opened 2024-07-14
[RPR] 7.0 Reaper Actions & Status Updates

Rjb99:
## Pull request type

- [x ] This is new functionality or an addition to existing functionality

## Pull request details
- [ x] The goal of this PR is detailed below:

	Adds 7.0 Reaper Actions and Statuses, and implements a coreProc module for Harpe to handle the CDR effect from enchanced harpe

## Testing / Validation
 I just grabbed this log:
 https://www.fflogs.com/reports/4gKT6n3NxcdB2fzM#fight=12&type=damage-done
 and made sure that all the icons looked correct.

## Job Maintenance
- [x ] I am active on the xivanalysis Discord and part of the job maintainers group for the job(s) in this PR


src/parser/jobs/rpr/modules/Harpe.tsx
Toastdeib, line 1: _Note:_ Unless I'm misunderstanding something in the core module, this doesn't seem like it'll have any output, which I assume is by design. Are you planning to add to it later? Is it going to track any other procs? If yes, it's probably worth naming it `Procs.tsx` instead of `Harpe.tsx` now so you don't have to refactor later. And if I _am_ misunderstanding something and this does have output, you should include a changelog entry for it.

Toastdeib (commented):
I double-checked all the data changes and they look good. I did have a couple of questions on the Harpe module, but once those are addressed I'm happy to approve and merge this.

src/parser/jobs/rpr/modules/Harpe.tsx
Toastdeib: Quick follow-up, because I saw your PR description right after posting this - I'm assuming the module is just intended to make sure the CDR stuff fires and we track Hell's Ingress/Egress correctly. I also see that RPR already has another module that overrides CoreProcs (which is just called Procs); I think this does require action, in that case, because neither that module nor this one override the handle, meaning one of them will clobber the other. You should override it in both modules for clarity and then this should be good to go.

Rjb99: Yeah Harpe is just going to be CDR. I kept it separate so I didn't have to worry about making dropped proc exceptions on the other Core Procs file we use for the soul reavers and enhanced actions. As you can and should drop Enchanced Harpe quite often. I felt this implementation would be cleaner.

I'll add the overrides to the handles later today.

Rjb99: handles have been added

AzariahAeroborn (approved):

PR #2009, opened 2024-07-14
RDM: Magic finishers under Magicked Swordplay are fine

Hugmeir:
## Pull request type

## Pull request details

Previously we were dinging people for this:

    `Enchanted Riposte`
        => `Enchanted Zwerchau`
            => `Enchanted Redoublement`
                => `Manafication`
                    => magic finishers
                        => second melee combo

But that's actually a pretty common way to use Manafication, in order to fit two sets of finishers under buffs, or three under pot.

This commir adds the magic finishers to the list of allowed GCDs when the Magicked Swordplay buff is active.

## Testing / Validation

https://www.fflogs.com/reports/6YT2JarVHFk8BbDA#fight=22&source=71
(apologies, am a console player, so the logs don't belong to me)


Cephid:
I checked with Mazz over on the balance, the use of this particular series of skills is more of an advanced choice now; so I'd agree with the change here to let Magicked Swordplay allow the use of a finisher combo before expending the sword play stacks.

AzariahAeroborn (approved):

PR #2014, opened 2024-07-18
Combo Breaks for Non Damage Combo Breakers

Cephid:
## Pull request type

- [X] This is a patch bump

## Pull request details

- [ ] This is in response to a discussion or thread on Discord: *[Link to thread or start of discussion]*

	Currently a non damaging action that is a combo breaker, such as the RDM Manafication action does not currently break combos.  This PR will give the ability for an implementing class to specify a non damaging action and pass it down to be fabricated into the expected EventDamage object if the action was used to break a combo.  Additionally this PR also increases the duration of the Manafication buff to 30 seconds that was changed in patch 7.01.  As the Combo Breaking function was the original blocker for marking the class supported I've also marked the class supported for 7.0-7.01.  There are currently 2 other RDM BugFixes I'd rather get attention before this PR however.

- [ ] The job(s) in this patch bump had only potency changes for this patch, which do not effect analysis
  - [ ] I have updated the data files for all potency changes for this patch
  - [ ] The skills and jobs affected by these potency changes do not track potency for xivanaylsis purposes

## Testing / Validation

   https://www.fflogs.com/reports/7MmY9PTKjG4avnQD - Manafication Breaking a Melee Combo

## Job Maintenance


src/parser/jobs/rdm/modules/Combos.tsx
```diff
@@ -36,6 +37,22 @@ export class Combos extends CoreCombos {
...
+	override initialise() {
+		super.initialise()
+
+		//Manafication breaks combos, so lets ensure we break when we see it, if we're in a combo!
+		this.addEventHook(
+			filter<Event>()
+				.type('action')
+				.source(this.parser.actor.id)
+				.action(ACTIONS.MANAFICATION.id),
+			this.onTrackedCast)
```
AzariahAeroborn: _Action:_ Starting a cast of any spell will break either combo for RDM too, so we also need to account for those.  Though starting and cancelling Jolt et al. doesn't waste your Red Mana stack(s) or drop or consume Magicked Swordplay.

src/parser/jobs/rdm/modules/Combos.tsx
```diff
@@ -36,6 +37,22 @@ export class Combos extends CoreCombos {
...
+	}
+
+	private onTrackedCast(event: Events['action']) {
+		this.onNonDamageCast(event)
```
AzariahAeroborn: _Consider:_ If you don't wind up needing to put special casing in here to treat starting a spell cast differently from using Manafication, you probably can just pass this.onNonDamageCast in to the addEventHook directly and not need this wrapper (I would expect so anyway, since onNonDamageCast is protected it should be accessible to implementers).

AzariahAeroborn (changes requested):

src/parser/jobs/rdm/modules/Combos.tsx
Cephid: Will those get caught by Events['damage'] filters?  Since they are damage based spells?  Or do cancelled actions not get picked up?  If they will get caught by Events['damage'] even if they didn't finish casting, then the Core module will handle them as normal; if not then how would one go about taking an action but only if it was cancelled via a filter on Event?  I wouldn't want to hook their action hit if a cast finished and the damage gets caught as we would double evaluate them.

src/parser/jobs/rdm/modules/Combos.tsx
Cephid: Oh good thought, I was doing something up there prior, but I discovered I could simplify what I was doing and removed it.  I'll make that particular change.

src/parser/jobs/rdm/modules/Combos.tsx
AzariahAeroborn: The 'prepare' event is the begin cast action and is what does the combo break for damaging spells.  No 'prepare' event is emitted for instants.
The 'action' event is when an action actually resolves and will always come after 'prepare' for channeled spells.
The 'damage' event corresponds to the "Player prepares Skill on Target" line in FFLogs (the calculateddamage) event and is when the damage is being snapshot.

Core combos won't double ding if it gets two recordBrokenCombo events off of the same cast, since the first one from the 'prepare' event will break the combo and clear out the active combo, so when the 'damage' event comes through, it won't see an active combo chain to break.  So we really do want to hook both - if someone breaks by casting a spell with Swiftcast / Acceleration, it'll be an instant and have an 'action' and 'damage' event at the same time but no 'prepare' event, while if someone starts a channel and then interrupts it because they realized they oopsed, there would be a 'prepare' event but no 'damage' event and waiting until whatever they did next will record the wrong combo breaking time (and action if they did something else after the interrupt).

ackwell:
merged as part of #2021

PR #2017, opened 2024-07-19
Dawntrail - BLM final Dawntrail support updates

AkairyuGestalter:
## Pull request type

- [x] This is a patch bump

## Pull request details

This PR adds/updates two RotationWatchdog evaluators, and includes some other minor cleanup notes to mark BLM as supported for Dawntrail:
- Updates the ExtraF1Evaluator to treat all usage of F1 as bad, and reworded the suggestion to explain why (it uses MP needed to get enough F4s for Flare Star)
- Adds an exaluator to check what Firestarter was used for, and to suggest using it for exiting Umbral Ice with Transpose, instead of extending the Astral Fire phase, unless it was effectively a freebie due to Manafont.
- - Note that this suggestion is always a minor one, since it's a pretty small DPS increase, and Firestarter as a phase extension/movement tool may be necessary on a fight-by-fight basis
- Updated the rotation table display to treat Flare Star carryover after Manafont as a 0/- neutral result instead of a 0/0 positive result.
- Swapped icons for the dropped Firestarter suggestion in Procs to the Firestarter status icon, to instead use the F3P icon for the FirestarterUsageEvaluator's suggestion

## Testing / Validation

- FirestarterUsage: http://localhost:3000/fflogs/tvf3nPAgm1hQFzwd/12/462
- Skip F1: http://localhost:3000/fflogs/Y1dVw8npCgbvZcjf/19/131

## Job Maintenance


src/parser/jobs/blm/changelog.tsx
```diff
@@ -8,6 +8,11 @@ export const changelog = [
 	// 	Changes: () => <>The changes you made</>,
 	// 	contributors: [CONTRIBUTORS.YOU],
 	// },
+	{
+		date: new Date('2024-07-19'),
+		Changes: () => <>Update suggestion to avoid using <DataLink action="FIRE_I" />, add evaluation for the usage of <DataLink status="FIRESTARTER" />, and mark as supported for Dawntrial.</>,
```
AzariahAeroborn:
```suggestion
		Changes: () => <>Update suggestion to avoid using <DataLink action="FIRE_I" />, add evaluation for the usage of <DataLink status="FIRESTARTER" />, and mark as supported for Dawntrail.</>,
```

Spelling

src/parser/jobs/blm/modules/RotationWatchdog.tsx
```diff
@@ -347,6 +352,12 @@ export class RotationWatchdog extends RestartWindow {
...
 			if (windowMetadata.finalOrDowntime) { return window.data.filter(event => event.action.id === this.data.actions.FIRE_IV.id).length }
 
+			// Windows begun by Manafont by definition should contain the full 6 F4s
+			if (window.data[0].action.id === this.data.actions.MANAFONT.id) {
+				windowMetadata.expectedFire4s = MAX_POSSIBLE_FIRE4
+				return MAX_POSSIBLE_FIRE4
+			}
+
 			// Start off with the baseline assumption they've reached full MP in Umbral Ice, since MP events from the log source aren't reliably timed
 			adjustment = NO_UH_EXPECTED_FIRE4
```
AzariahAeroborn, line 362: _Nit:_ Seeing this in passing, but since now MP generation is deterministic based on casts, can we do without this fudge factor?

AzariahAeroborn (changes requested):

src/parser/jobs/blm/modules/RotationWatchdog.tsx
AkairyuGestalter: Potentially, but the opener window is still going to be susceptible to not knowing precisely how much MP was left, in things like 24 mans, if the player didn't umbral soul their way to full MP. Can put it on the future/make things better list, but I don't think it's worth holding up support on.

AzariahAeroborn (approved):

PR #2018, opened 2024-07-19
[WAR]: add new job actions

kwerey:
## Pull request type

- [ ] This is a patch bump

## Pull request details

- [x] This is in response to a discussion or thread on Discord: https://discord.com/channels/441414116914233364/470049767477608459/1263884995248918558

		* This PR adds action and status data for the new job actions introduced for Warrior with Dawntrail.
		* It will stop XIVA reporting incorrectly low GCD uptime and make the timeline more useful
		* I do **not** consider this "7.0 support" since it does not include any recommendations based on the new actions.

Questions: 
* Should Vengeance (war 30% mit, now replaced with 'Damnation') be removed from the existing action data?
* Should any of the new buffs be added to the buffs section of the ActionTimeline module? 

The new buffs Warrior gained are as follows:

* "Wrathful" and "Primal Ruination Ready" indicate there is a proc available to be used (oGCD Primal Wrath and GCD Primal Ruination) 
* "Burgeoning Fury" stacks need to be built to 3 to enable usage of Primal Wrath

These aren't damage buffs, there is no benefit to be gained from using anything under them - but Infuriate is listed in this section currently and the same is true of that.

## Testing / Validation

https://www.fflogs.com/reports/a:XTy9wVgHv8K1b6jc#fight=15

## Job Maintenance

Discussed the changes here in the dev-tank channel on discord, happy to discuss them either here or there. I'm sure exactly what the "part of the job maintainers group for the job" checkbox means though!


src/data/STATUSES/layers/patch7.01.ts
```diff
@@ -1,9 +1,19 @@
...
 import {StatusRoot} from '../root'
 
 export const patch701: Layer<StatusRoot> = {
 	patch: '7.01',
 	data: {
 		// Patch 7.01 statuses
+
+		// Warrior's Primal Ruination proc reduced from 30s to 20s
+		PRIMAL_RUINATION_READY: {
+			id: 3834,
```
Toastdeib: _Consider:_ You can remove everything besides the `duration` property from this; the layers coalesce into the root file, so you only need to include values that actually changed from 7.0 and can safely omit anything that doesn't.

src/parser/jobs/war/changelog.tsx
```diff
@@ -1,10 +1,10 @@
-// import CONTRIBUTORS from 'data/CONTRIBUTORS'
-// import React from 'react'
+import CONTRIBUTORS from 'data/CONTRIBUTORS'
+import React from 'react'
 
 export const changelog = [
-	// {
```
Toastdeib, line 5: _Nit:_ We usually keep the commented example changelog there and just add below it; see [NIN's changelog](https://github.com/xivanalysis/xivanalysis/blob/dawntrail/src/parser/jobs/nin/changelog.tsx) for an example.

Toastdeib (commented):
Looks good overall; I double-checked all the data imports and saw no errors there, so I'm good to move this forward once you have a chance to look at and address the two small comments I left.

src/data/STATUSES/layers/patch7.01.ts
kwerey: Updated, thanks!

src/parser/jobs/war/changelog.tsx
kwerey: Makes sense, having it there and copy/pasteable is nice. I've put it back where I found it :)

kwerey:
Thanks for reviewing! I've updated both those points.

Toastdeib (approved):

PR #2023, opened 2024-07-20
[WAR] Add suggestions for new procs, update IR to mention Primal Wrath

kwerey:
## Pull request type

- [x] This is a patch bump

## Pull request details

- [x] This is in response to a discussion or thread on Discord: [*[Link to thread or start of discussion]*](https://discord.com/channels/441414116914233364/470049767477608459/1263976148061257758)

		* Add tracking for the use of new Warrior job actions Primal Ruination and Primal Wrath.
		* Update dropped Primal Rend proc suggestion to point it now also loses a use of Primal Ruination.

- [ ] The job(s) in this patch bump had only potency changes for this patch, which do not effect analysis
  - [ ] I have updated the data files for all potency changes for this patch

## Testing / Validation

Dropped Primal Rend, Primal Wrath, missed using Inner Release stacks: https://www.fflogs.com/reports/a:A4vKFYHtRJ9DnQ6z#fight=1
Dropped Primal Ruination: https://www.fflogs.com/reports/a:NgqK8BcxP1yGXdap#fight=last&type=damage-done&source=2

This is missing one case (gained the Primal Ruination proc, but did not spend it). I will update this PR to add another log that tests all cases. Includes a second recorded fight there does use procs as expected and doesn't incorrectly receive suggestions.

## Job Maintenance
- [?] I am active on the xivanalysis Discord and part of the job maintainers group for the job(s) in this PR
- [?] I have collaborated with one or more job maintainers for the job(s) in this PR while developing it

Discussed with a Balance warrior person and with XIVA folks.


src/parser/jobs/war/index.tsx
```diff
@@ -16,12 +16,12 @@ export const WARRIOR = new Meta({
...
 	supportedPatches: {
-		from: '✖',
-		to: '✖',
+		from: '7.0',
+		to: '7.01',
 	},
 
 	contributors: [
-	// {user: CONTRIBUTORS.YOU, role: ROLES.YOUR_ROLE},
+		{user: CONTRIBUTORS.KWEREY, role: ROLES.DEVELOPER},
```
Toastdeib: I might be wrong here (and another old hand is welcome to correct me), but I _think_ we prefer to have someone flagged as the maintainer role when a job is officially supported as the main point of contact for any issues that come up. If you're willing, it'd be better to set `ROLES.MAINTAINER` here.

src/parser/jobs/war/modules/PrimalProcs.tsx
```diff
@@ -0,0 +1,70 @@
...
+	]
+
+	protected override addJobSpecificSuggestions() {
+		const missedRend = this.getDropCountForStatus(this.data.statuses.PRIMAL_REND_READY.id)
+		const missedRuination = this.getDropCountForStatus(this.data.statuses.PRIMAL_RUINATION_READY.id)
+		const missedWrath = this.getDropCountForStatus(this.data.statuses.WRATHFUL.id)
+
+		this.suggestions.add(new TieredSuggestion({
+			icon: this.data.actions.PRIMAL_REND.icon,
+			content: <Trans id="war.primalrend.suggestions.dropped.content">
```
Toastdeib: _Nit:_ The usual standard for keys is `{job}.{module}.{content type}.{specific issue}.{whatever else}`, so this (and the two below) would be better as something like `war.procs.suggestions.dropped-primal-rend.content`, etc. This also applies to the `.why` text in the suggestions.

src/parser/jobs/war/modules/PrimalProcs.tsx
```diff
@@ -0,0 +1,70 @@
...
+				Using <DataLink action="INNER_RELEASE" /> grants <DataLink status="PRIMAL_REND_READY" />. Use it for a strong attack and to unlock the even stronger <DataLink action="PRIMAL_RUINATION" />.
+			</Trans>,
+			tiers: SEVERITY_MISSED_PROCS,
+			value: missedRend,
+			why: <Trans id="war.primalrend.suggestions.dropped.why">
+				You missed <Plural value={missedRend} one="# Primal Rend use" other="# Primal Rend uses" />.
```
Toastdeib: _Nit:_ Phrasing might be a bit cleaner as:
`You missed <Plural value={missedRend} one="# use" other="# uses" /> of Primal Rend.`

src/parser/jobs/war/modules/PrimalProcs.tsx
```diff
@@ -0,0 +1,70 @@
...
+			</Trans>,
+		}))
+
+		this.suggestions.add(new TieredSuggestion({
+			icon: this.data.actions.PRIMAL_RUINATION.icon,
+			content: <Trans id="war.primalruination.suggestions.dropped.content">
+				Using <DataLink action="PRIMAL_REND" /> grants you <DataLink status="PRIMAL_RUINATION_READY" /> for 20 seconds which lets you use <DataLink action="PRIMAL_RUINATION" />. It is your highest damage GCD, so try to spend it before it expires.
```
Toastdeib: _Nit:_ The phrasing of the second sentence here reads a bit strangely to me - maybe you could phrase it as:
`It's your strongest GCD, so make sure to spend it before it expires.`
instead?

src/parser/jobs/war/modules/PrimalProcs.tsx
```diff
@@ -0,0 +1,70 @@
...
+			</Trans>,
+			tiers: SEVERITY_MISSED_PROCS,
+			value: missedRuination,
+			why: <Trans id="war.primalruination.suggestions.dropped.why">
+			You let <Plural value={missedRuination} one="# Primal Ruination proc" other="# Primal Ruination procs" /> time out.
```
Toastdeib: _Nit:_ This may just be a me thing, but I think it's nicer when similar suggestions in the same module use similar phrasing. Consider rewriting this and the one below to mirror the Primal Rend why text:
`You missed <Plural value={missedRuination} one="# use" other="# uses" /> of Primal Ruination.`
`You missed <Plural value={missedRwrath} one="# use" other="# uses" /> of Primal Wrath.`

Alternatively, you could rewrite the first one to mirror these two. Either way, the consistency would be nicer in my book. If you go that route, though, I would replace "time out" with "expire" so it mirrors the wording used in the content block.

src/parser/jobs/war/modules/PrimalProcs.tsx
```diff
@@ -0,0 +1,70 @@
...
+			</Trans>,
+		}))
+
+		this.suggestions.add(new TieredSuggestion({
+			icon: this.data.actions.PRIMAL_WRATH.icon,
+			content: <Trans id="war.primalwrath.suggestions.dropped.content">
+				Using <DataLink action="FELL_CLEAVE" /> 3 times under <DataLink status="INNER_RELEASE" /> lets you use <DataLink action="PRIMAL_WRATH" />. Try to spend it before it expires.
```
Toastdeib: _Nit:_ Similar to the previous suggestion, I think "Make sure to" is better phrasing than "Try to", considering how important these abilities are for DPS.

src/parser/jobs/war/modules/PrimalProcs.tsx
```diff
@@ -0,0 +1,70 @@
+import {Plural, Trans} from '@lingui/react'
+import {DataLink} from 'components/ui/DbLink'
+import {Procs as CoreProcs} from 'parser/core/modules/Procs'
+import {TieredSuggestion, SEVERITY} from 'parser/core/modules/Suggestions'
+import React from 'react'
+
+const SEVERITY_MISSED_PROCS = {
+	1: SEVERITY.MAJOR,
```
Toastdeib: _Note:_ From my armchair WAR position, I don't see anything wrong with flagging any drops of any of these procs as major, but I did want to check whether you poked a theorycrafter before setting the severity tiers (or are one yourself, in which case it's moot).

Toastdeib (commented):
PR looks good to me functionally, I just have several nits around the strings before I mark this as approved. I'll take another look this evening if you've gotten changes in by then.

src/parser/jobs/war/modules/PrimalProcs.tsx
kwerey: Naw I'm not a warrior theorycrafter, I just figured "SEVERITY.MAJOR" was a reasonable translation of Mox from the Balance categorising these all as "big mistake" in a summary of the changes?
 
> Inner Release :ir: doesn't contain 3 Fell Cleaves => no Primal Wrath opened, big mistake
> Primal Wrath :wrath: opened (3x IR FC), but wasn't used (whoops dropped), big mistake
> Primal Ruination :ruination: opened, but wasn't used, big mistake

Primal Wrath is a 700 potency oGCD, Primal Ruination is 780p as well as guaranteed crit+DH. Happy to ping the good folks of the Balance to check those!

src/parser/jobs/war/modules/PrimalProcs.tsx
kwerey: That make sense, will update. In keeping with that would it also be better to rename this module `Procs.tsx` so the translation keys match it cleanly?

src/parser/jobs/war/modules/PrimalProcs.tsx
kwerey: Yeah, thinking about it I agree that it's nicer if the phrasing matches. Since the "content" block already explains the player let the proc expire I will go with what you suggested and copy the Primal Rend why text.

src/parser/jobs/war/index.tsx
kwerey: Let me give the caveat that I don't actually know much at all about Typescript, so my ability to fix any complicated problem is pretty limited! But I've taken enough of a look around the warrior modules to understand what they currently do and pick off some low hanging fruit issues, happy to mark myself as maintainer if it'd be in a state of neglect otherwise.

src/parser/jobs/war/modules/PrimalProcs.tsx
Toastdeib: Doesn't make a huge difference either way. I think it most (or at least a lot of) cases, we tend to use the core name for the override and import the actual core module as `Core[ModuleName]` (like `CoreProcs` in this case), but it's not super important. The main exceptions to that rule (at least in my experience) are when you have multiple overrides for the same core module on one job (e.g. Procs.tsx and Harpe.tsx on RPR, both of which override the core Procs module), or when you have a meme filename (e.g. DirtyDancing.tsx and Technicalities.tsx on DNC).

src/parser/jobs/war/modules/PrimalProcs.tsx
kwerey: Yep, let's call this major - https://discord.com/channels/441414116914233364/470049767477608459/1264640782535622668

src/parser/jobs/war/modules/PrimalProcs.tsx
kwerey: Gotcha, ty. The meme names are fun but I'll go with the standard and rename this to `Procs` so it's obvious that it's just doing very standard proc tracking and there isn't anything special about these 3 WAR abilities.

Toastdeib (approved):

PR #2024, opened 2024-07-20
Dawntrail - PCT Starry Muse RaidBuffWindow

AkairyuGestalter:
## Pull request type

## Pull request details

Added a RaidBuffWindow implementation to evaluate Pictomancer's Starry Muse buff windows:
- By default, Starry Muse should contain 9 GCDs, including the hammer combo, Rainbow Drip and Star Prism, 1-2 Comets, and filling the remainder with the Subtractive Palette combo.
 - Expectations sourced from: https://www.thebalanceffxiv.com/jobs/casters/pictomancer/openers/
- It should also include one use each of a creature Muse, and either Mog or Madeen
- There is an option to fit three muses into the window by cutting other spells out of the window if required for alignment, so that is also handled
- Also included is a suggestion not to use the additive combo spells (and Holy) since they're a potency loss and the player will always have access to the Subtractive spells by way of free use granted by Starry Muse

This PR also includes a couple other minor tweaks:
- Altered the table header override property to be a ReactNode instead of just an action to construct the header from, for more flexibility
- DRK was just using the Delerium action as before, but I opted to use the Hammer Time and Subtractive Palette statuses for their columns
 - DRK's use of this override header is also updated to fit the new format
 - Used the Living Muse action that becomes the actual creature muses once painted as the header for that collection
 - Mog and Madeen use the default action group behavior
- Added a Tinctures override to ignore the cure "action" that comes along with the use of Star Prism
 - The Starry Muse RaidBuffWindow also ignores this

## Testing / Validation

 - Clean buff windows: http://localhost:3000/fflogs/VtR1hQCkZ7Yxnczr/11/7
 - Some mistakes: http://127.0.0.1:3000/fflogs/Fm8V6wTYPhRGZ34k/11/4
 - Triple Muse sample: http://127.0.0.1:3000/fflogs/Q3f2jaALzMRdVHKr/3/1
  - Very scuffed dummy log just to verify the expected counts switch when 2+ Muse uses are found in the window, I'm not 100 yet to produce a properly executed log

## Job Maintenance


src/parser/jobs/pct/modules/StarryMuse.tsx
```diff
@@ -0,0 +1,154 @@
...
+			severityTiers: {
+				1: SEVERITY.MINOR,
+				3: SEVERITY.MEDIUM,
+				5: SEVERITY.MAJOR,
+			},
+		}))
+	}
+
+	private adjustExpectedActionGroupCounts(window: HistoryEntry<EvaluatedAction[]>, action: TrackedActionGroup): number {
+		const attemptedTripleMuse = this.countActionsUsed(window, this.museActions) > 1
```
AzariahAeroborn: Are we straying too far to enforcing a rotation with this?  Is there a more general way to adjust based on the actual number of extra muse actions they used, if they only use a two Muse window?

AkairyuGestalter: Probably should have noted this in the PR template before, but the expectations were sourced from Eydis' Pictomancer guide on the Balance (https://www.thebalanceffxiv.com/jobs/casters/pictomancer/openers/). There isn't a 2-muse window right now, so having 2 is being treated as "tried and failed to do 3"

AkairyuGestalter: After further discussion, and watching Eksu's guide video, I did some GCD math, and 2 muse has basically the same GCD expectations as triple muse, so I'm comfortable changing this to "Default 1" for the pre-window prepped muse, then just requiring all motifs painted in the window to get used. Should be fixed now!

AzariahAeroborn (approved):

PR #2025, opened 2024-07-21
feature(pld): add support for 7.0 and 7.01

czyba:
## Pull request type
Kind of all?
- [x] This is a patch bump

## Pull request details

This PR adds rudimentary support for 7.0 and 7.01 Paladin. The following changes and decisions have been made to add support:

* Add new Actions and Statuses that Paladin received with 7.0
* Replace Sentinel with Guardian in defensive analysis
* Replace Requiescat with Imperator in offensive analysis
* Adjust texts for new analysis.
   * Actual wording may be up for debate. Feel free to offer improvements on that front.

More Complex Design Choices:

Since Sword Oath no longer exists. When being granted the "Atonement Ready" buff, I chose to already increment the totals for Supplication and Sepulchre as well. This may lead to issues with end of fight scenarios but is more accurate in case someone does something like: FB -> RB -> RA -> FB -> RB -> RA. The second RA technically only loses the Atonement stack, but since it also loses its follow up actions I opted to count it for all 3 instead of just the single Atonement.

See example Screenshot:

image

I opted to have a simpler FoF analysis instead of a more complex but marginally better one. For one, I added BoH as another ogcd to the list. Before 7.0 we wanted to have at least 1 Holy Spirit. I opted to just say: "I want 3 of RA, AT, SU, SE, HS", since depending on which buff is up and how you play your rotation you may have slightly more powerful actions available. While this may help in optimizing 0.1% or something, a general reminder to use higher damaging skills instead of low ones like FB and RB seemed more helpful.

Unfortunately the column for the "lower potency" actions is now a bit crowded:
image

Workarounds:

In order to get the Combo Analysis running without having to significantly extend the whole analysis, I've opted to no longer mark the confiteor combo as a combo. The analysis has issues since the normal FB -> RB -> RA combo and the C -> BoF -> BoT -> BoV are independent combos that do not break each other. The default combo analysis only keeps track of a single combo however. Even using the ability to mark BoF as non-breaking for the normal combo will then result in, for example RA being marked as "uncomboed", since the analyser then keeps track of the Confiteor combo instead, which ends with BoV. Thus it thinks RA is uncomboed.

Setting the Confiteor combo to not being a combo fixes this issue, even though it is technically incorrect and may result in incorrect analysis in non-standard uses.

The second workaround regards to Blade of Honor. As far as I'm aware it does not have a cooldown, since it is tied to the Blade of Honor Ready buff. Not giving it a cooldown, however will result in the skill not being shown on the timeline, so I just gave it a cooldown of ~~30s~~ 1s.

## Testing / Validation

  - Top PLD at the time: https://www.fflogs.com/reports/dNPcMawF3XzDTxZ4#fight=3&type=damage-done

## Job Maintenance

I have created this PR by myself and can be found in the xivanalysis discord as Chris, id: chriss9211. I'll link the PR in the discord and am open to feedback.


src/data/ACTIONS/root/PLD.ts
```diff
@@ -304,4 +328,28 @@ export const PLD = ensureActions({
...
+		cooldown: 60000,
+		statusesApplied: ['REQUIESCAT', 'CONFITEOR_READY'],
+	},
+	BLADE_OF_HONOR: {
+		id: 36922,
+		name: 'Blade of Honor',
+		icon: iconUrl(2956),
+		onGcd: false,
+		// Setting a cooldown is technically incorrect, but xivanalysis core modules require it for it to show up on the timeline
+		cooldown: 30000,
```
kwerey: Hey! I'm a fellow new arrival here (showed up the other day to look at adding warrior stuff for 7.0.), so not qualified to review properly. But about this section:

> The second workaround regards to Blade of Honor. As far as I'm aware it does not have a cooldown, since it is tied to the Blade of Honor Ready buff. Not giving it a cooldown, however will result in the skill not being shown on the timeline, so I just gave it a cooldown of 30s.

I think the correct cooldown to set for this would be 1000 (ie 1 second), since that's what's listed in the job guide as the recast time. It's effectively never going to be relevant since you can't gain the Blade of Honor Ready status more than once a minute, but setting that would matches what the tooltip says.

czyba: > I think the correct cooldown to set for this would be 1000 (ie 1 second), since that's what's listed in the job guide as the recast time. It's effectively never going to be relevant since you can't gain the Blade of Honor Ready status more than once a minute, but setting that would matches what the tooltip says.

Good point! Totally missed that. Adjusted accordingly.

src/data/ACTIONS/root/PLD.ts
```diff
@@ -125,48 +123,74 @@ export const PLD = ensureActions({
...
 		icon: iconUrl(2518),
 		onGcd: true,
 		speedAttribute: Attribute.SPELL_SPEED,
-		combo: {
-			start: true,
-		},
+		// Technically incorrect, but this way we don't get spurious broken combos
+		// combo: {
+		// 	start: true,
+		// },
```
Styrfire, line 159: The Confiteor combo is a combo that get's broken by PLD's Fast Blade, Riot Blade, and Royal Authority. I think we still need to classify it as a combo. The Atonement follow up abilities do not combo nor break combos, so if you put that in Atonement and it's follow up abilities, things will work as intended. @czyba

src/data/ACTIONS/root/PLD.ts
```diff
@@ -125,48 +123,74 @@ export const PLD = ensureActions({
...
+		statusesApplied: [
+			'SEPULCHRE_READY',
+		],
+	},
+	SEPULCHRE: {
+		id: 36919,
+		name: 'Sepulchre',
+		icon: iconUrl(2523),
+		onGcd: true,
+		speedAttribute: Attribute.SKILL_SPEED,
```
Styrfire: For code clarity, I would position these right after Atonement. Also, they should have the "Breaks Combo" set to false so they don't break the Fast Blade or Confiteor combo.

src/parser/jobs/pld/changelog.tsx
```diff
@@ -1,10 +1,17 @@
...
 	// 	date: new Date('2021-11-19'),
 	// 	Changes: () => <>The changes you made</>,
 	// 	contributors: [CONTRIBUTORS.YOU],
 	// },
+	{
+		date: new Date('2024-07-21'),
+		Changes: () => (<>Add rudimentary support for 7.0 and 7.01 Paladin</>),
+		contributors: [],
+	},
 ]
```
Styrfire, line 17: I would add yourself as a Developer contributor so you get credit for your work!

src/parser/jobs/pld/index.tsx
```diff
@@ -19,8 +19,8 @@ export const PALADIN = new Meta({
 	</>,
 
 	supportedPatches: {
-		from: '✖',
-		to: '✖',
+		from: '7.0',
+		to: '7.01',
 	},
 
 	contributors: [
```
Styrfire, line 26: I would add yourself as a Developer contributor so you get credit!

src/parser/jobs/pld/modules/FightOrFlight.tsx
```diff
@@ -60,18 +62,42 @@ export class FightOrFlight extends BuffWindow {
...
+				expectedPerWindow: 3,
+			}],
+			suggestionIcon: actions.FIGHT_OR_FLIGHT.icon,
 			suggestionContent: <Trans id="pld.fightorflight.suggestions.gcd_actions.content">
 				Try to land at least one cast of <DataLink action="GORING_BLADE" />
 				, <DataLink action="CONFITEOR" />, <DataLink action="BLADE_OF_FAITH" />, <DataLink action="BLADE_OF_TRUTH" />
-				, <DataLink action="BLADE_OF_VALOR" />, and a <DataLink status="DIVINE_MIGHT" /> empowered <DataLink action="HOLY_SPIRIT" />
+				, <DataLink action="BLADE_OF_VALOR" />, and the highest potency combination of <DataLink action="SEPULCHRE" />, <DataLink status="DIVINE_MIGHT" />
+				empowered <DataLink action="HOLY_SPIRIT" />, <DataLink action="SUPPLICATION" />, <DataLink action="ATONEMENT" /> and <DataLink action="ROYAL_AUTHORITY" />
 				during every <DataLink action="FIGHT_OR_FLIGHT" /> window.
```
Styrfire: Playing optimally, it's impossible to get an empowered Holy Spirit and Sepulchre every FoF window. When hitting a striking dummy (bursting after 3 gcds), the 5th minute rotation restricts you from using Sepulchre unless you cut an empowered Holy Spirit or Sepulchre earlier which isn't good.

I believe the FoF window from a simplicity standpoint should have 3 of the following actions (at it's simplest level): empowered Holy Spirit, Royal Authority, Atonement, Supplication, and/or Supulchre. I believe it's confusing to the user to just use the Supulchre as an icon and show 3 uses of it. It doesn't line up with what's expected.

locale/en/messages.json
```diff
@@ -765,10 +766,11 @@
...
   "pld.requiescat.title": "Requiescat Usage",
+  "pld.royal-authority.checklist.description": "<0/> generates <1/> to use on <2/> and follow up actions <3/> and <4/>. This is effectively the same as getting {0} uses of <5/> or stronger actions.",
+  "pld.royal-authority.checklist.name": "Use Follow Up Skills of Royal Authority",
   "pld.royal-authority.title": "Royal Authority",
-  "pld.sword-oath.checklist.description": "<0/> generates {swordOathStacks} stacks of <1/> to use on <2/>. This is effectively the same as getting {swordOathStacks} uses of <3/>.",
-  "pld.sword-oath.checklist.name": "Use Sword Oath stacks generated by Royal Authority",
-  "pld.sword-oath.checklist.requirement.name": "Uses of <0/> out of possible uses",
+  "pld.sepulchre.checklist.requirement.name": "Uses of <0/> out of possible uses",
+  "pld.supplication.checklist.requirement.name": "Uses of <0/> out of possible uses",
   "pld.timeouts.confiteor.content": "Try to consume <0/> before it expires as the <1/> combo is your strongest combo.",
```
Styrfire, line 773: I think the locale need cleaned up a bit. Just replacing sword oath with royal authority should work imo. I don't think there's a point for the atonement, supplication, and sepulchre checklists. I do like how you worded the FoF suggestions!

locale/de/messages.json
```diff
@@ -765,10 +766,11 @@
...
   "pld.requiescat.title": "Requiescat-Benutzung",
+  "pld.royal-authority.checklist.description": "",
+  "pld.royal-authority.checklist.name": "",
   "pld.royal-authority.title": "",
-  "pld.sword-oath.checklist.description": "",
-  "pld.sword-oath.checklist.name": "",
-  "pld.sword-oath.checklist.requirement.name": "",
+  "pld.sepulchre.checklist.requirement.name": "",
+  "pld.supplication.checklist.requirement.name": "",
   "pld.timeouts.confiteor.content": "",
```
Styrfire, line 773: I think the locale need cleaned up a bit. Just replacing sword oath with royal authority should work imo.

Styrfire (changes requested):
Please see my comments for suggestions and feel free to hit me up on Discord @Styrfire if you have any questions, comments or concerns!

locale/de/messages.json
AzariahAeroborn: Locale files should not be edited directly.  The naming of the entries is controlled by the "id" field of the Trans tags used within module files.

locale/en/messages.json
AzariahAeroborn: Locale files should not be edited directly. The naming of the entries is controlled by the "id" field of the Trans tags used within module files.

src/data/STATUSES/root/PLD.ts
```diff
@@ -118,4 +111,47 @@ export const PLD = ensureStatuses({
...
+		name: 'Sepulchre Ready',
+		icon: iconUrl(12524),
+		duration: 30000,
+	},
+	BLADE_OF_HONOR_READY: {
+		id: 3831,
+		name: 'Blade of Honor Ready',
+		icon: iconUrl(13052),
+		duration: 30000,
+		stacksApplied: 3,
```
Toastdeib: _Action:_ Unless I'm missing something, this shouldn't have stacks associated with it.

src/parser/jobs/pld/changelog.tsx
```diff
@@ -1,10 +1,17 @@
...
+
 export const changelog = [
 	// {
 	// 	date: new Date('2021-11-19'),
 	// 	Changes: () => <>The changes you made</>,
 	// 	contributors: [CONTRIBUTORS.YOU],
 	// },
+	{
+		date: new Date('2024-07-21'),
+		Changes: () => (<>Add rudimentary support for 7.0 and 7.01 Paladin</>),
```
Toastdeib: _Nit:_ _Super_ inconsequential, but you can get rid of the parentheses around the `<> </>` tags.

src/data/ACTIONS/root/PLD.ts
czyba: Adjusted accordingly. I was going by action id, but that makes more sense.

src/parser/jobs/pld/changelog.tsx
czyba: Done.

src/parser/jobs/pld/index.tsx
czyba: Done.

Toastdeib: Or maintainer; we've historically preferred to have at least one person marked with `ROLES.MAINTAINER` before we flag a job as supported, and that person (or persons) has the responsibility of being the lead point of contact for any bug reports/feature requests that come up (not to say other folks aren't allowed to address them, but still). It's not super important (and we're actually discussing whether to drop the maintainer role entirely), but I wanted to mention it all the same.

locale/en/messages.json
czyba: I believe the issue is that I had the same text 3 times but 3 different keys. For an example see the list of the screenshot of the Royal Authority Analysis in PR Description. I use the same key for all three now, which also cleans up the messages a bit.

Though I'm not entirely sure if it causes issues how I do it now. Is there a better way to "define" a Translation at one point and then "use" it at another?

locale/de/messages.json
czyba: I believe the issue is that I had the same text 3 times but 3 different keys. For an example see the list of the screenshot of the Royal Authority Analysis in PR Description. I use the same key for all three now, which also cleans up the messages a bit.

Though I'm not entirely sure if it causes issues how I do it now. Is there a better way to "define" a Translation at one point and then "use" it at another?

src/parser/jobs/pld/modules/FightOrFlight.tsx
czyba: Fair point. I've removed the overrideHeader icon of Sepulchre and adjusted the PR description accordingly. The column is just a bit crowded now. I agree that clarity is more important, however.

src/parser/jobs/pld/modules/FightOrFlight.tsx
```diff
@@ -60,18 +62,42 @@ export class FightOrFlight extends BuffWindow {
...
-				{action: this.data.actions.HOLY_SPIRIT, expectedPerWindow: 1},
+				{action: actions.GORING_BLADE, expectedPerWindow: 1},
+				{action: actions.CONFITEOR, expectedPerWindow: 1},
+				{action: actions.BLADE_OF_FAITH, expectedPerWindow: 1},
+				{action: actions.BLADE_OF_TRUTH, expectedPerWindow: 1},
+				{action: actions.BLADE_OF_VALOR, expectedPerWindow: 1},
 			],
-			suggestionIcon: this.data.actions.FIGHT_OR_FLIGHT.icon,
+			suggestionIcon: actions.FIGHT_OR_FLIGHT.icon,
+			suggestionContent: <Trans id="pld.fightorflight.suggestions.gcd_actions.content">
```
Toastdeib: _Action:_ It looks like you're using the same text (and translation key) in both suggestions; given how verbose and action-filled it is, I'd recommend paring them down and making each one unique and only about the actions the individual evaluator is tracking. For instance, this one could be shortened to:

`Try to land <DataLink action="GORING_BLADE" />, <DataLink action="CONFITEOR" />, <DataLink action="BLADE_OF_FAITH" />, <DataLink action="BLADE_OF_TRUTH" />, and <DataLink action="BLADE_OF_VALOR" /> during every <DataLink action="FIGHT_OR_FLIGHT" /> window.`

Or even simpler:

`Try to land your full <DataLink action="CONFITEOR" /> combo plus <DataLink action="BLADE_OF_VALOR" /> during every <DataLink action="FIGHT_OR_FLIGHT" /> window.`

src/parser/jobs/pld/modules/FightOrFlight.tsx
```diff
@@ -60,18 +62,42 @@ export class FightOrFlight extends BuffWindow {
...
+					actions.ATONEMENT,
+					actions.SUPPLICATION,
+					actions.SEPULCHRE,
+					actions.HOLY_SPIRIT,
+				],
+				overrideHeader: actions.SEPULCHRE,
+				expectedPerWindow: 3,
+			}],
+			suggestionIcon: actions.FIGHT_OR_FLIGHT.icon,
 			suggestionContent: <Trans id="pld.fightorflight.suggestions.gcd_actions.content">
```
Toastdeib: _Action:_ Same comment here as above; a simpler rewrite (also reordered for the "combo" order) could look like:

`Try to land at least three of <DataLink action="ROYAL_AUTHORITY" />, <DataLink action="ATONEMENT" />, <DataLink action="SUPPLICATION" />, <DataLink action="SEPULCHRE" />, and <DataLink status="DIVINE_MIGHT" /> empowered <DataLink action="HOLY_SPIRIT" /> during every <DataLink action="FIGHT_OR_FLIGHT" /> window.`

Given that the suggestion is only looking for 3 and not a specific 3, you can drop the "highest combination of" qualifier as well.

Toastdeib (commented):
Threw a few more comments onto the pile. I haven't dug through `RoyalAuthority.tsx` yet so I may do that later after giving my brain a rest from reading code for most of the morning.

src/data/ACTIONS/root/PLD.ts
czyba: Unfortunately, I don't think it'll work. When I mark Atonement, Supplication and Sepulchre as `breakCombo: false` and make the Confiteor Combo an actual combo again, I end up with this: 

image

This is an issue I had during development, and I don't quite understand why it doesn't occur normally. Probably some difference between development and production build?

Anyways when I circumvent that error with:

```diff
diff --git a/src/parser/core/modules/Combos.tsx b/src/parser/core/modules/Combos.tsx
index 1978221c..b6f3c5e8 100644
--- a/src/parser/core/modules/Combos.tsx
+++ b/src/parser/core/modules/Combos.tsx
@@ -326,7 +326,7 @@ export class Combos extends Analyser {
                        <Table.Body>
                                {
                                        data.map(issue => {
-                                               const startEvent = issue.context[0]
+                                               const startEvent = issue.context[0] ?? {}
                                                const brokenTime = issue.type !== 'timeout' ? issue.breaker.timestamp : startEvent.timestamp + COMBO_TIMEOUT

                                                return <Table.Row key={startEvent.timestamp}>


```

I get the following view, with the changes you mentioned:

image

The Confiteor Combo breaks the Fast Blade combo. And after the combo the Royal Authority is marked as uncomboed.

A different approach would be to implement

```ts
	/**
	 * Jobs MAY override this to indicate whether this broken combo is allowed.
	 * Return true to indicate an allowable combo break, which will result in this break not being recorded and the current combo will be cleared with no other side effects.
	 * Return false to record the combo break and display it to the user
	 * @param event The event that caused the combo break
	 * @param context The active combo chain that is being broken
	 * @returns true if this combo break should not be recorded
	 */
	protected isAllowableComboBreak(_breaker: ComboBreak, _context: Array<Events['damage']>): boolean {
		return false
	}
```

for the PLD Combos. I orignally did that for the Confiteor combo, but that didn't solve the problem that the Combo Analyzer still marked the Royal Authority / Riot Blade as uncomboed, since we essentially ended the previous combo.

I also tried to Mark the Confiteor Combo Skills as not breaking combo to no avail.

If you've got an idea why this occurs, I'd be glad to fix it. I couldn't figure it out without delving deep into the Core Combos Analyzer.

src/data/STATUSES/root/PLD.ts
czyba: You're completely correct. copy  & paste error. Fixed accordingly.

src/parser/jobs/pld/modules/FightOrFlight.tsx
czyba: I've gone with your first suggestion. While the second one would also work, currently Blade of Honor is in the ogcd Evaluator, not the gcd evaluator. Therefore the shortened suggestion would feel incorrect without shifting Blade of Honor around to the gcd section.

I've also added the Blade of Honor suggestion text for the ogcd abilities.

src/parser/jobs/pld/modules/FightOrFlight.tsx
czyba: Adjusted accordingly, though I've also cut the "at least" and just said "Try to land three of", since the remaining usages are already supposed to be taken with the confiteor combo and goring blade.

src/parser/jobs/pld/changelog.tsx
czyba: Adjusted accordingly. My knowledge about the javascript ecosystem is very ~~large~~ small (edit: brainfart), but I thought that eslint can do that for you automatically, if configured? Or is there no rule for this particular thing?

src/data/ACTIONS/root/PLD.ts
czyba: @kwerey Would you be so kind to resolve this conversation, if the change is in line with your expectations. :)

src/parser/jobs/pld/index.tsx
Styrfire: @Toastdeib I was the maintainer from last expansion and had planned on doing the update this weekend, but @czyba beat me to it! That being said, if they are more active, I don't mind handing the maintainer role off to them if they would prefer to have it, but but otherwise, I can add myself.

locale/en/messages.json
Styrfire: @AzariahAeroborn We kind of have to in order to adjust the English portion to be correct, right? I get that with the id fields though.

src/data/ACTIONS/root/PLD.ts
kwerey: Sorry @czyba  - I don't think I have the permissions on this repo to do that. Thanks for updating, please feel free to mark it resolved.

locale/en/messages.json
AzariahAeroborn: No, English translations should **only** be updated via `yarn extract`, after putting the desired copy in the Trans tag in the module that has the copy.

AzariahAeroborn: > I believe the issue is that I had the same text 3 times but 3 different keys. For an example see the list of the screenshot of the Royal Authority Analysis in PR Description. I use the same key for all three now, which also cleans up the messages a bit.
> 
> Though I'm not entirely sure if it causes issues how I do it now. Is there a better way to "define" a Translation at one point and then "use" it at another?

Judging by the documentation here: https://lingui.dev/tutorials/explicit-vs-generated-ids - as long as you have the exact same copy in the Trans tag, you could re-use the same id.  However, if the copy varies or you're using other properties in the tag to give extra context to the translators, then repeating IDs won't work because lingui is going to grab the first string it sees with a given ID and ignore the rest of them.

czyba: I did run `yarn extract`. I remember that we had a similar discussion when I marked atonement non-combo breaking in 6.4. I've not adjusted any texts manually.

src/parser/jobs/pld/changelog.tsx
Toastdeib: We probably don't have a rule for this, it's just the usual convention we follow.

src/data/ACTIONS/root/PLD.ts
```diff
@@ -92,8 +92,6 @@ export const PLD = ensureActions({
 		icon: iconUrl(2506),
 		onGcd: true,
 		speedAttribute: Attribute.SKILL_SPEED,
-		cooldown: 60000,
```
edreeseg, line 95: This may be in the wrong place - I haven't had the chance to get super familiar with the logic yet. It would appear on line 234 of CooldownDowntime.tsx, we calculate the number of opportunities to use the ability until the end of the fight using the action's cooldown. It seems that right now, it's expecting us to be able to use Goring Blade every 3.75s, even though it is locked behind FoF. This results in logs giving you very large expected max usages (e.g. 10/166).

I placed this comment here because I'm not sure if this change is the one that causes the behavior (as we attempt to read from gRep.cooldown when determining the step to increment currentTime with).

AzariahAeroborn: Yeah, removing the cooldown will cause "use your cooldowns" to assume you can constantly use Goring Blade.  Since we now get a "Goring Blade Ready" buff, we should move Goring Blade out of the "Use Your Cooldowns" module and into a Procs tracking module instead (every time you gained Goring Blade Ready, you used it)

czyba: Is that not already implicitly done via the Fight or Flight Analyzer?

* Fight or Flight is in the "Use Your Cooldowns" Module
* Fight or Flight analysis tells you to use Goring Blade within Fight or Flight

For now I'll remove it from the Use Your Cooldowns Module and not add an an additional analysis for it. Is that ok?

src/data/ACTIONS/root/PLD.ts
czyba: @Styrfire I think I figured it out. The EW Branch also marks the combo as uncomboed! We just missed it because it was added as a layer.

https://github.com/xivanalysis/xivanalysis/blob/a5371bb565b001cfa091f935beadcbf5518eab0a/src/data/ACTIONS/layers/patch6.3.ts#L28-L39

So essentially the layer did the same thing I did within this PR directly.

Styrfire: Ah, I forgot about that. Good catch. And because it has it's own analyzer it should be fine.

src/parser/jobs/pld/index.tsx
czyba: I do not believe I'm cut out for the role. For one, I'm mostly inactive on discord, unless someone actively pings me. Additionally, I'm not familiar with webdev, nor do I have any aspirations to be. You also mentioned that the removal of the maintainer role is also discussed, so...

Though I do wonder, given this PR now has 5 people commenting on it, how to actually go about getting it approved?Especially if there are no maintainers who can have the final say? Do I have to convince everyone who commented on the PR? Are there some general maintainers who can accept a PR, even if they are not particularly interested in the job in question?

src/data/ACTIONS/root/PLD.ts
AzariahAeroborn: You're correct in that Fight or Flight is in the "Use Your Cooldowns" Module (and sorry for potential ambiguity -- I use that term as the friendly name for CooldownDowntime, which you seem to have found correctly).  Last expansion, Goring Blade wasn't a direct "flip skill" after Fight or Flight, so it needed to be tracked separately within that Use Your Cooldowns module, and removing it from there is the correct way to fix the issue edreeseg noticed.

Since the Fight or Flight BuffWindow module is expecting to see Goring Blade within Fight or Flight, that may be enough to ensure people are using Goring Blade before the buff falls off.  The one potential advantage of a Procs based module is that it would provide an additional, more specific suggestion for "hey, don't let Goring Blade Ready fall off without using it", but if y'all don't feel that separate suggestion is meaningful, I don't have a strong opinion on it.

src/parser/jobs/pld/index.tsx
AzariahAeroborn: For our PRs as a whole, 1 repo level maintainer (which includes @Toastdeib and myself) will have to approve the PR in order for it to merge.  Generally the first person who picks it up for review will work with you to see it through to merging.  

I only jumped in with comments on this one because I saw a couple comments that were incorrect that I wanted to help get you the correct information for before you wasted time doing the wrong thing.  Usually, I'd let Toast keep working with you on this unless he stopped actively reviewing it.

We do also encourage job maintainers (of which @Styrfire was the PLD job maintainer last expansion) to help comment on PRs from the "this is accurate for job analysis" perspective, because sometimes our repo maintainers aren't up-to-speed on a given job.

czyba: Ahhh, got it. Thanks for the explanation.

src/data/ACTIONS/root/PLD.ts
czyba: I can implement the separate analysis if required. I think @Styrfire would be the one to make that call, though, given they're the PLD maintainer.

src/parser/jobs/pld/modules/RoyalAuthority.tsx
```diff
@@ -1,25 +1,31 @@
...
+}
+
+type RoyalAuthorityDependentUsages = {
+	divineMight: Usages,
+	atonement: Usages,
+	supplication: Usages,
+	sepulchre: Usages,
 }
 
 export class RoyalAuthority extends Analyser {
```
AzariahAeroborn, line 31: I think there would probably be future benefit to look at whether this module & it's analysis would fit into the Procs module type to be a simpler thing to maintain & be more consistent with output for other classes that have similar functionality, but not something to hold up this PR for.

AzariahAeroborn (approved):

PR #2032, opened 2024-07-22
[DRG] Extend CounterGauge to output table of overcap times

ebshimizu:
## Pull request type

## Pull request details

- [x] This is in response to a discussion or thread on Discord: https://discord.com/channels/441414116914233364/1263970313075556384
This PR extends `CounterGauge` to track the timestamps of overcap events. This allows the `CoreGauge` model to assemble a table (in the style of `Positionals`) which provides links to the tlimeline indicating when a `CounterGauge` overcapped. This feature must be enabled with an opt-in configuration flag set when the `CounterGauge` is created and presumably added to a `CoreGauge` derived analyzer. Developed in response to a feature request for DRG's Firstminds' Focus gauge to output overcap timestamp events for that gauge.

Since this affects core, let me know if this implementation would cause problems for other jobs / future maintenance.

## Testing / Validation
Jobs that use `CounterGauge` should be unaffected by this change, as the feature is opt-in by default.

  - DRG Overcap test: https://www.fflogs.com/reports/HMGVdtQTRb1jhLmA#fight=last, should report 2 overcap instances
  - Affected Jobs:  https://www.fflogs.com/reports/9LnCvJD7xQKPdzRF#fight=3&source=211, looked at DNC and RDM in here, no unexpected outputs present

## Job Maintenance


src/parser/jobs/drg/modules/Gauge.tsx
```diff
@@ -75,4 +79,10 @@ export class Gauge extends CoreGauge {
 			</Trans>,
 		}))
 	}
+
+	override output() {
```
AkairyuGestalter: Not sure if this is left over from an earlier iteration of this change, but since this is just outputting Core's stuff, you don't need this override.

AkairyuGestalter (commented):
Overall this looks good. One note about the DRG Gauge module' output that I think we should fix, but if you'd rather leave as-is it's not breaking anything.

src/parser/jobs/drg/modules/Gauge.tsx
ebshimizu: oh yeah that was leftover from when i was considering just implementing it in the drg gauge file. removed!

AkairyuGestalter (approved):

PR #2034, opened 2024-07-22
[RPR] Bugfix, Soul & Shroud gauge modifiers were incorrect

Rjb99:
## Pull request type

- [x ] This is a bugfix to existing functionality

## Pull request details

Plentiful harvest no longer generates gauge and has been removed from the gauge modifiers
Added Executioner skills as generators for gauge

## Testing / Validation

- [x ] I used the log(s) listed below to develop and test this bugfix:

Pre-Fix: 
image

Post-Fix:
image

https://www.fflogs.com/reports/6yNcHD7vhJFdnWPV#fight=16&source=5

## Job Maintenance
- [x ] I am active on the xivanalysis Discord and part of the job maintainers group for the job(s) in this PR

### DIFF
--- a/src/parser/jobs/rpr/modules/OtherGauges.tsx
+++ b/src/parser/jobs/rpr/modules/OtherGauges.tsx
@@ -21,9 +21,11 @@ const SUGGESTION_TIERS = {
 
 const SHROUD_ACTIONS: ActionKey[] = [
 	'GIBBET',
+	'EXECUTIONERS_GIBBET',
 	'GALLOWS',
+	'EXECUTIONERS_GALLOWS',
 	'GUILLOTINE',
-	'PLENTIFUL_HARVEST',
+	'EXECUTIONERS_GUILLOTINE',
 	'ENSHROUD',
 ]
 
@@ -112,10 +114,6 @@ export class OtherGauges extends CoreGauge {
 		switch (action.id) {
 		case this.data.actions.ENSHROUD.id:
 			this.shroudGauge.spend(HIGH_SHROUD_MOD)
-			break
-
-		case this.data.actions.PLENTIFUL_HARVEST.id:
-			this.shroudGauge.generate(HIGH_SHROUD_MOD)
 			break
 
 		default:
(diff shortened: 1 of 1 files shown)


src/parser/jobs/rpr/modules/OtherGauges.tsx
```diff
@@ -114,10 +116,6 @@ export class OtherGauges extends CoreGauge {
 			this.shroudGauge.spend(HIGH_SHROUD_MOD)
 			break
 
-		case this.data.actions.PLENTIFUL_HARVEST.id:
```
AkairyuGestalter, line 117: Correct me if I'm wrong, but Arcane Circle grants a "free use of Enshroud" status now, right? We should account for that in here as well.

AkairyuGestalter (changes requested):
One note about Enshroud's gauge spending that I think needs fixing.

src/parser/jobs/rpr/modules/OtherGauges.tsx
Rjb99: You are correct. 
image
I've added a check similar to reawaken on Viper for Enshroud consuming gauge.

src/parser/jobs/rpr/modules/OtherGauges.tsx
```diff
@@ -111,11 +116,7 @@ export class OtherGauges extends CoreGauge {
 
 		switch (action.id) {
 		case this.data.actions.ENSHROUD.id:
-			this.shroudGauge.spend(HIGH_SHROUD_MOD)
-			break
-
-		case this.data.actions.PLENTIFUL_HARVEST.id:
-			this.shroudGauge.generate(HIGH_SHROUD_MOD)
+			if (this.IsIdealHost === false) { this.shroudGauge.spend(HIGH_SHROUD_MOD) }
```
AkairyuGestalter: this.actors.current.hasStatus should be enough, then you don't need to hook the application/removal etc.

AkairyuGestalter (changes requested):
One comment to simplify the Ideal Host stuff

AkairyuGestalter (approved):

PR #2036, opened 2024-07-23
BLU: Initial support for Dawntrail

Hugmeir:
## Pull request type

- [X] This is a patch bump

## Pull request details

Initial support for BLU in DT.

Technically nothing has changed for BLU at 80 -- except that we now have a bit of an inconsistency, because at level 100, Swiftcast is 40s (down from 60s) and Addle lasts 15s (up from 10s).

So on top of bumping support, this also overrides those two for BLU back to their EW values.

There's still an inconsistency in the tooltips shown in the browser (xivapi rightfully shows the values for DT) but at least internally we do calculations using the right values for BLU.

That all being said, I am not a Node dev, my solution is probably just awful -- it was the first thing I could cobble together -- so do chime in with suggestions to change that to anything else.

## Testing / Validation

We actually don't use the Swiftcast cooldown / Addle duration anywhere in the analysis, so nothing can be tested as-is.  I did an ad-hoc test where I added Swiftcast to "Defensives" and saw the "possible uses" calculation change depending on whether the patch was applied or not.

## Job Maintenance


AkairyuGestalter (approved):
Since we don't rely on the role action changes for BLU analysis, I was going to suggest maybe we just mark it as supported with no other updates. This seems sane though, and isn't affecting other jobs so we can probably let it be for now and revisit later if necessary.

PR #2043, opened 2024-07-26
[MNK] Add fury gauge

hintxiv:
## Pull request type

- [X] This is new functionality or an addition to existing functionality

## Pull request details

Adds a graph and overcap suggestions for the new balls gauge. The implementation using SetGauge might not be 100% perfect but in my opinion it's perfectly functional and would be good to get out before savage.

image
image

## Testing / Validation

http://localhost:3000/fflogs/a:mwc4Nf67v1WYzKPp/13/131 (7 overcaps expected)
http://localhost:3000/fflogs/PkCxp62L1jMm3q8Z/12/8 (no overcap expected)

## Job Maintenance


src/parser/jobs/mnk/modules/Balls.tsx
```diff
@@ -0,0 +1,128 @@
...
+			{
+				value: 'coeurl' + 0,
+				color: COEURL_GAUGE_COLOR.fade(GAUGE_FADE),
+				label: <Trans id="mnk.balls.coeurlFury.label">Coeurl Fury 1</Trans>,
+			},
+			{
+				value: 'coeurl' + 1,
+				color: COEURL_GAUGE_COLOR.fade(GAUGE_FADE),
+				label: <Trans id="mnk.balls.coeurlFury2.label">Coeurl Fury 2</Trans>,
+			},
```
AzariahAeroborn: What's the benefit of making these two separate Set Gauges rather than using a single Counter Gauge with a max value of 2?

AzariahAeroborn (changes requested):

src/parser/jobs/mnk/modules/Balls.tsx
hintxiv: Currently (to my knowledge) there's no way to combine a counter gauge with a set gauge. I tried a few combinations (3 counter gauges, 2 enum gauges + 1 counter gauge, 1 set gauge + 1 counter gauge) and this one looks the best. I think the ideal solution would be to allow multiple values for each item in the set, but it's kind of a one off for Monk and nobody else.

AzariahAeroborn (approved):

PR #2046, opened 2024-07-26
Add a core "disengage gcds" analyser

kwerey:
## Pull request type

- [ ] This is a patch bump

## Pull request details

- [x] This is in response to a discussion or thread on Discord: *[Link to thread or start of discussion]*
- [ ] The goal of this PR is detailed below:

* All tanks and several melee jobs have a low value ranged attack that they can use when disengaging from an enemy.
* Gunbreaker had a module tracking this, since the only job specific input is the action ID, we could move it to core so that other jobs can easily implement it as well by passing their disengage action in.

## Testing / Validation

- [] The changes being implemented with this bugfix include comments that contain example log(s) to test with

https://www.fflogs.com/reports/a:GxgMcjNHyVmqFb89#fight=last&type=damage-done

## Job Maintenance


src/parser/core/modules/DisengageGcds.tsx
```diff
@@ -0,0 +1,53 @@
...
+	override initialise() {
+		super.initialise()
+
+		const playerFilter = filter<Event>().source(this.parser.actor.id)
+		this.addEventHook(playerFilter.type('action').action(this.trackedAction.id), this.onUse)
+		this.addEventHook('complete', this.onComplete)
+	}
+
+	// Override this to suggest job-specific ways to handle downtime
+	protected addDisengageSuggestion() {
```
Rjb99: Consider; I'm not liking that the stat maker is labelled as a suggestion, I feel it muddies the water about what the function does, but the module is small and simple enough that it can be read and understood quickly. 

I think it would be better if this was called addDisengageStatisitic() and a 2nd function addJobSpecificSuggestions() be added that is called onComplete() that can be overrode for suggestions, just to align it with other core module setups.

src/parser/core/modules/DisengageGcds.tsx
```diff
@@ -0,0 +1,53 @@
...
+		this.statistics.add(new SimpleStatistic({
+			title: <Trans id="core.disengage.statistic.title">
+				{this.trackedAction.name} Uses
+			</Trans>,
+			icon: this.trackedAction.icon,
+			value: `${this.disengageGcds}`,
+			info: <Trans id="core.disengage.statistic.info">
```
Rjb99: Action; It would be better if the info i18n string was declared outside of the protected function so that you can just override the string, instead of having to override the entire function just to change some text. This would keep it inline with other core modules.

Rjb99 (changes requested):
My word means less than the people with merge powers, but 2 comments, only one is something I'd consider blocking though.

src/parser/core/modules/DisengageGcds.tsx
kwerey: Thanks again for reviewing, I have updated to make the info string configurable.  Is it useful to do the same with the title, and if so how?

My instinct here would be that we'd want to default to using the name of thedisengage ability we're telling people to avoid in the title, since that's shorter and more specific than something like "avoid your basic ranged attack". 

From what I can tell though, we need to provide the title as an override to the Analyser module, and that this must be a text string rather than a ReactNode. Is there a way to pass in a variable to the t() function? I couldn't find an example with a little bit of rummaging around, and this doesn't seem to work:

```
	static override title = t('core.disengage.statistic.title')`{this.trackedAction.name} Uses`
```

Would we want to:

1. provide an "[action name] uses]" title within the function?
2. provide a title string like "Ranged Attack Uses" which can be overridden, but which is less specific by default?
3. a secret third thing that hasn't occurred to me or that I don't know how to do as a Typescript newbie?

src/parser/core/modules/DisengageGcds.tsx
kwerey: Yes, that makes sense about terminology. In fact this the output from this is not a suggestion but a statistic, I've added an addJobSpecificStatistics() function that works in the same way as the one in the Combos module - setting it to true disables the default statistic output, setting it to false adds an additional statistic alongside the one in the core module.

I'm v new to using Typescript and unaware of best practices, advice very welcome on whether or not what I've done here is sane & in line with normal code style.

kwerey:
Since there were no potency or action changes in 7.05 for warrior I have also bumped the supported patches value in the changelog.

src/parser/core/modules/DisengageGcds.tsx
```diff
@@ -0,0 +1,65 @@
...
+export class DisengageGcds extends Analyser {
+	static override handle = 'disengagegcds'
+
+	@dependency data!: Data
+	@dependency statistics!: Statistics
+
+	private disengageGcds: number = 0
+
+	// Pass in the GCD you want to track as an Action.
+	protected trackedAction: Action = this.data.actions.UNKNOWN
```
AzariahAeroborn: Probably better for a generic class like this to be marked as abstract and then to mark required items like this as abstract.  See BuffWindow or Procs for examples (I think Procs is probably the closer example for this here)

Also wondering if this should take an Action[] (array) instead, to allow for jobs that have multiple disengage GCDs they'd want to count, but I don't have a good mental image of what that would look like in the suggestion output.

src/parser/core/modules/DisengageGcds.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+		this.statistics.add(new SimpleStatistic({
+			title: <Trans id="core.disengage.statistic.title">
+				{this.trackedAction.name} Uses
+			</Trans>,
+			icon: this.trackedAction.icon,
+			value: `${this.disengageGcds}`,
+			info: this.disengageInfo,
+		}))
+	}
```
AzariahAeroborn: I'd prefer to see this more like how Procs does the Icon & Content for its suggestions (and like how you already have the override point for `disengageInfo`).  

Otherwise, if we want this to be "override the statistic with your own", this should just be one `protected addStatistic(disengageGcds: number)` function.

src/parser/core/modules/DisengageGcds.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+	private onUse(): void {
+		this.disengageGcds++
+	}
+
+	private onComplete() {
+		this.addDisengageStatistic()
+	}
```
AzariahAeroborn: If you go with the Procs-style overrides for Statistic title, icon, and info, then you don't need this function since you could just directly call addDisengageStatistic with your onComplete event hook.

If you go with the "overrideable `addStatistic` function" approach, then this should pass in the disengageGcds number as noted.

src/parser/core/modules/DisengageGcds.tsx
AzariahAeroborn: I would probably vote for Ranged Attack Uses and allow override, seems simple enough for implementers.

src/parser/jobs/war/index.tsx
```diff
@@ -17,7 +17,7 @@ export const WARRIOR = new Meta({
 
 	supportedPatches: {
 		from: '7.0',
-		to: '7.01',
+		to: '7.05',
```
AzariahAeroborn: Can you please separate this patch bump into a separate PR?  Would rather not hold up WAR 7.05 support over this statistic.

AzariahAeroborn (dismissed):

src/parser/jobs/war/index.tsx
kwerey: Makes sense - here is a seperate PR: https://github.com/xivanalysis/xivanalysis/pull/2080

src/parser/core/modules/DisengageGcds.tsx
kwerey: All right, I'll take a look at procs and update.

It's a good question about passing an array in rather than a single action. I can't think of any jobs where we'd want to aggregate multiple disengage attacks together and provide a single statistic as output. 

Some examples I can think of that have different ranged options would be:

1. Paladin, where they have both a ranged attack and hardcasting Holy Spirit as an option if they disengage. In that case we'd probably want to treat them differently, AFAIK preferring to hardcast Holy Spirit.
2. Ninja being able to use Raiton from range, but then maybe being short on mudras for their burst - I'm betting this would be more usefully analysed by pointing out if it causes them to miss uses of recommended actions in buff window
3. Reaper being able to use Harpe or Harvest Moon - Harvest Moon is better, hardcasting Soulsow in combat to get it is bad, is good, they already have a module that evaluates Harvest Moon usage

For the jobs I know about where there are multiple disengage options that we may want to track, it'd probably be better to implement two different instances of a class which have different info text and icons. Can you think of any classes where we may want to combine output?

src/parser/core/modules/DisengageGcds.tsx
kwerey: Thanks for reviewing. Doing what the Procs module does makes sense to me, I've taken a stab at that, let me know if it looks good.

It doesn't seem like we can simply override the title using the `t('name.title')`Title Goes Here`` syntax that the Procs module is using since the title is a required argument for a statistic, so I have added that as a React node.

src/parser/core/modules/DisengageGcds.tsx
kwerey: I've updated the class and overrides but left this supporting a single action rather than an array since the discussion here didn't come up with any examples that'd benefit from multiple ranged actions having the same info text: https://discord.com/channels/441414116914233364/441424599310270464/1269323101703901279 

It seems that generally when melee/tanks have multiple disengage actions one is preferrable to another so we'd want to give different feedback for each.

kwerey:
@AzariahAeroborn I've updated this per review comments, please take a look when you have the opportunity and let me know if there's any more to do here.

AkairyuGestalter (approved):
Looks good to me, and all requested changes appear to have been made. Marking approved.

PR #2052, opened 2024-07-27
feat: upgraded Procs module to be stacks aware

AzariahAeroborn:
## Pull request type

- [ ] This is a patch bump

## Pull request details

- [x] This is in response to a discussion or thread on Discord: https://discord.com/channels/441414116914233364/470050747720138753/1266746905526931458

Create a new version of the core Procs module that is able to handle statuses that apply multiple stacks, and streamline the interface & logic accordingly.  This PR intentionally does not change the logic for dropped proc forgiveness due to downtime (all procs that fall off during downtime are forgiven by default without trying to consider whether it could have been used before downtime or not) or due to end of fight rushing (no forgiveness for dropping at end of fight).

## Testing / Validation

- [X] The changes being implemented with this bugfix include comments that contain example log(s) to test with
- [ ] I used the log(s) listed below to develop and test this bugfix:

https://www.fflogs.com/reports/y9bc6qpf4KgVnvTX - RDM MagickedSwordplay 2 dropped stacks

https://www.fflogs.com/reports/a:cm4L9dR2gMrPDQCw#fight=4&source=31 - BLM 1 dropped Thunderhead, 1 dropped Firestarter
https://www.fflogs.com/reports/a:KAcpNLH6BaQ3RF4n#fight=31&source=1481 - BLM 4 dropped Thunderhead (with 1 on at end of fight), 3 dropped Firestarter

https://www.fflogs.com/reports/a:gfdkBAHXYQxTZMDL#fight=3&source=9 - BRD no dropped or overwritten Hawk's Eye
https://www.fflogs.com/reports/a:bDBLTYwP3ANmz7h9#fight=23&source=103 - BRD 23 overwritten, 2 dropped Hawk's Eye

https://www.fflogs.com/reports/a:mk261xdvB3gyw4aY#fight=17&source=35 - DNC 1 dropped proc
https://www.fflogs.com/reports/a:WbJ2mhyBR17KDkZ4#fight=3&source=4 - DNC 7 dropped, 4 overwritten procs

DRG doesn't have any of the Procs output enabled to test against

https://www.fflogs.com/reports/a:ZJdwrykgW7jFXYCa#fight=10&source=90 - GNB no dropped procs, 100% continuation skills usage
https://www.fflogs.com/reports/a:aAJDzPyT63h8cZkt#fight=2&source=152 - GNB all continuation procs dropped.  Seems to resolve a bug that was resulting in counting Ready To Raze as both dropped and overwritten at 8:02.346 (we're getting both a refresh and remove buff event there so it's a little weird but now it seems to be counting more as expected).  I'm not sure why we're only getting 35 drops in production and 34 drops with this fix when there are 36 applies, so there may still be weird happening with Ready To Raze here.

https://www.fflogs.com/reports/a:jPH3BmGdcQ7htxW4#fight=50&source=4249 - MNK no replies dropped
https://www.fflogs.com/reports/a:yqmwtMJ7YR4CW2Hb#fight=6&source=15 - MNK 1 Fire and 3 Wind dropped

https://www.fflogs.com/reports/a:q8TVh3JtPHZQ6BAX#fight=9&source=2 - NIN no procs dropped
https://www.fflogs.com/reports/gWpnqRyfFY1TZm7D#fight=1 - NIN Procs module test log (2 dropped Phantom Kamaitachi, 1 dropped Tenri Jindo)

https://www.fflogs.com/reports/a:xGPB41tJLd3znbQj#fight=2&source=4 - PCT no procs dropped
https://www.fflogs.com/reports/a:xQ8VW6mZKFa3zfRr#fight=13&source=361 - PCT 4 Starstruck drops, 4 Aetherhues drops (couldn't find a log with Rainbow Bright drops)

https://www.fflogs.com/reports/a:2YF7gcjmhpTDwVzy#fight=2&type=damage-done&source=1 - RDM no Thorned Flourish or Prefulgence dropped, 2 Verfire dropped, 10 Verfire/Verstone overwritten
https://xivanalysis.com/fflogs/NJyZkQAtVwKYMXfC/1/2 - RDM test log dropped Prefulgence and Thorned Flourish, 1 Verfire, 1 Verstone
https://xivanalysis.com/fflogs/RAqxQcVp23WgnfKM/1/2 - RDM test log dropped Grand Impact

https://www.fflogs.com/reports/a:W6cDr8HQK4RthNCL#fight=1&source=128 - RPR 1 harpe cast (RPR makes no suggestions regarding enhanced harpe usage), 420 potency loss to Enhanced Gibbet overwrites (the Executioner's Gallows bug still exists in both cases because it's a data problem)
https://www.fflogs.com/reports/a:mhX9wjVMtzkbK413#fight=30&source=1603 - RPR 0 harpe casts, 420 potency overwriting Gibbet, 120 potency expired Gibbet

https://www.fflogs.com/reports/a:GM9BZ41xQzr7q6LT#fight=6&source=5 - SAM 10/10 Ogi use, 22/22 Tendo use, no dropped or overwritten procs
https://www.fflogs.com/reports/a:3GHJwAXrxcg6Vq4j#fight=8&source=9 - SAM 4/6 Ogi use, 23/26 Tendo use, 2 dropped procs

https://www.fflogs.com/reports/a:z392wBCRFv4kXPLn#fight=23&source=5 - SCH no dropped procs
https://www.fflogs.com/reports/a:n48kmrPaXFDdBgqT#fight=34&source=1452 - SCH 1 dropped proc

https://www.fflogs.com/reports/a:QAdBnt4jTRWNFXZ6#fight=2&source=26 - VPR 100% venom usage, 0 dropped or overwritten procs
https://www.fflogs.com/reports/a:1hRmBratbwW7kK2V#fight=1&source=7 - VPR 92.8% venom usage, 6 dropped and 2 overwritten procs

https://www.fflogs.com/reports/a:PjGaA13R9rvzgcKb#fight=361&source=9340 - WAR 0 dropped or overwritten procs
https://www.fflogs.com/reports/a:KrbYD364FhRHLk1j#fight=27&source=660 - WAR 2 dropped Primal Rend, 2 dropped Primal Ruination, 1 dropped Primal Wrath
The only log I could find for WAR having a drop or overwrite of Nascent Chaos didn't actually emit a refresh buff event for Nascent Chaos when the second infuriate was used (1.027 seconds after the first one)

https://www.fflogs.com/reports/a:NhLFTctjM7nkJ8mq#fight=14&source=9 - WHM 0 dropped procs 
https://xivanalysis.com/fflogs/bMaRNfz3TyhCg7Z4/1/2 - WHM test log dropped Divine Caress

## Job Maintenance


AzariahAeroborn:
After further consideration, determined this would be better as a full upgrade to the Procs module rather than splitting out a separate function that's a superset of the base handling.  

I'm still working my way through checking consumers (looks like I'm about halfway through), but updated with test logs I've grabbed so far.

AzariahAeroborn:
Added links to logs testing as many current implementations of Procs as I could find.  Tried to include one log with no/minimal detections and one that covered all of the procs suggestions emitted by each job.

kwerey:
Warrior https://www.fflogs.com/reports/a:GFPXfWphgLmrMY89#fight=last

Drop detection: working
Immediate overwrite: not working - appears to be an issue with ACT mistakenly deduplicating the status apply event.
Delayed overwrite: working

src/parser/core/modules/Procs.tsx
```diff
@@ -288,12 +296,15 @@ export abstract class Procs extends Analyser {
 	private checkConsumeProc(procGroup: ProcGroup, event: Events['action']): boolean {
 		// If we don't currently have this proc active, we can't possibly consume it
 		if (!this.currentWindows.has(procGroup)) { return false }
+		const action = this.data.getAction(event.action)
 		// If we have no idea what this action is, it doesn't consume it
-		if (this.data.getAction(event.action) == null) { return false }
+		if (action == null) { return false }
+		// If we're in downtime, we don't consume procs (don't credit using an AoE-around-self Proc during downtime)
+		if (this.downtime.isDowntime(event.timestamp)) { return false }
```
kwerey, line 311: I am wondering if there are any cases where jobs might want to override this and still flag procs dropped during downtime, for example healer abilities. The only ones I can think of are things like Liturgy of the Bell or Macrocosmos which just execute their effect when they expire, but it might be worth asking in healer channel if there's a use case for exposing an "ignore downtime" flag which defaults to true?

AzariahAeroborn: I don't know that Liturgy or Macro would really be something to track with Procs anyway, since they don't require re-activation to gain their effect, just give an early activation of their on-expires effect if you do reactivate.

The intent with this PR is to be 100% consistent with the behavior of the current module, and then take a look at potential changes in future PRs if necessary.

kwerey: Yea that makes sense! I mentioned it because I think White Mage now gets a proc-like thing off Temperance and I'm not sure if it's possible for that to be wasted on expiration, but it's definitely something that could be added later.

kwerey: Doesn't sound like there currently are, please disregard this comment :)

src/parser/core/modules/Procs.tsx
```diff
@@ -20,31 +20,26 @@ const ICON_MUSCLE_MEMORY = 1994
 
 export interface ProcBuffWindow {
 	start: number,
-	stop?: number
+	stop: number, // All windows will eventually be closed with stopAndSave, either due to a remove event or EOF.  Force not-null to make reliable for consumers and initialize with stop = start
```
ackwell: `stop = start` is technically invalid (a consumer needs to know that the `stop` value could be incorrect, but unlike nullable, does not force them to check it).

in saying this; there already seems to be a separation between current and historical - possibly worth defining

```ts
interface CurrentWindow {
	start: number,
	// ...
}

interface HistoricalWindow extends CurrentWindow {
	end: number
}
```

this would allow the current- logic to operate without encoding an `end` time, but still ensure that code working with historical data has a guaranteed `end`.

src/parser/core/modules/Procs.tsx
```diff
@@ -217,7 +215,35 @@ export abstract class Procs extends Analyser {
...
+	 */
+	protected considerDroppedProcs(window: ProcBuffWindow): boolean {
+		// Don't count procs that were dropped due to death
+		if (this.actors.get(this.parser.actor).at(window.stop).hp.current === 0) { return false }
+
+		// Don't count procs that were dropped due to downtime
+		if (this.downtime.isDowntime(window.stop)) { return false }
+
+		// Don't count procs that were dropped due to end of fight
+		if (window.stop === this.parser.pull.timestamp + this.parser.pull.duration) { return false }
```
ackwell: just in case there's ever funky numbers
```suggestion
		if (window.stop >= this.parser.pull.timestamp + this.parser.pull.duration) { return false }
```

src/parser/core/modules/Procs.tsx
```diff
@@ -479,76 +506,19 @@ export abstract class Procs extends Analyser {
...
-		}
-
-		// If this was an overwrite event, and overwrites are disallowed for this proc, save a record of that
-		if (event && type === 'overwrite' && !procGroup.mayOverwrite) {
-			this.tryAddEventToOverwrites(procGroup, event)
-		}
-
-		if (!this.currentWindows.has(procGroup)) { return }
-
+	private stopAndSave(procGroup: ProcGroup, event?: Event, isOverwrite?: boolean): void {
```
ackwell: even if only two possible types, would rather keep the `type` - booleans in function args are really obtuse at the callsite.

ackwell (approved):
focused primarily on core update.

mostly seems sane. few nitpicks, mostly interested in type safety on the window interface

PR #2053, opened 2024-07-27
feat: adjust pull timestamp to account for events from pre-combat

AzariahAeroborn:
## Pull request type

- [ ] This is a patch bump

## Pull request details

- [x] This is in response to a discussion or thread on Discord: https://discord.com/channels/441414116914233364/441414116914233366/1266513815143583908

Since FFLogs has refined their rules for determining start of combat and started including some pre-pull events in their timeline with a negative relative timestamp, our timeline and the relative times we show in most areas in the analysis are now out of sync with the relative timestamps in the FFLogs timeline.  This adjusts that handling accordingly.

- Adjusts our prepull synth offsets to be based on the first event in the timeline, not the start of combat
- Adjusts our "timestamp" value to match the start of combat
- Adjusts the debug Event Viewer to show negative timestamps

## Testing / Validation

https://www.fflogs.com/reports/VNJxHbhMFkaPptXB#fight=10

## Job Maintenance

### DIFF
--- a/src/parser/core/modules/EventsView/EventsView.tsx
+++ b/src/parser/core/modules/EventsView/EventsView.tsx
@@ -94,7 +94,7 @@ function EventItem({data: {events, pull}, index, style}: EventItemProps) {
 
 	const timestamp = formatDuration(
 		event.timestamp - pull.timestamp,
-		{secondPrecision: 3},
+		{secondPrecision: 3, showNegative: true},
 	)
 
 	const formatter = eventFormatters.get(event.type)
--- a/src/report.ts
+++ b/src/report.ts
@@ -53,8 +53,6 @@ export interface Pull {
 	 */
 	id: string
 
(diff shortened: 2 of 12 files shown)


hintxiv (approved):
lgtm, I think showing negative timestamps in the timeline would be ideal but I imagine that involves a non-trivial refactor.

image

AzariahAeroborn:
> lgtm, I think showing negative timestamps in the timeline would be ideal but I imagine that involves a non-trivial refactor.
> 
> image

For pre-cast actions, FFLogs is only picking up the cast event and the prepares event ('action' and 'damage' in our translation), and not picking up begincast events.  So we don't have a "known" start time for the begincast event of channeled casts.  We could potentially back-calculate for actual channels, though it would get weird for DNC's Standard or Technical Step, but my main goal here is to get the "in combat" timing of the timeline to match up, not really to change the way our pre-pull handling deals with negative time events.

PR #2059, opened 2024-07-28
[DRG] Enable suggestion output for procs

ebshimizu:
## Pull request type

## Pull request details
- [x] This is in response to a discussion or thread on Discord: https://discord.com/channels/441414116914233364/1267146030005227602
  - Enable output for dropped procs. DRG procs can't be overwritten due to the CD between proc granting casts being longer than the proc duration. Because of this, any drop is at least medium severity, with more than one dropping off being marked as major.

## Testing / Validation
  - [x] I have submitted the example log(s) to the xivanalysis static on FFLogs for record keeping
  - https://www.fflogs.com/reports/6zHNFQRG3tjkMh2d#fight=1

## Job Maintenance

### DIFF
--- a/src/parser/jobs/drg/modules/Procs.tsx
+++ b/src/parser/jobs/drg/modules/Procs.tsx
@@ -1,6 +1,17 @@
 import {Procs as CoreProcs} from 'parser/core/modules/Procs'
+import {SEVERITY} from 'parser/core/modules/Suggestions'
 
 export default class Procs extends CoreProcs {
+	protected override showDroppedProcSuggestion = true
+	// DRG wants to use every proc, and dropping any of them is a major issue
+	protected override droppedProcSeverityTiers = {
+		1: SEVERITY.MEDIUM,
+		2: SEVERITY.MAJOR,
+	}
+
+	// we'll show invuln stuff too, though some of the drg actions are cleaves so will have to monitor for spurious suggestions
+	protected override showInvulnProcSuggestion = true
+
 	trackedProcs = [
 		{
 			procStatus: this.data.statuses.NASTROND_READY,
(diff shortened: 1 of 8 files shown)


src/parser/jobs/drg/modules/Procs.tsx
```diff
@@ -1,6 +1,17 @@
 import {Procs as CoreProcs} from 'parser/core/modules/Procs'
+import {SEVERITY} from 'parser/core/modules/Suggestions'
 
 export default class Procs extends CoreProcs {
+	protected override showDroppedProcSuggestion = true
+	// DRG wants to use every proc, and dropping any of them is a major issue
+	protected override droppedProcSeverityTiers = {
+		1: SEVERITY.MEDIUM,
+		2: SEVERITY.MAJOR,
+	}
```
AzariahAeroborn: The default icon is kind of a meme icon (hasty touch for dropped procs, hallowed ground for invuln procs) and the copy is super generic.  Standard practice has been to implement a class specific ProcContent override to be a little more specific than "Avoid letting your procs fall off without using them. Proc actions are generally stronger than other actions and should not be wasted."... do you want to do that here?

Looks like you're the first job to leverage showInvulnProcSuggestion so I don't take a strong opinion on overriding the icon, but the copy for that content is also kind of vague.

ebshimizu: ah yeah i can be actually helpful and write a job specific suggestion. I'm not super attached to the invuln proc suggestions, i suspect DRG might have more issues than other jobs due to cleaves with that. I turned it off in this commit, but can leave it on if we want to test it and see if it's relevant this tier.

AzariahAeroborn (approved):

PR #2060, opened 2024-07-28
7.0 support: Scholar overheal bucketing changes and fairy gauge suggetions

nonowazu:
## Pull request type

## Pull request details

- [x] This is in response to a discussion or thread on Discord: [*Link to relevant conversation*](https://discord.com/channels/441414116914233364/470049935077933076/1264318988830249097)

* Changed overheal bucketing to new categories (4 displayed categories are a bridge too far for the stat box)
* Added the seraphism status effect right beside the seraphism status effect (they both have the [same name](https://beta.xivapi.com/api/1/search?&sheets=Status&query=Name~%22Seraphism%22) but this one is the HoT, so I labelled the status as such)
* Changed fairy gauge overcap to always be a minor suggestion since overcap isn't as important (only tether uses it)

## Testing / Validation

* https://www.fflogs.com/reports/a:kbTLt8Dja6hKxq3n - testing seraphism status effect ignoring
* I do not have one for the seraph changes just yet, but the logic hasn't changed and I'm not at the computer right now. However, I will test when I get home later to double-check with a crafted log for overcap.

## Job Maintenance
- [x] I am active on the xivanalysis Discord and part of the job maintainers group for the job(s)


src/parser/jobs/sch/modules/Overheal.js
```diff
@@ -6,6 +6,7 @@ import React from 'react'
 
 export default class Overheal extends CoreOverheal {
 	static handle = 'overheal'
+	static debug = true
```
AzariahAeroborn: We don't allow committing debug = true
```suggestion
	static debug = false
```

AzariahAeroborn (changes requested):

src/parser/jobs/sch/modules/Overheal.js
nonowazu: I know; I missed it. One minute and I'll fix.

AzariahAeroborn (approved):

PR #2069, opened 2024-07-31
RDM 7.05 Patch Bump

Cephid:
## Pull request type

- [X] This is a patch bump

## Pull request details

- [X] The job(s) in this patch bump had only potency changes for this patch, which do not effect analysis
  - [X] I have updated the data files for all potency changes for this patch
  - [X] The skills and jobs affected by these potency changes do not track potency for xivanaylsis purposes

Potency and Status duration for this patch, both Manafication and Magicked Swordplay are now 30 second durations.

## Job Maintenance


src/data/STATUSES/root/RDM.ts
```diff
@@ -44,7 +44,7 @@ export const RDM = ensureStatuses({
 		id: 1971,
 		name: 'Manafication',
 		icon: iconUrl(17491),
-		duration: 15000,
+		duration: 30000,
```
AzariahAeroborn: This definitely needs to be done in the layers file so that it doesn't effect pre-7.05 log timelines.  Also, shouldn't this be for the Magicked Swordplay status and not the Manification status?

AzariahAeroborn (changes requested):

src/data/ACTIONS/root/RDM.ts
```diff
@@ -366,7 +366,7 @@ export const RDM = ensureActions({
 		speedAttribute: Attribute.SPELL_SPEED,
 		cooldown: 2500,
 		castTime: 5000,
-		potency: 420, // Increase White by 6
+		potency: 440, // Increase White by 6
```
AzariahAeroborn: Potency changes should also be done in the layer file, to maintain consistency with other jobs.

AzariahAeroborn (changes requested):

src/data/STATUSES/root/RDM.ts
Cephid: It would seem when 7.01 came out I mistakenly hit magicked swordplay instead of manafication.  Also what do you mean layers, changing the duration on a status is done via its duration properties is it not?

src/data/ACTIONS/root/RDM.ts
Cephid: Same question as before, that wasn't something when I made changes for 7.01; there's also no pins in the discord describing what you're referring to either.

src/data/STATUSES/root/RDM.ts
AzariahAeroborn: There are files in the layers directory (there's a corresponding one for STATUSES, but this one is for ACTIONS: https://github.com/xivanalysis/xivanalysis/blob/dawntrail/src/data/ACTIONS/layers/patch7.05.ts) that contain just the changes to given actions or statuses in a specific patch.  The layers files allows us to support both pre- and post-patch logs, since the root data values are as of 7.0, and then we can add each subsequent patch's changes from the layers files.  

Which is where the Manafication change from 7.01 was already previously made: https://github.com/xivanalysis/xivanalysis/blob/dawntrail/src/data/STATUSES/layers/patch7.01.ts

So you'll just need to add a patch7.05 layer for STATUSES to cover the Magicked Swordplay duration change here.  And then fix Magicked Swordplay in the root file to be the 15000ms duration it had in 7.0 and 7.01.

AzariahAeroborn (approved):

PR #2076, opened 2024-08-02
RDM - BugFix - Changed Manafication to allow for rushing

Cephid:
## Pull request type

- [ ] This is a patch bump

## Pull request details

- [X] This is in response to a discussion or thread on Discord: https://discord.com/channels/441414116914233364/1268753420303597660/1268753420303597660

		Changed Manafication to accept casted GCDs for the purpose of allowing Manafication rushing, which is popping it on cooldown misaligned with buffs but holding Prefulgence for the height of the 2 Minute buff window.

## Testing / Validation

- [X] The changes being implemented with this bugfix include comments that contain example log(s) to test with

   https://www.fflogs.com/reports/K7TyWcQwvpGH1B8X#fight=20&type=damage-done

## Job Maintenance


AzariahAeroborn (changes requested):
The suggestion content for both the expectedGcd and allowedGcd evaluators no longer matches what we're tracking here.  Since we're shifting to track any valid consuming cast under Manafication, would this be better shifted to just be a Procs-based analyser?

Cephid:
Hrm you're right on the text; but I feel it could still be useful to know what was used within each window.  Would modifying the suggestions to:

Ensure you consume all stacks of <StatusLink status="MANAFICATION" /> so you don't lose out on <ActionLink action="PREFULGENCE" /> your single most powerful action.

and

GCDs used during <StatusLink status="MANAFICATION"/> should be limited to actions that consume <StatusLink status="MANAFICATION"/>

Be acceptable?

AzariahAeroborn:
Since Manafication lasts 30 seconds, there's room to fit in quite a few non-consuming GCDs without losing the buff, so I'm not 100% sure the allowable GCDs tracking & suggestion even makes sense any more.  But if you're comfortable with this behavior and copy I won't block on it.

Cephid:
Hrm, let me think on that today; looking at things in reality only unenchanted moves, vercure, verraise cannot consume Manafication stacks in the GCD space.  Realistically I cannot think of a route where you wouldn't consume them unless the boss disengaged or you were trying not to consume them upon contemplation.  It might be better served to just throw it under procs.

Cephid:
Yeah for now lets leave it with the window; if folks would like it gone we can just remove it later I think.  I've been told it's somewhat helpful for checking to see if they put the right potencies in the window though I don't know how many would actually do that kind of checking.

AzariahAeroborn (approved):

PR #2097, opened 2024-08-24
[Core] adding charges information to defensives + [AST] minor updates (change log, contributors, defensive, draw wording, patches)

otocephaly:
## Pull request type

- [x] This is a patch bump

## Pull request details

- [ ] This is in response to a discussion or thread on Discord: *[[Link to thread or start of discussion]]*
		Adding charges information to Core Defensives to allow users to see what charges are out there when. The inspiration came from Lightspeed as a user could theoretically hold a charge throughout the fight even if they could use it theoretically anywhere.
		Adding AST defensives for lightspeed/swiftcast, splitting out changelog, adding sushi as contributor, update minor wording in draw, prep patches for eventual support.

- [x] Job had changes, but is not quite yet ready to be marked as supported.

## Testing / Validation

		Example of poor usage: https://www.fflogs.com/reports/kYW9D4KX3LjJcbTa#fight=2&source=26
		Example of high-parse usage: https://www.fflogs.com/reports/a:grKqwYdk7GCZXWaf#fight=12&source=10

## Job Maintenance

## Screenshots
These screenshots are intended to show the charges update and the prepull fix.

Poor lightspeed/celestial intersection usage example:
image

High-parse usage example:
image


src/parser/core/modules/Defensives.tsx
```diff
@@ -124,15 +130,16 @@ export class Defensives extends Analyser {
 							content: {
 								content: <Table compact unstackable celled>
 									<Table.Body>
-										{
-											this.tryGetAdditionalUseRow(defensive)
-										}
+										{firstAdditionalUsageTry}
```
AkairyuGestalter: It seems like the changes we made recently to better clamp around the actual pull start and end times might be related to the prepull issues we're seeing here. There's some code in getAdditionalUsageData that is trying to adjust the availableTimestamp based on whether there was a prepull use of the action or not, and that's now causing the problems we're seeing as a side effect.

Instead of this change to check for prepull and skip the tryGetAdditionalUseRow, I think we should remove that availableTimestamp adjustment in getAdditionalUsageData. In my testing, that fixes both the prepull row duplication issue, and the other issue reported around mid-fight available windows not showing properly.

src/parser/jobs/ast/modules/Defensives.tsx
```diff
@@ -12,5 +12,7 @@ export class Defensives extends CoreDefensives {
 		this.data.actions.HOROSCOPE,
 		this.data.actions.NEUTRAL_SECT,
 		this.data.actions.COLLECTIVE_UNCONSCIOUS,
+		this.data.actions.LIGHTSPEED,
+		this.data.actions.SWIFTCAST,
```
AkairyuGestalter, line 16: It _feels_ weird to me to call either of these "Defensives" but I get why they're here. It might make more sense long term to have a separate "utility" cooldowns report, or update the description/titling on Defensives to account for both, but I'm not going to hold up this PR for that :)

src/parser/jobs/ast/changelog.tsx
```diff
@@ -1,10 +1,21 @@
+import {DataLink} from 'components/ui/DbLink'
 import CONTRIBUTORS from 'data/CONTRIBUTORS'
 import React from 'react'
 
 export const changelog = [
+	{
+		date: new Date('2024-08-24'),
+		Changes: () => <>Added <DataLink action={'LIGHTSPEED'} /> and <DataLink action={'SWIFTCAST'} /> to defensives. </>,
```
AkairyuGestalter, line 8: Probably worth a mention that we're fixing the prepull display bugs here too

src/parser/jobs/ast/index.tsx
```diff
@@ -34,7 +34,7 @@ export const ASTROLOGIAN = new Meta({
 	},
```
AkairyuGestalter, line 34: Is there something else still extant to finish wrt supporting AST for 7.0 to 7.05? If not, should we be marking the job supported now?

AkairyuGestalter (changes requested):
Couple comments, mainly a suggestion for improving the prepull report fix since it seems to be running afoul of other core changes we've made since we last tried to deal with it!

src/parser/jobs/ast/index.tsx
otocephaly: technically an update to arcana logs that sushi has started and I think she wanted to do a more thorough review, but otherwise, I think it's good

src/parser/jobs/ast/modules/Defensives.tsx
otocephaly: That's fair -- would be good to have a utility module in a similar sense
I was loosely thinking they are "defensives" in a way that keeps ABC for healing or raises

src/parser/core/modules/Defensives.tsx
otocephaly: oooh okie! would you like me to include that in this PR or do a separate one since it's more of a fix for both?

otocephaly: after looking at removing the available timestamp adjustment, it doesn't quite fix the duplicate issue
let me do a separate PR for this specific issue so we can address both

(e.g. if I comment out that adjustment, I see the fix for the missing suggestion, but I don't see a fix for the duplicate noted by that other commentor)
https://discord.com/channels/441414116914233364/470049935077933076/1275859622196150423
Here is what it looks like in dev after commenting out the available timestamp adjustment
image

otocephaly: nvm I don't know what I'm doing lol I see something about clamps and got lost lol

otocephaly: fyi, this will be removed from this PR as it's addressed independently in #2100

src/parser/jobs/ast/changelog.tsx
otocephaly: no longer applicable for this PR

otocephaly:
Hi @AkairyuGestalter ,

PR is now good to go again
I removed the prepull fix since we are looking at that in another PR
Only part is if you are good with the information wording for charges in defensives (I think it's fine, but I wrote it so maybe it comes across funny or something lol)
Thank you for the help with this and I appreciate the comments; always keeping me on my toes!

cc: @suShirow

AkairyuGestalter (approved):

AkairyuGestalter (approved):

PR #2100, opened 2024-08-26
[Core] defensives prepull fix

otocephaly:
Fixes bugs relating to available timestamp adjustments for prepulls and prepulls never used again duplicates

## Pull request type

## Pull request details

- [x] This is in response to a discussion or thread on Discord: *[Link to thread or start of discussion]*
    - https://discord.com/channels/441414116914233364/470049935077933076/1275859622196150423
    - https://discord.com/channels/441414116914233364/470049935077933076/1277048869619765403
    - fix prepull shenanigans in defensives

## Testing / Validation

    Log 1 from Protraction duplication https://www.fflogs.com/reports/v2VzJLCMwrxA1mPF#fight=1&source=9
    Log 2 for missing available suggestions https://www.fflogs.com/reports/762kPgLXJDp1wCAR#fight=6&source=2
    Log 3 for weird shake it off https://www.fflogs.com/reports/7WVCf1XMdnzATHQy#fight=11&source=3

## Job Maintenance

## Screenshots for bug fix
Log 1
image

Log 2
image

Log 3
image


src/parser/jobs/sge/modules/Gauge.tsx
```diff
@@ -210,7 +210,7 @@ export class Gauge extends CoreGauge {
 	//#endregion
 
 	private onComplete() {
-		const addersgallLeniency = ADDERSGALL_TIME_REQUIRED / 2
```
AkairyuGestalter: I'm not seeing the issue you mentioned in the PR notes in production at the moment, and I don't think this change is accurate. Can we drop it?

AkairyuGestalter (changes requested):
I don't think the SGE part is necessary, and we need to fix up the changelog to pass the linter, but otherwise this seems fine.

src/parser/jobs/sge/modules/Gauge.tsx
otocephaly: okie removed!
(it's been a while since I looked at this, but I think it was a dev only issue, so if you happen to see it again, at least it won't come as a surprise lol)

otocephaly:
Hi @AkairyuGestalter , test seems to have an issue with removing the clamps
what would you recommend?
image

AzariahAeroborn:
That test failure is occurring because removing the clamps on getEpochTimestamp breaks the logic of the test -- now the very first action in the test with a -Infinity timestamp isn't getting clamped to the pull time of the fight and as such it fails the test which expects that it should be clamped.  

Overall, from looking at the actual change being made in the Defensives module, I'm pretty sure we don't need to remove the clamps at all.  Instead of the proposed changes, can we just:
1) Remove the call to tryGetAdditionalUsageRow at lines 127-129 at the beginning of the table
2) Update the code for tryGetUsageRow to check if the first usage is pre-pull.  
3) If the first usage is not pre-pull, have tryGetUsageRow output this.tryGetAdditionalUsageRow(defensive) before the current output, then continue on with the rest of the output.
4) If the first usage is pre-pull, only output the current output (don't check for additional usages before the first use, since the first use was prepull)

For log #2, the issue may be coming from line 171 in the current production code -- that line seems to assume pre-pull events will have a timestamp _equal to_ the pull time, and that isn't true any more... the timestamp will be less than the pull timestamp.  Maybe just changing that to a <= check will fix it?

AzariahAeroborn (dismissed):
See comment in discussion thread.

otocephaly:
Hi @AzariahAeroborn @AkairyuGestalter 

I reverted clamps back to normal as recommended (I wonder if we should clamp to the lowest expected prepull e.g. something like this.parser.pull.timestamp - 30 seconds just for display / calculation purposes. for now, I leave as is)

For the prepull in defensives stuff, I used the logic I had to remove the duplicates in the table and added it to the getAdditionalUsage since you recommended looking there. I think when I added that prepull available timestamp adjustment line way back when, I made it a lil too convoluted that I am not sure what I was thinking lol... anywho, I updated the former line 162 to include a check for prepull since it assumed if the timestamp was the same then it was available from the beginning. I left a comment to show my thought process on that line. I removed the old convoluted logic for whatever it was doing

Thank you

AkairyuGestalter (approved):
This is looking good now! And just in time for FRU to really make people think about defensives!

Issue #2091, opened 2024-08-22
Dancer Closed Position uptime over 100%


keysmashes:
![image](https://github.com/user-attachments/assets/dee6ffe7-3b57-4a7c-adcd-57daf8f015d1)

it's nice to be reassured that i'm doing well, but i don't think i'm _that_ good

sample log: https://xivanalysis.com/fflogs/a:jZCa4GFzRM19VHXt/18/7

PR #2101, opened 2024-08-26
[DNC] fixed issue #2091

megurte:
## Pull request type

- [ ] This is a patch bump

## Pull request details

- [x] This is in response to a GitHub issue: *#2091*

## Testing / Validation

   *https://xivanalysis.com/fflogs/a:jZCa4GFzRM19VHXt/18/7*
   *https://www.fflogs.com/reports/4C1XGMcfxDJr7hqp#fight=25&type=auras&ability=1000048&source=10*

## Job Maintenance

### DIFF
--- a/src/parser/jobs/dnc/modules/DirtyDancing.tsx
+++ b/src/parser/jobs/dnc/modules/DirtyDancing.tsx
@@ -156,7 +156,7 @@ export class DirtyDancing extends ActionWindow {
 		const statusTime = Math.max(this.statuses.getUptime(statusKey, this.actors.friends) - this.downtime.getDowntime(), 0)
 		const uptime = Math.max(this.parser.currentDuration - this.downtime.getDowntime(), 0)
 
-		return (statusTime / uptime) * 100
+		return Math.min((statusTime / uptime) * 100, 100)
 	}
 
 	override onComplete() {


megurte:
Explanation: All logs has a delta in few ms that contains precast of skills:
image
image

So in case of buffs that active during the whole fight like food or close position the ration between uptime and statusTime will be more then 1 (more then 100%)

src/parser/jobs/dnc/modules/DirtyDancing.tsx
```diff
@@ -156,7 +156,7 @@ export class DirtyDancing extends ActionWindow {
 		const statusTime = Math.max(this.statuses.getUptime(statusKey, this.actors.friends) - this.downtime.getDowntime(), 0)
 		const uptime = Math.max(this.parser.currentDuration - this.downtime.getDowntime(), 0)
 
-		return (statusTime / uptime) * 100
+		return Math.min((statusTime / uptime) * 100, 100)
```
AzariahAeroborn, line 159: We intentionally don't do clamps like this, so that the higher number shows in development when the calculations go haywire (and the green bar number before you "unroll" the checklist item to view the total does clamp in production).

What's going on under the hood here is that FFLogs changed their definition of "0.00s" to be the start of combat to level out some speed racing hijinx people were doing, and they're now including some pre-pull events along with that change, which has changed the definition of "currentDuration" because now there's a little more event time in the pull than combat time.  But it's most visible with Dance Partner, since that's generally set completely external to the pull and thus should "always" get the over 100% total of including the pre-pull event timing.

AzariahAeroborn (changes requested):

AkairyuGestalter:
Closing this PR, since the problem is a deeper one than just clamping the values. We're going to end up exposing this issue on SGE as well after #2106 merges, so we'll be better served looking at making a change at the core "how long did this fight really last" calculation level.

PR #2122, opened 2024-10-13
[BRD] Remove buff window GCD trackers filter for more intuitively display the correct GCD wave.

YuzukiTsuru:
## Pull request type

## Pull request details

For BRD, the buff window GCD trackers can indicate whether Muse is being maximized. However, it has a filter set to only display cases with errors, which can be misleading as it doesn't clarify what is correct. This submission removes that filter and will show all GCDs utilizing Muse

## Testing / Validation
- https://www.fflogs.com/reports/haQbJVKyWZtY4w3c

## Job Maintenance
- [x] I prefer to receive any communications through GitHub only


AkairyuGestalter (approved):
When Yumiya originally developed this, we were concerned that showing all of the windows would get too cluttered, so we opted to hide properly executed windows similar to BLM's Rotation Outliers module.

Since this won't end up showing literally the entire fight the way BLM's module would have, and it promotes being able to see at a glance what a properly executed window looks like, we both agree this is a fine change to make for BRD!

PR #2126, opened 2024-10-19
Support browser launching for developers on Linux and Mac systems

AkairyuGestalter:
## Pull request type

## Pull request details

We previously updated the webpack devServer configuration to handle some change on Windows systems that was preventing the "open the browser to localhost:3000" behavior of `yarn start` from working. That change was causing `yarn start` to crash on Linux systems however, which wasn't immediately apparent until I started migrating my desktop over to Fedora.

I found an example that recommended changing the `open` property of the devServer config from the localhost URI to the OS command that should be used to launch the browser, and moving the URI to the `openPage` property instead.

I haven't yet tested if this change retains the expected behavior on Windows, but it does resolve the problem on my Linux system.

## Testing / Validation

Verify that `yarn start` correctly opens the default browser to localhost:3000 on Window systems.

### DIFF
--- a/webpack.config.ts
+++ b/webpack.config.ts
@@ -12,6 +12,17 @@ import {calculateLocaleCompletion} from './config/locale'
 
 import './config/loadDotenv'
 
+function browserOpenCommand() {
+	switch (os.platform()) {
+	case 'win32': // Windows systems
+		return 'explorer'
+	case 'darwin': // Mac Systems
+		return 'open'
+	default: // Linux and other Unix-compliant systems
+		return 'xdg-open'
+	}
+}
+
 /* eslint-disable @typescript-eslint/no-magic-numbers */
 
 interface Environment {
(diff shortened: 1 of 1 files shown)


webpack.config.ts
```diff
@@ -43,7 +44,8 @@ export default (env: Environment, {
 	devServer: {
 		host: 'localhost',
 		port: 3000,
-		open: 'http://localhost:3000',
+		open: os.platform() === 'win32' ? 'explorer' : 'xdg-open',
```
nonowazu: While this covers windows and linux use cases, it doesn't work for the mac use case (which is just `open`). So you might need a double ternary there.

AkairyuGestalter: Pulled the selection of the command string out to a separate function, rather than nested ternary nonsense. I'd originally opted not to handle mac since I had no way of personally validating it, but don't mind fixing it if someone else can!

AzariahAeroborn (approved):

PR #2130, opened 2024-10-20
Fix xivapi not support language zh

YuzukiTsuru:
## Pull request type

## Pull request details

if only set `language={i18nStore.gameLanguage}` it will lead to xivapi `language=zh`, which is not supported by xivapi.

```
{"code":400,"message":"invalid request: Failed to deserialize query string: invalid or unsupported language \"zh\""}
```

Add a redirect to EN when using Chinese Client
```ts
const gameLanguage = i18nStore.gameLanguage === Language.CHINESE
	? Language.ENGLISH
	: i18nStore.gameLanguage
```

image

### DIFF
--- a/src/components/ui/DbLink.tsx
+++ b/src/components/ui/DbLink.tsx
@@ -34,8 +34,8 @@ export function Provider({children}: ProviderProps) {
 
 	// fix for xivapi not support language zh
 	const gameLanguage = i18nStore.gameLanguage === Language.CHINESE
-	? Language.ENGLISH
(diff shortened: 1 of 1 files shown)


ackwell:
Hm, this shouldn't be possible - https://github.com/xivanalysis/xivanalysis/pull/2129/files#diff-7a4265bb108fa9584f8d2efb0b4fc2a373a4504e50bd07378a54673868e9a827R8 prevents the i18n store from exposing CN as a game language. Possibly related to local storage>

PR #2138, opened 2024-11-10
Patch and layer definitions for 7.1

AkairyuGestalter:
Patch 7.1 housekeeping to let devs get started on job updates. Added patch definition and action/status layer files. Core updates to happen on a separate PR after we have game data files and some early parses to check for odd invuln stuff.

Technically okay to merge before the servers go down. It won't change anything immediately other than starting to display the "xivanalysis supports up to patch 7.05 of 7.1" message on the home page a bit early.


hintxiv (approved):
🐐

PR #2160, opened 2024-11-21
Updated potencies and Bumped RDM Support

Cephid:
## Pull request type

- [x] This is a patch bump

## Pull request details

- [x] The job(s) in this patch bump had only potency changes for this patch, which do not effect analysis
  - [x] I have updated the data files for all potency changes for this patch

## Job Maintenance


src/data/ACTIONS/root/RDM.ts
```diff
@@ -204,7 +204,7 @@ export const RDM = ensureActions({
 		onGcd: true,
 		speedAttribute: Attribute.SPELL_SPEED,
 		cooldown: 2500,
-		potency: 620, // Gains 11 Black Mana, if Black is lower 100% Verfire ready
+		potency: 650, // Gains 11 Black Mana, if Black is lower 100% Verfire ready
```
AkairyuGestalter: These changes ought to get done in the patch layer file, rather than root, so pre 7.1 analysis doesn't get messed up because of the difference

Cephid: I've rolled back the changes to root and re-applied them to the 7.1 layers file.

AkairyuGestalter (approved):

PR #2182, opened 2025-01-26
Replace yarn with pnpm, update node to 22.x

ackwell:
# for developers

this update _will_ necessitate at least minimal interaction to get things running. sorry!

## updating tools

there's a few options available. the below are listed in order of recommendation, however any of the below should function correctly. see troubleshooting below if you hit issues with any of these approaches, and reach out in \#dev-general if you get stuck.

after updating, you should have versions _similar_ to the below (minor/patch may differ, but major should match):

```console
$ node -v
v22.13.1
$ pnpm -v
10.0.0
```

### version managers

a configuration file is provided for [mise](https://mise.jdx.dev/).

running `mise install` in your checkout should automatically install and configure the appropriate environment, and scripts should work as expected.

if you are using another version manager such as `asdf`, `nvm`, or similar, i'd _personally_ recommend checking out mise - it's really great - but you can also continue using it by installing node 22 with your manager of choice, and proceeding with the instructions below for corepack.

### corepack

[corepack](https://nodejs.org/docs/latest-v22.x/api/corepack.html) can be used to manage the package manager. you'll need to install node 22 yourself, either from [node's website](https://nodejs.org/en), or your package manager of choice.

1. ensure `node -v` prints something starting `v22`.
2. `corepack enable`
3. `corepack install`

## installing dependencies

while both `yarn` and `pnpm` manage the `node_modules` folder, they take different approaches for managing the resulting file structure. `pnpm` will try its best to work with a mixed environment, however i'd strongly recommend allowing it to start fresh.

1. delete `node_modules` completely<br>`pnpm run clean` is recommended, or the ol' `rm -rf node_modules` if clean doesn't work.
2. `pnpm install`

## running commands

scripts need to be run with `pnpm` instead of `yarn` - but the script names are the same. example changes:

|task|old|new|
|---|---|---|
|install dependencies|`yarn`|`pnpm install`|
|start dev server|`yarn start`|`pnpm run start`|
|scripts in general|`yarn <x>`|`pnpm run <x>`|

## troubleshooting

<table><tbody>
<tr><td><details>
<summary><code>UsageError: This project is configured to use pnpm because /.../package.json has a "packageManager" field</code></summary>

<br>This error will be thrown if trying to install dependencies or run scripts using `yarn` or `npm`. Use `pnpm` for commands, see "running commands" above.

</details></td></tr>
<tr><td><details>
<summary><code>ERR_PNPM_UNSUPPORTED_ENGINE  Unsupported environment (bad pnpm and/or Node.js version)</code></summary>

<br>`pnpm` is being used, but it has detected that either its own, or node's version is incorrect. the error message is fairly descriptive, and should tell you what is at the wrong version, and what version is expected. refer to the "updating" section above for instructions specific to xiva.

</details></td></tr>
</tbody></table>

if you hit an issue not mentioned above, and you have not already done so since setting up `pnpm`, delete `node_modules`, and reinstall it with `pnpm install`. if the issue persists, please reach out in \#dev-general.

# implementation

this section covers this pr. if you're here to check how to _use_ the results of this pr being merged, check the sections above.

## what & why

this pr replaces [yarn](https://classic.yarnpkg.com/) with [pnpm](https://pnpm.io/)

we're currently 3 major versions behind `yarn`'s latest, however notably, yarn >=2 is a completely rewrite (which made some decisions I'm not a huge fan of).

given it's a new package manager either way, cutting over to pnpm seems worthwhile. it offers a more powerful featureset, and some nice sanity checks that help keep dependencies aligned with usage.

...also i'm used to it because of work. won't pretend that's not the case.

## pr content and decisions

the changes in this pr are intended to be as minimal as possible (well, the _results_ are. new lockfile means lmao tier line count change). package lock file was imported across, so resolved versions should match exactly to what we're already using from the old `yarn.lock`.

notably, pnpm builds `node_modules` in a manner that is significantly stricter than `yarn`'s approach, which lead to some failed imports / scripts:
- a few imports of `react-router` have been replaced with `react-router-dom`. xiva doesn't depend directly on `react-router`, but the `-dom` package, so using the re-exports from that
- new packages in the package.json: these are transitive dependencies we were using without explicit version ranges, this just encodes the version ranges

everything else is just `yarn <x>` -> `pnpm run <x>` / `yarn` -> `pnpm install` and such forth

as with yarn, `pnpm run <x>` can be shorthanded as `pnpm <x>`, however `pnpm test` risks footguns, so i'm keeping to `pnpm run <x>` for docs.


otocephaly:
Seems to work fine for me! I had to run `npm install -g pnpm` as suggested by akairyu

pnpm install worked
pnpm run start worked
pnpm run extract worked

example change for fun was to change a suggestion pld
image

PR #2189, opened 2025-02-07
Update typecript, eslint

ackwell:
Breaking up the updates because ho boy there's a lot to do.

Per title, this PR updates typescript and eslint, and a few bits and pieces to support those. Shitload of files changed, but it's mostly small changes.

typescript:
- initialisers on abstract properties have been removed
- some of the type yoga in `core/filter` has been rejigged slightly so it stops complaining
- caught errors are now `unknown`, and need to be narrowed with runtime type verification

eslint:
- jsx use does not require `React` to be in scope, removed all unused imports
- for consistency, also removes all other usages of the React namespace import
- refactored bare expressions
	- `x || y()` -> `if (!x) { y() }`
	- `x && y` -> `if (x) { y() }`
	- `x ? y() : z()` -> `if (x) { y() } else { z() }` 
- removed unused booleans in conditions
	- `true && x` -> `x`
	- `false || x` -> `x`
	- etc
- a variety of existing rules that hadn't been catching some instances for some reason


ackwell:
collect'n labels like a christmas tree out here

PR #2197, opened 2025-02-13
Bucket event filter predicates by type

ackwell:
this'd really benefit from testing on like any log you can think of.

---

currently; event hooks are filtered by arbitrary predicate functions - `filter<T>()` is really just a predicate builder, and any `(event) => event is T` is a valid predicate.

this is great for composibility!
this is shit for optimisations!

for every event, _every_ hook predicate needs to be called to work out which hooks should actually be executed. given a vast majority of events will trigger a minority of hooks, this is pretty wasteful.

the most common property of events (perhaps other than their timestamp) is their type - and conveniently, it's also a very-commonly-filtered property. if we can bucket predicated by the type they filter for, and then bypass any predicates that don't have a matching type _before_ calling them, we can save on that execution time.

---

this pr implements the above through the introduction of a(n optional) special key on filter predicates, and a side channel within the filter implementation to provide that key in straight-forward consumption.

more concretely;

- `Dispatcher` will now check predicates for a `[FILTER_KEY]` (symbol) property.
	- If it exists, it will be used to bucket the predicate - it will only be called if the event's type matches the key (map lookup).
	- If it _does not_ exist, it will be called for every event.
- `filter<T>()` checks the value provided to any `.type(...)` calls, and exposes it in `[FILTER_KEY]`
	- this is currently only checking for literal strings, which catches the vast majority of filters that i've checked.
	- any non-literal type filters (i.e. `.type(oneOf(...))`), or filters with no type at all, will not expose a `[FILTER_KEY]`.

it'd be possible to support `oneOf` and similar as well, but I don't think the gains are worth the additional effort at this point in time.

before:
image

sick gains:
firefox_zVWf24L7Ts

### DIFF
--- a/src/parser/core/Dispatcher.ts
+++ b/src/parser/core/Dispatcher.ts
@@ -3,8 +3,14 @@ import {Injectable} from './Injectable'
 
 type Handle = (typeof Injectable)['handle']
 
+export const FILTER_TYPE = Symbol('filter.type')
+
 /** Signature of event filter predicate functions. */
-export type EventFilterPredicate<T extends Event> = (event: Event) => event is T
+export type EventFilterPredicate<T extends Event> = {
+	(event: Event): event is T
+	/** If set, will be used to bucket the predicate as a first-pass optimisation. */
+	[FILTER_TYPE]?: string,
+}
 
 /** Callback signature for event hooks. */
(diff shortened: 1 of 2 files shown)


ackwell:
what's the worst that could happen

PR #2203, opened 2025-03-22
fix: replace custom Inner Release analyser with core procs module

kwerey:
## Pull request type

## Pull request details

- [x] This is in response to a discussion or thread on Discord: https://discord.com/channels/441414116914233364/470049767477608459/1337810619293765735

Warrior was using a custom analyser for its Inner Release buff which was calculating percentages and displaying them to many many significant figures. It looks like we can replace this with the core procs module since the only specific logic it was doing was counting number of stacks used or expired. 

Two questions:

-  This moves the guidance from the 'Checklist' section to a 'Suggestion'. I'm not sure the intended difference between those two - is it OK to move this to be a suggestion so it is provided alongside other proc related guidance? 
- I renamed the element for the localisation text to match the format that someone suggested when adding other new warrior procs, but there is existing localisation text under the old name - should I add the old description field to the new generated localisation files, or is it better to leave that for language maintainers?

## Testing / Validation

Unused stacks: https://www.fflogs.com/reports/a:GJZz76gBnvaYb3PL?fight=6&source=23

On live XIVA:

image

With this change:
image

## Job Maintenance


AkairyuGestalter:
Can you add a changelog for this? Changes seem pretty sane otherwise, will give them a test over the weekend.

kwerey:
Ah yep of course, thanks for the reminder. I've added one now and rebased this branch.

AkairyuGestalter (approved):

PR #2206, opened 2025-03-24
[AST] 7.2 bump

otocephaly:
Preemptively bumping this since changes are minor

## Pull request type

- [x] This is a patch bump

## Pull request details

- [x] The job(s) in this patch bump had only potency changes for this patch, which do not effect analysis
  - Oracle is oGCD and gravity doesn't change analysis even if it's a gain on two

## Job Maintenance


AkairyuGestalter (approved):
I don't think I'd noticed before that we didn't have AoEUsages overrides for the healers. Not worth holding up a patch bump for it though...

PR #2228, opened 2025-04-17
[AST] update card CD management plus wording

otocephaly:
## Pull request type

## Pull request details
- [x] This is in response to a discussion or thread on Discord: *[Link to thread or start of discussion]*(https://www.fflogs.com/reports/a:LhGYBmQVjwpJWKqb?fight=20&source=75)
    - Update amount of cards to align with divination burst windows rather than keeping always on CD

## Testing / Validation
    - https://www.fflogs.com/reports/a:LhGYBmQVjwpJWKqb?fight=20&source=75
    - (overwritting example) https://www.fflogs.com/reports/a:N418xZWmrDX2aKBR?fight=25&source=359

   *[Provide a list of log links - either direct to FFLogs or the xivanalysis log link for development or production]*

## Job Maintenance

## Screenshots
Updated wording and logic
image

Overwrite example and updated macrocosmos wording
image

Old for reference
image
image


AkairyuGestalter:
I still feel like there's a better way to represent the intent we're trying to get at here. The cooldown fudging to only ask for a certain number of Draw uses based on alignment with Divination's CD makes sense, but the checklist requirements for using Play 1 and Lord are already being handled by the ExpectedCardEvaluator as part of the Divination RaidBuffWindow analysis. I feel like what we're actually missing to account for the "you might intentionally drift outside of the window because of fight timeline reasons" case, is a warning against overwriting Lord/Play 1 if not used when the next Draw is pressed, rather than trying to stuff to much into this checklist's requirements.

If you'd rather we keep this as-is at least for now, let me know and I'll go ahead and merge it in since it is doing what it was designed to do. I just wanted to bring up the alternative while I was thinking about it :)

otocephaly:
> I still feel like there's a better way to represent the intent we're trying to get at here. The cooldown fudging to only ask for a certain number of Draw uses based on alignment with Divination's CD makes sense, but the checklist requirements for using Play 1 and Lord are already being handled by the ExpectedCardEvaluator as part of the Divination RaidBuffWindow analysis. I feel like what we're actually missing to account for the "you might intentionally drift outside of the window because of fight timeline reasons" case, is a warning against overwriting Lord/Play 1 if not used when the next Draw is pressed, rather than trying to stuff to much into this checklist's requirements.
> 
> If you'd rather we keep this as-is at least for now, let me know and I'll go ahead and merge it in since it is doing what it was designed to do. I just wanted to bring up the alternative while I was thinking about it :)

hmm I see what you're getting at. the only edge case it covers that raidbuffwindow does not is when a boss is killed between the odd minute and even minute where in that case, you'd be better off just using the melee card and lord during the 1 min burst. The overwriting is a good suggestion, let me take that back and put something in.
Do you think it'd be better to remove the checklist portion and only put a recommendation for when the boss is killed within 1 min of the expected divination cast and the balance/lord is held? (an example of this happening is M5S since the boss even during enrage is always killed just before raidbuffs can go out, could put as a medium or minor suggestion since sometimes you can't predict if the boss will be killed before raid buffs)

AkairyuGestalter:
> hmm I see what you're getting at. the only edge case it covers that raidbuffwindow does not is when a boss is killed between the odd minute and even minute where in that case, you'd be better off just using the melee card and lord during the 1 min burst. The overwriting is a good suggestion, let me take that back and put something in. Do you think it'd be better to remove the checklist portion and only put a recommendation for when the boss is killed within 1 min of the expected divination cast and the balance/lord is held? (an example of this happening is M5S since the boss even during enrage is always killed just before raidbuffs can go out, could put as a medium or minor suggestion since sometimes you can't predict if the boss will be killed before raid buffs)

I think keeping the checklist with just the "did you use Draw enough" requirement, but leaving off the "use Play 1" and "use Lord" requirements since those are covered by Divination is good. I know we're usually hesitant about being too strict about ending the fight with resources still available since killtimes are variable, so if you choose to add something there, I'd probably make it only ever a Minor suggestion.

otocephaly:
Hi @AkairyuGestalter ,
I've updated per your suggestions (the screenshots above have been updated)
Additionally, I updated macrocosmos usages to have higher thresholds and better wording to not imply it should be used on CD (it's mostly just to suggest that people use it more often in the fight)

src/parser/jobs/ast/modules/Draw.tsx
```diff
@@ -139,33 +159,24 @@ export class Draw extends Analyser {
 		this.checklist.add(new Rule({
 			displayOrder: DISPLAY_ORDER.DRAW_CHECKLIST,
 			name: <Trans id="ast.draw.checklist.name">
-				Play as many cards as possible
+				Play as many cards as possible considering <DataLink action="DIVINATION" /> windows.
```
AkairyuGestalter: Probably want to revise the checklist name and description now that it's truly focusing on Draw usage. Otherwise lgtm.

otocephaly: good catch! thank you
updated

AkairyuGestalter (approved):
I like it. Away we go.

PR #2232, opened 2025-05-08
Capitalize 'Brotherhood'

wfeli:
## Pull request type

## Pull request details

When analyzing a MNK log the 'b' in Brotherhood was not capitalized.

## Testing / Validation

any MNK log with Brotherhood used is suitable. I created a local build and verified it.
https://xivanalysis.com/fflogs/ZNmkhxyHrDYzdvV8/9/423
## Job Maintenance
- [x] I prefer to receive any communications through GitHub only


src/parser/jobs/mnk/modules/Brotherhood.tsx
```diff
@@ -44,7 +44,7 @@ const BROTHERHOOD_DRIFT_TOLERANCE = 1500
  * said windows to provided RoF windows.
  */
 export class Brotherhood extends RaidBuffWindow {
-	static override handle = 'brotherhood'
+	static override handle = 'Brotherhood'
```
AkairyuGestalter: Rather than capitalizing the handle, we should be setting a title here. That allows it to be translated for other languages as well. Should be able to follow examples from other modules!

AkairyuGestalter (changes requested):
Thanks for the interest in contributing! One minor comment about the way this is getting fixed to resolve.

src/parser/jobs/mnk/modules/Brotherhood.tsx
wfeli: ok, I got back from a trip and will do this in the next few days. thanks for pointing out

AkairyuGestalter (approved):

PR #2264, opened 2025-12-06
[DRK] Fixed minor issue with non-standard Esteem evaluator

VioletStardustXIV:
## Pull request type

## Pull request details

Fixes an issue where double Abyssal Drain was always marked as incorrect for Living Shadow, even when it was replacing an ability of the same potency.

## Testing / Validation

Example log:
https://xivanalysis.com/fflogs/a:XxAFJpLgP7ZYQ62m/5/134

image

In the above set of Esteem ability usages, the last one (where an Edge of Shadow was replaced by an Abyssal Drain) should not have been marked against the user (but is in live), since they are equivalent. While a non-standard Esteem it's equivalent-to-a-gain. The other three (marked with an X) are still incorrect though.

## Job Maintenance

I'm active on the discord but not part of the official job maintainer group. Feel free to reach out to me there or here.


AkairyuGestalter (approved):
Food for thought: Consider adding a header for the EsteemUsageEvaluator so it shows the column in the table to help users see at a glance which windows are the ones that the suggestions are calling out. Not worth blocking this fix on it, but could be useful.

PR #2267, opened 2025-12-12
[DRK] Include pre-pull Living Shadows for 'Use your cooldowns' and Esteem evaluator

VioletStardustXIV:
## Pull request type

## Pull request details

Indexes Living Shadow usage over Scorn being applied to the player instead of Living Shadow being used if the player is level 100. fflogs does not include Living Shadows used pre-pull, but it does always include Scorn.

- Fixes prepull Living Shadows not being counted for 'Use your cooldowns'
- Fixes actions used by Esteem not including prepull Esteems

Also visually changes the actions used by living shadow to remove Living Shadow

image

Prepull Living Shadows are still absent for sub-100 logs, but everything displays as it did before.

## Testing / Validation
This log (rank 1 Necron) has a -6s Living Shadow
https://xivanalysis.com/fflogs/D31Mm8xw4yWBZfVt/2/12

Before:
image

image

After:
image

image

## Job Maintenance
Not a maintainer but I am active on discord.

### DIFF
--- a/src/data/ACTIONS/root/DRK.ts
+++ b/src/data/ACTIONS/root/DRK.ts
@@ -296,7 +296,7 @@ export const DRK = ensureActions({
 		speedAttribute: Attribute.SKILL_SPEED,
 		breaksCombo: false,
 		potencies: [{
-			value: 620,
+			value: 600,
 			bonusModifiers: [],
 		}],
 	},
@@ -308,7 +308,7 @@ export const DRK = ensureActions({
 		speedAttribute: Attribute.SKILL_SPEED,
 		breaksCombo: false,
 		potencies: [{
-			value: 720,
+			value: 700,
 			bonusModifiers: [],
 		}],
 	},
@@ -320,7 +320,7 @@ export const DRK = ensureActions({
 		speedAttribute: Attribute.SKILL_SPEED,
 		breaksCombo: false,
 		potencies: [{
-			value: 820,
+			value: 800,
 			bonusModifiers: [],
 		}],
 	},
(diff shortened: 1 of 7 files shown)


src/data/ACTIONS/root/DRK.ts
```diff
@@ -344,7 +344,7 @@ export const DRK = ensureActions({
 		speedAttribute: Attribute.SKILL_SPEED,
 		breaksCombo: false,
 		potencies: [{
-			value: 800,
+			value: 1000,
```
AkairyuGestalter: Is this fixing a mistake with the potency values or did this get a buff during a patch at some point? If it's the latter, we should leave this at 800 in the root file and put the potency change in the appropriate patch layer file instead.

VioletStardustXIV: It was updated to 1000 in 7.05 (800 in 7.0), though, curiously, Torcleaver's potency in this file is set to 820, though it was only updated to 800 in 7.1

VioletStardustXIV: Let me update this for both :)

VioletStardustXIV: Oh, it looks like Disesteem was properly updated in the patch, but the Delirium combo wasn't, which confused me. Reverting this, and fixing the Delirium combo (should be potency upped in the 7.1 patch layer file).

VioletStardustXIV: Good catch, thanks. Fixed the 7.1 potencies at the base layer and added the potency updates to the 7.1 patch file

AkairyuGestalter (approved):

PR #2269, opened 2025-12-18
[AST] Bump 7.4 and remove unnecessary CO from overheal

otocephaly:
bumped 7.4 and removed the unnecessary status from overhealing for collective unconsciousness

## Pull request type

- [x] This is a patch bump

## Pull request details

		*[For new functionality / added functionality, describe what is being added]*
		*[For bugfixes, please provide reproduction steps and what the expected vs actual behavior is in production right now]*
    - not count the non-heal, non-mit portion of collective unconscious and bump
    

- [x] The job(s) in this patch bump had no changes for this patch
    - the change to Collective Unconscious has no additional effect on analysis since it just changed one status from 5 to 10 seconds. The other change was just to remove the non-heal, non-mit portion from being counted unnecessarily in overheal.

## Testing / Validation

- Any log where an AST cast collective unconscious at least once will show this. I used this one as an example https://www.fflogs.com/reports/t9X8rHvczjymMTgq?fight=3&source=6

   *[Provide a list of log links - either direct to FFLogs or the xivanalysis log link for development or production]*

## Job Maintenance

## Screenshots
Before
image

After
image

### DIFF
--- a/src/parser/jobs/ast/modules/Overheal.tsx
+++ b/src/parser/jobs/ast/modules/Overheal.tsx
@@ -1,7 +1,6 @@
 import {Trans} from '@lingui/react/macro'
 import {DataLink} from 'components/ui/DbLink'
 import {Action} from 'data/ACTIONS'
-import {Status} from 'data/STATUSES'
 import {Events} from 'event'
 import {Overheal as CoreOverheal, TrackedOverhealOpts} from 'parser/core/modules/Overheal'
 import {DISPLAY_ORDER} from './DISPLAY_ORDER'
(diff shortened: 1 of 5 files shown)


AkairyuGestalter (changes requested):
The adjustment to the CU status duration should probably be in the patch 7.4 status layer file rather than in root. If you rebase this onto the current dawntrail branch, it should be there, and that should also fix the checks error now that the 7.4 patch is actually defined too.

otocephaly:
> The adjustment to the CU status duration should probably be in the patch 7.4 status layer file rather than in root. If you rebase this onto the current dawntrail branch, it should be there, and that should also fix the checks error now that the 7.4 patch is actually defined too.

oop you are correct -- I wasn't thinking
I don't recall when it was originally bumped to 5s, so I'll add it into there as well

src/parser/jobs/ast/modules/Overheal.tsx
```diff
@@ -10,6 +11,12 @@ const NEUTRAL_SECT_APPLICATION_BUCKET_ID = 1
 export class Overheal extends CoreOverheal {
 	protected override checklistDisplayOrder = DISPLAY_ORDER.OVERHEAL_CHECKLIST
 
+	private ignoredStatuses: Array<Status['id']> = [
```
AkairyuGestalter: Rather than special-case handling this in considerHeal, you might want to look at how I handled this for Kardia in SGE. If you make a trackedHealCategories bucket for it, and set that bucket's ignore prop to false, the rest is taken care of for you by the core module.

AkairyuGestalter (commented):
One more comment that you can take or leave, but seems like would be a simple fix.

src/parser/jobs/ast/modules/Overheal.tsx
otocephaly: appreciate the heads up and I do like consistency
(I probably should've checked other jobs but I wasn't thinking lol it's just one of those days/weeks/months/years/decades/lifetimes)
I've updated that in the most recent commit
Thank you!

AkairyuGestalter (approved):

PR #2274, opened 2025-12-21
[GNB] 7.4 Support

Rjb99:
## Pull request type

- [ x] This is a patch bump

## Pull request details
- [ X] This is the last PR after all analysis changes for this job have been merged, and now the job is ready to be marked supported for this patch

GNB changes:

Bloodfest -> 60s, resulting in all burst windows being the same. It also now temporarily doubles the gauge limit, allowing us to not need to hold it for alignment anymore.

Gnashing Fang -> 2 charges, no longer affected by SKS, resulting in a new sub 1% dps increase technique that is noted, but not checked for in No Mercy.

Tincture window adjusted to expect 2 Gnashing Fangs.
Sonic Break lowered to 15 seconds, resulting in only 5 ticks instead of 10 ticks.

All changes have been implemented with before/after patch logic to keep gunbreaker fully supported.

## Testing / Validation
https://www.fflogs.com/reports/aBGqYLnWkpj2mb1h?fight=6&type=damage-done
https://www.fflogs.com/reports/P2JaxCmA8nbNp6kd?fight=3&source=22

## Job Maintenance
- [ x] I am active on the xivanalysis Discord and part of the job maintainers group for the job(s) in this PR


Rjb99:
Additionally, Akairyu helped me troubleshoot a bug in Countergauge where 0 deltas in gauge modify resulted in a miscalculation of spent/generated resources.

A simple If Amount !== 0 guardrail has been added to the modify function.

locale/en/messages.json
```diff
@@ -631,6 +633,11 @@
   "gnb.nomercy.chart.notes.no": "No",
   "gnb.nomercy.chart.notes.yes": "Yes",
   "gnb.nomercy.notes.header": "Bloodfest Used",
+  "gnb.nomercy.prepend-message.disclaimer": "This technique is not always possible or correct in the event that <0/> is interrupted, and as such the <1/> module does not expect it, but this technique should be used whenever you can.",
+  "gnb.nomercy.prepend-message.fast-gcd": "If you are running at a faster GCD (2.46 or faster), you can do a full 9 GCDs under <0/> so you will use <1/> right before a weaponskill in the combo and its follow up <2/>, ideally <3/>.",
+  "gnb.nomercy.prepend-message.infographic": "Please see the rotation infographic by Krom at the top of this page for more information.",
+  "gnb.nomercy.prepend-message.intro": "Gunbreaker has a small dps optimization technique that can be done by using the extra charge of <0/> from the half minute of your rotation in order to enter <1/> at the following times:",
+  "gnb.nomercy.prepend-message.slow-gcd": "If you are running at a slower GCD (2.47 or slower), you can only do 8 GCDs under <0/> so you will use <1/> right before a <2/>, ideally <3/>.",
```
VioletStardustXIV: I don't necessarily know that xiva should be encouraging people to do this, but you can (and it's actually a gain) force a 9 GCD No Mercy at 2.48. I'm not sure what the middle ground is between what's here and what's possible but it definitely is possible!

VioletStardustXIV: this feels like nitpicking so feel free to ignore it but maybe:
> you can only do 8 GCDs

->

> you will only have 8 GCDs

src/data/STATUSES/layers/patch7.4.ts
```diff
@@ -7,5 +8,16 @@ export const patch740: Layer<StatusRoot> = {
 		COLLECTIVE_UNCONSCIOUS_MITIGATION: {
 			duration: 10000,
 		},
+
+		BLOODFEST: {
+			id: 5051,
+			name: 'Bloodfest',
+			icon: iconUrl(213623),
+			duration: 8000,
```
AkairyuGestalter: The new Bloodfest status appears to be a 30s duration for me, mistype or is there a reason this is only set to 8s?

src/parser/jobs/gnb/modules/Ammo.tsx
```diff
@@ -55,12 +55,45 @@ export class Ammo extends CoreGauge {
 	override initialise() {
 		super.initialise()
 
-		if (this.parser.patch.before('7.1')) {
-			DoubleDownCost = 2
+		if (this.parser.patch.before('7.1') || this.parser.patch.after('7.3')) {
+			this.ammoModifiers.set(this.data.actions.DOUBLE_DOWN.id, {action: -2})
```
AkairyuGestalter: Most of the other cases that I can recall for having patch-wise const changes did it by making two different const variables for the different values, and conditional logic for assigning the functional variable to those, rather than an assign-and-conditional-modify like this. It still works, and either way is kinda clunky, so I'll leave it up to you if you want to change it.

src/parser/jobs/gnb/modules/Ammo.tsx
```diff
@@ -71,16 +104,24 @@ export class Ammo extends CoreGauge {
...
 	}
+	private onBloodfestApply() {
+		this.ammoGauge.setMaximum(BLOODFEST_AMMO_CAP)
+		// Based on test logs, Bloodfest status and ammo gain comes from the status apply.
+		// This causes ammo to not properly be added when processed on bloodfest cast.
+		this.ammoGauge.modify(MAX_AMMO)
+	}
+	private onBloodfestRemove() {
+		const droppedOverloadAmmo = this.ammoGauge.value - MAX_AMMO
+		if (droppedOverloadAmmo > 0) {
```
AkairyuGestalter, line 117: More of a heads-up than anything else: reading through how CounterGauge deals with the maximum changes, I think ammo you lost due to the ceiling falling on you gets counted as "overcap". GNB doesn't appear to use the built-in CounterGauge overcap stuff, but something you'll want to be aware of if you do in the future.

src/parser/jobs/gnb/modules/Cooldowns.tsx
```diff
@@ -25,8 +25,8 @@ ________________________________________________
 Quick Maths, 2500 * GCD # = Offset, take the higher number for the offset.
 */
 
-const FIRST_USE_OFFSET_BLOODFEST = 5000
-const ALLOWED_BLOODFEST_HOLDTIME = 10000 // Bloodfest is allowed to be held in order to delay it back under buffs, this should be a maximum of 4 GCDs, 2 ahead of NM, 2 after NM cast to dump ammo.
+const FIRST_USE_OFFSET_BLOODFEST = 2500
+let ALLOWED_BLOODFEST_HOLDTIME = 10000 // Bloodfest is allowed to be held in order to delay it back under buffs, this should be a maximum of 4 GCDs, 2 ahead of NM, 2 after NM cast to dump ammo.
```
AkairyuGestalter: I feel like in this case it'd be better to go the ALLOWED_BLOODFEST_HOLDTIME_[BEFORE/AFTER]_[PATCH] route, since we really shouldn't be having re-assignable variables outside of classes like this.

src/parser/jobs/gnb/modules/SonicBreak.tsx
```diff
@@ -43,6 +43,10 @@ export class SonicBreak extends Analyser {
 	}
 
 	override initialise() {
+
+		if (this.parser.patch.after('7.3')) {
+			MAX_TICKS = 5
```
AkairyuGestalter: Patch-wise const stuff again. Also lint is complaining about the magic number here.

AkairyuGestalter (changes requested):
Mostly coding style things, and one question about the bloodfest status duration.

src/parser/jobs/gnb/modules/Ammo.tsx
Rjb99: This logic was implemented due to the fact that overcap would also include regular overcapping.
This metric is a "hey dummy, you stayed overloaded too long and dropped ammo" rather than a total overcap.

AkairyuGestalter: Cool, just wanted to make sure you were aware (also food for thought for me while I'm looking at improvements to the reason defaulting, maybe we could be better about that... hmmm....)

src/data/STATUSES/layers/patch7.4.ts
Rjb99: it is 30s, mistype indeed.

src/parser/jobs/gnb/modules/Ammo.tsx
Rjb99: I'm not sure if the ariable names are great, but updated to consts.
const DOUBLE_DOWN_COST_7_2_AND_7_3 = 1
const DOUBLE_DOWN_PRE_7_1 = 2 // also 7.4 and beyond!

locale/en/messages.json
Rjb99: changed both can's to wills. feels less optional.

src/parser/jobs/gnb/modules/SonicBreak.tsx
```diff
@@ -44,9 +46,8 @@ export class SonicBreak extends Analyser {
 
 	override initialise() {
 
-		if (this.parser.patch.after('7.3')) {
-			MAX_TICKS = 5
-		}
+		MAX_TICKS = this.parser.patch.before('7.4') ? MAX_TICKS_PRE_7_4 : MAX_TICKS_POST_7_4
```
AkairyuGestalter: One final nit: Can we move MAX_TICKS inside of the Sonic Break class as a private maxTicks variable, to separate the consts from the assignable variables?

Rjb99: MAX_TICKS changed to a private MaxTicks value that is assigned the values of the 2 consts based on patch.

AkairyuGestalter (approved):
Thanks for humoring me, looks good now! .shipit

PR #2241, opened 2025-06-30
[DRK] Add Missed Dark Arts info module, including timestamps/reasons for dropped Dark Arts

VioletStardustXIV:
## Pull request type

- [X] This is new functionality or an addition to existing functionality
- [X] This is a bugfix to existing functionality
- [ ] This is a patch bump

## Pull request details

This does a few things:
1. Fixes an issue where TBNs that don't break on allies don't get marked against the player in the 'missed dark arts' suggestion
2. Adds a new module `Missed Dark Arts`, that covers missed Dark Arts from TBN uses. There is information in the above module about which dark arts was lost in the warning, addressing https://discord.com/channels/441414116914233364/1328735109481758813

image

^ The above goes with the existing
image

- [ ] This is in response to a GitHub issue: *[Link to issue]*
- [X] This is in response to a discussion or thread on Discord:  https://discord.com/channels/441414116914233364/1328735109481758813
- [X] The goal of this PR is detailed below:

## Testing / Validation

- [x] The changes being implemented with this bugfix include comments that contain example log(s) to test with
  - [ ] I have submitted the example log(s) to the xivanalysis static on FFLogs for record keeping
- [x] I used the log(s) listed below to develop and test this bugfix:

Level 100:
https://xivanalysis.com/fflogs/fxrmAT2nVHPZRa61/2/3

Level 80:
https://xivanalysis.com/fflogs/k8KRYNLZJPnB6yfg/136/4

Level 100 with no TBN mistakes:
https://xivanalysis.com/fflogs/8kd1FRA9GxKagWMz/4/70

## Job Maintenance
- [ ] I am active on the xivanalysis Discord and part of the job maintainers group for the job(s) in this PR
- [ ] I have collaborated with one or more job maintainers for the job(s) in this PR while developing it
- [ ] I prefer to receive any communications through GitHub only

I'm not technically part of the job maintainer's group but I'm active on the discord and happy to talk through anything.

PR #2277, opened 2025-12-29
[DRK] Fix bug related to incorrectly logged dark arts for multiple dark knights in party

VioletStardustXIV:
## Pull request type

## Pull request details

There appears to be a bug with the DRK module when multiple Dark Knights are in the party for the lost Dark Arts module.

There is validity to checking TBNs from multiple Dark Knights --- if I apply TBN to myself and then someone applies TBN to me before it popped, I lose my Dark Arts charge, but it appears the logic is flawed. [This log](https://xivanalysis.com/fflogs/a:pjy6xdNvazDkW1Cm/29/43) seems to show the linked Dark Knight having a suggestion due to a different Dark Knight using a TBN on a different party member at 5:57 for example.

It is probably possible to catch every edge case for lost dark arts, but also it seems incorrect to mark 'against' someone for a secondary DRK overwriting their TBN anyway. As a result, updated this to only care about TBNs from self to self (which it actually did before).

This is a regression I introduced in https://github.com/xivanalysis/xivanalysis/pull/2241 --- and the behaviour now matches what it was before that regression.

## Testing / Validation

Example log with incorrect Dark Arts missed logs (two DRKs in party):
https://xivanalysis.com/fflogs/a:pjy6xdNvazDkW1Cm/29/43

Example log with correct Dark Arts missed logs:
https://xivanalysis.com/fflogs/fxrmAT2nVHPZRa61/2/3

(After the change, both appear correctly)

## Job Maintenance
Not an official maintainer (I'd be happy to get more involved, though) but active in discord. Let me know if you have any concerns.

### DIFF
--- a/src/parser/jobs/drk/changelog.tsx
+++ b/src/parser/jobs/drk/changelog.tsx
@@ -8,6 +8,11 @@ export const changelog = [
 	// },
 	//
 	{
+		date: new Date('2025-12-29'),
+		Changes: () => <>Fixed bug related to missed Dark Arts being incorrectly logged with multiple Dark Knights in the party.</>,
+		contributors: [CONTRIBUTORS.VIOLET],
+	},
+	{
 		date: new Date('2025-12-20'),
 		Changes: () => <>Patch 7.4 bump, minor bug fix related to Esteem evaluator, and severities adjusted.</>,
 		contributors: [CONTRIBUTORS.VIOLET],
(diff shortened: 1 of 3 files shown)


AkairyuGestalter (approved):
lgtm as a bugfix. If you decide you want to be able to track "the other DRK overwrite yours"  as a separate informational thing later, you might consider looking at RaidBuffWindow and the RaidBuffOverwriteEvaluator which does basically that.

PR #2290, opened 2026-01-17
restore Chinese language support with Cafemaker V2 API

hikarievo:
- Enable Chinese game edition in language configuration
- Route Chinese requests to Cafemaker V2 API with 'chs' language code
- Keep other languages on XIVAPI v2
- Fix language code mapping between internal 'zh' and API-specific 'chs'

### DIFF
--- a/src/components/ui/DbLink.tsx
+++ b/src/components/ui/DbLink.tsx
@@ -26,15 +26,18 @@ export interface ProviderProps {
 /** Tooltip data provider. Wrapping here to supply i18n language context. */
 export function Provider({children}: ProviderProps) {
 	const {i18nStore} = useContext(StoreContext)
+	const gameLang = i18nStore.safeGameLanguage
+	const apiLanguage = gameLang === Language.CHINESE
+        ? 'chs'  // Cafemaker using 'chs' instead of 'zh'
+        : gameLang  
 
-	// const baseUrl = i18nStore.gameLanguage === Language.CHINESE
-	// 	? 'https://cafemaker.wakingsands.com'
-	// 	: undefined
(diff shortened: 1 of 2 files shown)


src/components/ui/DbLink.tsx
```diff
@@ -26,15 +26,18 @@ export interface ProviderProps {
 /** Tooltip data provider. Wrapping here to supply i18n language context. */
 export function Provider({children}: ProviderProps) {
 	const {i18nStore} = useContext(StoreContext)
+	const gameLang = i18nStore.safeGameLanguage
+	const apiLanguage = gameLang === Language.CHINESE
+        ? 'chs'  // Cafemaker using 'chs' instead of 'zh'
```
AkairyuGestalter, line 31: I get an uncaught ReferenceError about the 'chs' language not being defined when I try to run this. I also briefly tried reverting most of these changes aside from the baseUrl selection change, and instead updating LANGUAGES.ts' definition for Language.Chinese to have that be 'chs' instead, and I got the same issue. I'm not sure what the solution needs to be, but this will prevent the side from loading for any language as-is.

src/components/ui/DbLink.tsx
```diff
@@ -26,15 +26,18 @@ export interface ProviderProps {
 /** Tooltip data provider. Wrapping here to supply i18n language context. */
 export function Provider({children}: ProviderProps) {
 	const {i18nStore} = useContext(StoreContext)
+	const gameLang = i18nStore.safeGameLanguage
+	const apiLanguage = gameLang === Language.CHINESE
```
AkairyuGestalter, line 30: Missing the import of the Language object

AkairyuGestalter (changes requested):

PR #2302, opened 2026-04-02
Add a warning for not-level-cap content

Toastdeib:
## Pull request type

- [ ] This is a patch bump

## Pull request details

- [x] This is in response to a discussion or thread on Discord: https://discord.com/channels/441414116914233364/1488926686912319649
- [ ] The goal of this PR is detailed below:

The PR adds a warning into the About box at the top of the page when the user tries to analyze content below the current level cap, warning that analysis **will** (not may) be inaccurate, with the goal of reducing the number of confused new users wondering why the page is yelling at them about missing skills that they don't have yet (among other things).

## Testing / Validation

* Log provided by the user who made the request (level 80 content, shows warning): https://www.fflogs.com/reports/TN7rFJ3xzWkdqtYD?fight=1&source=2
* Log of my own (level 100 content, doesn't show warning): https://www.fflogs.com/reports/QXk6CDqj2tWT9bnA?fight=6&type=summary&source=8
* Log provided by otocephaly on Discord (level 60 BLU content, doesn't show warning): https://www.fflogs.com/reports/a:7JPTyHrg4KVMBhfp?fight=3&type=summary&source=4

## Job Maintenance

### DIFF
--- a/src/data/PATCHES/patches.ts
+++ b/src/data/PATCHES/patches.ts
@@ -15,6 +15,7 @@ export interface PatchInfo {
 	branch?: PatchBranch
 }
 
+export const LEVEL_CAP = 100
 // This is all right from /PatchList - should be easy to sync Eventually™
 export const FALLBACK_KEY = '✖'
 export const PATCHES = ensureRecord<PatchInfo>()({
--- a/src/data/PATCHES/utilities.ts
+++ b/src/data/PATCHES/utilities.ts
@@ -1,7 +1,7 @@
 import _ from 'lodash'
 import {Report} from 'report'
 import {GameEdition} from '../EDITIONS'
-import {FALLBACK_KEY, PATCHES, PatchInfo, PatchNumber} from './patches'
+import {FALLBACK_KEY, LEVEL_CAP, PATCHES, PatchInfo, PatchNumber} from './patches'
 
 interface PatchData {[key: string]: PatchInfo}
 const patchData: PatchData = PATCHES
@@ -46,4 +46,13 @@ export function patchSupported(
 		: Infinity
 
 	return _.inRange(at, fromDate, toDate)
+}
+
+export function contentSupported(loggedLevel: number | undefined) {
+	if (!loggedLevel) {
+		// If the log doesn't include a level, assume it's supported to preserve the status quo.
+		return true
+	}
+
+	return loggedLevel < LEVEL_CAP
 }
--- a/src/data/PATCHES/utilities.ts
+++ b/src/data/PATCHES/utilities.ts
@@ -2,7 +2,7 @@ import _ from 'lodash'
 import {Report} from 'report'
 import {GameEdition} from '../EDITIONS'
 import {FALLBACK_KEY, LEVEL_CAP, PATCHES, PatchInfo, PatchNumber} from './patches'
-import {JobKey} from "../JOBS"
+import {JobKey, JOBS} from "../JOBS"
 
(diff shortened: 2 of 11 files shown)


src/data/PATCHES/utilities.ts
```diff
@@ -47,3 +47,12 @@ export function patchSupported(
...
 	return _.inRange(at, fromDate, toDate)
 }
+
+export function contentSupported(loggedLevel: number | undefined) {
+	if (!loggedLevel) {
+		// If the log doesn't include a level, assume it's supported to preserve the status quo.
+		return true
+	}
+
+	return loggedLevel < LEVEL_CAP
```
xiashtra: This does not take into account jobs that have a maximum level less than 100.  For instance, BLU content is supported at BLU's maximum level cap, which is currently 80.

xiashtra (changes requested):

src/data/PATCHES/utilities.ts
Toastdeib: A very good point, and not one I considered. I can bake an exception into this function that ignores limited jobs.

Toastdeib: Fixed this, and also fixed the logic because I was originally returning the opposite of what the function name said. 🙃

xiashtra (approved):

src/data/PATCHES/utilities.ts
```diff
@@ -47,3 +48,17 @@ export function patchSupported(
...
 	return _.inRange(at, fromDate, toDate)
 }
+
+export function contentSupported(loggedLevel: number | undefined, job: JobKey) {
+	if (!loggedLevel) {
+		// If the log doesn't include a level, assume it's supported to preserve the status quo.
+		return true
+	}
+
+	if (job === "BLUE_MAGE") {
```
AkairyuGestalter: As opposed to hard-coding blue mage here, can we add another nullable prop to the Job interface to indicate whether the job is a limited job? That'd follow the pattern I used for indicating that PLD and DRK use MP despite not being the Caster/Healer role for the purposes of displaying the MP bar in the timeline, and would mean BST (or any other future ones) just need to set that in their Job definition and it'll be handled for free.

src/data/PATCHES/patches.ts
```diff
@@ -15,6 +15,7 @@ export interface PatchInfo {
 	branch?: PatchBranch
 }
 
+export const LEVEL_CAP = 100
```
AkairyuGestalter, line 18: This is probably fine as a constant. It feels weird, but I can't think of anywhere else better to put it... If you have a brainwave while fixing the other comment then cool, otherwise I'm not blocking on this one.

AkairyuGestalter (commented):
One thing we could make better, one thing I can't think of a better for!

src/data/PATCHES/utilities.ts
Toastdeib: That's a significantly better solution than mine, I'm happy to implement it.

src/data/PATCHES/patches.ts
Toastdeib: I admittedly didn't love putting it here, but I couldn't think of a better place for it. I'll sit on it for a day and push the change for your other comment tomorrow evening, with or without an update to this depending on whether I have any brilliant ideas.

Edit: Didn't have any brainwaves, so I'm just going to mark this as resolved now that I've addressed the other comment.

AkairyuGestalter (approved):

PR #2310, opened 2026-06-02
Add UMAD utaStatus IDs

AkairyuGestalter:
## Pull request type

- [x] This is a patch bump

## Pull request details

UMAD adds a couple new statuses that prevent players from acting. Ran the regenerate script to capture those so we respect them for uptime calculations.

Yes, one of them has a weirdly encoded "name". Apparently this is normal-ish per a response in the XIVAPI Discord.

## Testing / Validation

Nothing in particular.

## Job Maintenance
- [x] I am active on the xivanalysis Discord


ackwell (approved):
confirming the `_rsv` name is expected behavior. surprised we've not run into it before in UTA.

